PR #70, opened 2013-12-15
Update TNonblockingServer.cpp

zmldndx:
if failed to notify pipe, the TConnection will be leak, which cause memory getting larger, and the most important, cause fd getting larger, so eventually it will cause "cannot get new fd" error.
if not fixed, it will cause fd leak, you can observ by using "lsof |grep can" to see whether there are fds with "cannot identify protocol"


logicouter:
this pull request hasn't been merged into master?

Jens-G:
Thank you for issuing a pull request for your patch. Currently we do not accept pull requests from Github, please see http://thrift.apache.org/docs/HowToContribute/ for instructions on submitting your patch back to Apache Thrift. Thanks

PR #84, opened 2014-03-21
Tree/Recursive struct support in thrift

djwatson:
Patches to implement tree/list/co-recursive structures in thrift, a la protobufs.


ringerc:
I see a bunch of mutual recursive structure support in the tests, but nothing for an optional single self-referential member, like

```
struct RecSelf {
   1: i16 item
   2: optional RecSelf 
 }
```

Thift does not appear to generate sensible code for this. The example I am working with is from the opentracing jaeger toolset, namely the IDL

```
struct Downstream {
    1: required string serviceName
    2: required string serverRole
    3: required string host
    4: required string port
    5: required Transport transport
    6: optional Downstream downstream
}
```

This works fine in Java, Go, etc. But in C++ it produces the nonsensical code

```
class Downstream {
 public:
 
  /* blah elided blah */

  virtual ~Downstream() throw();
  std::string serviceName;
  std::string serverRole;
  std::string host;
  std::string port;
  Transport::type transport;
  Downstream downstream;

  _Downstream__isset __isset;

  /* blah elided blah */
};
```

in which the class is a by-value member of its self.

I think the ideal solution here would be to adopt something like `std::optional` (c++17 proposal) as a `thrift::optional` and use that instead of the `__isset` stuff for optional members. But backwards compatibility could be a problem.

ringerc:
Created as https://issues.apache.org/jira/browse/THRIFT-4484

ringerc:
Anyone else facing that issue: workaround is to declare a reference:

```
struct RecSelf {
   1: i16 item
   2: optional RecSelf &self
 }
```

Note the `&`.

PR #85, opened 2014-03-25
fix unicode bug in readString&writeString

wmttom:


jfarrell:
Thanks for your patch @wmttom, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #93, opened 2014-04-02
THRIFT-2429

djwatson:
Original diff summary:

```
Add "terse_writes" c++ gen option to thrift compiler.

Summary:
Add "terse_writes" c++ generator option to thrift compiler, to
suppress writing non-optional/non-required fields if they haven't
changed from the default value.

 - This is default-off, whitelist-only.
   It may be whitelisted per-.thrift, or per struct.

 - On a Unicorn loadgen benchmark appeared to save roughly 2/3rds of
   the SearchRequest deserialization cost (fields are only sparsely set).

 - This will not affect embedded structs, or strings with non-""
   default values, since we only want this change to do inexpensive
   integer-like comparisons.

 - Also, the user is advised that:
   - default values for whitelisted structs shouldn't be changed.
     (or, if they are, upgraded to optional/required)
   - if one re-read()s into an old struct, one should __clear() it first.
     however, this is not a new problem for anybody with optional fields.

This diff introduces "safe" mode for `terse_writes` in order to
prevent us from producing stupid bugs in PHP (and Python) land.

If we don't require explicit default value to be set, we have to
make PHP / Python logic to correctly understand `null` / `None` and
treat them as default field values.
```

davejwatson:

It sounds like the proposal in THRIFT-2429 is to make this the default for optional fields as well, with no compiler flag, which is just minor changes to this diff.

### DIFF
--- a/compiler/cpp/src/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/generate/t_cpp_generator.cc
@@ -42,6 +42,20 @@ using std::vector;
 
 static const string endl = "\n";  // avoid ostream << std::endl flushes
 
+namespace {
+
+                                                                                                                                                                                                                                      
+enum TerseWrites { TW_DISABLED = 0, TW_SAFE = 1, TW_ALL = 2 };                                                                                                                                                                        
+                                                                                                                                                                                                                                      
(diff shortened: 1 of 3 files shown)


jeking3:

PR #95, opened 2014-04-06
Fix TExtensibleServlet to be able to add custom headers at init() stage.

abashev:
Right now with initialization of customHeaders property at init(ServletContext) is not possible to add custom headers at own init() stage. This small patch will fix it.


jfarrell:
Thanks for your patch @abashev, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #113, opened 2014-04-30
THRIFT-2471

djwatson:
three changes as discussed in THRIFT-2471:
- Update parser to use & instead of cpp.ref
- Recursion depth limit, a la protobufs
- shared_ptr instead of raw pointer w/deletes in the destructor.


jeking3:
Sorry, #1195 has nothing to do with this one.  I was off by one.

PR #122, opened 2014-05-17
0.9.x

romange:
Eliminate redundant allocations for temporary args during RPC processing. 
The change is suitable for ThreadPoolServer where each thread processes its own connection.

We create thread local request/response arguments instead of allocating them on stack in TDispatchProcessor. This way, the next  RPC reuses already allocated strings and arrays.
In order to make sure that the objects are semantically empty we added a new method "clear" that resets all its fields to their initial values (0 for basic types, empty for containers) and unsets optional fields.

On our system total CPU load during high QPS periods (18000 QPS and more ) was cut by factor of 2 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ gen-*
 /compiler/cpp/thriftl.cc
 /compiler/cpp/thrifty.cc
 /compiler/cpp/thrifty.h
+/compiler/cpp/thrifty.hh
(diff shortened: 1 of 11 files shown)


jfarrell:
hey @romange, we do not merge changes into our branches once they have been released. If you would like to submit a patch to trunk for this issue then someone would be happy to review. Thank you

PR #123, opened 2014-05-18
Update t_cpp_generator.cc

fidelchen:
for std::greater


jfarrell:
Thanks for your patch @fidelchen, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #124, opened 2014-05-19
THRIFT-2522 TypeScript extension for Thrift Compiler

ghost:
Adds the possibility to generate TypeScript Definition Files in addition to the generated JavaScript code (new parameter 'ts').
Supports JQuery, NodeJS support will come soon.


beejjacobs:
> NodeJS support will come soon

Any update on NodeJS support?

kamek-pf:
I was wondering the same thing.
For now I'm writting the libdefs myself, but it would be a nice addition for sure.

Jens-G:
We have a ts target?

kamek-pf:
Apparently :p
It generates definition files that matches the `js` target, but it looks like there is no equivalent for `js:node`.

PR #126, opened 2014-05-23
Add extract file

samt42:
add extract file for make dist


jfarrell:
Thanks for your patch @samt42, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #129, opened 2014-05-27
THRIFT-2551 OutOfMemoryError kills serve thread

Fuud:
THRIFT-2551 OutOfMemoryError "unable to create new native thread" kills serve thread.

If executor can not create new thread (because of exceed descriptors limit for example) it throws OutOfMemoryError. This case is same as if RejectedExecutionException is thrown from executor: we can wait and retry.


andre77:
Hi,
in our environment we use librthrift version 0.9.0, where we face the problem with OutOfMemoryError "unable to create new native thread"
ok, with the fix in the newer version the serve thread will "survive" such an exception, but still a big question for me, why there are so many threads running?
even after increasing the max processes to 4096, we face such a problem...
someone an idea?

PR #134, opened 2014-06-06
Update simple_json_protocol.go

hpgood:
fix bug:when read the end of message ,don't need ParsePostValue().


jfarrell:
Thanks for the patch

PR #140, opened 2014-06-13
rb:namespaced: generate Ruby files in idiomatic namespaced directories.

aughr:
Idiomatic Ruby expects that namespaces are mirrored on the filesystem. This new
namespaced option lets that happen.

(https://issues.apache.org/jira/browse/THRIFT-2574, patch there also)


bracki:
Hi,
this [test](https://github.com/Jimdo/thrift/commit/47278d6e66987b375200982e0218657b4c0af62d) exposes the regression.

PR #144, opened 2014-06-22
Python 3 support

tbartelmess:
Changed to python library to work both on Python 2 and Python 3 using the six library.


jfarrell:
Thanks for the initial work on this patch, closing as #213 appears to be a more complete version of this patch

PR #146, opened 2014-06-24
add cpp lib that support epoll server, thus we can use epoll and multi p...

anthonywei:
this version support thrift with c++ server use epoll and multiprocess

add four files as follows:

server/TEpollServer.h 
server/TEpollServer.cpp

transport/TServerEpollSocket.h 
transport/TServerEpollSocket.cpp

Advantags: we use epoll to watch the connections events, not use recv() and timeout. more effective. And also it is even more effective than recv() + timeout + multi-thread and also it is of simlity. you do not need to care about thread-safe class, objects and anything.


jfarrell:
Thanks for your patch @anthonywei, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #148, opened 2014-06-27
changed the dict() to dict:dict() in the type spec of types and generate...

cjimison:
...d the -type spec for each record in the _types.hrl file

This is needed for Erlang R17 support as well as being able to run the code in dialyzer


jfarrell:
Thanks for your patch @cjimison, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #169, opened 2014-07-31
Add LoggingRejectedExecutionHandler on TThreadPoolServer ThreadPoolExecutor()

darionyaphet:
Add LoggingRejectedExecutionHandler on TThreadPoolServer ThreadPoolExecutor() to avoid some uncached-exception to break server from running .


jfarrell:
Thanks for your patch @darionyaphet, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

darionyaphet:
@jfarrell thank you ！

PR #183, opened 2014-08-10
Cleanup python generation

jamesbroadhead:
- Remove some trailing whitespace
- Slightly sort imports
- Replace is_oneway() calls with a bool


jfarrell:
Thanks for your patch @jamesbroadhead, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #184, opened 2014-08-10
python: Make message-writing more generic

jamesbroadhead:
Make the use of oprot & self._oprot generic, to remove some complexity
from the generator


jfarrell:
Thanks for your patch @jamesbroadhead, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #185, opened 2014-08-11
Seems the wrong prop name (should be recv_buf)

vitalpoltava:
Seems the name of the property is not correct -- it is read_buf, yet it should be recv_buf.


jfarrell:
Thanks for your patch @vitalpoltava, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #201, opened 2014-08-21
JavaScript: set Accept: and Content-Type: headers in request

stigsb:
When a JS client communicates with a backend supporting several encodings, knowing which protocol to use is a bit kludgy. This patch tries to fix that by always setting the Accept: and Content-Type: request headers to "application/json".


jfarrell:
Thanks for your patch @stigsb, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #203, opened 2014-08-26
Update main path in bower.json

ejkj:


jfarrell:
Thanks for your patch @ejkj, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #213, opened 2014-09-11
Thrift-1857: Python 3 support, redux

eevee:
As mentioned on [the Jira ticket](https://issues.apache.org/jira/browse/THRIFT-1857?focusedCommentId=14104848&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14104848), there are still some minor problems with `fastbinary.c`, which uses a private API that no longer exists in Python 3.  I think it's worth just removing outright.

But the test suite passes completely on Python 3, now.

----

FYI: I left the company I worked for and no longer have any need for Thrift, so I won't be working on this further.  Anyone else, feel free to pick it up.  Anyone who desperately needs Thrift on Python 3 may want to try [thriftpy](https://github.com/eleme/thriftpy), which is pure-Python, works everywhere, has a similar API, and doesn't require a manual code generation step.

### DIFF
--- a/compiler/cpp/src/generate/t_py_generator.cc
+++ b/compiler/cpp/src/generate/t_py_generator.cc
@@ -802,7 +802,7 @@ void t_py_generator::generate_py_struct_definition(ofstream& out,
     out <<
       indent() << "def __repr__(self):" << endl <<
       indent() << "  L = ['%s=%r' % (key, value)" << endl <<
-      indent() << "    for key, value in self.__dict__.iteritems()]" << endl <<
+      indent() << "    for key, value in self.__dict__.items()]" << endl <<
       indent() << "  return '%s(%s)' % (self.__class__.__name__, ', '.join(L))" << endl <<
       endl;
 
@@ -2202,7 +2202,7 @@ void t_py_generator::generate_deserialize_container(ofstream &out,
   // For loop iterates over elements
   string i = tmp("_i");
   indent(out) <<
-    "for " << i << " in xrange(" << size << "):" << endl;
+    "for " << i << " in range(" << size << "):" << endl;
 
     indent_up();
 
--- a/compiler/cpp/src/generate/t_py_generator.cc
+++ b/compiler/cpp/src/generate/t_py_generator.cc
@@ -1987,7 +1987,7 @@ void t_py_generator::generate_process_function(t_service* tservice,
       indent_down();
       for (x_iter = xceptions.begin(); x_iter != xceptions.end(); ++x_iter) {
         f_service_ <<
-          indent() << "except " << type_name((*x_iter)->get_type()) << ", " << (*x_iter)->get_name() << ":" << endl;
+          indent() << "except " << type_name((*x_iter)->get_type()) << " as " << (*x_iter)->get_name() << ":" << endl;
         if (!tfunction->is_oneway()) {
           indent_up();
           f_service_ <<
--- a/compiler/cpp/src/generate/t_py_generator.cc
+++ b/compiler/cpp/src/generate/t_py_generator.cc
@@ -373,7 +373,7 @@ void t_py_generator::init_generator() {
   f_consts_ <<
     py_autogen_comment() << endl <<
     py_imports() << endl <<
-    "from ttypes import *" << endl <<
+    "from .ttypes import *" << endl <<
     endl;
 }
 
@@ -1044,7 +1044,7 @@ void t_py_generator::generate_service(t_service* tservice) {
   }
 
   f_service_ <<
-    "from ttypes import *" << endl <<
+    "from .ttypes import *" << endl <<
     "from thrift.Thrift import TProcessor" << endl <<
     render_fastbinary_includes() << endl;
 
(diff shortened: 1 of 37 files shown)


jfarrell:
hey @eevee, thanks for the work on this patch, any chance you could rebase this against the current trunk please, would like to see it make it in for THRIFT-1857

helgridly:
I'm happy to pick this up if @eevee isn't around or doesn't have the time.

eevee:
Feel free, @helgridly, and thanks  :)  ♥

helgridly:
I did the rebase, ran tests on python3, and fixed up a few issues over (in my own fork)[https://github.com/broadinstitute/thrift/tree/eevee/python3]

However, the tests fail on python2 with the same issue documented on JIRA (here)[https://issues.apache.org/jira/browse/THRIFT-1857?focusedCommentId=14046961&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14046961]

@jfarrell How would you like me to proceed? I can open a separate PR with my changes, but I don't know what to do about the failing tests on python2.

jeroenvlek:
The commit 3b4b03e introduces a conversion to bytearray in TBinaryProtocol.writeString():


```python
# after commit 3b4b03e and 3d912d7
def writeString(self, str):
  encoded = bytearray(str, 'utf-8')
  self.writeI32(len(encoded))
  self.trans.write(encoded)

# before
def writeString(self, str):
  self.writeI32(len(str))
  self.trans.write(str)
```

This causes problems for people trying to write byte strings. See for example here: https://github.com/wbolster/happybase/issues/40

c4urself:
@jfarrell - (bump) I've been using @helgridly's version of the code with Python 2 and 3 and it seems to work well, @jeroenvlek's changes seem useful as well -- is there anything that can be done to get this pulled in // what are next steps?

jfarrell:
@c4urself working on the next RC now, if we can get all the changes in this patch and the comments from @jeroenvlek into one patch updated against trunk and passing tests then happy to get it in

jeroenvlek:
I'd be happy to make a new pull request with the adapted writeString and readString, but do you want to have the assumption that the incoming string is either encoded with UTF-8 or a bytestring hard-coded?

Maybe you could keep writeString for byte strings and create a writeUtf8?

lababidi:
@jfarrell @jeroenvlek Any progress on this? I'd be happy to help out in any way.

jeroenvlek:
Well, I'm still waiting for an answer on my question above on UTF-8. Other than that, my patch should be easy to apply. If work permits, I'll rebase my patch on the current master soon.

jfarrell:
sorry, had missed the question on this. like the idea of pairing it as TType.UTF8 and TBinaryProtocol.writeUtf8()

jeroenvlek:
@lababidi I'm kind of swamped with work and my weekends are packed too. I don't see myself doing this before the end of next week at the earliest. If you want to pick it up, by all means, go for it. 

The idea is to keep writeString()/readString() for byte arrays and introduce a writeUtf8()/readUtf8() (and make sure it's called for TType.UTF8). You can deduce how to implement it from the try/catch below my second edit above. Should be pretty straightforward :)

Please notify this thread if you pick it up, then I know if my help is still needed or not.

c4urself:
Note: I've tried this branch on Python3 and Python2 and while things seem to work correctly on Python 3; In Python2, the `six.BytesIO` object is not recognised correctly by the Python 2 `PycStringIO_InputCheck` in https://github.com/apache/thrift/blob/0.9.2/lib/py/src/protocol/fastbinary.c#L624 causing the `TBinaryProtocolAccelerated` protocol to fail.

This may save someone else some digging around -- not sure where the fix should be, a hacky workaround is to import StringIO from cStringIO for Python2... -- or maybe someone else here has a better idea?

eevee:
Yeah, like I said in the OP, `fastbinary.c` uses a private C interface to cStringIO that was removed in Python 3, and I'm not familiar enough with it to fix it.

FYI: I left the company I worked for and no longer have any need for Thrift, so I won't be working on this further.  Anyone else, feel free to pick it up.  Anyone who desperately needs Thrift on Python 3 may want to try [thriftpy](https://github.com/eleme/thriftpy), which is pure-Python, works everywhere, has a similar API, and doesn't require a manual code generation step.

c4urself:
I've rebased this against the master branch incorporating @jeroenvlek's suggestions -- tried to be careful with the rebase but would love a set of eyes to take a look at it -- also not sure if we want to open this as a new PR?

 https://github.com/apache/thrift/compare/master...Tubular:feature/python-3-support?expand=1

jeroenvlek:
This does miss the write-/readUtf8(). I'd suggest keeping py2 code as is, and the py3 code in write-/readUtf8()

Jens-G:
> would love a set of eyes to take a look at it 

Absolutely. There have been a number of people asking for Python 3 support so - at least in theory - it should not be that hard to find more eyes. @Pythonians - we need your opintion!

nsuke:
I did some research on this.
TType.UTF8 was added years ago but currently not used by any language.
So using this now will result in incompatibility against any other implementation.

I rebased @c4urself 's rebase and fixed both fastbinary (py2) and string/binary (py2 & py3) issues : #680

PR #235, opened 2014-10-03
THRIFT-1618: synchronize access to hashtable in FieldMetaData

dhelder:


dhelder:
Replaced with https://github.com/apache/thrift/pull/348

PR #238, opened 2014-10-06
Work around MemoryError when using an SSL endpoint

jbq:


jfarrell:
Thanks for your patch @jbq, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #253, opened 2014-11-07
Pull TAsyncProcessor interface out of TBaseAsyncProcessor

m50d:
As briefly mentioned on IRC (I'm lmm). This makes it possible to use e.g. TNonblockingServer with a custom async processor that doesn't extend TBaseAsyncProcessor.

(Use case: I have an implementation in scrooge that generates such async processors, for use in scala).

Async processors still need to "implement" TProcessor so that they can be configured as a processor (the implementation can be simply `return false` as it is in `TBaseAsyncProcessor`); this is somewhat awkward but a change to make that unnecessary would be much more intrusive.


bsideup:
:+1:

PR #271, opened 2014-11-17
Add moveable_types option to C++ generator

flandr:
Setting this option enables generation of MoveConstructible types,
allowing types to be bound via move in std::bind. This is especially
pleasant when using Thrift container types.


jfarrell:
Thanks for your patch @flandr, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

flandr:
Of course, and it was https://issues.apache.org/jira/browse/THRIFT-2836 & was merged. Not sure exactly how, tho clearly not through this interface ;)

PR #273, opened 2014-11-17
Fixes THRIFT-2817 and THRIFT-2839

ghost:
Current implementation of json protocols uses buffer peeks extensively.
But in certain situations this causes the buffer to read beyond the json message,
which stalls the whole connection. (THRIFT-2817)

To test this fix I ported the integration tests from the nodejs thrift library to go.
Those are comprehensive integration tests that tests the whole thrift stack.
The tests revealed another problem in TFramedTransport.

When calling TFramedTransport.Read, it should return remaining bytes,
but it doesn't when passing a []byte that is bigger than the current frame_size. (THRIFT-2839)

This PR fixes THRIFT-2817 and THRIFT-2839.

### DIFF
--- a/lib/go/test/tests/protocols_test.go
+++ b/lib/go/test/tests/protocols_test.go
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


bivas:
We're still experiencing read errors with this fix merged (`Not enough frame size 4085 to read 4096 bytes`)

The commit is missing a `return` statement in case of `framesize` smaller than the `buffer` and `err == nil ` looks strange

Jens-G:
@bivas Could you provide a PR addressing these issues? Thanks!

bivas:
@Jens-G unfortunately, no. Was pointing this out so maybe @wingedkiwi can have a quick look (and maybe a fix) on this issue.

wingedkiwi:
I'm turning my attention to thrift again in a few weeks. Then I can have a look.

PR #299, opened 2014-12-02
THRIFT-2869: Run JSON schema validator from test

stigsb:
This patch makes "make check" test the JSON generator by running a schema validator on generated files.
Jira: THRIFT-2869

### DIFF
--- a/lib/json/schema.json
+++ b/lib/json/schema.json
@@ -1,11 +1,12 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
 
-  "id": "http://thrift.apache.org/program-schema#",
+  "id": "http://thrift.apache.org/schema.json#",
   "description": "Schema for Apache Thrift protocol descriptors",
 
   "definitions": {
     "type-id": {
+      "title": "Any type id (name)",
       "enum": [
         "void",
         "string",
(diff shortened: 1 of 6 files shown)


lib/json/schema.json
```diff
@@ -20,66 +21,78 @@
...
+            "type": "object",
+            "properties": {
+                "typeId":      { "type": "string" },
+                "keyTypeId":   { "$ref": "#/definitions/type-id" },
+                "keyType":     { "$ref": "#/definitions/type-desc" },
+                "valueTypeId": { "$ref": "#/definitions/type-id" },
+                "valueType":   { "$ref": "#/definitions/type-desc" },
+                "class":       { "type": "string" },
+                "elemTypeId":  { "$ref": "#/definitions/type-id" },
+                "elemType":    { "$ref": "#/definitions/type-desc" }
```
nsuke: @stigsb Not sure you're still around but would you mind elaborating on why you moved key/value/class/elem here ?
Thanks for sharing this BTW. It's pity this one escaped our eyes for so long.

stigsb: It's been a while, so my memory may be off here, but I seem to remember that it was because the validator wouldn't catch failures otherwise.

nsuke: Thanks for the explanation.
I manually modified generated code but it seems it works fine.
I understand it may have been another issue but based on information we have, I'm excluding this part for now.

You can find update PR here: #749 
It already found one regression (outdated schema) introduced after this PR :smile:

PR #304, opened 2014-12-03
THRIFT-2878 - Go validation support of required fields

cvlchinet:
Adds required fields validation to Go.


cvlchinet:
Rebased and changed to use OptionalRequiredTest.thrift.

PR #311, opened 2014-12-09
THRIFT-2882 Lua Generator: using string.len funtion to get

winsweet:
struct(map,list,set) size


jeking3:
@jfarrell this pull request can be closed - I verified the code was merged into master.

PR #317, opened 2014-12-12
erl lib: add recv_timeout configuration for client tcp connection

ericliang:
Tcp recv timeout is only used in server socket but in client, and thus client could block there to wait for the server response, which is not satisfied in time-critical situations.


jfarrell:
Thanks for your patch @ericliang, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #318, opened 2014-12-12
Thrift 2850 CMake for Apache Thrift

bachp:
These two commits add CMake support for C++, C and Java.
They also add a CMake job to travis


bachp:
I think theis is obsolete now. I will close it.

PR #322, opened 2014-12-17
THRIFT-2900 Wrap Python THttpClient withTimeout in a try/finally

ascandella:
Otherwise the original value of the socket timeout is not restored
properly.


breerly:
@sectioneight looks like the build failed?

sectioneight:
Indeed it did. Seems to be unrelated...

PR #325, opened 2014-12-17
THRIFT-2561: Typedef enums using NS_ENUM.

jimspeth:


jeking3:
@jfarrell this pull request is not necessary as the work was done in a duplicate item.

PR #326, opened 2014-12-17
THRIFT-2896: Added support for a custom base class.

jimspeth:


jeking3:
This needs to be rebased against the current master and squashed.

jeking3:

PR #343, opened 2015-01-02
THRIFT-2924 TNonblockingServer crashes when user-provided event_base is ...

nsuke:
...used


nsuke:
Merged as 8016af800ff19e2b67f691453c573f3d8f34211d

PR #344, opened 2015-01-02
THRIFT-2925 CMake build does not work with OpenSSL nor anything installe...

nsuke:
...d in non-system location

This diff fixes them by correcting invalid CMake variable names:
* OPENSSL_... rather than OpenSSL_...
* LIBEVENT_... rather than Libevent_...
* Boost_INCLUDE_DIRS rather than Boost_INCLUDE_DIR
* LIBEVENT_INCLUDE_DIRS rather than LIBEVENT_INCLUDE_DIR
* ZLIB_INCLUDE_DIRS rather than ZLIB_INCLUDE_DIR

Note:
* OPENSSL_INCLUDE_DIR is correct (rather than ..._DIRS)
* Boost_INCLUDE_DIR exists and actually works for most cases but
  Boost_INCLUDE_DIRS is the one desinged to be included

Also, library headers are now included as SYSTEM headers.


nsuke:
Merged as c444fb581d2157c260f36db091a7cdd4511c6e32, thanks :)

PR #345, opened 2015-01-02
THRIFT-2926: JS: Binary protocol with unit tests.

radekg:


radekg:
This build has failed because of some jslint issues. Any hints how can I help resolving?

bufferoverflow:
jslint is used via lib/js/test/build.xml which also runs the unit tests test.js/test.html

My favorite would be integration of your tests into test.js/test.html and use guessProtocolFactory within the server part https://github.com/apache/thrift/blob/master/lib/js/test/src/test/Httpd.java#L217

radekg:
Thank you. I'll take a look at this.

radekg:
@bufferoverflow I am trying to understand where the problem is. It seems that travis is using a different method of testing than the one documented in the README for js tests. The README assumes grunt as a main way of testing while travis seems to be using ant?
I am also trying to figure out how does one run the main set of tests. Do I have to follow this write up https://thrift.apache.org/docs/BuildingFromSource ?
Regarding the failing tests, travis complains about some invalid bitwise operations while bitwise is allowed by build.xml jslint. Not sure where to start looking for stuff.

radekg:
I've done a little bit more research into this. There's a major issue which will surface in the unit test once I can get it integrated. I've got the test properly integrated in the grunt task. What happens it that PhantomJS doesn't like the following expression:

    String.fromCharCode.apply(null, new Uint8Array( buffer ) )

This has been described here: https://github.com/mozilla/pdf.js/issues/1955
The problem is, if I apply the "fix" mentioned in the comments, the test doesn't pass because unicode characters are decoded incorrectly. So, one way or another, the test will fail. I can see only one solution to this issue. I am going to close this pull request and provide the ThrifBinaryProtocol for JS as an external module to be added on top of Thrift library. Once this is done, I will post a link to the project here.

radekg:
And the code: https://github.com/radekg/thrift-js-binary-protocol

bufferoverflow:
Thanks @radekg for this udate.

any idea about the jslint issues?

    cd lib/js/test
    make check
    lib/js/test/build.xml:216: JSLint: 39 errors in 1 file

They might be related to the unit test issues:
```
cd lib/js/test
ant unittest
...
unittest:
     [echo] Running Unit Tests with headless browser!
     [exec] Xlib:  extension "RANDR" missing on display ":99".
     [java] Timeout: killed the sub-process
     [java] Java Result: -1
     [exec] 'waitFor()' finished in 403ms.
     [exec] Tests completed in 267 milliseconds.
     [exec] 4 assertions of 41 passed, 37 failed.
```

You can also remove ThriftTest.thrift file from this commit it is a duplication.

radekg:
On it and a broader discussion here: https://issues.apache.org/jira/browse/THRIFT-2926

ebremer:
Is there any further work being done to support TBinaryProtocol and TCompact in Javascript?

nsuke:
@ebremer none that I'm not aware of.
You can use lib/nodejs code with browserify (or probably webpack), though.

jeking3:

emmenlau:
Dear all, I'm curious if the issue described in mozilla/pdf.js#1955 still exists and is relevant in this context? Or could this PR be reconsidered?

emmenlau:
I've checked the PR in more detail and found that the code for the binary protocol is not up to date for current thrift anymore. It may be simpler to start over based on the nodejs thrift implementation than to update this PR. Since the original code for this PR seems not maintained anymore, I suggest to leave the PR closed.

PR #346, opened 2015-01-04
THRIFT-2928 Rename the erlang test_server module

bosky101:
Submitted the fix for the issue I filed earlier today.

Avoiding clashes with the in-built erlang app bundled since November '09
Used git mv, then renamed the module in the source.
Verified that git blame continues to work as expected.

More Details on Jira/THRIFT-2928


jfarrell:
THRIFT-2928 closed, thanks for the patch

PR #349, opened 2015-01-09
Pass error through callback

sh1mmer:
Injects the seqid to allow the connection to reference the callback and then associates as many request specific errors with the callback as possible.

Also wraps the callback so that it only fires once.

TODOs logging, more unit tests

I need to find an a better unit testing library for Node that is Apache 2.0 compatible (e.g. upstreamable).

cc @Raynos @kriskowal @sectioneight @breerly @willyham


sh1mmer:
Closing. Meant to do on the Uber repo this isn't ready for upstream yet.

PR #352, opened 2015-01-11
THRIFT-2938 Add headers to Ajax calls in TXHRTransport

degupta:


jeking3:

PR #353, opened 2015-01-11
THRIFT-2939: Js on par with nodejs

radekg:
The generated code for regular JS differs from NodeJS code. NodeJS uses direct return values from `read<Type>()` while regular JS code expects a value from `read<Type>()` to be an object with `.value` property.

This makes it impossible to share generated code between browser JS and NodeJS. The fix is rather simple:
- do not append `.value` in generated code while accessing values read from protocol / transport
- do not return an object with the value property from read, return result read

I am not sure why such difference. Both environment implementations take an input, which is a protocol, and execute the same read methods. There is no need to have an additional complexity in regular JS.

### DIFF
--- a/compiler/cpp/src/generate/t_js_generator.cc
+++ b/compiler/cpp/src/generate/t_js_generator.cc
@@ -1503,6 +1503,12 @@ void t_js_generator::generate_deserialize_field(ofstream& out,
     }
 
     out << ";" << endl;
+    
+    if (!gen_node_) {
+      // allow either .value, prior to 1.0.0 or direct result - 1.0.0 and up.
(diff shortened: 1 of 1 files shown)


malandrew:
Good to see these changes. I encountered some of the same issues. I hacked around for a few days and ported all the browser features to the nodejs codebase for use with browserify. There is a massive diff here, https://github.com/uber/thrift/commit/9ca44b40504cc8d45d0c1e47709128e474d991df , but I'm going to work over the next day or so to break this monolithic commit up into smaller commits capturing all the changes I made.

jeking3:
Tag THRIFT-2939 so the PR shows up in github as a linked PR.

jeking3:

PR #357, opened 2015-01-16
THRIFT-2423: Facebook's THeader protocol and transport for cpp

djwatson:
Pulled from a pre-c++11 / folly / iobuf version.  Removed some unused stuff.

This adds:
* headers
* dynamic protocol selection
* dynamic zlib compression
* backwards compat with unframed / framed

Added it to TestServer / TestClient.  Tested various combinations, seems to work.

There were enough changes made vs. the fbthrift version that this could probably use a review.

### DIFF
--- a/lib/cpp/src/thrift/protocol/THeaderProtocol.cpp
+++ b/lib/cpp/src/thrift/protocol/THeaderProtocol.cpp
@@ -1,0 +1,263 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef THRIFT_PROTOCOL_THEADERPROTOCOL_CPP_
+#define THRIFT_PROTOCOL_THEADERPROTOCOL_CPP_ 1
+
+#include <thrift/protocol/THeaderProtocol.h>
+#include <thrift/protocol/TCompactProtocol.h>
+#include <thrift/TApplicationException.h>
+
+#include <limits>
+#include <boost/static_assert.hpp>
+
+
+namespace apache { namespace thrift { namespace protocol {
+
+void THeaderProtocol::resetProtocol() {
+  if (proto_ && protoId_ == trans_->getProtocolId()) {
+    return;
+  }
+
+  protoId_ = trans_->getProtocolId();
+
+  switch (protoId_) {
+    case T_BINARY_PROTOCOL:
+      proto_ = boost::make_shared<TBinaryProtocolT<THeaderTransport>>(trans_);
+      break;
+
+    case T_COMPACT_PROTOCOL:
+      proto_ = boost::make_shared<TCompactProtocolT<THeaderTransport>>(trans_);
+      break;
+
+    default:
+      throw TApplicationException(TApplicationException::INVALID_PROTOCOL,
+                                  "Unknown protocol requested");
+  }
+}
+
+uint32_t THeaderProtocol::writeMessageBegin(const std::string& name,
+                                            const TMessageType messageType,
+                                            const int32_t seqId) {
+  resetProtocol(); // Reset in case we changed protocols
+  trans_->setSequenceNumber(seqId);
+  return proto_->writeMessageBegin(name, messageType, seqId);
+}
+
+uint32_t THeaderProtocol::writeMessageEnd() {
+  return proto_->writeMessageEnd();
+}
+
+uint32_t THeaderProtocol::writeStructBegin(const char* name) {
+  return proto_->writeStructBegin(name);
+}
+
+uint32_t THeaderProtocol::writeStructEnd() {
+  return proto_->writeStructEnd();
+}
+
+uint32_t THeaderProtocol::writeFieldBegin(const char* name,
+                                          const TType fieldType,
+                                          const int16_t fieldId) {
+  return proto_->writeFieldBegin(name, fieldType, fieldId);
+}
+
+uint32_t THeaderProtocol::writeFieldEnd() {
+  return proto_->writeFieldEnd();
+}
+
+uint32_t THeaderProtocol::writeFieldStop() {
+  return proto_->writeFieldStop();
+}
+
+uint32_t THeaderProtocol::writeMapBegin(const TType keyType,
+                                        const TType valType,
+                                        const uint32_t size) {
+  return proto_->writeMapBegin(keyType, valType, size);
+}
+
+uint32_t THeaderProtocol::writeMapEnd() {
+  return proto_->writeMapEnd();
+}
+
+uint32_t THeaderProtocol::writeListBegin(const TType elemType,
+                                         const uint32_t size) {
+  return proto_->writeListBegin(elemType, size);
+}
+
+uint32_t THeaderProtocol::writeListEnd() {
+  return proto_->writeListEnd();
+}
+
+uint32_t THeaderProtocol::writeSetBegin(const TType elemType,
+                                        const uint32_t size) {
+  return proto_->writeSetBegin(elemType, size);
+}
+
+uint32_t THeaderProtocol::writeSetEnd() {
+  return proto_->writeSetEnd();
+}
+
+uint32_t THeaderProtocol::writeBool(const bool value) {
+  return proto_->writeBool(value);
+}
+
+uint32_t THeaderProtocol::writeByte(const int8_t byte) {
+  return proto_->writeByte(byte);
+}
+
+uint32_t THeaderProtocol::writeI16(const int16_t i16) {
+  return proto_->writeI16(i16);
+}
+
+uint32_t THeaderProtocol::writeI32(const int32_t i32) {
+  return proto_->writeI32(i32);
+}
+
+uint32_t THeaderProtocol::writeI64(const int64_t i64) {
+  return proto_->writeI64(i64);
+}
+
+uint32_t THeaderProtocol::writeDouble(const double dub) {
+  return proto_->writeDouble(dub);
+}
+
+uint32_t THeaderProtocol::writeString(const std::string& str) {
+  return proto_->writeString(str);
+}
+
+uint32_t THeaderProtocol::writeBinary(const std::string& str) {
+  return proto_->writeBinary(str);
+}
+
+/**
+ * Reading functions
+ */
+
+uint32_t THeaderProtocol::readMessageBegin(std::string& name,
+                                           TMessageType& messageType,
+                                           int32_t& seqId) {
+  // Read the next frame, and change protocols if needed
+  try {
+    trans_->resetProtocol();
+    resetProtocol();
+  } catch (const TApplicationException& ex) {
+    writeMessageBegin("", T_EXCEPTION, 0);
+    ex.write((TProtocol*)this);
+    writeMessageEnd();
+    trans_->flush();
+    
+    // The framing is still good, but we don't know about this protocol.
+    // In the future, this could be made a client-side only error if
+    // connection pooling is used.
+    throw ex;
+  }
+  return proto_->readMessageBegin(name, messageType, seqId);
+}
+
+uint32_t THeaderProtocol::readMessageEnd() {
+  return proto_->readMessageEnd();
+}
+
+uint32_t THeaderProtocol::readStructBegin(std::string& name) {
+  return proto_->readStructBegin(name);
+}
+
+uint32_t THeaderProtocol::readStructEnd() {
+  return proto_->readStructEnd();
+}
+
+uint32_t THeaderProtocol::readFieldBegin(std::string& name,
+                                         TType& fieldType,
+                                         int16_t& fieldId) {
+  return proto_->readFieldBegin(name, fieldType, fieldId);
+}
+
+uint32_t THeaderProtocol::readFieldEnd() {
+  return proto_->readFieldEnd();
+}
+
+uint32_t THeaderProtocol::readMapBegin(TType& keyType,
+                                       TType& valType,
+                                       uint32_t& size) {
+  return proto_->readMapBegin(keyType, valType, size);
+}
+
+uint32_t THeaderProtocol::readMapEnd() {
+  return proto_->readMapEnd();
+}
+
+uint32_t THeaderProtocol::readListBegin(TType& elemType,
+                                        uint32_t& size) {
+  return proto_->readListBegin(elemType, size);
+}
+
+uint32_t THeaderProtocol::readListEnd() {
+  return proto_->readListEnd();
+}
+
+uint32_t THeaderProtocol::readSetBegin(TType& elemType,
+                                       uint32_t& size) {
+  return proto_->readSetBegin(elemType, size);
+}
+
+uint32_t THeaderProtocol::readSetEnd() {
+  return proto_->readSetEnd();
+}
+
+uint32_t THeaderProtocol::readBool(bool& value) {
+  return proto_->readBool(value);
+}
+
+uint32_t THeaderProtocol::readByte(int8_t& byte) {
+  return proto_->readByte(byte);
+}
+
+uint32_t THeaderProtocol::readI16(int16_t& i16) {
+  return proto_->readI16(i16);
+}
+
+uint32_t THeaderProtocol::readI32(int32_t& i32) {
+  return proto_->readI32(i32);
+}
+
+uint32_t THeaderProtocol::readI64(int64_t& i64) {
+  return proto_->readI64(i64);
+}
+
+uint32_t THeaderProtocol::readDouble(double& dub) {
+  return proto_->readDouble(dub);
+}
+
+uint32_t THeaderProtocol::readString(std::string& str) {
+  return proto_->readString(str);
+}
+
+uint32_t THeaderProtocol::readBinary(std::string& binary) {
+  return proto_->readBinary(binary);
+}
+
+template<typename StrType>
+uint32_t THeaderProtocol::readStringBody(StrType& str,
+                                         int32_t size) {
+  return proto_->readStringBody<StrType>(str, size);
+}
+
+}}} // apache::thrift::protocol
+
+#endif // #ifndef THRIFT_PROTOCOL_THEADERPROTOCOL_CPP_
--- a/lib/cpp/src/thrift/protocol/TProtocol.h
+++ b/lib/cpp/src/thrift/protocol/TProtocol.h
@@ -659,6 +659,11 @@ public:
   virtual ~TProtocolFactory() {}
 
   virtual boost::shared_ptr<TProtocol> getProtocol(boost::shared_ptr<TTransport> trans) = 0;
+  virtual boost::shared_ptr<TProtocol> getProtocol(boost::shared_ptr<TTransport> inTrans,
+						   boost::shared_ptr<TTransport> outTrans) {
+    (void)outTrans;
+    return getProtocol(inTrans);
+  }
 };
 
 /**
--- a/lib/cpp/src/thrift/protocol/TProtocolTypes.h
+++ b/lib/cpp/src/thrift/protocol/TProtocolTypes.h
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef THRIFT_PROTOCOL_TPROTOCOLTYPES_H_
+#define THRIFT_PROTOCOL_TPROTOCOLTYPES_H_ 1
+
+namespace apache { namespace thrift { namespace protocol {
+
+enum PROTOCOL_TYPES {
+  T_BINARY_PROTOCOL = 0,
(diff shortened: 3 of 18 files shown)


ben-craig:
I'll review this... it's going to take a couple of days though.

lib/cpp/src/thrift/protocol/THeaderProtocol.cpp
```diff
@@ -0,0 +1,260 @@
...
+void THeaderProtocol::resetProtocol() {
+  if (proto_ && protoId_ == trans_->getProtocolId()) {
+    return;
+  }
+
+  protoId_ = trans_->getProtocolId();
+
+  switch (protoId_) {
+    case T_BINARY_PROTOCOL:
+      proto_ = boost::shared_ptr<TProtocol>(
```
ben-craig: One of the many things I've been meaning to do is to make a big sweep over the cpp code and change boost::shared_ptr creations to use boost::make_shared.

If you have time, you can make that change in your new files.  If not, that is fine also.  It can wait until someone goes through and changes them across the board.

lib/cpp/src/thrift/protocol/THeaderProtocol.cpp
```diff
@@ -0,0 +1,260 @@
...
+  // Read the next frame, and change protocols if needed
+  try {
+    trans_->resetProtocol();
+    resetProtocol();
+  } catch (const TApplicationException& ex) {
+    writeMessageBegin("", T_EXCEPTION, 0);
+    ex.write((TProtocol*)this);
+    writeMessageEnd();
+    trans_->flush();
+  }
```
ben-craig: The above try / catch block effectively swallows the exception after writing out a response.  I would expect it to re-throw the exception, or at least to do something like return 0.  That way calling code can know that something is wrong, and that a new message shouldn't be read.
Alternatively, if there is a strong reason to not propagate that information, a supporting comment would do.

lib/cpp/src/thrift/protocol/TProtocolTypes.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef THRIFT_PROTOCOL_TPROTOCOLTYPES_H_
+#define THRIFT_PROTOCOL_TPROTOCOLTYPES_H_ 1
+
+namespace apache { namespace thrift { namespace protocol {
+
+enum PROTOCOL_TYPES {
+  T_BINARY_PROTOCOL = 0,
+  T_JSON_PROTOCOL = 1,
+  T_COMPACT_PROTOCOL = 2,
+};
```
ben-craig: One day (probably not today), we should figure out how to share constants like these across all the languages.

lib/cpp/src/thrift/protocol/TProtocol.h
```diff
@@ -659,6 +659,11 @@ class TProtocolFactory {
   virtual ~TProtocolFactory() {}
 
   virtual boost::shared_ptr<TProtocol> getProtocol(boost::shared_ptr<TTransport> trans) = 0;
+  virtual boost::shared_ptr<TProtocol> getProtocol(boost::shared_ptr<TTransport> inTrans,
```
ben-craig: I'm not sure why this getProtocol overload is here.  All the servers have an inputProtocolFactory and outputProtocolFactory.  Maybe this is a little bit more usable... but it adds unusual noise to the review.  I would like to have some justification for this overload in regards to THeader... or I would like to see this pulled into its own issue.

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -0,0 +1,308 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef THRIFT_TRANSPORT_THEADERTRANSPORT_H_
+#define THRIFT_TRANSPORT_THEADERTRANSPORT_H_ 1
+
+#include <tr1/functional>
```
ben-craig: This needs to be thrift/cxxfunctional.h, and usages of ::std::tr1::function need to be replaced with tcxx::function.  Clang doesn't have the tr1 headers... but they do have the C++11 headers, and thrift/cxxfunctional.h deals with that.

Or maybe it just needs to be removed?  I don't see a tr1::function lying around anywhere.

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -0,0 +1,308 @@
...
+
+#include <thrift/protocol/TBinaryProtocol.h>
+#include <thrift/protocol/TProtocolTypes.h>
+#include <thrift/transport/TBufferTransports.h>
+#include <thrift/transport/TTransport.h>
+#include <thrift/transport/TVirtualTransport.h>
+
+#include <bitset>
+#include "boost/scoped_array.hpp"
```
ben-craig: very minor... this should use < > instead of " ".

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -0,0 +1,308 @@
...
+ * and need similar buffers.  Major changes are readFrame & flush.
+ */
+class THeaderTransport
+    : public TVirtualTransport<THeaderTransport, TFramedTransport> {
+ public:
+
+  static const int DEFAULT_BUFFER_SIZE = 512u;
+
+  /// Use default buffer sizes.
+  explicit THeaderTransport(const boost::shared_ptr<TTransport> transport)
```
ben-craig: pass by const &.  Your other ctor also needs some &s sprinkled on its args.

If we were using C++11 everywhere, then _maybe_ we would pass by value and then std::move into the member variable... but we aren't using C++11 everywhere.

lib/cpp/src/thrift/protocol/TProtocol.h
djwatson: THeader detects the client transport type - unframed, framed, header - and responds in kind.  To make this work, the input/output transport need to be the *same* THeader instance.  

The transports don't necessarily need to be the same though (and in the case of TNonblocking, need to be different due to the way the MemoryBuffers work).  So THeader here takes in two transports, but uses a single protocol.

Actually in the fbthrift code this wasn't added to the interface, but just special cased in TNonblockingServer with a dynamic_cast to TheaderProtocol, which we could do instead

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -0,0 +1,308 @@
...
+   * Read an i32 from the wire as a varint. The MSB of each byte is set
+   * if there is another byte to follow. This can read up to 5 bytes.
+   */
+  uint32_t readVarint32(uint8_t const* ptr, int32_t* i32,
+			uint8_t const* boundary);
+
+  /**
+   * Write an i32 as a varint. Results in 1-5 bytes on the wire.
+   */
+  uint32_t writeVarint32(uint32_t n, uint8_t* pkt);
```
ben-craig: If this is writing an i32... 'n' should be an int32_t instead of a uint32_t.

Similar commends on writeVarint16.

lib/cpp/src/thrift/protocol/TProtocol.h
ben-craig: This makes sense.  I would like that information captured in a comment somewhere.  Maybe in THeaderTransport.h?

lib/cpp/src/thrift/protocol/TProtocolTypes.h
Jens-G: True words. There are a bunch of candidates.

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -0,0 +1,308 @@
...
+#include <thrift/transport/TTransport.h>
+#include <thrift/transport/TVirtualTransport.h>
+
+#include <bitset>
+#include "boost/scoped_array.hpp"
+#include <pwd.h>
+#include <unistd.h>
+
+// Don't include the unknown client.
+#define CLIENT_TYPES_LEN 4
```
ben-craig: This seems wrong.  The preceding comment says "Don't include the unknown client", which is fine, but the last known client type has a value of '2'.  That suggests that this define should be for '3'.  I suspect that there used to be a '3' in some super-secret facebook stuff :)

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -0,0 +1,308 @@
...
+
+using apache::thrift::protocol::T_COMPACT_PROTOCOL;
+
+/**
+ * Header transport. All writes go into an in-memory buffer until flush is
+ * called, at which point the transport writes the length of the entire
+ * binary chunk followed by the data payload. This allows the receiver on the
+ * other end to always do fixed-length reads.
+ *
+ * Subclass TFramedTransport because most of the read/write methods are similar
```
ben-craig: I don't think I like subclassing TFramedTransport.  Some of your members call TBufferBase methods, while others call TFramedTransport.  
In addition, by subclassing TFramedTransport, you allow THeaderTransports to be passed to functions expecting a TFramedTransport.  That could cause some hard to diagnose bugs, as TFramedTransport's wire layout is different from THeaderTransport's.

If you want to continue to get code reuse from TFramedTransport, then that's great, and you can do so by using containment instead of inheritance.  Make a member TFramedTransport, and forward calls to the member.  In the case where you want to call into TBufferBase versions of methods, you can do so like this:
framedTransport.TBufferBase::readAll(buf, len);

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+
+uint16_t THeaderTransport::getProtocolId() const {
+  if (clientType == THRIFT_HEADER_CLIENT_TYPE) {
+    return protoId;
+  } else {
+    return T_BINARY_PROTOCOL; // Assume other transports use TBinary
+  }
+}
+
+void THeaderTransport::allocateReadBuffer(uint32_t sz) {
```
ben-craig: not super-thrilled with this name, as sometimes it allocates, and sometimes it doesn't.  Maybe 'ensureReadBufferAllocated'?  'reserveReadBuffer'?

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+  ptr += strLen;
+}
+
+void THeaderTransport::readHeaderFormat(uint16_t headerSize, uint32_t sz) {
+  readTrans_.clear(); // Clear out any previous transforms.
+  readHeaders_.clear(); // Clear out any previous headers.
+
+  // skip over already processed magic(4), seqId(4), headerSize(2)
+  uint8_t* ptr = reinterpret_cast<uint8_t*>(rBuf_.get() + 10);
+  headerSize *= 4;
```
ben-craig: potential integer overflow here

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+  const uint8_t* const headerBoundary = ptr + headerSize;
+  if (headerSize > sz) {
+    throw TTransportException(TTransportException::CORRUPTED_DATA,
+                              "Header size is larger than frame");
+  }
+  uint8_t* data = ptr + headerSize;
+  ptr += readVarint16(ptr, &protoId, headerBoundary);
```
ben-craig: I'm not thrilled with VarInts here.  I suspect that for the Facebook use case, you are stuck with them, because you are probably already running lots of services with VarInts in the headers.

Why am I not fond of VarInts?  They save space on the wire, so that must be good, right?  Well, for most use cases, the slow part of networking is because of latency, not throughput.  Stated another way, round trips are expensive, but extra bytes in the same round trip tend not to be.  On a gigabit network, a byte costs roughly 8 nanoseconds, and VarInt16 will only save one byte in its best case.  So let's translate this to processor time now.  Assume a 3 GHz processor.  3 cycles fit into one nanosecond.  Does the entirety of the code required to deal with VarInts in the good case fit in 24 cycles?  If it takes more cycles, then you have slowed things down.  Now, this is a gross simplification in many ways, as it doesn't account for caches, or the work that the network stacks have to perform to send one more byte... but it does mean that VarInts aren't a clear win.

Now maybe this doesn't apply to Facebook.  I don't know whether they are more bandwidth bound or CPU bound, and I don't know whether CPU time or network time consumes more power.  Either way though, I am skeptical of this design choice.  Maybe we can get the best for Facebook and the best for Apache by using a different version / magic number to distinguish the cases.  Or maybe we call this TCompactHeader and something else TBinaryHeader.

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+      }
+    }
+    sizeBytesRead += bytesRead;
+  }
+
+  sz = ntohl(szN);
+
+  allocateReadBuffer(minFrameSize + 4);
+
+  if ((sz & TBinaryProtocol::VERSION_MASK) == (uint32_t)TBinaryProtocol::VERSION_1) {
```
ben-craig: I am having a hard time figuring out which bytes on the wire mean what in the context of THeaderTransport.  I'll describe what I know of TBinaryProtocol and TFramedTransport, and I'm hoping that you can fill in the blanks for THeaderTransport.

For ancient, unframed binary protocol, the wire layout was something like the following:
int32_t sizeOfStringBody, char stringBody[], int8_t messageType, int32_t seqID

For the modern, unframed binary protocol, the wire layout takes advantage of the int-ness of the old style, and does something like this:
int16_t version = 0x80010000, int8_t pad, int8_t messageType, int32_t sizeOfStringBody, char stringBody[], int32_t seqID

Modern framed looks like this:
int32_t sizeOfFrame, int16_t version = 0x80010000, int8_t pad, int8_t messageType, int32_t sizeOfStringBody, char stringBody[], int32_t seqID

Note that modern framed and ancient unframed are difficult to distinguish from just looking at the size.  Also note that at least one other version number has been taken.  See TBinaryProtocol.h
// VERSION_2 (0x80020000)  is taken by TDenseProtocol.

I think what you are doing is the following... but I'm not sure:
int32_t sizeOfFrame, int16_t newVersion = 0x0FFF0000, int16_t flags, int32_t seqID, int16_t sizeOfHeader, vint16_t protoID, vint16_t numTransforms, and then more header stuff.

Couldn't we just use 0x80030000 instead of 0x0FFF0000?  That seems like it fits in with the current architecture better, and we won't have 'magic' numbers running around as much.  Alternatively, since this is a 'big' change... maybe go with a version 0xC0000000.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -547,12 +553,18 @@ void TNonblockingServer::TConnection::transition() {
   case APP_READ_REQUEST:
     // We are done reading the request, package the read buffer into transport
     // and get back some data from the dispatch function
-    inputTransport_->resetBuffer(readBuffer_, readBufferPos_);
-    outputTransport_->resetBuffer();
-    // Prepend four bytes of blank space to the buffer so we can
-    // write the frame size there later.
-    outputTransport_->getWritePtr(4);
-    outputTransport_->wroteBytes(4);
+    if (!server_->getOutputProtocolFactory()) {
```
ben-craig: If I understand this correctly, then this is basically asking "are we in the THeader case".  I think that this information should be stored in a well named member enum, and then we can query that.  The current query doesn't really express intent to me.
Alternatively, comment the reason for the difference.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -547,12 +553,18 @@ void TNonblockingServer::TConnection::transition() {
...
+      inputTransport_->resetBuffer(readBuffer_, readBufferPos_);
+      outputTransport_->resetBuffer();
+    } else {
+      inputTransport_->resetBuffer(readBuffer_ + 4, readBufferPos_ - 4);
```
ben-craig: I believe this is the 'legacy' branch.  Is there a compelling reason to change this code from how it was?  I'm not sure why the +/- 4 is necessary now, but for some reason wasn't necessary before.  Maybe this isn't a legacy branch and I'm just misunderstanding things?

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+ *
+ * @param   str                  output string
+ * @throws  CORRUPTED_DATA  if size of string exceeds boundary
+ */
+void THeaderTransport::readString(uint8_t* &ptr, /* out */ string &str,
+                uint8_t const* headerBoundary) {
+  int32_t strLen;
+
+  uint32_t bytes = readVarint32(ptr, &strLen, headerBoundary);
+  if (ptr + strLen > headerBoundary) {
```
ben-craig: pointer overflow risk.  If strLen is huge on a 32-bit system, then ptr will wrap around and skip this check.
This construct will avoid any overflow possibilities, because you know that headerBoundary and ptr start out relatively close to one another:
if(strLen > headerBoundary - ptr) {throw ...}

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+      break;
+    }
+    if (infoId >= infoIdType::END) {
+      // cannot handle infoId
+      break;
+    }
+    switch (infoId) {
+      case infoIdType::KEYVALUE:
+        // Process key-value headers
+        uint32_t numKVHeaders;
```
ben-craig: The type mismatch with the following readVarInt32 is likely to make Java unhappy.  Remember that Java doesn't have unsigned types.  Data on the wire should generally be treated as signed in order to appease that language.

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+  if (headerSize > sz) {
+    throw TTransportException(TTransportException::CORRUPTED_DATA,
+                              "Header size is larger than frame");
+  }
+  uint8_t* data = ptr + headerSize;
+  ptr += readVarint16(ptr, &protoId, headerBoundary);
+  int16_t numTransforms;
+  ptr += readVarint16(ptr, &numTransforms, headerBoundary);
+
+  uint16_t macSz = 0;
```
ben-craig: This appears unused... except as a more convoluted no-op with 'sz -= macSz;'.  I'm guessing that the 'mac' (message authentication code?) feature was dropped in this push.

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+      stream.opaque = (voidpf)0;
+      err = deflateInit(&stream, Z_DEFAULT_COMPRESSION);
+      if (err != Z_OK) {
+        throw TTransportException(TTransportException::CORRUPTED_DATA,
+                                  "Error while zlib deflateInit");
+      }
+      stream.next_out = tBuf_.get();
+      stream.avail_out = tBufSize_;
+      err = deflate(&stream, Z_FINISH);
+      if (err != Z_STREAM_END || stream.avail_out == 0) {
```
ben-craig: This implementation seems to assume that deflating something will add 512 bytes at most to the message size.  I think a loop and some reallocation of tBuf_ would be appropriate here.

lib/cpp/src/thrift/transport/THeaderTransport.cpp
```diff
@@ -0,0 +1,606 @@
...
+
+  if (haveBytes > MAX_FRAME_SIZE) {
+    throw TTransportException(TTransportException::CORRUPTED_DATA,
+                              "Attempting to send frame that is too large");
+  }
+
+  if (clientType == THRIFT_HEADER_CLIENT_TYPE) {
+    // header size will need to be updated at the end because of varints.
+    // Make it big enough here for max varint size, plus 4 for padding.
+    int headerSize = (2 + getNumTransforms()) * 5 + 4;
```
ben-craig: a lot of magic numbers sprinkled about...
At least '5' should be easy enough to define as THRIFT_MAX_VARINT32_BYTES (or something similar)

lib/cpp/src/thrift/transport/THeaderTransport.cpp
djwatson: In Facebook's testing, varints in CompactProtocol were always faster (by at least 20%) than BinaryProtocol (we didn't test theader directly).  It did involve some gross unrolling of loops though:

https://github.com/facebook/fbthrift/blob/master/thrift/lib/cpp/util/VarintUtils.tcc

Protocolbufs does similar unrolling.

lib/cpp/src/thrift/transport/THeaderTransport.cpp
djwatson: I'll add this doc to the commit: https://github.com/facebook/fbthrift/blob/master/thrift/doc/HeaderFormat.md

Yea I can't think of any reason the magic couldn't be 8003 vs 0FFF

lib/cpp/src/thrift/server/TNonblockingServer.cpp
djwatson: THeader expects the first four bytes of framing to still be in the message.  Nonblocking server previously always stripped it off

lib/cpp/src/thrift/transport/THeaderTransport.cpp
ben-craig: I could see arrays of varints being faster, since you would reduce data cache pressure.  I don't have specific numbers though, and numbers rightfully beat out thought-experiments.

Protocolbufs always does the varint thing.  One of the main designers of protocolbufs went on to make this: https://capnproto.org/
So he's going with a mem-mappable layout.  That would be awesome, but would require some pretty significant rearchitecture.

Regardless, I won't hold up this commit over varints.  I'm still skeptical, but not skeptical enough to block progress on something that adds as much value as THeader.

lib/cpp/src/thrift/transport/THeaderTransport.h
djwatson: Having a TFramedTransport as a member is harder than it seems - THeaderTransport should probably inherit from TBufferedTransport, but then we'd have two TBufferedTransports, one for header and one for framed, to make work together somehow.  I could just copy in TFramedTransport::readSlow I guess

ben-craig: Within the Thrift code base, nothing passes around pointers to TBufferedTransport objects.  Some of the async classes pass around pointers to TBufferBase objects though.  If you are trying to make THeaderTransport fit in the async classes, then inheriting from TBufferBase should be enough.

If you don't want / don't care about the async classes, then I don't see the benefit in inheriting from the TBuffer classes.

djwatson:
Just FYI additional commit added to PR covering review comments

ben-craig:
Looks pretty good.  Just the one last zlib comment and I think everything will be good to go.

Thanks for putting up with my nitpicky comments :)

bufferoverflow:
could you rebase to master and I commit

djwatson:
fixed zlib check to be  == Z_OK.  Rebased to apache/master

nsuke:
What happened to this ?
Although it does not cleanly apply at all anymore, with "git merge" it is still straightforward to merge.
If there isn't any concern, I'll merge this.

bufferoverflow:
@nsuke Yes, please do so.

nsuke:
There were several build issues.
I'll commit the fixed version shortly after #677.

PR #363, opened 2015-01-29
Fix namespace regression for Ruby

bracki:
Fixes a bug introduced by #140.


jfarrell:
THRIFT-2960

jfarrell:
committed with modification to make test cases pass, thanks for the patch

PR #364, opened 2015-01-29
Service inheritance does not work with namespaced Ruby code.

bracki:
This PR fixes `thrift --gen rb:namespaced` to work with inheritance.

See also #140 and #363.


aughr:
@bracki Thanks for cleaning up after my mistakes!

PR #365, opened 2015-01-30
cmake // build // fixed java library build enabler

snikulov:
Hello, 

Here the fix for case when jdk is installed, but no Ant tool present.
Since Ant is required to build java library, I move Ant search to  cmake/DefineOptions.cmake and add dependency for WITH_JAVA option.

Without those changes we'll pass JAVA_FOUND test and will fail later in find_package(Ant REQUIRED)


snikulov:
Related to https://issues.apache.org/jira/browse/THRIFT-2850

snikulov:
Merged to master https://github.com/apache/thrift/commit/9d8c1bf6f62c22c13c5b6b4da51b1de1a9d27339

PR #368, opened 2015-01-31
THRIFT-2835 Add possibility to distribute generators separately from thr...

nsuke:
...ift core, and load them dynamically


jeking3:
Lots of work in here!  You might want to close and re-open this pull request to kick a new build.  It needs to pass CI before anyone can merge it.

nsuke:
Hi, you're so diligent :)
It's quite old so I'll rather rebase it.

jeking3:
The pull request list would be a lot easier to manage if it was always clean/empty, so I want to work towards that.

nsuke:
I rebased it to the current master.
The original first commit is extracted as THRIFT-3138 (ef0a8fa62a8a86b23999f9a9ad31478852474854) because it was an unrelated issue.
The other commit remains strictly unchanged except that I had to resolve several trivial conflicts in build scripts.

nsuke:
TODO:
* Remove `sleep 0.2` in cpp_plugin_test.sh without breaking it
* Build with static boost unittest (need to do the same as some tests in lib/cpp/test)

PR #369, opened 2015-02-02
THRIFT-2189 add union support for go generator

antonlindstrom:
This makes it possible to check if a field  is set and skips adding new ones.
Fields in unions are pointers by setting them as `t_field::T_OPTIONAL`.

To be sure that exactly one and only one field is set in a union, we count the
number of fields set and return an error if not exactly one field is set.

Thanks!

### DIFF
--- a/compiler/cpp/src/generate/t_go_generator.cc
+++ b/compiler/cpp/src/generate/t_go_generator.cc
@@ -132,6 +132,10 @@ public:
                               t_struct* tstruct,
                               const string& tstruct_name,
                               bool is_result = false);
+  void generate_countsetfields_helper(std::ofstream& out,
+                              t_struct* tstruct,
+                              const string& tstruct_name,
+                              bool is_result = false);
   void generate_go_struct_reader(std::ofstream& out,
                                  t_struct* tstruct,
                                  const string& tstruct_name,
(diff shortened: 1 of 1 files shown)


compiler/cpp/src/generate/t_go_generator.cc
```diff
@@ -1233,6 +1249,46 @@ void t_go_generator::generate_isset_helpers(ofstream& out,
...
+void t_go_generator::generate_countsetfields_helper(ofstream& out,
+                                            t_struct* tstruct,
+                                            const string& tstruct_name,
+                                            bool is_result) {
+  (void)is_result;
+  const vector<t_field*>& fields = tstruct->get_members();
+  vector<t_field*>::const_iterator f_iter;
+  const string escaped_tstruct_name(escape_string(tstruct->get_name()));
+
+  out << indent() << "func (p *" << tstruct_name << ") CountSetFields" << tstruct_name
```
lsjbd, line 1264: What is the purpose to use the struct name as a suffix for the method? To avoid name confliction?

antonlindstrom: Hmm, this was quite some time ago. I haven't been using thrift for a while so I can't say for sure.

lsjbd: OK

PR #370, opened 2015-02-04
THRIFT-2975 Graceful handling of server-side errors in Python

abhinav:
Change `process_*` functions to handle unexpected exceptions from
user-supplied handlers by throwing a `TApplicationException(INTERNAL_ERROR)`
to the client.


abhinav:
This needs to log the actual exceptions before throwing TApplicationException to the clients and the clients need control of where the exceptions are logged. The current design doesn't cover that. Abandoning this diff to evaluate whether that's possible.

PR #372, opened 2015-02-06
THRIFT-2180: Cocoa-Library: addresses integer type issues on 64-bit platforms

jeremy-w:
No easy way to test, since there's no shared Xcode project to generate warnings. For reference, I was building for iOS and for Mac with flags:

    OTHER_CFLAGS = -Wall -Wextra -Weverything -Wno-objc-missing-property-synthesis -Wno-unused-parameter -Wno-auto-import -Wno-documentation-unknown-command -Wno-direct-ivar-access -Wno-objc-interface-ivars

The remaining warnings should be addressed by issues THRIFT-2982 through THRIFT-2987.


randomstep:
I'd also like to see these changes merged in. In the meantime we'll need to maintain them downstream.

PR #377, opened 2015-02-13
THRIFT-3000: Issue with IPv6 an IPv4 on windows machines

sharpdevel:
When using TcpListener and TcpClient it depends on the network configuration if IPv4 or IPv6 is used. By upgrading the framework to .NET 4.5 the DualMode can be set on the sockets of the listener and the client. The sockets then try to estabild IPv6 sockets before they fallback to IPv4. For example IPv4 is preferred on Windows 7 machines while on Windows 8 and 8.1 machines IPv6 is preferred. This patch works on Window 7, 8 and 8.1.

Additionally the AssemblyFileVersion must not contain an asterix. By removing the AssemblyFileVersion it alsways corresponds the the generated value of AsemblyVersion

In the PreBuild command of the ThriftTest project I added "" around the MSBuildToolPath to make the script work on english windows machine wheren the folder equals to sth like "C:\program files\....."

### DIFF
--- a/lib/csharp/ThriftMSBuildTask/Properties/AssemblyInfo.cs
+++ b/lib/csharp/ThriftMSBuildTask/Properties/AssemblyInfo.cs
@@ -56,5 +56,4 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("0.9.0.*")]
-[assembly: AssemblyFileVersion("0.9.0.*")]
+[assembly: AssemblyVersion("0.9.0.*")]//Generates AssemblyFileVersion accordingly.
--- a/lib/csharp/ThriftMSBuildTask/ThriftMSBuildTask.csproj
+++ b/lib/csharp/ThriftMSBuildTask/ThriftMSBuildTask.csproj
@@ -28,7 +28,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>ThriftMSBuildTask</RootNamespace>
     <AssemblyName>ThriftMSBuildTask</AssemblyName>
-    <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <FileUpgradeFlags>
     </FileUpgradeFlags>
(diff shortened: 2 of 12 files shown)


Jens-G:
May I ask you to [create a JIRA ticket for that](thrift.apache.org/docs/HowToContribute)? 

Thanks,
JensG

sharpdevel:
You a right: Mono 3.12.0 has no support for the Socket.DualMode property which I used in my patch. Sorry for not double checking that. As requested I opened a jira issue: https://issues.apache.org/jira/browse/THRIFT-3000

Thanks
sharpdevel

Jens-G:
Can we do something with #ifdefs to protect against when it is not available?

sharpdevel:
I could imagine something like this:


            if (typeof(TcpClient).GetProperty("DualMode") == null)
            {
                client = new TcpClient(AddressFamily.InterNetwork);
            }
            else
            {
                client = new TcpClient(AddressFamily.InterNetworkV6);
                client.Client.DualMode = true;
            }?

but I must admit that I'm not that fit in Mono and cornercasing it with ifdefs


Von: Jens Geyer <notifications@github.com>
Gesendet: Montag, 16. Februar 2015 22:13
An: apache/thrift
Cc: Tobias Mundt
Betreff: Re: [thrift] Issue with IPv6 an IPv4 on windows machines (#377)


Can we do something with #ifdefs to protect against when it is not available?

-

Jens-G:
I just gave it a shot and it doesn't compile. That's more or less what I expected, because the compiler still wants to resolve the symbol:

      src/Transport/TSocket.cs(64,18): error CS1061: Type `System.Net.Sockets.Socket' does
      not contain a definition for `DualMode' and no extension method `DualMode' of type 
      `System.Net.Sockets.Socket' could be found. Are you missing an assembly reference?

sharpdevel:
You are right. Die DualMode also has to be set using reflection. If you agree with me to head in this direction I will fix it and make another pull request - d'accord?

Jens-G:
I don't have a better idea, so yes.

jfarrell:
Thanks for your patch @sharpdevel, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

Jens-G:
@sharpdevel: Didn't recognize that PR, it somehow got lost in the mails. Could you file a JIRA ticket and reopen the PR mentioning the ticket number? I'll take a look at the stuff then. Thanks!

sharpdevel:
I'm afraid I don't know how to reopen this pull request. Do I have to create a new one? I already create a Jira-Ticket https://issues.apache.org/jira/browse/THRIFT-3000. Do I have to recreate that too?

PR #378, opened 2015-02-13
THRIFT-2998: Set Content-Type for HTTP request in Node.js wrapper. Done similar to th...

tstanev:
...e Python wrapper here: 
https://github.com/apache/thrift/blob/master/lib/py/src/transport/THttpClient.py#L127
Without this header, connecting to some servers results in error response. (THRIFT-2998)

### DIFF
--- a/lib/nodejs/lib/thrift/http_connection.js
+++ b/lib/nodejs/lib/thrift/http_connection.js
@@ -214,10 +214,13 @@ util.inherits(HttpConnection, EventEmitter);
  */
 HttpConnection.prototype.write = function(data) {
   var self = this;
-  self.nodeOptions.headers["Content-length"] = data.length;
+  var opts = self.nodeOptions;
+  opts.headers["Content-length"] = data.length;
+  if (!opts.headers["Content-Type"])
+    opts.headers["Content-Type"] = "application/x-thrift";  
   var req = (self.https) ?
-      https.request(self.nodeOptions, self.responseCallback) :
-      http.request(self.nodeOptions, self.responseCallback);
+      https.request(opts, self.responseCallback) :
+      http.request(opts, self.responseCallback);
   req.on('error', function(err) {
     self.emit("error", err);
   });


larsxschneider:
@RandyAbernethy Looks like a good workaround for now. Any reason not to merge? :smile:

RandyAbernethy:
No worse than no header I suppose, however Thrift media types are now per protocol and registered with IANA:
http://www.iana.org/assignments/media-types/media-types.xhtml#application
 - vnd.apache.thrift.binary
 - vnd.apache.thrift.compact
 - vnd.apache.thrift.json	

Would be nice if we could mount a Thrift wide effort to get these into Thrift v1.0. If you need application/x-thrift in the mean time I suppose we could commit it with a TODO.

Thoughts?

stigsb:
IMHO client code should always send the IANA-registered types, while servers should accept `application/x-thrift` as an alias for `application/vnd.apache.thrift.binary` (probably for a long time).

tstanev:
I am only porting the fix from the python wrapper in order to make the node wrapper actually work. I don't know enough about Thrift to have a thought about what it should or should not do instead, but I am certain that with the server I attempted to use this with, only the application/x-thrift header worked. The IANA header did not work, so using the IANA header will not work in my case.

jfarrell:
@RandyAbernethy agree with your suggested approach of committing this and creating a backlog story to tackle the larger IANA registered media type as a follow up

tstanev:
It's rebased now.

tstanev:
By the way -- it's been nearly three years and I have since switched to another way of doing things. You can close the PR if nobody else needs it.

jeking3:
@tstanev Thank you for putting in the time.  The PR backlog on the project is getting better - there were over 100 PRs at the beginning of the year and we're down to 50, so every little bit helps.  Of course today of all days Travis CI is having some pretty nasty issues with Docker Hub. :)

jeking3:
To properly test this fix I need to add http transport support to the nodejs cross programs, so I'll do that as part of the effort.

PR #383, opened 2015-02-24
THRIFT-2885: Add a 'fullcamel' option for obj-c

toddgardner:
This allows you to write your idl files in snake_case and have your objective c code be in the (stylistic) camelCase. It makes the java option of the same name.

It also fixes a minor warning about isEqual methods with no members (which caused an unused variable "other").

Tested by reading and compiling generated code for tutorial; didn't see any tests for compiler outputs.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,14 @@
+machine:
+  java:
+    version: oraclejdk8
+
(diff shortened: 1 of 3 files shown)


jeking3:

PR #395, opened 2015-03-06
THRIFT-3060: NodeJS Client - Fix connection retry logic

cjthompson:
**Fix**: Retry variables were not initialized until after the connection succeeded. This fix properly calculates the retry timer even when the initial connection cannot be made.

**Fix**: The `offline_queue` array never gets cleared.  On reconnect, if requests ended up in the offline_queue, they would be sent again on each reconnect. This fix flushes the queue after the requests have been written to the connection.

**Change**: both connect handlers use the same code, so I pulled out the function and pass it into both the `connect` and `secureConnect` event listeners.

**Change**: Add an on "end" listener for cases where the connection may be closed by the server and left in a half-open state.

### DIFF
--- a/lib/nodejs/lib/thrift/connection.js
+++ b/lib/nodejs/lib/thrift/connection.js
@@ -32,6 +32,20 @@ var binary = require('./binary');
 
 var Connection = exports.Connection = function(stream, options) {
   var self = this;
+  var connectHandler = function () {
+    self.connected = true;
+
+    this.setTimeout(self.options.timeout || 0);
+    this.setNoDelay();
+    this.frameLeft = 0;
+    this.framePos = 0;
+    this.frame = null;
+    self.initialize_retry_vars();
+
+    self.flushQueue();
+    self.emit("connect");
+  };
+
   EventEmitter.call(this);
 
   this.seqId2Service = {};
@@ -41,6 +55,7 @@ var Connection = exports.Connection = function(stream, options) {
   this.protocol = this.options.protocol || TBinaryProtocol;
   this.offline_queue = [];
   this.connected = false;
+  this.initialize_retry_vars();
 
   this._debug = this.options.debug || false;
   if (this.options.max_attempts &&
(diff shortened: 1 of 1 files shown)


jfarrell:
hey @cjthompson, thanks for the patch. mind creating a ticket in our jira for this so it can be tracked and reviewed. Details are available at https://thrift.apache.org/docs/HowToContribute

cjthompson:
I submitted a patch to issue THRIFT-3060.

cjthompson:
Is there any reason to not include this patch?

nsuke:
Hi @cjthompson, thanks for the reminder.

SSL does not seem to work with this change:
https://travis-ci.org/apache/thrift/jobs/56435212#L7581
https://travis-ci.org/apache/thrift/jobs/56435202#L5448
Could you look into this  ?

lib/nodejs/lib/thrift/connection.js
```diff
@@ -209,10 +205,9 @@ Connection.prototype.connection_gone = function () {
   this.connected = false;
   this.ready = false;
 
+  this.retry_delay = Math.floor(this.retry_delay * this.retry_backoff);
```
nsuke: Is there good reason to change the initial timeout value and also `retry_backoff` 1.7 -> 1.5 ?

lib/nodejs/lib/thrift/connection.js
```diff
@@ -182,10 +170,18 @@ Connection.prototype.initialize_retry_vars = function () {
   this.retry_timer = null;
   this.retry_totaltime = 0;
   this.retry_delay = 150;
-  this.retry_backoff = 1.7;
+  this.retry_backoff = 1.5;
   this.attempts = 0;
 };
 
+Connection.prototype.flushQueue = function () {
+  var current_queue = this.offline_queue.slice();
```
nsuke: I don't think we need a copy here since we're throwing away the old one right away.

nsuke:
I don't understand well the original motivation for clearing the queue.

It seems to me, if the user calls some RPC methods it should be either done or error.
They should not be silently ignored on random automatic reconnects.
If the disconnect is because of erroneous RPC input, the caller is responsible for properly re-initializing client.

Am I missing something ?

jeking3:

PR #396, opened 2015-03-11
Update thrift.js

forrestc:
Error.message and stack are non-enumerable properties, which cannot be inherited by util.inherits. It is annoying that an Exception object doesn't contain a message, so it is a critical bug I think.

### DIFF
--- a/lib/nodejs/lib/thrift/thrift.js
+++ b/lib/nodejs/lib/thrift/thrift.js
@@ -138,9 +138,13 @@ var TProtocolExceptionType = exports.TProtocolExceptionType = {
 
 
 var TProtocolException = exports.TProtocolException = function(type, message) {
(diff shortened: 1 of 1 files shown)


RandyAbernethy:
Hello Forrest,

The Thrift project accepts pull requests but they have to be attached to a Jira Issue. If you create a Jira Issue for this here:
https://issues.apache.org/jira/browse/THRIFT/?selectedTab=com.atlassian.jira.jira-projects-plugin:issues-panel

and attach the patch someone will review it.

Best,
Randy

jfarrell:
Thanks for your patch @forrestc, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #398, opened 2015-03-15
bower.json: fix "main" path

uvwxy:
The path "lib/js/thrift.js" is broken. "lib/js/src/thrift.js" seems to be valid.

With this fix I can include the thrift.js lib via wiredep again.


uvwxy:
https://issues.apache.org/jira/browse/THRIFT-3040

PR #402, opened 2015-03-20
0.9.2

StephaneP:
There was a typo in protocol.js, it was put as Thift.TProtocolExceptionType when it should be Thrift.


jfarrell:
This has been taken care of in trunk, no references to 'Thift' can be found

PR #403, opened 2015-03-20
[THRIFT-3048] - Make TCompactProtocol always return an object for i64

Willyham:
This changes the TCompactProtocol to always return an Int64 object in the same way that TBinary and TJSON do.

This is definitely a breaking change and will likely cause application problems for many people using TCompact and i64 types.


Willyham:
Same test cases from the issue run against this code:
```
compact small: 123, object
binary small: 456, object
compact: 4294967297, object
binary: 4294967297, object
compact large: Infinity, object
binary large: Infinity, object
```

Willyham:
Hey Randy,

Thank you. I was having trouble building the entire test suite on my distro. I should note that the NodeJS specific tests which run with `npm test` all passed after I patched them, so there must be some missing coverage in those. Thanks for finishing this up.

PR #404, opened 2015-03-21
Fix issues when using map or set types in structs in python

lucasmarshall:


jfarrell:
Thanks for your patch @lucasmarshall, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #406, opened 2015-03-22
Multiplexer in Ruby

akelmanson:


Jens-G:
Tracked in THRIFT-1125

PR #407, opened 2015-03-27
Bool transport issue and more transport

aceLren:
Bool wasn't working correctly so I updated that.  I also added transport for Warp WS and Post requests to make it easier to build servers.


jfarrell:
Thanks for your patch @aceLren, in order for us to accept a pull request we need to have a jira ticket associated with it. Please see our contributing guide at http://thrift.apache.org/docs/HowToContribute

PR #412, opened 2015-03-30
Add compiler/cpp/lex.yythriftl.cc to gitignore.

huerlisi:
I've accidentially checked that file in in another PR. So this commit
adds that file to the .gitignore.

Might be that this is only a temporary file, as I got some compiling issues.
Should be ignored anyway.


Jens-G:
Tracked in THRIFT-3074
Please have a look at our [contribution guide](http://thrift.apache.org/docs/HowToContribute)

PR #413, opened 2015-03-31
[THRIFT-2820] Implement an alternative Transport for JS using the AngularJS $http-service

uvwxy:
I added the target --gen js:angular to generate angularjs targeted thrift service clients. The thrift.js has been modified simliar to the --gen js:jquery target.


uvwxy:
See https://issues.apache.org/jira/browse/THRIFT-2820?focusedCommentId=14389241&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14389241 for explanation

henrique:
Hi Paul,
Thanks for the PR. It looks pretty good, but have tried to remove the jQuery dependency?
As you mentioned on your comments, I also think it make more sense if we didn't need it but I am not completely sure if you can run $http only with jqLite...

uvwxy:
Hi Henrique,

the dependecy is only needed when $http is not initialised correctly. In this case I fall back to the jqRequest mechanism to complete the tests.

lib/js/test/test-angular.html
```diff
@@ -0,0 +1,58 @@
...
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+  <title>Thrift Javascript Bindings: Unit Test</title>
+
+  <script src="build/js/thrift.js"         type="text/javascript" charset="utf-8"></script>
+  <script src="gen-js/ThriftTest_types.js" type="text/javascript" charset="utf-8"></script>
+  <script src="gen-js/ThriftTest.js"       type="text/javascript" charset="utf-8"></script>
+
+  <!-- AngularJS + Dependency-->
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
```
henrique: Could we remove the jq dependency here?

uvwxy: The dependency is required to run the tests that do not use the angular js service.

jeking3:

PR #414, opened 2015-03-31
THRIFT-3064 added default handler for undeclared exceptions in php gener...

regeda:
PHP app crashes on every undeclared exception.
But it isn't necessary declare all internal exceptions (like RedisException, DoctrineException, etc.) in .thrift file. They should be catched and serialized by Thrift into TApplicationException (aka INTERNAL_ERROR).
It will be more informative than plane 500 HTTP code.


Jens-G:
What was that?

PR #421, opened 2015-04-03
[THRIFT-3071] check minimum required version of automake in bootstrap.sh

jeking3:
~/thrift$ sh bootstrap.sh 
automake version 1.11.3 is too old (need 1.13 or later)


jeking3:
Going to fix some warnings at the same time.

PR #422, opened 2015-04-03
THRIFT-3080: fix connection leak of C++ Nonblocking Server while huge nu...

abadcafe:
...mber connections are accepted and unix socket stream fd is busy.

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -28,7 +28,10 @@
 #include <thrift/transport/PlatformSocket.h>
 
 #include <iostream>
-#include <poll.h>
+
+#ifdef HAVE_SYS_SELECT_H
+#include <sys/select.h>
+#endif
 
 #ifdef HAVE_SYS_SOCKET_H
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1393,9 +1394,39 @@ bool TNonblockingIOThread::notify(TNonblockingServer::TConnection* conn) {
...
-  if (send(fd, const_cast_sockopt(&conn), kSize, 0) != kSize) {
-    return false;
+  int ret = -1;
+  struct pollfd pfd = {fd, POLLOUT, 0};
+  int kSize = sizeof(conn);
+  const char * pos = (const char *)const_cast_sockopt(&conn);
+
+  while (kSize > 0) {
+    pfd.revents = 0;
+    ret = poll(&pfd, 1, -1);
```
jeking3: Should this be using the THRIFT_POLL and other THRIFT_ macros for portability?

abadcafe: Sure.

jeking3: TServerSocket has a macro based implementation that is portable for reference.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1393,9 +1397,42 @@ bool TNonblockingIOThread::notify(TNonblockingServer::TConnection* conn) {
...
+  int ret = -1;
+  int kSize = sizeof(conn);
+  const char * pos = (const char *)const_cast_sockopt(&conn);
+
+  while (kSize > 0) {
+    FD_ZERO(&wfds);
+    FD_ZERO(&efds);
+    FD_SET(fd, &wfds);
+    FD_SET(fd, &efds);
+    ret = select(fd + 1, NULL, &wfds, &efds, NULL);
```
jeking3: Seems like a dangerous assumption that the file descriptor you are interested is +1 from the obtained file descriptor.  Did you mean to use `getNotificationRecvFD()` instead of `fd + 1` here?

abadcafe: fd + 1 is select()'s convention which is the highest-numbered file descriptor in any of the three sets plus 1. There is only one fd here, so just plus 1 is safe.

jeking3: Fair enough, I read the first page of the manual and saw nfds and made an assumption!

Should we be using THRIFT_POLL here similar to how TSocket does so that it is [more] compatible cross-platform?

abadcafe: In early windows(<= XP), there is no poll() function. The implement of THRIFT_POLL in windows XP was simulated by select(). The problem is the fake poll() lacks POLL_HUP/POLL_ERR state, I am afraid it was not function completed?

On the other side, select() is common enough on every platform supported by thrift.

xiaosuo:
This patch may introduce a more serious problem in theory.

No standard defines writing to a socket is an atomic operation, so if writing is interleaved by other threads, the reading end will end up with an invalid connection pointer, and the following operation on connection will cause segment fault.

In order to fix the above issue, we must turn to pipe(2), which guarantees that writing data less than PIPE_BUF is atomic . And pipes are more efficient than sockets, because no skb is required to save small pointers.

BTW, if we don't set the writing end to nonblocking mode, we don't need to use select to emulate it.

lyhuzi:
为什么不用ul_swriteo_ms_ex2

lyhuzi:
why not use ul_swriteo_ms_ex2

PR #424, opened 2015-04-04
THRIFT-2441 prevent client connections from delaying server stop

jeking3:


lib/cpp/src/thrift/server/TThreadPoolServer.h
```diff
@@ -105,19 +105,16 @@ class TThreadPoolServer : public TServer {
       timeout_(0),
       taskExpiration_(0) {}
 
-  virtual ~TThreadPoolServer();
+  virtual ~TThreadPoolServer() {}
```
ben-craig: This is nit-picky... but I would prefer you revert this particular change.  Please leave the virtual dtor out-of-lined.  An inlined virtual destructor can cause significant object size bloat.  This doesn't end up leaking into final binaries, but it can make build times and file copy times longer.

lib/cpp/test/TServerSocketTest.cpp
```diff
@@ -46,7 +40,8 @@ BOOST_AUTO_TEST_CASE( test_bind_to_address )
     accepted->close();
     sock1.close();
 
-    TServerSocket sock2("this.is.truly.an.unrecognizable.address.", m_serverPort);
+    std::cout << "An error message from getaddrinfo on the console is expected:" << std::endl;
+    TServerSocket sock2("257.258.259.260", m_serverPort);
```
ben-craig: Are you sure this IP address isn't real?  I think I saw someone use it on a TV show.

/kidding

lib/cpp/test/TSocketInterruptTest.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+BOOST_AUTO_TEST_CASE( test_interruptable_child_read )
+{
+    TestTServerSocket sock1("localhost", m_serverPort);
+    sock1.listen();
+    TSocket clientSock("localhost", m_serverPort);
+    clientSock.open();
+    boost::shared_ptr<TTransport> accepted = sock1.acceptImpl();
```
ben-craig: This should really be 'accept()' and not 'acceptImpl()'.

lib/cpp/test/TestTServerSocket.h
```diff
@@ -0,0 +1,31 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <thrift/transport/TServerSocket.h>
+
+class TestTServerSocket : public apache::thrift::transport::TServerSocket
```
ben-craig: I am confused.  Why is this class necessary?  It doesn't seem like it is adding anything of use.

I recommend either commenting why you need / want it in the code, or deleting it.

ben-craig:
All of the concerns I have with this request are minor.  Once these are fixed, I give it a +1.  Note that the out-of-line destructor comment applies to the other destructors that got inlined.

I will not personally be able to do the commit for another week or two, but if someone else is able to (after these fixes), then I will be much appreciative.

lib/cpp/src/thrift/server/TThreadPoolServer.h
jeking3: I did this so that it would be the same as other classes in the project, as TSimpleServer and TThreadedServer do this.  If I change this I would change all three servers to be the same.

Personally, I don't like having implementations in headers at all even if they are trivial, because when you need to change the implementation (and not the signature of the call) you end up rebuilding too much.

lib/cpp/test/TServerSocketTest.cpp
jeking3: Initially I used a bogus DNS name however it took 15 seconds to fail.  This was much faster. :>

lib/cpp/test/TSocketInterruptTest.cpp
jeking3: I am specifically trying to test the implementation here, and not TServerTransport::accept.

lib/cpp/test/TestTServerSocket.h
jeking3: It allows acceptImpl() to be called directly by the test.

lib/cpp/src/thrift/server/TSimpleServer.cpp
```diff
@@ -88,7 +88,7 @@ void TSimpleServer::serve() {
       string errStr = string("Some kind of accept exception: ") + tx.what();
       GlobalOutput(errStr.c_str());
       continue;
-    } catch (string s) {
+    } catch (const string& s) {
```
jeking3: I'm curious, is there anything actually throwing a bare string?  I don't see anything, and in general concerned about the inability to debug errors caused by catching too much stuff.

ben-craig: I don't know of any code in thrift that intentionally throws a string.  Throwing a string (or anything without a vtable really) is certainly a poor practice.

jeking3:
As part of work on THRIFT-3084 I added a TServer integration test, and found that the shutdown behavior and the rules for when you can call setInterruptableChildren() had to be tightened up for a couple cases.  I am going to close this pull request until this is resolved so that it doesn't get merged in until the test passes.

jeking3:
Added an integration test to prove behavior is correct; re-opening for CI build.

jeking3:
Closing and reopening because three builds failed with "build.ninja not found".  Seems unrelated to changes made.

jeking3:
Nevermind, it is my fault - fixing it..

jeking3:
Wow, this really isn't my morning...

lib/cpp/src/thrift/server/TSimpleServer.cpp
jeking3: I removed this logic in THRIFT-3083.

lib/cpp/src/thrift/transport/TSocket.cpp
```diff
@@ -493,29 +585,9 @@ uint32_t TSocket::read(uint8_t* buf, uint32_t len) {
...
     if (errno_copy == THRIFT_ECONNRESET) {
-      /* shigin: freebsd doesn't follow POSIX semantic of recv and fails with
-       * THRIFT_ECONNRESET if peer performed shutdown
-       * edhall: eliminated close() since we do that in the destructor.
-       */
       return 0;
     }
-#endif
-
-#ifdef _WIN32
```
bufferoverflow: This sequence is needed for some platforms, did you some tests on windows?

jeking3: Hi Roger, thanks for inspecting this closely.  I determined that if one called read() on a closed socket, it was not returning 0 like it should.  It turns out that the block seen above which was limiting the logic for handling ECONNRESET to __FreeBSD__ and __MACH__ should be ubiquitous.  If a read() errors with ECONNRESET the proper response is to return 0 indicating EOF.

The WIN32 block is the same, given THRIFT_ECONNRESET == WSAECONNRESET when _WIN32 is enabled.  As such, now all platforms respond to connection reset during read with 0, meaning EOF.

Does the CI build test framework test Windows and run the CMake tests?  If so, then the TServerIntegrationTest will test your concern.  I did not run this code on Windows to make sure it builds or passes tests, but I can do that on Monday if needed.

jeking3: I spent a few hours trying to get a MinGW environment set up, but I wasn't able to so I am switching to Cygwin instead.  Is there a MinGW 64-bit environment out there that has everything you need for Thrift?  MinGW-w64 seemed like it had promise but it was missing the autotool chain.

bufferoverflow: Thanks Jim, I'm fine with your explanations, CMake build on appveyor currently fails and I try to get it up and runnning again. I think Windows platform users shall use Visual Studio instead of cygwin and MinGW or as a better choice clang;-)

jeking3:
Getting a windows build going has been really difficult, which is bizarre because it used to be easy.  I've tried MinGW, and cygwin, and native with cmake.  Each time I have run into some roadblock that keeps me from making any progress.  I want to run the cmake unit tests (make check) on Windows to assist in verifying this change, but it hasn't been easy to do, yet...

jeking3:
To make the (hopefully) impending merge easier, I merged master into this branch and there was one conflict to resolve so I took care of it.

jeking3:
I rebased with master after THRIFT-3000 was reverted to ensure that in Travis CI, job #1 passes.

jeking3:
Committed - closing pull request.

PR #431, opened 2015-04-07
THRIFT-3087 Pass on errors like "connection closed"

NOMORECOFFEE:
Currently client received 'badmatch' error when the thrift server teardown. I'd like information about errors in the thrift clinet before the process exits.
I've attached an implementation which fixes this good enough for my own and add test.


NOMORECOFFEE:
The failing test (https://travis-ci.org/apache/thrift/jobs/57468277) seems to be unrelated to my changes.

PR #432, opened 2015-04-09
[THRIFT-3088] TThreadPoolServer with Sasl auth may leak CLOSE_WAIT socket

punwinger:
Start TThreadPoolServer to server with TSaslServerTransport.Factory as transportFactory. While using nc to test the specified port whether reachable, it will leak CLOSE_WAIT socket.That's because nc will close socket at once while successful connect TThreadPoolServer, but the server still try using sasl protocol to build an inputTransport which of course failed at once. However inputTransport is null which makes it can't close socket properly which lead to CLOSE_WAIT socket.

### DIFF
--- a/lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
+++ b/lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
@@ -303,6 +303,10 @@ public class TThreadPoolServer extends TServer {
       if (outputTransport != null) {
         outputTransport.close();
       }
+
+      if (client_.isOpen()) {
+        client_.close();
+      }
     }
   }
 }


lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
```diff
@@ -303,6 +303,10 @@ public void run() {
       if (outputTransport != null) {
         outputTransport.close();
       }
+
+      if (client_.isOpen()) {
```
jeking3: This was also addressed in THRIFT-3083 where the Simple, Threaded, and ThreadPool server serve() routine was consolidated; some servers closed the client and some did not.

jeking3: Does TSimpleServer in java need the same change?

jeking3:
This was committed on Apr 12 and can be closed, however please check the TServer variants in Java and see that they use the same logic to close all resources.

jeking3:
@jfarrell this can be closed - it was fixed a while back.

PR #433, opened 2015-04-09
THRIFT-3081 consolidate client processing loops

jeking3:
This pull request consolidates the client processing loops contained within TSimpleServer, TThreadedServer, TThreadPoolServer that were all similar but not functionally identical. This will improve maintainability.

### DIFF
--- a/lib/cpp/src/thrift/server/TConnectedClient.cpp
+++ b/lib/cpp/src/thrift/server/TConnectedClient.cpp
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <thrift/server/TConnectedClient.h>
+
+namespace apache { namespace thrift { namespace server {
+
+using apache::thrift::TProcessor;
+using apache::thrift::protocol::TProtocol;
+using apache::thrift::server::TServerEventHandler;
+using apache::thrift::transport::TTransport;
+using apache::thrift::transport::TTransportException;
+using boost::shared_ptr;
+using std::string;
+
+TConnectedClient::TConnectedClient(const string& serverType,
+                                   shared_ptr<TProcessor> processor,
+                                   shared_ptr<TProtocol> inputProtocol,
+                                   shared_ptr<TProtocol> outputProtocol,
+                                   shared_ptr<TServerEventHandler> eventHandler,
+                                   shared_ptr<TTransport> client)
+                        
+  : serverType_(serverType),
+    processor_(processor),
+    inputProtocol_(inputProtocol),
+    outputProtocol_(outputProtocol),
+    eventHandler_(eventHandler),
+    client_(client),
+    opaqueContext_(0)
+{}
+
+TConnectedClient::~TConnectedClient()
+{}
+
+void TConnectedClient::run()
+{
+  if (eventHandler_) {
+    opaqueContext_ = eventHandler_->createContext(inputProtocol_, outputProtocol_);
+  }
+
+  for (;;) {
+    if (eventHandler_) {
+      eventHandler_->processContext(opaqueContext_, client_);
+    }
+
+    try {
+      if (!processor_->process(inputProtocol_, outputProtocol_, opaqueContext_)) {
+        break;
+      }
+    } catch (const TTransportException& ttx) {
+      if (ttx.getType() != TTransportException::END_OF_FILE &&
+          ttx.getType() != TTransportException::INTERRUPTED) {
+            string errStr = (serverType_ + " client died: ") + ttx.what();
+            GlobalOutput(errStr.c_str());
+      }
+      break;
+    }
+  }
+
+  cleanup();
+}
+
+void TConnectedClient::cleanup()
+{
+  if (eventHandler_) {
+    eventHandler_->deleteContext(opaqueContext_, inputProtocol_, outputProtocol_);
+  }
+
+  try {
+    inputProtocol_->getTransport()->close();
+  } catch (const TTransportException& ttx) {
+    string errStr = string(serverType_ + " input close failed: ") + ttx.what();
+    GlobalOutput(errStr.c_str());
+  }
+  try {
+    outputProtocol_->getTransport()->close();
+  } catch (const TTransportException& ttx) {
+    string errStr = string(serverType_ + " output close failed: ") + ttx.what();
+    GlobalOutput(errStr.c_str());
+  }
+  try {
+    client_->close();
+  } catch (const TTransportException& ttx) {
+    string errStr = string(serverType_ + " client close failed: ") + ttx.what();
+    GlobalOutput(errStr.c_str());
+  }
+}
+
+}}} // apache::thrift::server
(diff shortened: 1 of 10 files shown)


lib/cpp/src/thrift/server/TConnectedClient.h
```diff
@@ -0,0 +1,114 @@
...
+     * @param[in] outputProtocol the output TProtocol
+     * @param[in] eventHandler   the server event handler
+     * @param[in] client         the TTransport representing the client
+     */
+    TConnectedClient(const std::string& serverType,
+                     boost::shared_ptr<apache::thrift::TProcessor> processor,
+                     boost::shared_ptr<apache::thrift::protocol::TProtocol> inputProtocol,
+                     boost::shared_ptr<apache::thrift::protocol::TProtocol> outputProtocol,
+                     boost::shared_ptr<apache::thrift::server::TServerEventHandler> eventHandler,
+                     boost::shared_ptr<apache::thrift::transport::TTransport> client);
```
ben-craig: minor / can be ignored:  I would prefer these be passed in by const ref.  Atomic increments and decrements aren't free, and making the arguments const ref can avoid a little bit of ref count twiddling.

lib/cpp/src/thrift/server/TThreadPoolServer.cpp
```diff
@@ -146,9 +73,12 @@ void TThreadPoolServer::serve() {
       shared_ptr<TProcessor> processor = getProcessor(inputProtocol, outputProtocol, client);
 
       // Add to threadmanager pool
-      shared_ptr<TThreadPoolServer::Task> task(
-          new TThreadPoolServer::Task(*this, processor, inputProtocol, outputProtocol, client));
-      threadManager_->add(task, timeout_, taskExpiration_);
+      shared_ptr<TConnectedClient> pClient(
```
ben-craig: use make_shared.  It shaves off an allocation.  For bonus terseness... don't even make a local shared_ptr<TConnectedClient>.  Just pass the make_shared result to threadManager_->add.

lib/cpp/src/thrift/server/TThreadedServer.cpp
```diff
@@ -166,21 +74,19 @@ void TThreadedServer::serve() {
...
-      TThreadedServer::Task* task
-          = new TThreadedServer::Task(*this, processor, inputProtocol, outputProtocol, client);
-
-      // Create a task
-      shared_ptr<Runnable> runnable = shared_ptr<Runnable>(task);
+      shared_ptr<TConnectedClient> pClient(
+              new TConnectedClient("TThreadedServer",
+                      getProcessor(inputProtocol, outputProtocol, client),
+                      inputProtocol, outputProtocol, eventHandler_, client),
+              boost::bind(&TThreadedServer::disposeClient, this, _1));
```
ben-craig: (not an issue)...and here, you can't use make_shared, because you have a custom deleter.

lib/cpp/src/thrift/server/TThreadedServer.h
```diff
@@ -20,120 +20,86 @@
...
     stop_ = true;
     serverTransport_->interrupt();
   }
 
 protected:
-  void init();
+  virtual void disposeClient(TConnectedClient *pClient);
 
-  boost::shared_ptr<ThreadFactory> threadFactory_;
+  boost::shared_ptr<concurrency::ThreadFactory> threadFactory_;
```
ben-craig: /sigh
I curse Microsoft that I have to bring this up...

I've had issues with this spelling with Visual Studio 2012.  Visual Studio 2012 has a concurrency namespace.  Not 'microsoft::concurrency', but just 'concurrency'.  Please put this back to just 'ThreadFactory', or qualify it further with an 'apache::thrift::'.

ben-craig:
Again, just minor issues.  +1 after they get fixed.

As with the THRIFT-2441 review, avoid inlined virtual dtors.  Try to scrub all the 'concurrency::' usages.

lib/cpp/src/thrift/server/TConnectedClient.h
jeking3: This is fine.  I have found the use of references to smart pointers have been confusing to folks with less C++ experience, so I tend to avoid them.  I would hope the compiler could see the constructor is doing nothing and optimize this to a single instantiation, however I have no problem changing it as requested.

lib/cpp/src/thrift/server/TThreadPoolServer.cpp
jeking3: I thought that was supposed to be used when making a shared pointer from within the code of an existing smart pointer, i.e. a class that is boost::enable_shared_from_this.  Wasn't aware it could be used in that manner.  I'll look into this and I can roll it into the subsequent call, I do this normally but was erring on the side of caution on the amount of fanciness.

lib/cpp/src/thrift/server/TThreadedServer.h
jeking3: My personal belief is that namespace aliases and using clauses lead to really nasty compiler issues over time which is why I removed them, or in this case it seems that is was simply assumed someone else would do this before including the header.  To be safe, perhaps they should all be fully qualified.  I struggled with whether or not to do this because of the verbosity.

jeking3: Any objections to adding the following to each class that is typically wrapped in a shared ptr?  We'll take TTransport as an example, instead of:
const boost::shared_ptr<apache::thrift::transport::TTransport>&
it could be
const apache::thrift::transport::TTransport::Ptr&
This is achieved by adding a typedef of
typedef boost::shared_ptr<TTransport> Ptr;
in the TTransport class header.

jeking3: I also think it would make sense to follow best practices on the using clauses and remove all of them from all headers so that namespaces are not contaminated outside the project.  This makes for longer declarations in headers but it is far more reliable especially when switching language revisions or running into things that you mentioned with Microsoft ::concurrency.

jeking3:
Okay this one is ready for re-review now.  I'll remind folks that this change and THRIFT-2441 have an interaction, so whichever one is merged first, I will need to refresh the other to line up with it.  The changes are minor, mostly in stop() being moved to the .cpp file(s).

lib/cpp/src/thrift/server/TConnectedClient.cpp
```diff
@@ -0,0 +1,105 @@
...
+      }
+    } catch (const TTransportException& ttx) {
+      if (ttx.getType() != TTransportException::END_OF_FILE &&
+          ttx.getType() != TTransportException::INTERRUPTED) {
+            string errStr = (serverType_ + " client died: ") + ttx.what();
+            GlobalOutput(errStr.c_str());
+      }
+      break;
+    }
+  }
```
jeking3: Previously there were exception handlers for std::exception and ..., and both would log.  I noticed that TMultiplexedProcessor treats an unknown service name as an exception - it replies to the client and then throws a TException, so I guess we need to handle TException here by logging and then continue, however I would like to specifically not handle std::exception or ... here because catching these and logging makes finding the root cause of errors created by code outside of thrift very difficult, and I don't think the standard servers should be catching things that thrift doesn't throw.  Therefore this needs to change.  Given I want THRIFT-2441 to be merged first I am going to close this pull request and rebase these changes on top of THRIFT-2441, and will re-submit once it is merged.  It will include this change of catching TException and logging it, and continuing the loop.

lib/cpp/src/thrift/server/TThreadPoolServer.cpp
ben-craig: I think you are referring to shared_from_this.  Here are some of the boost docs talking about the difference:
http://www.boost.org/doc/libs/1_57_0/libs/smart_ptr/enable_shared_from_this.html
http://www.boost.org/doc/libs/1_57_0/libs/smart_ptr/make_shared.html

lib/cpp/src/thrift/server/TConnectedClient.cpp
ben-craig: On the client side of things, I agree with catching very little.

For the server side, my preferred policy would be to catch just about everything, log as best we can, then close the connection to the client that caused the exception.  

This code is run pretty close to "threadMain", so any uncaught exceptions will bring down the entire process, resulting in a DoS for other clients.  In addition, some exceptions are sort-of expected, like std::bad_alloc.  If you don't catch bad_alloc, then the client that puts your server one over capacity will make the server as a whole go away.

PR #434, opened 2015-04-10
Fix THRIFT-3090 (cmake build is broken)

marco-m:
Problem found and fixed on Mac OS X 10.10.3, but at least the link problems must happen on every OS.

### DIFF
--- a/lib/cpp/CMakeLists.txt
+++ b/lib/cpp/CMakeLists.txt
@@ -156,7 +156,7 @@ if(WITH_LIBEVENT)
     include_directories(SYSTEM ${LIBEVENT_INCLUDE_DIRS})
 
     ADD_LIBRARY_THRIFT(thriftnb ${thriftcppnb_SOURCES})
-    TARGET_LINK_LIBRARIES_THRIFT(thriftnb ${SYSLIBS} ${LIBEVENT_LIBRARIES})
+    TARGET_LINK_LIBRARIES_THRIFT(thriftnb thrift ${SYSLIBS} ${LIBEVENT_LIBRARIES})
 endif()
 
 if(WITH_ZLIB)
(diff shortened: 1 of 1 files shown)


lib/cpp/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if(WITH_BOOSTTHREADS)
 else()
   find_package(Boost 1.53.0 REQUIRED)
 endif()
-include_directories(SYSTEM "${Boost_INCLUDE_DIRS}")
```
nsuke: Why do you need to remove "SYSTEM" ?

build/cmake/DefinePlatformSpecifc.cmake
```diff
@@ -58,8 +58,9 @@ if(MSVC)
     endif(WITH_MT)
 endif(MSVC)
 
-# GCC Specific
-if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
+# GCC or Clang specific
+if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
bufferoverflow: does CMAKE_COMPILER_IS_CLANG work as well?

lib/cpp/CMakeLists.txt
marco-m: I don't need to remove SYSTEM to fix the cmake build on Mac OS X. On the other hand, I don't see why SYSTEM is needed :-) In any case, this change can be reverted.

build/cmake/DefinePlatformSpecifc.cmake
marco-m: Unfortunately CMAKE_COMPILER_IS_CLANG doesn't exist (or at least didn't exist last time I researched the topic).

lib/cpp/CMakeLists.txt
nsuke: SYSTEM implies -isystem for gcc/clang thus can avoid polluting log with warnings from boost.

PR #440, opened 2015-04-13
THRIFT-1857 Replace deprecated "," with "as" in python exception generation code.

gwax:


jfarrell:
Hi @gwax, thanks for the patch. Can you please create a jira issue for this so we can better track your contribution. details available here: http://thrift.apache.org/docs/HowToContribute

gwax:
I have added the relevant (existing) Jira issue to the description. Is that sufficient or should I do something further?

jfarrell:
perfect, thanks

PR #441, opened 2015-04-14
THRIFT-227 Pretty print binary collections in Java

roshan:
`set<binary>` and `list<binary>` don't have pretty printers corresponding to `binary`. This adds them.


bpodgursky:
nice

PR #446, opened 2015-04-16
THRIFT-3099 cmake build is broken on FreeBSD

marco-m:
See https://issues.apache.org/jira/browse/THRIFT-3099


jeking3:
This was merged on 16/Apr and can be closed.

PR #448, opened 2015-04-17
THRIFT-2877 Generate hashCode using primitives, static utility methods

roshan:
This is pretty much the List.hashCode() except without any list. It takes about a third the time on some rudimentary benchmarks. I tried it out with

```
typedef i32 SomeId
typedef binary BinId
typedef string StringId

struct NonTrue {}

union MaybeAThing {
    1: NonTrue nt
    2: bool bl
}

enum Nomnom {
    EAT=31
    LIVE=515
}

struct AllPrims {
    1: bool boole
    2: byte single_byte
    3: i16 shrt
    4: i32 integ
    5: i64 longue
    6: double f64
    7: string str
    8: binary bin
    9: Nomnom en
    10: NonTrue stru
    11: MaybeAThing un
    12: SomeId intid
    13: BinId binid
    14: StringId strid
}
```

generating [this hashCode()](https://gist.github.com/roshan/7b7fff349e3ed06322c3).

Populating these structs with some values has it match the AbstractList.hashCode() result but maybe we could try a different multiplicative factor.

Sorry about the other one https://github.com/apache/thrift/pull/447. I rebased to one commit, but couldn't seem to reuse the old PR (it compiled locally on gcc but failed on clang).

### DIFF
--- a/compiler/cpp/src/generate/t_java_generator.cc
+++ b/compiler/cpp/src/generate/t_java_generator.cc
@@ -1874,9 +1874,12 @@ void t_java_generator::generate_java_struct_equality(ofstream& out, t_struct* ts
   scope_down(out);
   out << endl;
 
+  const int MUL = 31; // HashCode multiplier
+  const int B_YES = 1231;
+  const int B_NO = 1237;
   out << indent() << "@Override" << endl << indent() << "public int hashCode() {" << endl;
   indent_up();
-  indent(out) << "List<Object> list = new ArrayList<Object>();" << endl;
+  indent(out) << "int hashCode = 1;" << endl;
 
   for (m_iter = members.begin(); m_iter != members.end(); ++m_iter) {
     out << endl;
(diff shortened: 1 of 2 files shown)


roshan:
Ah, I used Java 8 only methods. I'll fix this and reopen later.

nsuke:
Hi @roshan, sorry for the delay, I've just discovered this.
While this change makes sense in general, I have one concern.

> The TBaseHelper.hashCode methods are the Java 8 implementations of hashCode for
those types.

Does this involve any code-level transplant, like copy-paste ?
In that case I'm afraid we have no way to incorporate this, as OpenJDK is GPL (let alone Oracle JDK disassemble).

roshan:
Ha, thanks for that @nsuke . I only used the JDK 8 methods so we could keep `hashCode` identical in value to its current value. Is that not a concern? If so, it should be easy to write a trivial implementation that does not resemble the JDK 8 methods in functionality or value.

nsuke:
@roshan great, thanks for working on it after such a long break.
I would suggest rewriting merely targeting "boxing value not involved" and "reasonably good" hash functions "from scratch."

If the JDK-identical return values are important, we may probably ask https://issues.apache.org/jira/browse/LEGAL if it's feasible first, but it's not the case, I suppose ?

jsirois:
My 2 nonbinding cents: No reasonable client should care about hashCode values, only that the values meet the hashCode contract (this is true of any language).  I don't think preserving them should ever be a goal.

A second comment, not as helpful since it requires work, is that we should enshrine perf improvements using a tool like [jmh](http://openjdk.java.net/projects/code-tools/jmh/).  We use this in [Apache Aurora](https://github.com/apache/aurora/tree/master/src/jmh/java/org/apache/aurora/benchmark) for example to ensure we have a shared history of performance improvements and regressions.

roshan:
@nsuke, I switched it so we now consider a `long` like two `int`s (the high bits and low bits) and we combine their values like we combine `hashCodes` (just with a different multiplicative factor). For `double`s we get the `long` represented by their bytes and treat as before. Since we're changing the values, I also took the chance to pick arbitrary constants to combine with.

For what it's worth, though, the previous `hashCode` helper functions are pretty much what are described in Effective Java. Anyway, I'll rebase and squash after you have a look.

@jsirois That's a good idea. Unfortunately, I don't have the time at the moment to do that :(

compiler/cpp/src/generate/t_java_generator.cc
```diff
@@ -1886,24 +1889,57 @@ void t_java_generator::generate_java_struct_equality(ofstream& out, t_struct* ts
...
+          break;
+        case t_base_type::TYPE_I64:
+          indent(out) << "hashCode = hashCode * " << MUL << " + org.apache.thrift.TBaseHelper.hashCode(" << name << ");" << endl;
+          break;
+        case t_base_type::TYPE_DOUBLE:
+          indent(out) << "hashCode = hashCode * " << MUL << " + org.apache.thrift.TBaseHelper.hashCode(" << name << ");" << endl;
+          break;
+        default:
+          throw "compiler error: the following base type has no hashcode generator: " +
+                 t_base_type::t_base_name(((t_base_type*)t)->get_base());
```
nsuke: We can take this opportunity to stop throwing string here.
As I believe it should never happen (right ?), std::logic_error would be appropriate.

compiler/cpp/src/generate/t_java_generator.cc
```diff
@@ -1886,24 +1889,57 @@ void t_java_generator::generate_java_struct_equality(ofstream& out, t_struct* ts
...
-    indent(out) << "boolean present_" << name << " = " << present << ";" << endl;
-    indent(out) << "list.add(present_" << name << ");" << endl;
-    indent(out) << "if (present_" << name << ")" << endl;
     if (t->is_enum()) {
-      indent(out) << "  list.add(" << name << ".getValue());" << endl;
+      indent(out) << "hashCode = hashCode * " << MUL << " + " << name << ".getValue();" << endl;
+    } else if (t->is_base_type()) {
+      switch(((t_base_type*)t)->get_base()) {
+        case t_base_type::TYPE_VOID:
+          break;
```
nsuke: As this is a struct field, there should never be void ?
I you agree, let's put similar exception to the one for `default:` below.

compiler/cpp/src/generate/t_java_generator.cc
nsuke: Just a style nit, could you indent switch in the same way as the other part of this file (less indentation) ?

compiler/cpp/src/generate/t_java_generator.cc
```diff
@@ -1886,24 +1889,57 @@ void t_java_generator::generate_java_struct_equality(ofstream& out, t_struct* ts
...
+        case t_base_type::TYPE_I8:
+          indent(out) << "hashCode = hashCode * " << MUL << " + (int) (" << name << ");" << endl;
+          break;
+        case t_base_type::TYPE_I16:
+        case t_base_type::TYPE_I32:
+          indent(out) << "hashCode = hashCode * " << MUL << " + " << name << ";" << endl;
+          break;
+        case t_base_type::TYPE_I64:
+          indent(out) << "hashCode = hashCode * " << MUL << " + org.apache.thrift.TBaseHelper.hashCode(" << name << ");" << endl;
+          break;
```
nsuke: It seems it can be merged to `TYPE_DOUBLE:` below.

nsuke:
@roshan I've added comments on minor points. Other than that I think we're good to proceed :+1:

roshan:
Thanks for the comments, @nsuke . I will squash after builds pass.

PR #455, opened 2015-04-21
THRIFT-1857: Added some missing leading periods for the ttypes import

jeroenvlek:
This should fix the ttypes import.


jeroenvlek:
Oops, build failed. I'll look into it.

PR #456, opened 2015-04-21
THRIFT-2850 cmake win cpp library and unit tests

jeking3:
I made changes to the cmake environment so that cmake can build the compiler, static libraries and unit tests on Windows.  In my testing I found that the shared thrift library build is not exporting any symbols so no IMPLIB is being generated (that would be thrift.lib or thriftd.lib), so none of the unit tests would link, nor would thriftz.

Here's the command I used to generate the environment:

mkdir thrift-build
cd thrift-build
"C:\Program Files (x86)\CMake\bin\cmake.exe" ..\thrift -G"NMake Makefiles" -DWITH_SHARED_LIB=OFF -DWITH_BOOSTTHREADS=ON -DBOOST_ROOT=C:\Boost -DZLIB_ROOT=C:\win3p\tools\x64\zlib-1.2.5 -DOPENSSL_ROOT_DIR=C:\win3p\tools\x64\openssl-1.0.1b -DFLEX_EXECUTABLE=C:\Users\Jim\workspace\winflexbison\win_flex.exe -DBISON_EXECUTABLE=C:\Users\Jim\workspace\winflexbison\win_bison.exe

You can also use -G"Visual Studio 10 Win64" to get a thrift.sln file that you can use to load visual studio, this example is for Visual Studio 2010.  Visual Studio 2013 would be -G"Visual Studio 12 Win64".

I was not able to get the nonblocking server to build because libevent requires autoconf to generate a "event-config.h", and the libevent nmake project does not produce one of these, so I could not compile the non-blocking server; further the libevent header "event.h" conflicts with the Windows Platform SDK header Event.h, so FindLibevent.cmake always finds the Platform SDK as an include directory, so I did not attempt to get it working right now.

Tested this on Windows and on Ubuntu 12.04 using CMake-3.2.2.

When running the tests on Windows you need to put the library paths into your PATH environment for boost, openssl so that they can be found.  The CMake environment does not copy required libraries into the "bin" directory at this time.


jeking3:
Darnit, a last minute change busted Windows - closing to fix. :|

jeking3:
csharp build failed and I believe it is unrelated, recycling pull request to rebuild

jeking3:
Need help - all of the "make cross" csharp tests appear to be failing in Travis CI but I do not believe the changes I made would have impacted them at all (this pull request contains CMake and trivial C++ changes).  Are any build artifacts stored from Travis, for example the log files of each make cross child?

nsuke:
It seems that the failure is introduced by THRIFT-3000 https://travis-ci.org/apache/thrift/builds/58981770.

Unfortunately, there are no way (that I known of) to retrieve cross test failure log from Travis.
Adding an `after_failure` entry to `cat` the `unexpected_failures.log` should do, though.

lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -576,12 +576,12 @@ int TSSLSocketFactory::passwordCallback(char* password, int size, int, void* dat
   TSSLSocketFactory* factory = (TSSLSocketFactory*)data;
   string userPassword;
   factory->getPassword(userPassword, size);
-  int length = userPassword.size();
+  std::string::size_type length = userPassword.size();
   if (length > size) {
```
henrique: Don't you get a signed vs unsigned warning here?

jeking3: std::string::size_time is the same as the return value of std::string::size(), which is potentially different than "int" (on Windows this was a compiler warning), which is why I changed it to quiet the warnings.  It is more correct now.

lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -319,7 +319,7 @@ void TSSLSocket::checkHandshake() {
     return;
   }
   ssl_ = ctx_->createSSL();
-  SSL_set_fd(ssl_, socket_);
+  SSL_set_fd(ssl_, (int)socket_);
```
nsuke: Maybe `static_cast` (and the other a few places) ?

jeking3: That's reasonable; the other signedness cases these changes created should also be cleaned up elsewhere.  Thank you for reviewing!

PR #459, opened 2015-04-23
THRIFT-3111 C# library requries .NET 4.5 compatible Mono

nsuke:


nsuke:
Closing since we're not going .NET 4.5 yet.

PR #462, opened 2015-04-25
Thrift 3113 c++11 check issue

RomainNaour:
The ax_cxx_compile_stdcxx_11 macro bundled in thrift package is broken [1] and out of date. Since Thrift actually doesn't use that particular C+11 feature, make C+11 support optional.
Also, add two patches added by Gustavo Zacarias used to cross-compile thrift.
[1] https://savannah.gnu.org/patch/index.php?8287


bufferoverflow:
committed, please close the issue

RomainNaour:
Thanks !

PR #463, opened 2015-04-25
Thrift 3114 Using local temp variables to not pollute the global table

xli2012:
Using local temp variables to not pollute the global table.


Jens-G:
Tracked in THRIFT-3114

PR #465, opened 2015-04-26
THRIFT-3117 Java TSSLTransportFactory can't load certificates within JAR...

Smyatkin-Maxim:
Added getStoreAsStream function to look for certificates in filesystem paths, URLs or classpath resources just like it is done in ActiveMQ.


Smyatkin-Maxim:
Well, looks like some of the most recent commits have broken the build :)

PR #469, opened 2015-04-28
Thrift 727 - support C++ native exception message

mreve:
Added code generating implementation of std::exception::what() function for exceptions.

I needed to modify the tutorial's IDL file, as well as CppClient and CppServer files, because member field of the exception defined & used there is named "what", and that made the compiler complain.

The modification in CppClient contains only a comment - it can be uncommented and used as a test, to see the result of my change.

Please view the discussion in Jira:
https://issues.apache.org/jira/browse/THRIFT-727

### DIFF
--- a/compiler/cpp/src/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/generate/t_cpp_generator.cc
@@ -1643,11 +1643,11 @@ void t_cpp_generator::generate_struct_ostream_operator_decl(std::ofstream& out, 
 void t_cpp_generator::generate_exception_what_method_decl(std::ofstream& out,
                                                           t_struct* tstruct,
                                                           bool external) {
-  std::string prefix = "";
+  out << "const char* ";
   if (external) {
-    prefix = string(tstruct->get_name() + "::");
+    out << tstruct->get_name() << "::";
   }
-  out << "const char* " << prefix << "what() const throw()";
+  out << "what() const throw()";
 }
 
 namespace struct_ostream_operator_generator {
--- a/compiler/cpp/src/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/generate/t_cpp_generator.cc
@@ -1124,6 +1124,7 @@ void t_cpp_generator::generate_struct_declaration(ofstream& out,
   
   // std::exception::what()
   if (is_exception) {
+    out << indent() << "mutable std::string _TExceptionMessageHolder;" << endl;
     out << indent();
     generate_exception_what_method_decl(out, tstruct, false);
     out << ";" << endl;
@@ -1732,13 +1733,34 @@ void t_cpp_generator::generate_exception_what_method(std::ofstream& out, t_struc
   out << " {" << endl;
 
   indent_up();
-
+  out << indent() << "bool dump_successful = true;" << endl;
   out << indent() << "std::stringstream ss;" << endl;
+
+  out << indent() << "try {" << endl;
+
+  indent_up();
   out << indent() << "ss << \"TException - service has thrown: \" << *this;" << endl;
-  out << indent() << "return ss.str().c_str();" << endl;
+  indent_down();
+
+  out << indent() << "} catch (const std::exception& e) {" << endl;
+
+  indent_up();
+  out << indent() << "dump_successful = false;" << endl;
+  indent_down();
+
+  out << indent() << "}" << endl;
+  out << indent() << "if (!dump_successful || !ss.good()) {" << endl;
+
+  indent_up();
+  out << indent() << "return \"TException - service has thrown: " << tstruct->get_name() << "\";" << endl;
+  indent_down();
+
+  out << indent() << "}" << endl;
+  out << indent() << "this->_TExceptionMessageHolder = ss.str();" << endl;
+  out << indent() << "return this->_TExceptionMessageHolder.c_str();" << endl;
 
   indent_down();
-  out <<"}" << endl << endl;
+  out << "}" << endl << endl;
 }
 
 /**
--- a/compiler/cpp/src/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/generate/t_cpp_generator.cc
@@ -138,6 +138,7 @@ public:
   void generate_struct_result_writer(std::ofstream& out, t_struct* tstruct, bool pointers = false);
   void generate_struct_swap(std::ofstream& out, t_struct* tstruct);
   void generate_struct_ostream_operator(std::ofstream& out, t_struct* tstruct);
+  void generate_exception_what_method(std::ofstream& out, t_struct* tstruct);
 
   /**
    * Service-level generation functions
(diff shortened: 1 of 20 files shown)


compiler/cpp/src/generate/t_cpp_generator.cc
```diff
@@ -1626,6 +1640,16 @@ void t_cpp_generator::generate_struct_ostream_operator_decl(std::ofstream& out,
...
 }
 
+void t_cpp_generator::generate_exception_what_method_decl(std::ofstream& out,
+                                                          t_struct* tstruct,
+                                                          bool external) {
+  std::string prefix = "";
+  if (external) {
+    prefix = string(tstruct->get_name() + "::");
+  }
+  out << "const char* " << prefix << "what() const throw()";
```
hcorg: throw() is deprecated (especially that you concatenate strings there and you cannot guarantee "no throw" behavior)

tutorial/tutorial.thrift
```diff
@@ -113,7 +113,7 @@ struct Work {
  * Structs can also be exceptions, if they are nasty.
  */
 exception InvalidOperation {
-  1: i32 what,
+  1: i32 what_op,
```
hcorg: 1. this naming convention is inconsistent with rest of tutorial.thrift file (camelCase)
2. changing it here requires update in all tutorials, not only C++

compiler/cpp/src/generate/t_cpp_generator.cc
```diff
@@ -1699,6 +1723,24 @@ void t_cpp_generator::generate_struct_ostream_operator(std::ofstream& out, t_str
...
+void t_cpp_generator::generate_exception_what_method(std::ofstream& out, t_struct* tstruct) {
+  out << indent();
+  generate_exception_what_method_decl(out, tstruct, true);
+  out << " {" << endl;
+
+  indent_up();
+
+  out << indent() << "std::stringstream ss;" << endl;
+  out << indent() << "ss << \"TException - service has thrown: \" << *this;" << endl;
+  out << indent() << "return ss.str().c_str();" << endl;
```
hcorg: this can cause Segmenation Fault, as ss will be destroyed on function exit but pointer to it external contents is returned...

Please do try making a unit test for that functionality

compiler/cpp/src/generate/t_cpp_generator.cc
mreve: It may be, but TException implements what() using this definition. Removing "throw()" results in compilation error: "overriding 'virtual const char* apache::thrift::TException::what() const throw ()'".

However, I believe I can just implement it without concatenating strings specifically.

tutorial/tutorial.thrift
mreve: done.

compiler/cpp/src/generate/t_cpp_generator.cc
mreve: It may be, but TException implements what() using this definition. Removing "throw()" results in compilation error: "overriding 'virtual const char* apache::thrift::TException::what() const throw ()'".

However, I can just implement it without concatenating strings specifically.

compiler/cpp/src/generate/t_cpp_generator.cc
mreve: I can't find where & how to add unit test. Would you mind giving me any hint on that subject, or pointing me to the right resources?

hcorg: see how unit testing of generated code is done in lib/cpp/test

for example: https://github.com/apache/thrift/blob/master/lib/cpp/test/ToStringTest.cpp

mreve: Hey, thanks for help! However, before I begin to write it - I believe it can cause sometimes SegFault even if it would pass the unit test. And, what more, we'll encounter this problem no matter what we return here (any string created in what() will be destroyed the same way as the ss, won't it?). 

So, the only way to make it work is to store the returned text in a field of the custom exception class, am I right? Then we can return a pointer to this field, and it will remain until the exception is not destroyed. BUT - what is a const function. So, I can't modify the object itself inside what(). 

There is a tricky way to get around it...we can keep a mutable variable as a data member field. Then in what() we can generate the message, store it in mutable field (which we can modify no matter const function), and return the pointer to this field. But again, if somebody calls what() twice, the second call will override the message so the first obtained pointer can become invalid... And another thing, this mutable field, what type should it be? String, and we'd return .c_str() from it?

What do you think would be a solution? I understand that the scenarios I'm talking about aren't very typical for the exceptions, but I think they still should be discussed...

hcorg: I think mutable std::string field named texceptionMessageHolder (or even longer name) is a good solution - mutable s a keyword so it's not that much a trick - we still do not change exception contents, but helper field.
It seems very unlikely (and quite ugly) for someone to store result of what() as a const char* and call what() again... "what()" is usually used for logging, and as such is rarely called twice, nor someone stores it direct result.

Whenever in doubt - choose std::string over const char* - it's better to have handle object to manage memory etc. In very rare situations memory array may be better (efficiency, memory consumption etc.) yet it has to be first proven to be so.

tutorial/tutorial.thrift
```diff
@@ -113,7 +113,7 @@ struct Work {
  * Structs can also be exceptions, if they are nasty.
  */
 exception InvalidOperation {
-  1: i32 what,
+  1: i32 whatOp,
```
jeking3: "what" needs to be added to the compiler's list of keywords as part of this work.

compiler/cpp/src/generate/t_cpp_generator.cc
jeking3: You could store the content of the string in a std::string member variable on construction then return mString.c_str(), holding the string variable.  The specification (even C++11) is very flimsy about specifying the valid lifetime of this pointer.

Also we will need to revisit this when we enable a feature (documented in other Jira tickets) to set the minimum required C++ version of the resulting generated code.  Before C++11, what is specified as throw(), with C++11 this changes to noexcept;

tutorial/tutorial.thrift
hcorg: no, it should not - as a rule: Thrift does not keep track of language specific keywords.
This decision was made some time ago mostly because of two reasons:
 - with current language and platform coverage list of such keywords would be very long and hard to maintain
 - someone using Thrift to connect Java with Python should not be bothered with keywords coming from C++ library specifics

Proper solution was proposed: each generator should provide list of keywords extending "core Thrift keywords" and when generating code should check identifiers against those keywords (by some generic mechanisms). That would require some refactoring of generators in compiler, yet maybe someday..

compiler/cpp/src/generate/t_cpp_generator.cc
hcorg: C++11 will be covered with cpp_v2 lib, which will generate this method with noexcept.
Yet stringstream and string does not give noexcept guarantee (for example - they probably allocate memory). So this will require some more thought in C++11

jeking3: So in this case you would prefer a buffer allocated in the constructor, filled in, so what() cannot throw - sounds fair.

PR #473, opened 2015-04-30
THRIFT-3084 add optional concurrent client limit enforcement to lib/cpp

jeking3:
This allows users of TThreadedServer and TThreadPoolServer to declare the maximum number of concurrent clients allowed in at any given time.  When the limit is reached, the serve() thread is blocked from accept()ing another connection until one of the other connections closes.


jeking3:
Closing as it was merged into external repo.

PR #474, opened 2015-04-30
THRIFT-2850 get windows cmake working again and building the unit tests

jeking3:
I made changes to the cmake environment so that cmake can build the compiler, static libraries and unit tests on Windows. In my testing I found that the shared thrift library build is not exporting any symbols so no IMPLIB is being generated (that would be thrift.lib or thriftd.lib), so none of the unit tests would link, nor would thriftz.

Here's the command I used to generate the environment:

mkdir thrift-build
cd thrift-build
"C:\Program Files (x86)\CMake\bin\cmake.exe" ..\thrift -G"NMake Makefiles" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWITH_SHARED_LIB=OFF -DWITH_STATIC_LIB=ON -DWITH_BOOSTTHREADS=ON -DBOOST_ROOT=C:\Boost -DZLIB_ROOT=C:\win3p\tools\x64\zlib-1.2.5 -DOPENSSL_ROOT_DIR=C:\win3p\tools\x64\openssl-1.0.1b -DFLEX_EXECUTABLE=C:\Users\Jim\workspace\winflexbison\win_flex.exe -DBISON_EXECUTABLE=C:\Users\Jim\workspace\winflexbison\win_bison.exe

You can also use -G"Visual Studio 10 Win64" to get a thrift.sln file that you can use to load visual studio, this example is for Visual Studio 2010. Visual Studio 2013 would be -G"Visual Studio 12 Win64".

I was not able to get the nonblocking server to build because libevent requires autoconf to generate a "event-config.h", and the libevent nmake project does not produce one of these, so I could not compile the non-blocking server; further the libevent header "event.h" conflicts with the Windows Platform SDK header Event.h, so FindLibevent.cmake always finds the Platform SDK as an include directory, so I did not attempt to get it working right now.

Tested this on Windows and on Ubuntu 12.04 using CMake-3.2.2.

When running the tests on Windows you need to put the library paths into your PATH environment for boost, openssl so that they can be found. The CMake environment does not copy required libraries into the "bin" directory at this time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,6 @@
 # under the License.
 #
 
-
 cmake_minimum_required(VERSION 2.8.12)
 
 project(thrift)
--- a/build/cmake/ConfigureChecks.cmake
+++ b/build/cmake/ConfigureChecks.cmake
@@ -31,6 +31,7 @@ endif(NOT HAVE_AI_ADDRCONFIG)
 
 check_include_file(arpa/inet.h HAVE_ARPA_INET_H)
 check_include_file(fcntl.h HAVE_FCNTL_H)
+check_include_file(getopt.h HAVE_GETOPT_H)
 check_include_file(inttypes.h HAVE_INTTYPES_H)
 check_include_file(netdb.h HAVE_NETDB_H)
 check_include_file(netinet/in.h HAVE_NETINET_IN_H)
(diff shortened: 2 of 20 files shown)


jeking3:
Closing as it was merged into external repo.

PR #476, opened 2015-05-01
THRIFT-3122 Convert plain js objects given as thrift struct constructor arguments to thrift structs

itkach:


bufferoverflow:
nice idea, but the introduction of new structs within test/ThriftTest.thrift breaks many other language tests, do you really need new structs? e.g.  BonkMap is available as MapType
There is also a inconsistent naming: ListBonks, BonkSet, BonkMap

itkach:
> nice idea, but the introduction of new structs within test/ThriftTest.thrift breaks many other language tests, do you really need new structs? e.g. BonkMap is available as MapType

`MapType` is a typedef for `map<string,Bonk>`, `BonkMap` is a struct containing member of type `map<string,Bonk>` - not the same thing.

The unit tests cover several basic cases:
- struct member is a struct (`Xtruct2` fits)
- struct member is a container that contains a struct
 * list (`ListBonks` fits)
 * set (no suitable existing type, `BonkSet` added)
 * map (no suitable existing type, `BonkMap` added)
- struct member is a container that contains container (no suitable existing type, `Bonkers` added - `CrazyNesting` doesn't fit because it has a map with a struct as a key which can't work in Javascript and no values that are structs)

It does appear that new structs are necessary. I understand that adding new service methods would break other tests, but adding struct definitions shouldn't break anything, it's really surprising (and wrong) 
if it does. If that is indeed the case I can move new struct definitions to a separate .thrift file to be used just in the deep constructor unit test.

> There is also a inconsistent naming: ListBonks, BonkSet, BonkMap

I can rename `BonkSet` -> `SetBonks`, `BonkMap` -> `MapBonks` to match `ListBonks` which was there before, although if I need to move structs for the deep constructor unit test to a separate file anyway I'd probably use different names altogether. Let me know.

henrique:
Hi Itkach,
I also had a look at your patch before and it looks valid but the problem here is the BonkSet, which in C++ will require comparison operators, as you can see in:
https://travis-ci.org/apache/thrift/jobs/60874797#L1480
As you said an ,alternative would be to create a separated test for that.
I also wonder if the compiler is the right place to add the __thriftCopyXXX functions, I think they should probably belong to the library code.

itkach:
> BonkSet, which in C++ will require comparison operators,

That is a strange requirement and it's not clear how would one go about providing such comparison operators for generated classes - looks like an issue with C++ code to me. But this is another discussion. I will restore ThriftTest.thrift to the way it was and will create a new .thrift file for my test .

> I also wonder if the compiler is the right place to add the __thriftCopyXXX functions, I think they should probably belong to the library code.

That's easy for nodejs, but in a browser environment user would have to explicitly add dependency on this new library via <script> tag (which is internal implementation detail and is not to be used for the client code directly), otherwise generated code would be broken "out of the box". So I think compiler is, unfortunately, the right place for these functions. 

Ideally there would be no compiler for js and no generated javascript at all, I think it should all be library code that works off of the spec (say spec'c json representation), but that's not the current approach and is a discussion for another day.

henrique:
Thanks for fixing it Igor but the acceptance tests "make cross" fail here:
https://travis-ci.org/apache/thrift/jobs/61195454#L5395

itkach:
I see that the build failed again: https://travis-ci.org/apache/thrift/jobs/61506900#L5435
but it doesn't look like this is related to my changes. When I run tests java-nodejs and java-py tests locally I sometimes get more failures then this due a timeouts and sometimes everything passes.

itkach:
Are there any other concerns I need to address to have this change merged?

henrique:
Hi Igor,
Thanks for pinging and sorry for the delay here. I have been pretty busy recently but can have a look at it later tonight.
Tests for the browser side would be awesome and, as I said before, putting the copy functions in the libs instead of copying them on every service would be nice.

itkach:
> Tests for the browser side would be awesome 

Existing browser tests pass when executed locally (make check with Java server is successful and grunt test fails only two tests involving the very long string because my node, or rather iojs, server pukes for some reason - doesn't appear to be related to my changes), in addition to unit tests running under node. If I understand correctly browser tests run in PhantomJS, which is basically some version of WebKit, which is ECMA5 js environment - not that the new code assumes ECMA5, but if concern is about js versions compatibility then "browser" tests don't really give a lot of extra assurances compared to tests under node.  So I'm not quite sure what additional tests are you looking for. Can you elaborate?

> and, as I said before, putting the copy functions in the libs instead of copying them on every service would be nice.

When you say `in the libs` do you actually mean two places - `nodejs/lib/thrift/` and `js/src/thrift.js`? Just to clarify - you would like the new list/map copy functions to be added in (copied into) these two places instead of a single place in `t_js_generator.cc`?

henrique:
1. e.g. lib/js/test/deep-constructor.test.js ? If you have the time it will be perfect.
2. Yes, we have this duplication but at least it would be only loaded once, in case there are multiple services. An alternative would be to add a compiler option in t_js_generator.cc

itkach:
> e.g. lib/js/test/deep-constructor.test.js ? 

So more duplication, except it also needs to be rewritten for the different test runner js uses

> An alternative would be to add a compiler option in t_js_generator.cc

What would the option do?

henrique:
e.g. lib/js/test/deep-constructor.test.js ?

    So more duplication, except it also needs to be rewritten for the different test runner js uses

It looks like we are going nowhere here, but yes they are two different libraries and need to be tested twice; I don't mind how. If you have another idea just let us know.

        An alternative would be to add a compiler option in t_js_generator.cc

    What would the option do?

I meant to make these changes (THRIFT-3122) optional. e.g thrift --gen js:jquery --gen node:convert-plain. But this is not a requirement.

bufferoverflow:
What about Jasmine, Karma and PhantomJS as unit test suite?
and Protractor if we do e2e, what's best for nodejs? or node.js io.js and javascript all in one?

itkach:
Copy functions are now in libs, for js and node, tests for js are added. While porting tests to use json protocol for js/browser I noticed a fix for list values in maps was present in js version but not in nodejs (f9b1ca8) and nested lists where not properly read (35b6163199ee2dd8d30689fdb429656aaa09330d). Is there anything else that needs to be addressed here?

henrique:
I should really had trusted Travis. This also broke my fork and I will have to revert it unless someone provides a quick fix.
It seems to be breaking "make check" just after the successful testAll.sh? A bit weird to be honest...
https://api.travis-ci.org/jobs/64377830/log.txt?deansi=true

itkach:
Relevant error is

```
Error: Cannot find module './gen-nodejs/JsDeepConstructorTest_types'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/home/travis/build/henrique/thrift/lib/nodejs/test/deep-constructor.test.js:1:76)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
```
and

https://github.com/itkach/thrift/commit/4502c457cf523fbbc8816b40925bb006a316b808

should fix it

henrique:
Hi Igor,

Yes, we missed that too, I fixed another syntax error for the travis version of node but I see now that your unit-test is actually failing (I initially thought it was only a CI problem).
https://travis-ci.org/henrique/thrift/jobs/64895282
Thanks again for your PR but I had to revert it, as I don't have time to fix it myself. I would suggest that you make sure it runs on Travis before resubmitting the pull request like described in:
https://github.com/itkach/thrift/blob/master/CONTRIBUTING.md#contributing-via-github-pull-requests

Cheers,
Henrique

itkach:
> I fixed another syntax error for the travis version of node but I see now that your unit-test is actually failing

It appears that test is failing because of this incorrect [change](https://github.com/henrique/thrift/commit/2a867c67fe631ae82d4697c2e8cdcf711b518a01).

henrique:
Thanks for pointing out Igor but it was broken before that change already

itkach:
> Thanks for pointing out Igor but it was broken before that change already

How? What was the error? All tests in `nodejs/test/deep-constructor.test.js` pass in my THRIFT-3122 branch

henrique:
it doesn't pass in Travis CI (nor on my laptop), please setup the github travis webhook on your fork so you can check it.
https://travis-ci.org/henrique/thrift/builds/64881077

itkach:
> setup the github travis webhook on your fork

Are there instructions on how to do it? Looks like builds are triggered automatically only on open pull requests?

> https://travis-ci.org/henrique/thrift/builds/64881077

I see... 
```
NodeJS Library:
   Using NodeJS .............. : /home/travis/.nvm/v0.10.36/bin/node
   Using NodeJS version....... : v0.10.36
```
`Buffer` doesn't have `equals()` method in Node 0.10.

henrique:
http://docs.travis-ci.com/user/getting-started/#Step-one%3A-Sign-in

itkach:
I enabled travis hook for my fork and pushed test fix (https://github.com/itkach/thrift/commit/861d262c66f21ad85343ef419d0081ba294e158a) to trigger the build (https://travis-ci.org/itkach/thrift/jobs/65926385). 

It immediately failed:
```
$ sh build/travis/installCXXDependencies.sh
sudo: must be setuid root
sudo: must be setuid root
sudo: must be setuid root
dpkg-query: no path found matching pattern /usr/include/boost/version.hpp.

The command "sh build/travis/installCXXDependencies.sh" failed and exited with 1 during .

Your build has been stopped.

```

So I guess my travis setup is incomplete. What am I missing?

PR #483, opened 2015-05-06
THRIFT-3130 finish removal of THRIFT_OVERLOAD_IF

jeking3:
Testing proved it was unnecessary.


jeking3:
I made some small changes to the TServerIntegrationTest to support testing THRIFT-1944 and rebased the fixup so there's still only one commit tag.  The resulting build failed early so I recycled the pull request.

PR #486, opened 2015-05-10
THRIFT-1248 fix TMemoryBuffer pointer arithmetic and add unit test

jeking3:


HuaisiXu:
Could you describe the bug? We saw a heap corruption bug that may be caused by Thrift. We see wBound_ is less than wBase_ in core dump.

PR #488, opened 2015-05-11
THRIFT-3143

wilfrem:
Current typescript support is only work for browser, and generated d.ts uses internal module.
So, it's hard to use for typescript in node.js.

To solve probrem, generate typescript code directory.

### DIFF
--- a/compiler/cpp/Makefile.am
+++ b/compiler/cpp/Makefile.am
@@ -86,6 +86,7 @@ thrift_SOURCES += src/generate/t_c_glib_generator.cc \
                   src/generate/t_xsd_generator.cc \
                   src/generate/t_html_generator.cc \
                   src/generate/t_js_generator.cc \
+                  src/generate/t_ts_generator.cc \
                   src/generate/t_javame_generator.cc \
                   src/generate/t_delphi_generator.cc \
                   src/generate/t_go_generator.cc \
--- a/compiler/cpp/src/generate/t_ts_generator.cc
+++ b/compiler/cpp/src/generate/t_ts_generator.cc
@@ -167,6 +167,7 @@ public:
                                  std::string prefix = "",
                                  bool include_callback = false,
                                  bool ignore_returntype = false);
+  std::string handler_function_signature(t_function* tfunction);
   std::string argument_list(t_struct* tstruct, bool include_type = true);
   std::string type_to_enum(t_type* ttype);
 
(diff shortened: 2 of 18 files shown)


LoungeFlyZ:
This would be great to get in.  Is anyone able to take a look at it?

endel:
Thanks for your work @wilfrem. I think this pull-request should be reviewed and maybe accepted, since JavaScript is evolving so fast - which makes the current JavaScript/Node generators unusable in modern build tools for the browser, for example. Having a TypeScript generator could solve this problem.

I did some changes in the `import` section, which seems to be broken in the current version. Here's the patch to import the actual references needed:

```patch
diff --git a/compiler/cpp/src/generate/t_ts_generator.cc b/compiler/cpp/src/generate/t_ts_generator.cc
index 2100f7f..44fbeb5 100644
--- a/compiler/cpp/src/generate/t_ts_generator.cc
+++ b/compiler/cpp/src/generate/t_ts_generator.cc
@@ -343,9 +341,36 @@ string t_ts_generator::render_includes() {
   if (gen_node_) {
     const vector<t_program*>& includes = program_->get_includes();
     for (size_t i = 0; i < includes.size(); ++i) {
-      result += "import " + includes[i]->get_name() + "_ttypes = require('./" + includes[i]->get_name()
-                + "_types')\n";
+
+      // Import all exposed enums/structs/types from defined module
+      std::vector<std::string> imports;
+
+      const std::vector<t_typedef*>& typedefs = includes[i]->get_typedefs();
+      for (size_t j = 0; j < typedefs.size(); ++j) {
+        imports.push_back(typedefs[j]->get_type()->get_name());
+      }
+
+      const std::vector<t_enum*>& enums = includes[i]->get_enums();
+      for (size_t j = 0; j < enums.size(); ++j) {
+        imports.push_back(enums[j]->get_name());
+      }
+
+      const std::vector<t_struct*>& structs = includes[i]->get_structs();
+      for (size_t j = 0; j < structs.size(); ++j) {
+        imports.push_back(structs[j]->get_name());
+      }
+
+      std::stringstream modules;
+      for(size_t j = 0; j < imports.size(); ++j)
+      {
+        if(j != 0)
+          modules << ", ";
+        modules << imports[j];
+      }
+
+      result += "import { " + modules.str() + " } from './" + includes[i]->get_name() + ".ts'\n";
     }
+
     if (includes.size() > 0) {
       result += "\n";
     }
```

endel:
My approach is not very clean as it simply ignores node.js existance, it is meant only to run in the browser: https://github.com/endel/thrift/commit/32357443b5bfb5663ef3c45036c1680c08d04953

HIRANO-Satoshi:
Does anyone merge it?

I think I need this.

jeking3:
I'd recommend that you rebase your changes on the latest origin/master and push to make sure they are compatible with the development tip and pass all tests.

estk:
Bump.
Is anyone working on this?

jeking3:
I'd recommend that you rebase your changes on the latest origin/master and push to make sure they are compatible with the development tip and pass all tests.

jeking3:

PR #500, opened 2015-05-16
THRIFT-3112: Typed AsyncMethodCallback with return type

mreve:
In Java generator, a few occurrences of AsyncMethodCallback were typed with function return type, and a few were not. Please correct me if it should be typed differently.

See:
https://issues.apache.org/jira/browse/THRIFT-3112

### DIFF
--- a/compiler/cpp/src/generate/t_java_generator.cc
+++ b/compiler/cpp/src/generate/t_java_generator.cc
@@ -4144,7 +4144,8 @@ string t_java_generator::async_function_call_arglist(t_function* tfunc,
   }
 
   if (include_types) {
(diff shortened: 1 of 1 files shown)


jsirois:
@mreve I've added some explanation on https://issues.apache.org/jira/browse/THRIFT-3112, but this patch should in-fact be typed differently (today).  FWICT the typing needed though is fairly broken and your typing - which matches server side processor typing - should be the way we move forward.  I may move the discussion in https://issues.apache.org/jira/browse/THRIFT-3112 to a PROPOSAL thread that proposes fixing the client side typing which will be a breaking change for users.

jeking3:
@jfarrell this is closed/fixed in Jira.  This PR can be closed.

PR #501, opened 2015-05-17
THRIFT-3154: tornado tutorial fail to work

tryer3000:


jeking3:
@jfarrell this is closed fixed in Jira.  This PR can be closed.

PR #508, opened 2015-05-30
THRIFT-3170: Add a flag to allow the ignoring of common initialisms in Go

Magrath:
As discussed in https://issues.apache.org/jira/browse/THRIFT-3170


Magrath:
@Jens-G I've done the necessary merging, this should be good to go now.

PR #509, opened 2015-05-30
THRIFT-3174: Modify initialism code in Go compiler to check first word

Magrath:
https://issues.apache.org/jira/browse/THRIFT-3174


Magrath:
Thanks!

PR #511, opened 2015-06-02
Limit lists to 10,000 items in fastbinary decoding to avoid crashing …

dvirsky:
…servers from huge allocations on junk/mallicious input

https://issues.apache.org/jira/browse/THRIFT-3175

### DIFF
--- a/lib/py/src/protocol/fastbinary.c
+++ b/lib/py/src/protocol/fastbinary.c
@@ -32,7 +32,7 @@
 # if defined(_MSC_VER) && _MSC_VER < 1600
    typedef int _Bool;
 #  define bool _Bool
-#  define false 0 
+#  define false 0
 #  define true 1
 # endif
 # define inline __inline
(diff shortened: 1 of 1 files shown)


dvirsky:
The build failure seems to be unrelated to the patch:

```
...
vector-0.10.12.3 failed during the building phase. The exception was:
ExitFailure 9
This may be due to an out-of-memory condition.
```

lib/py/src/protocol/fastbinary.c
```diff
@@ -936,7 +951,7 @@ decode_struct(DecodeBuffer* input, PyObject* output, PyObject* spec_seq) {
 }
 
 
-/* --- MAIN RECURSIVE INPUT FUNCTION --- */
+/* --- MAIN RECURSIVE INPUT FUCNTION --- */
```
bufferoverflow: What's this?

lib/py/src/protocol/fastbinary.c
```diff
@@ -1028,7 +1043,7 @@ decode_val(DecodeBuffer* input, TType type, PyObject* typeargs) {
     }
 
     len = readI32(input);
-    if (!check_ssize_t_32(len)) {
+    if (!check_list_length(len)) {
       return NULL;
```
bufferoverflow: I guess we need to inform the user about reaching the limit somehow, e.g. via PyErr_SetString

lib/py/src/protocol/fastbinary.c
dvirsky: I'll fix the typo

lib/py/src/protocol/fastbinary.c
dvirsky: check_list_length already does that:
```c
   PyErr_SetString(PyExc_OverflowError, "list size out of the sanity limit (10000 items max)");
```

lib/py/src/protocol/fastbinary.c
dvirsky: fixed in 9994c6e55f3c11acc523545c1ca67deb05cd0910

PR #512, opened 2015-06-05
THRIFT-3182: Close transport after reading an invalid frame size

marshalium:
This is necessary because after reading the invalid
frame size from the underlying transport the transport
is left in a bad state. Any following reads will
likely throw more invalid frame size exceptions and
if they don’t they will be reading corrupted messages.

Closing forces the caller to discard the connection and
get a new one.

### DIFF
--- a/lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
+++ b/lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
@@ -19,12 +19,12 @@
 package org.apache.thrift.transport;
 
 /**
- * This transport is wire compatible with {@link TFramedTransport}, but makes 
+ * This transport is wire compatible with {@link TFramedTransport}, but makes
  * use of reusable, expanding read and write buffers in order to avoid
  * allocating new byte[]s all the time. Since the buffers only expand, you
  * should probably only use this transport if your messages are not too variably
  * large, unless the persistent memory cost is not an issue.
- * 
+ *
  * This implementation is NOT threadsafe.
  */
 public class TFastFramedTransport extends TTransport {
@@ -91,7 +91,7 @@ public class TFastFramedTransport extends TTransport {
   }
 
   /**
-   * 
+   *
    * @param underlying Transport that real reads and writes will go through to.
    * @param initialBufferCapacity The initial size of the read and write buffers.
    * In practice, it's not critical to set this unless you know in advance that
@@ -141,11 +141,14 @@ public class TFastFramedTransport extends TTransport {
     int size = TFramedTransport.decodeFrameSize(i32buf);
 
     if (size < 0) {
-      throw new TTransportException("Read a negative frame size (" + size + ")!");
+      close();
+      throw new TTransportException(TTransportException.CORRUPTED_DATA, "Read a negative frame size (" + size + ")!");
     }
 
     if (size > maxLength) {
-      throw new TTransportException("Frame size (" + size + ") larger than max length (" + maxLength + ")!");
+      close();
+      throw new TTransportException(TTransportException.CORRUPTED_DATA,
+          "Frame size (" + size + ") larger than max length (" + maxLength + ")!");
     }
 
     readBuffer.fill(underlying, size);
--- a/lib/java/src/org/apache/thrift/transport/TFramedTransport.java
+++ b/lib/java/src/org/apache/thrift/transport/TFramedTransport.java
@@ -130,11 +130,14 @@ public class TFramedTransport extends TTransport {
     int size = decodeFrameSize(i32buf);
 
     if (size < 0) {
-      throw new TTransportException("Read a negative frame size (" + size + ")!");
+      close();
+      throw new TTransportException(TTransportException.CORRUPTED_DATA, "Read a negative frame size (" + size + ")!");
     }
 
     if (size > maxLength_) {
-      throw new TTransportException("Frame size (" + size + ") larger than max length (" + maxLength_ + ")!");
+      close();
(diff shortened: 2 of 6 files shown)


Jens-G:
Hi @marshalium,

could you just [file a JIRA ticket](http://thrift.apache.org/docs/HowToContribute) and link it to this PR by mentioning the number somewhere? 

Re the patch: Do we really need that new exception type? What about non-Java Clients? How are they supposed to handle that case?

Thanks!

marshalium:
Jira opened: https://issues.apache.org/jira/browse/THRIFT-3182

I added the new exception as a subclass of TTransportException exception type because I think it's more user friendly to catch specific exception subclasses when the user of the api cares to. Is it more the Thrift style to add another type constant to TTransportException and not subclass? That way a user has to catch TTransportException and then switch on type.

I think users may want to handle this exception differently because it can mean that the client has a misconfigured frame size or is requesting too much data.

Looking at some of the other clients (for example: https://github.com/apache/thrift/blob/master/lib/csharp/src/Transport/TFramedTransport.cs) they don't seem to even enforce that the frame size is valid. But I think they could all be updated to both validate the frame size and properly close the transport when they read an invalid one.

lib/java/src/org/apache/thrift/transport/TInvalidFrameSizeException.java
```diff
@@ -0,0 +1,29 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.thrift.transport;
+
+public class TInvalidFrameSizeException extends TTransportException {
```
jeking3: This would be simpler as a TTransportException(TTransportExceptionType::CORRUPTED_DATA) (from C++).  That's the convention currently in the code, to use an exception type, not to derive.  I agree with you in principle as it is easier to unit test changes with specific exception types, just want to point out if doesn't follow the project's currently established conventions.  

At the very least, the TInvalidFrameSizeException should initialize super with that exception type code.

Jens-G: I always understood Thrift as being as consistent as possible across languages, and I think this is a benefit that we should not give up. I already said that earlier, but I say it again: Introducing a new exception typ is nothing I'm against per se, quite the contrary. But if we do it only in one language, we start to break this convention and open doors that we don't want open. [Recommended reading: Broken Window theory](https://en.wikipedia.org/wiki/Broken_windows_theory).

marshalium: Fair enough. I'll add CORRUPTED_DATA to TTransportException and start throwing a TTransportException with that type instead of a TInvalidFrameSizeException.

marshalium:
I've pushed a commit that uses TTransportException instead of a new subclass. This should be ready to review again.

jeking3:
BUILD FAILED
/home/travis/build/apache/thrift/lib/java/build.xml:202: Test org.apache.thrift.transport.TestTFastFramedTransport failed

marshalium:
This new commit should take care of that test. Sorry about that.

marshalium:
That build failure was different. Java tests passed, but it failed building the Haskell code. It seem unrelated:

> cabal: Error: some packages failed to install:
> HTTP-4000.2.20 depends on text-1.2.1.1 which failed to install.
> attoparsec-0.13.0.0 depends on text-1.2.1.1 which failed to install.
> hashable-1.2.3.3 depends on text-1.2.1.1 which failed to install.
> network-uri-2.6.0.3 depends on text-1.2.1.1 which failed to install.
> parsec-3.1.9 depends on text-1.2.1.1 which failed to install.
> scientific-0.3.3.8 depends on text-1.2.1.1 which failed to install.
> text-1.2.1.1 failed during the building phase. The exception was:
> ExitFailure 9
> This may be due to an out-of-memory condition.
> thrift-1.0.0 depends on text-1.2.1.1 which failed to install.
> unordered-containers-0.2.5.1 depends on text-1.2.1.1 which failed to install.

marshalium:
I believe all issues have been addressed. Is this blocked on something, or is there something I need to do to get this merged?

jeking3:
I've been kicking off new builds until it passes.  I agree it doesn't look like the failure is related, but if you could keep kicking it until it does, that would be ideal.  I'd like to continue to promote a culture of passing the CI build before merge, given there is so much goodness in the result of that environment.

marshalium:
How do I kick off a new a build? I don't see any UI for that in Travis. Maybe it needs to be done by an administrator of some sort?

I'm not optimistic that it will pass if I were able to kick it off again. Looking at the other failing builds, many of them are failing on the same Cabal "depends on text-1.2.1.1 which failed to install" errors. For example: https://travis-ci.org/apache/thrift/jobs/70311882. Is someone investigating the cause of that problem?

jfarrell:
hey @marshalium, i'm working on cutting the 0.9.3 rc right now and using the docker containers in build/docker/ to do so. Travis is working to enable docker support and we are in the beta for this, once 0.9.3 is cut i'll be working to switch us over to that, which should help make the CI runs more consistent across environments

marshalium:
Hey @jfarrell. Are things better with CI now? Should I rebase off of master and try again? Thanks!

jfarrell:
@marshalium I cut 0.9.3 off master, so everything in master or on the 0.9.3 branch should work for your testing

PR #513, opened 2015-06-06
THRIFT-3286: Apache Ant is a necessary dependency

user640378:
adding to apt-get command

### DIFF
--- a/doc/install/debian.md
+++ b/doc/install/debian.md
@@ -1,7 +1,7 @@
 ## Debian or Ubuntu setup
 The following command install all the required tools and libraries to build and install the Apache Thrift compiler on a Debian/Ubuntu Linux based system.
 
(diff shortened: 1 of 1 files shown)


Jens-G:
Hi user<some_random_number>,

thanks for the patch. Unfortunately, and as little the changes are, for a number of reasons we have certain (very minimal) requirements on how to contribute. If you just could [file a JIRA ticket](http://thrift.apache.org/docs/HowToContribute) for the issue (or pick an matching existing one if there is any) and link this PR to it by mentioning the JIRA ticket number somewhere? 

That would be really awesome!

Thanks!

user640378:
Hi Jens,

Sure, no problem! I'll do that. Since it was a minor change I had thought jira might not be necessary. I'll do it now.

PR #516, opened 2015-06-09
Add UNIX Domain Socket support

saleyn:
Non-blocking server support for listening on unix domain sockets (UDS).

Add UDS support to Erlang client

### DIFF
--- a/contrib/fb303/cpp/Makefile.am
+++ b/contrib/fb303/cpp/Makefile.am
@@ -38,7 +38,8 @@ endif
 
 AM_CPPFLAGS = -I..
 AM_CPPFLAGS += -Igen-cpp
-AM_CPPFLAGS += -I$(thrift_home)/include/thrift 
+AM_CPPFLAGS += -I$(thrift_home)/include
+AM_CPPFLAGS += -I../../../lib/cpp/src
 AM_CPPFLAGS += $(BOOST_CPPFLAGS)
 AM_CPPFLAGS += $(FB_CPPFLAGS) $(DEBUG_CPPFLAGS)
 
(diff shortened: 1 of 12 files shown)


jfarrell:
hey @saleyn, thanks for the patch. We try to keep our dependencies to a minimum to make it easier for the client libraries to be integrated into a wide variety of projects. One quick thing i see with this patch is that euds does not have a license, which makes it a little harder for us to even look at integrating with Apache Thrift. Can the same result be achieved using native support without too much additional lifting?

saleyn:
I will add a license file.  Would Apache 2.0 license do?

jfarrell:
Thanks, I like that you have submitted euds upstream, would be nice to not have that as a dependency and just use it natively. Is there any traction on that pull request to OTP?

saleyn:
The latest I heard it was put on the Raimo's (from the OTP team) plate of things to do.  Hopefully he'll manage to get it done in the next several months.  It's a shame OTP is missing native support for UDS - likely a design oversight, but I am certain it'll get done as people express strong support for addressing the issue (through my patch or its derivative).

jeking3:
@jfarrell recommend closing this PR.  It never got a thrift ticket, and has an external dependency to a library that has been discontinued:
```
UPDATE: this project will be deprecated beginning with Erlang 19RC2, which will include native support for Unix Domain Sockets.
```

Any implementation of unix sockets for erlang would need to be re-done at this point.  I opened a wish ticket in Jira for this: THRIFT-4169

saleyn:
If I get spare time for this I'll migrate this to using the native Unix Domain Sockets support in Erlang.  Though I no longer use Thrift, and currently for me this is a very low priority.

PR #517, opened 2015-06-11
Always set response on THttpClient

i:
Due to an early return, not setting response leads to a file descriptor leak. This fixes that by letting the client close its transport every time.


prashantv:
@i Just started a PR for the same issue, didn't see your fix.
I think it's better to not set p.Response, since it affects other code (IsOpen returns true, Read will allow reading bytes which most likely won't match the protocol, etc). I think it might be better to just close the Body on an error: #518

Jens-G:
I agree with @prashantv and just committed the other patch. If you do not share our opinion please feel free to rebase and improve the patch, otherwise you can probably close the PR. Either way, thanks for pointing this out, looking forward to your next contributions ;-)

PR #519, opened 2015-06-11
THRIFT-3122 Convert plain js objects given as thrift struct constructor arguments to thrift structs

itkach:


itkach:
Any objections to merging these changes?

PR #527, opened 2015-06-24
THRIFT-3053: Added perl SSL Socket support, split SSLSocket and SSLSe…

jeking3:
…rverSocket out from their base classes, fixed THRIFT-3191 generated perl compiler exception handling code, added perl to make cross, fixed THRIFT-3189 allowing perl to listen on a specific interface through construction arguments.   Did not add support in the perl client SSLSocket to verify server certificate authenticity at this time.


jeking3:
The "make cross" omitted perl from the cross check, likely because of:
checking for perl... /usr/bin/perl
checking for perl module Bit::Vector... no
checking for perl module Class::Accessor... no

Sadly, the travis build succeeded even though it should have failed because perl wasn't tested.  I updated the perl deps and the perl doc (README.md) and we'll see if it gets things right this time around.  If not, I'll look into why those are failing.

jeking3:
I need some help understanding why this is failing in the Travis CI environment when it passes locally.  The "-ssl" tests that involve Perl are all failing.  The non-ssl ones involving perl that are failing like nodejs are not ones I've run locally so they may need a suppression which I can add.

I still don't like that I cannot see the results of each "make cross" test in Travis CI... can we fix that?

jeking3:
SUCCESS: Integrated in Thrift #1629 so I am closing.

PR #528, opened 2015-06-25
THRIFT-406 Convert C++ Test to Boost::Test

cmhe:
This converts all C++ test to Boost::Test, integrates the DenseProtocolTest as a test like the others and removes C asserts from the TMemoryBufferTest.

Two test cases from DenseProtocolTest have to be commented, because they use the private vlqRead and vlqWrite Methods of apache::thrift::protocol::TDenseProtocol.


cmhe:
Merged in commit:
https://github.com/apache/thrift/commit/5ef662b9046d18aee224018061413ad7ee5d80a5

PR #529, opened 2015-06-26
[THRIFT-3204] SimpleServer processRequests method nerver handle some errors

wangjild:
SimpleServer processRequests method nerver handle some errors which is not  thrift.TTransportException type

https://issues.apache.org/jira/browse/THRIFT-3204


jeking3:
@jfarrell Recommend we close this as won't fix.  It never passed and the original author's github repository isn't around any more.

PR #532, opened 2015-06-29
THRIFT-3211: add php client compact protocol to make cross test

jeking3:
Note compact passes but json fails; tracking that failure in a related defect in Jira.


jeking3:
Looks like go server, php client cannot talk TCompactProtocol however php was able to talk to all other servers.

nsuke:
Maybe you want to run
```
test/test.py --update-expected-failures=merge
```
to add them to known failures as documented [here](https://github.com/apache/thrift/blob/master/test/README.md#list-of-known-failures).

jeking3:
I removed all PHP make cross exclusions except for "go-php" because they all fail.  "go" in general seems to have some fundamental issues playing nice with the other implementations.  There are a total of 156 test exclusions involving "go".

jeking3:
Given the rather large set of known failing tests, we should have Jira items tracking each language to language test failure.  I'm concerned these are somewhat accepted as normal rather than issues to resolve.

test/tests.json
```diff
@@ -332,7 +332,9 @@
         "ip"
       ],
       "protocols": [
-        "binary"
+        "binary",
+        "compact",
+        "binary:accel"
```
jeking3: Note: JSON is broken against every server, should get a Jira ticket.

jeking3:
The "make cross" travis job is passing now, however another job failed for reasons I cannot tell, so I merged with master to see if this was something that someone else fixed.

PR #535, opened 2015-07-01
THRIFT-3214 Add Erlang option for using maps instead of dicts

thecodeboss:
Erlang OTP 18 introduces significant performance improvements when using maps instead of dicts. Here is a benchmark showing the gains:

https://gist.github.com/BinaryMuse/bb9f2cbf692e6cfa4841

Thrift should support an option for the Erlang compiler to produce maps instead of dicts, to take advantage of these performance improvements.

### DIFF
--- a/compiler/cpp/src/generate/t_erl_generator.cc
+++ b/compiler/cpp/src/generate/t_erl_generator.cc
@@ -55,6 +55,7 @@ public:
     out_dir_base_ = "gen-erl";
 
     legacy_names_ = (parsed_options.find("legacynames") != parsed_options.end());
+    maps_ = (parsed_options.find("maps") != parsed_options.end());
   }
 
   /**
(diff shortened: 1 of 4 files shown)


compiler/cpp/src/generate/t_erl_generator.cc
```diff
@@ -55,6 +55,7 @@ class t_erl_generator : public t_generator {
     out_dir_base_ = "gen-erl";
 
     legacy_names_ = (parsed_options.find("legacynames") != parsed_options.end());
+    use_maps_ = (parsed_options.find("usemaps") != parsed_options.end());
```
jparise: I think `maps` would be a more consistent option name.  Similar existing options are `slots` (for Python) and `templates` (for C++), which don't include a **use** prefix.

thecodeboss: Great catch @jparise, the PR has been modified with your suggestion.

PR #538, opened 2015-07-03
THRIFT-3219: Provide a C++ tutorial on server-side IP logging and per-connection state

ben-craig:
This also adds the C++ tutorial builds to the cmake build by default.

The tutorial demonstrates per-connection state as a side effect.


CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ endif()
 
 if(BUILD_CPP)
     add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/lib/cpp)
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/tutorial/cpp)
     if(BUILD_TESTING)
```
bufferoverflow: we probably need to add a BUILD_TUTORIALS option in the future

tutorial/cpp/CppLogIPServer.cpp
```diff
@@ -0,0 +1,138 @@
...
+    return new CalculatorHandler;
+  }
+  virtual void releaseHandler( ::shared::SharedServiceIf* handler) {
+    delete handler;
+  }
+};
+
+int main() {
+  TThreadedServer server(
+    boost::make_shared<CalculatorProcessorFactory>(boost::make_shared<CalculatorCloneFactory>()),
```
bufferoverflow: I would say we can add this to the regular Tutorial server and add some

ben-craig: Ok, combined the old and the new tutorials in the latest commit.

bufferoverflow:
Thanks Ben!
Could you commit this?

PR #545, opened 2015-07-08
THRIFT-3230: transform typedef when getting type name

lifei:
Python compiler generates wrong code if there is function throwing a typedef of exception with another namespace. Use the real type name instead of  the name of typedef.


lifei:
https://issues.apache.org/jira/browse/THRIFT-3230

lifei:
https://github.com/apache/thrift/commit/01a77ab01e7459d96059a2b49d9885d14a360ef1

PR #550, opened 2015-07-14
THRIFT-3243: Fix connection leak in THttpClient

abhinav:


abhinav:
Actually, there's a major problem in that. Fix upcoming.

abhinav:
This isn't the right way to do this

PR #555, opened 2015-07-16
(CSharp) Adding lifecycle management for service handlers through TProcessorFactory

diogojc:
I want to be able to control the lifecycle of the instances implementing my thrift service handlers (TProcessor).

Before this change the behaviour for every Thrift server is that a service handler will always be the same instance throughout all requests and therefore would either need to be stateless or require thread safety.

I've replaced the TProcessor parameter in the constructors of TServer and subclasses, with a TProcessorFactory so this decision is postponed to the person creating the object graph by specifying a factory for the processors.

Please help me in figuring out what supporting things I need to change to reflect this that I haven't done yet.

Diogo.

### DIFF
--- a/lib/csharp/Makefile.am
+++ b/lib/csharp/Makefile.am
@@ -38,6 +38,8 @@ THRIFTCODE= \
             src/Protocol/TProtocolDecorator.cs \
             src/Protocol/TMultiplexedProtocol.cs \
             src/Protocol/TMultiplexedProcessor.cs \
+            src/Protocol/TProcessorFactory.cs \
+            src/Protocol/TProcessorSingletonFactory.cs \
             src/Protocol/TType.cs \
             src/Protocol/TField.cs \
             src/Protocol/TMessage.cs \


Jens-G:
Nice idea. I'll take a look at it. 

In the meantime, you may want to read our [contribution Guidelines](http://thrift.apache.org/docs/HowToContribute), if you did not do that already.

Jens-G:
I think that's a good addition. But ... I'd like to have it even less breaking. Could you add some more CTORs with the **old** signature? Internally these would just wrap the passed processor into a singleton factory and then call the appropriate new CTOR version. What do you think?

This would automatoically resolve all issues with the tests and the tutorial code that we have right now.

And oh yes, [please create a JIRA issue](http://thrift.apache.org/docs/HowToContribute) for it.

Jens-G:
Hi @diogojc, any news here?

jeking3:
@jfarrell recommend closing this PR.  It was abandoned; no Jira ticket; author's fork has been destroyed.

jeking3:

PR #556, opened 2015-07-21
Add http transport for server in Golang

hhkbp2:
Hi,

This PR adds a http transport for server in Golang. It contains http protocol parsing codes, which works with the go/c++ client side http related logic.

PR #557, opened 2015-07-21
Thrift 3251 Add http transport for server to Go lib

hhkbp2:
Hi,

This PR is a reopen for previous PR #556, in order to follow the contribute guide.

It adds http transport for server to Go lib.
For more info please refer to Jira
[https://issues.apache.org/jira/browse/THRIFT-3251](https://issues.apache.org/jira/browse/THRIFT-3251)

### DIFF
--- a/lib/go/thrift/http_client.go
+++ b/lib/go/thrift/http_client.go
@@ -21,6 +21,7 @@ package thrift
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -177,6 +178,7 @@ func (p *THttpClient) Flush() error {
 		return NewTTransportExceptionFromError(err)
 	}
 	p.header.Add("Content-Type", "application/x-thrift")
+	p.header.Add("Content-Length", fmt.Sprintf("%d", p.requestBuffer.Len()))
 	req.Header = p.header
 	response, err := client.Do(req)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


dimiro1:
I implemented with a handler func. https://gist.github.com/dimiro1/d38aaf4b7d84c2d5c8f0

Jens-G:
@dimiro1  could you please make a PR out of it?

dimiro1:
@Jens-G The current test infrastructure for the go lib explicitly assume that the server is a SocketServer /test/go/src/server.go. I think it will be a lot of work write the test to conform with the current structure. In my gist [https://gist.github.com/dimiro1/d38aaf4b7d84c2d5c8f0] I wrote a simple test based on a simple Calculator client. What do you think about this approach? Do have any other idea? I am open to suggestions.

Jens-G:
All languages are encouraged 1) to implement the standard Thrift test. This test is the basic test to ensure interoperability across languages for all features implemented. So if we have a http server, that thing should be testable. The [command line syntax is documented](http://thrift.apache.org/test/#test-executable-specification) and works the same way for all languages.

So yes, it would be good to have this integrated into the ThriftTest server and controlled via the ```--transport http(s)``` parameter. The client already supports it, [the server throws an error right now](https://github.com/apache/thrift/blob/master/test/go/src/common/server.go#L92).
____
1) one could also say "required" and it would not be wrong

PR #560, opened 2015-07-24
THRIFT-3263: PHP Compiler always casts scalar types in jsonSerialize()

stigsb:
In PHP, values from databases and such are usually always strings, so it is useful to cast values during JSON serialization to maintain the Thrift schema's types.

THRIFT-3263


jeking3:
@jfarrell this can be closed.  It was merged.

PR #562, opened 2015-07-24
Thrift 3265--servers for pipe applications (stdin, stdout)

emma58:
Adds server classes to C#, Python, and Java libraries for server/client communication over stdin and stdout.

### DIFF
--- a/lib/java/Makefile.am
+++ b/lib/java/Makefile.am
@@ -37,7 +37,7 @@ precross: $(THRIFT)
 	$(ANT) $(ANT_FLAGS) compile-test
 
 check-local: all
-	$(ANT) $(ANT_FLAGS) -debug test
+	$(ANT) $(ANT_FLAGS) test
 
 EXTRA_DIST = \
 	build.xml \


Jens-G:
Hi, great Addition, but please clean it up before I take a look at it. For starters, all those "generated files" are not to be included.

lib/csharp/test/TestIOStreamServer/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,55 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("TestIOStreamServer")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Sandia National Laboratories")]
+[assembly: AssemblyProduct("TestIOStreamServer")]
+[assembly: AssemblyCopyright("Copyright © Sandia National Laboratories 2015")]
```
Jens-G: Whats that copyright notice? Are you sure your are legally allowed to put that code into Open Source at all? How can we check that?

lib/csharp/test/TestIOStreamServer/TestIOStreamClient.cs
```diff
@@ -0,0 +1,425 @@
...
+            Assert.AreEqual(setout.Count, setin.Count);
+            foreach (int i in setout)
+            {
+                Assert.IsTrue(setin.Contains(i));
+            }
+        }
+
+        [TestMethod]
+        public void testEnum()
+        {
```
Jens-G: I'd rather like the test being included into the existing standard ThriftTest test client, instead of writing a duplicated one. Having two of them does not make much sense, do you agree?

lib/java/test/org/apache/thrift/test/JavaBeansTest.java
```diff
@@ -29,33 +29,32 @@ public static void main(String[] args) throws Exception {
     // Test isSet methods
```
nsuke: Included this file by mistake ? Only indentations are messed.

lib/csharp/test/TestIOStreamServer/Properties/AssemblyInfo.cs
emma58: That is a good catch... I am not sure that we can contribute it, and I am working to find out. I will probably take it off github for the time being because I should have thought of this sooner!

PR #569, opened 2015-07-28
THRIFT-3164: disable insecure SSLv3 protocol in cpp lib unless asked for

jeking3:
unless SSLv3 is specifically asked for this will make the default more secure; added a matrix security unit test for SSL in the cpp lib


jeking3:
Merged, closing.

PR #575, opened 2015-08-01
c_glib: Compiler: Omit processing-function defs for empty services

simonsouth:
With this change the compiler omits the processing-function-definition array for a service with no methods, corresponding to the checks on lines 2555 and 2562 that omit the code that would use this array under the same condition.


jfarrell:
Committed as part of THRIFT-3284

PR #577, opened 2015-08-03
THRIFT-3288 c glib build unit tests with all warnings enabled

simonsouth:
**Note:** This pull request includes the changes from request #576 (to fix THRIFT-3285), which should be applied first.

These changes
- Update the C (GLib) unit tests' Makefile so
  - Compiler-flag variables are used correctly according to automake's documentation and
  - The same comprehensive set of compiler warnings is enabled as is used for the C++ unit tests,
- Update the compiler for C (GLib) so the code it generates builds without any compiler warnings, and
- Update the unit tests themselves so they build without warnings.

As a bonus, the final commit gently refactors several compiler functions changed as part of this work, primarily to improve readability and to more consistently use helper functions for formatting rather than embed formatting directly in the output.

### DIFF
--- a/compiler/cpp/src/generate/t_c_glib_generator.cc
+++ b/compiler/cpp/src/generate/t_c_glib_generator.cc
@@ -968,25 +968,38 @@ void t_c_glib_generator::generate_const_initializer(string name,
         }
       }
       if (field_type == NULL) {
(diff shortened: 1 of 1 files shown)


jeking3:
@jfarrell this is marked closed fixed in Jira.  This PR can be closed.

PR #579, opened 2015-08-05
THRIFT-3290: Making from as a reserved language keyword

prathik:


jfarrell:
Hi @prathik, Thanks for the patch. would you mind adding an issue for this on our jira so we can better track it. Details on our process are available here http://thrift.apache.org/docs/HowToContribute

prathik:
Hi @jfarrell, Thank you for pointing that out. Here is the issue ticket https://issues.apache.org/jira/browse/THRIFT-3290.

Thanks,
Prathik

PR #585, opened 2015-08-14
THRIFT-3294 Fix buffer overrun in Java implementation of Zlib transport

Magrath:
The TZlibTransport did not check the write buffer length when flushing. Instead, it assumed that the entire write buffer byte array was the compressed input.

The unit test is based on TestTFramedTransport and verifies the fix. 

https://issues.apache.org/jira/browse/THRIFT-3294


Magrath:
Closing as this has now been merged by @RandyAbernethy https://github.com/apache/thrift/commit/bb98e97fd3c82117c87d23e3fb6b8bbd800784f2 (thanks!)

PR #587, opened 2015-08-16
THRIFT-3296 - Go cross test does not conform to spec

nsuke:


nsuke:
merged as 2fab3deb09e492cbcb2b1e4ed1c1d302787a9abd
Thanks @henrique !

PR #588, opened 2015-08-18
0.9.2

akash3456:
hi,

i seem to be getting a Class 'Thrift\Exception\TException' not found in Types.php, however all the main thrift are not in the root thrift directory?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,117 @@
 Thrift Changelog
 
+Version 0.8.0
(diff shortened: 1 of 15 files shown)


jfarrell:
hi @akash3456, not sure what the problem is here, would you mind opening a ticket for your issue or emailing our dev@thrift.apache.org mailing list with the issue you are seeing

PR #589, opened 2015-08-18
THRIFT-3324 - Update `go get` command to include all packages

zzmp:


kostya-sh:
Why? There is only one package: `git.apache.org/thrift.git/lib/go/thrift`

PR #590, opened 2015-08-19
THRIFT-3300 Reimplement TZlibTransport in Java using streams

Magrath:
https://issues.apache.org/jira/browse/THRIFT-3300


Magrath:
Thanks @RandyAbernethy!

PR #592, opened 2015-08-23
THRIFT-3303 Disable concurrent cabal jobs on Travis to avoid GHC crush

nsuke:


nsuke:
Although an unrelated cross test case failed, the build in question "all(automake)" succeeded:
https://travis-ci.org/apache/thrift/jobs/76861567

See also build for identical commit:
https://travis-ci.org/nsuke/thrift/builds/76861417

PR #596, opened 2015-08-25
THRIFT-3306: Java: TBinaryProtocol: Use a single temp byte[] buffer

evanj:
Reduces allocations and GC when TBinaryProtocol is allocated repeatedly.

Originally separate byte[] arrays were allocated for each integer type, and
a separate array for input and output. However, one buffer of the largest
size (8 bytes for i64) can be used for all purposes.


evanj:
Note: the Travis CI build may have failed, but it failed to compile some parts of the C++ code, so I suspect this is due to a flaky test and is not legitimate, since I didn't touch any of the C++ stuff.

PR #598, opened 2015-09-02
cmake // fixes for std/boost::thread detection; security test; qt test

snikulov:
Hello All!

Relates to https://issues.apache.org/jira/browse/THRIFT-2850

1. Will use BOOST if not explicitly provided -DWITH_STDTHREADS=ON and Boost_FOUND.
2. Will not build SecurityTest if NOT OPENSSL_FOUND (SecurityTest uses SSL headers which is not available)
3. Fix for TQTcpServerTest_Qt5.

Build successful on Windows (MSVC 2013) and CentOS 7

### DIFF
--- a/build/cmake/DefineOptions.cmake
+++ b/build/cmake/DefineOptions.cmake
@@ -62,8 +62,10 @@ endif()
 find_package(OpenSSL QUIET)
 CMAKE_DEPENDENT_OPTION(WITH_OPENSSL "Build with OpenSSL support" ON
                        "OPENSSL_FOUND" OFF)
-option(WITH_BOOSTTHREADS "Build with Boost thread support" OFF)
 option(WITH_STDTHREADS "Build with C++ std::thread support" OFF)
+CMAKE_DEPENDENT_OPTION(WITH_BOOSTTHREADS "Build with Boost threads support" OFF
+    "NOT WITH_STDTHREADS;Boost_FOUND" OFF)
+
 
 # C GLib
 option(WITH_C_GLIB "Build C (GLib) Thrift library" ON)
--- a/lib/cpp/test/qt/CMakeLists.txt
+++ b/lib/cpp/test/qt/CMakeLists.txt
@@ -18,11 +18,15 @@
 #
 
 set(CMAKE_AUTOMOC ON)
-find_package(Qt5 REQUIRED COMPONENTS Test)
+find_package(Qt5 REQUIRED COMPONENTS Test Network)
 set(TQTcpServerTest_Qt5_SOURCES
     TQTcpServerTest.cpp
 )
 add_executable(TQTcpServerTest_Qt5 ${TQTcpServerTest_Qt5_SOURCES})
-target_link_libraries(TQTcpServerTest_Qt5 testgencpp_cob thriftqt5 thrift Qt5::Test)
+target_link_libraries(TQTcpServerTest_Qt5 testgencpp_cob)
+LINK_AGAINST_THRIFT_LIBRARY(TQTcpServerTest_Qt5 thriftqt5)
+LINK_AGAINST_THRIFT_LIBRARY(TQTcpServerTest_Qt5 thrift)
+target_link_libraries(TQTcpServerTest_Qt5 Qt5::Test Qt5::Network)
+
 add_test(NAME TQTcpServerTest_Qt5 COMMAND TQTcpServerTest_Qt5)
 
--- a/lib/cpp/test/CMakeLists.txt
+++ b/lib/cpp/test/CMakeLists.txt
@@ -117,17 +117,6 @@ target_link_libraries(TServerIntegrationTest -lrt)
 endif ()
 add_test(NAME TServerIntegrationTest COMMAND TServerIntegrationTest)
 
-add_executable(SecurityTest SecurityTest.cpp)
-target_link_libraries(SecurityTest
-    testgencpp
-    ${Boost_LIBRARIES}
-)
-LINK_AGAINST_THRIFT_LIBRARY(SecurityTest thrift)
-if (NOT MSVC AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-target_link_libraries(SecurityTest -lrt)
-endif ()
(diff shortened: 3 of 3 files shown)


build/cmake/DefineOptions.cmake
```diff
@@ -62,8 +62,10 @@ endif()
 find_package(OpenSSL QUIET)
 CMAKE_DEPENDENT_OPTION(WITH_OPENSSL "Build with OpenSSL support" ON
                        "OPENSSL_FOUND" OFF)
-option(WITH_BOOSTTHREADS "Build with Boost thread support" OFF)
 option(WITH_STDTHREADS "Build with C++ std::thread support" OFF)
+CMAKE_DEPENDENT_OPTION(WITH_BOOSTTHREADS "Build with Boost threads support" ON
```
nsuke: Looks like a good idea for Windows.

For *nix, pthread has been the default, and IMO it should remain so because:
* *configure.ac* option states boost thread is experimental.
* In my experience, Thrift pthread implementation has better performance (although negligible for most cases) than that of boost or std-thread on Linux.

lib/cpp/test/qt/CMakeLists.txt
```diff
@@ -18,11 +18,15 @@
 #
 
 set(CMAKE_AUTOMOC ON)
-find_package(Qt5 REQUIRED COMPONENTS Test)
+find_package(Qt5 REQUIRED COMPONENTS Test Network)
```
nsuke: Network module should be already found either by non-test part of cpp/CMakeLists.txt or qt5/CMakeLists.txt before reaching here.
I'm not particularly against being explicit if it's your intention, though.

lib/cpp/test/qt/CMakeLists.txt
```diff
@@ -18,11 +18,15 @@
...
 set(TQTcpServerTest_Qt5_SOURCES
     TQTcpServerTest.cpp
 )
 add_executable(TQTcpServerTest_Qt5 ${TQTcpServerTest_Qt5_SOURCES})
-target_link_libraries(TQTcpServerTest_Qt5 testgencpp_cob thriftqt5 thrift Qt5::Test)
+target_link_libraries(TQTcpServerTest_Qt5 testgencpp_cob)
+LINK_AGAINST_THRIFT_LIBRARY(TQTcpServerTest_Qt5 thriftqt5)
+LINK_AGAINST_THRIFT_LIBRARY(TQTcpServerTest_Qt5 thrift)
+qt5_use_modules(TQTcpServerTest_Qt5 Test Network)
```
nsuke: According to [Qt5 documentation](http://doc.qt.io/qt-5/cmake-manual.html), *qt5_use_modules* is the way to go with CMake **prior to 2.8.11**.
Since root CMakeLists.txt requires 2.8.12, I believe imported targets (*Qt5::Test Qt5::Network*) are more appropriate than the macro here.

lib/cpp/test/qt/CMakeLists.txt
```diff
@@ -18,11 +18,15 @@
...
-find_package(Qt5 REQUIRED COMPONENTS Test)
+find_package(Qt5 REQUIRED COMPONENTS Test Network)
 set(TQTcpServerTest_Qt5_SOURCES
     TQTcpServerTest.cpp
 )
 add_executable(TQTcpServerTest_Qt5 ${TQTcpServerTest_Qt5_SOURCES})
-target_link_libraries(TQTcpServerTest_Qt5 testgencpp_cob thriftqt5 thrift Qt5::Test)
+target_link_libraries(TQTcpServerTest_Qt5 testgencpp_cob)
+LINK_AGAINST_THRIFT_LIBRARY(TQTcpServerTest_Qt5 thriftqt5)
+LINK_AGAINST_THRIFT_LIBRARY(TQTcpServerTest_Qt5 thrift)
```
nsuke: +1
You're absolutely right for using LINK_AGAINST... macro (I think I wrote the original line).
These should fix static lib build at the very least.

build/cmake/DefineOptions.cmake
snikulov: @nsuke  Well, then we can change in following way:

CMAKE_DEPENDENT_OPTION(WITH_BOOSTTHREADS "Build with Boost threads support" ON
-    "NOT WITH_STDTHREADS;Boost_FOUND" OFF)
+    "NOT WITH_STDTHREADS;Boost_FOUND;WITH_BOOSTTHREADS" OFF)

Then, this feature will not be enabled if not explicitly set to on using -DWITH_BOOSTTHREADS=ON in command line.

What do you think?

BTW, I've found that this option will set add_definitions, but not update config.h
Is it normal?

lib/cpp/test/qt/CMakeLists.txt
snikulov: @nsuke Yep. But this was related to my laziness :)
qt5_use_modules also set includes. Error was related to include issue and LINK_AGAINST...

build/cmake/DefineOptions.cmake
nsuke: According to http://www.cmake.org/cmake/help/v3.0/module/CMakeDependentOption.html

> This macro presents an option to the user only if a set of other conditions are true. 

This would hide the option itself unless the user set it in the first place, like a chicken out of no egg for GUI users.

How about literally changing the default value only:

    CMAKE_DEPENDENT_OPTION(WITH_BOOSTTHREADS "Build with Boost threads support" OFF
         "NOT WITH_STDTHREADS;Boost_FOUND" OFF)


About the config.h, I think your concern is right.
Long story short, it originates from my past patch, and in its current state I don't see any reason not to replace *add_definitions* with *set*.

lib/cpp/test/qt/CMakeLists.txt
nsuke: The imported target *Qt5::Network* should handle includes for you too.
It automatically propagates *INTERFACE_INCLUDE_DIRECTORIES* and others.

snikulov:
@nsuke updated pull request per your comments.

nsuke:
@snikulov sorry if there was any misunderstanding but I'm just reviewing as a peer contributor and I have no way to "approve."
That said, I believe it's good to merge.

@bufferoverflow and others, could you help with this ?
The patch looks good and Travis **CMake** builds succeeded.

PR #601, opened 2015-09-07
THRIFT-557 charset problem with file Autogenerated by Thrift

collinmsn:
to fix the issue https://issues.apache.org/jira/browse/THRIFT-557


Jens-G:
Correct me if I'm wrong, but isn't the "magic comment" [expected to have a very specific format](https://www.python.org/dev/peps/pep-0263/)? The patch seems not to provide that ...

collinmsn:
Do you mean the comment this commit added is not any form of the following?
# coding=<encoding name>
# -*- coding: <encoding name> -*-

PR #602, opened 2015-09-07
THRIFT-3318 - Implement TSimpleJSONProtocol for PHP

edvakf:
I've implemented a SimpleJSON writer for PHP. 

I made it similar to TJSONProtocol.php and TSimpleJSONProtocol.java so it should comply with the coding standard.


edvakf:
I don't think the test failure is due to my change.

```
make[4]: Entering directory `/home/travis/build/apache/thrift/lib/cpp/test'
make[4]: *** No rule to make target `tests.xml', needed by `distdir'.  Stop.
```

edvakf:
Jira issue here https://issues.apache.org/jira/browse/THRIFT-3318

PR #603, opened 2015-09-11
there is no 'What' in InvalidOperation, maybe is 'WhatOp'

youtsea:


Jens-G:
Hi, could you please file a JIRA ticket [as described in our contrib guide](http://thrift.apache.org/docs/HowToContribute)? Thanks!

PR #606, opened 2015-09-16
THRIFT-2918 Race condition in Python TProcessPoolServer test

bufferoverflow:
This makes python test pass on my machine.


bufferoverflow:
committed

PR #608, opened 2015-09-17
THRIFT-3299 Dart language bindings in Thrift

stevenosborne-wf:
Description: Add bindings for [dart](https://www.dartlang.org/).

Ticket: https://issues.apache.org/jira/browse/THRIFT-3299


trentgrover-wf:
@evanweible-wf FYI - it would be awesome to get w_transport in as the transport layer for this so that client and server APIs can match up 1-to-1

Jens-G:
Hi, some comments.

* Version numbers should match the Thrift version number.
* They also need to be  mentioned in the "how to release" docs, see website, but that can be done later.
* All (textual) files must have an ASF license header. This also includes shell scripts etc.

markerickson-wf:
@Jens-G I'll update the versions of static Dart libraries to `0.9.3`.  But for generated Dart libraries, such as what you're seeing in `tutorial 0.0.1`, it is really just an arbitrary placeholder, which developers would modify to match their own versioning scheme for a service.  If there is a way to define a service version in a `.thrift` definition file, then I could use that, which would be ideal.

Jens-G:
I see, didn't recognize that the second one was generated when I wrote that. I saw it later in the compiler code. I think that's fine. Lib should be at whatever Thrift version is (the compiler tells you when called with -v) and as I said we need to document this [here](https://thrift.apache.org/docs/committers/HowToVersion).

Thanks for the good, keep pushing!
Von: Mark Erickson
Gesendet: 24.09.2015 04:39
An: apache/thrift
Cc: Jens Geyer
Betreff: Re: [thrift] THRIFT-3299 Dart language bindings in Thrift (#608)

@Jens-G I'll update the versions of static Dart libraries to `0.9.3`.  But for generated Dart libraries, such as what you're seeing in `tutorial 0.0.1`, it is really just an arbitrary placeholder, which developers would modify to match their own versioning scheme for a service.  If there is a way to define a service version in a `.thrift` definition file, then I could use that, which would be ideal.

---

Jens-G:
No, there is no such mechanism as a version number for services. Maybe IDL annotations could be used for this, never tried that.

http://stackoverflow.com/questions/25631235/what-is-an-annotation-in-apache-thrift-and-what-is-it-used-for


Von: Mark Erickson
Gesendet: 24.09.2015 04:39
An: apache/thrift
Cc: Jens Geyer
Betreff: Re: [thrift] THRIFT-3299 Dart language bindings in Thrift (#608)

@Jens-G I'll update the versions of static Dart libraries to `0.9.3`.  But for generated Dart libraries, such as what you're seeing in `tutorial 0.0.1`, it is really just an arbitrary placeholder, which developers would modify to match their own versioning scheme for a service.  If there is a way to define a service version in a `.thrift` definition file, then I could use that, which would be ideal.

---

markerickson-wf:
@Jens-G I think I have addressed your comments now.  Automake files are in, cross tests are passing with a Dart client and Python server, versions and license headers are updated.  I tried to identify the files that will need to be updated on release here - https://github.com/apache/thrift/pull/608/files#diff-d9d9844c5f034229628ba1f2e341ce6bR44.

PR #617, opened 2015-09-22
fix PHP TCompactProtocol Int64 trans error when number less then zero

Saerdna:
when I use TCompactProtocol in PHP client, Int64 less then zero convert error
server is Go 
some case:
        -1 convert to 1.844674407371E+19


nsuke:
Hello, please create a JIRA issue for it.

I think it is not a right solution because it would break many other cases like values larger than 2^32.

Here's the test failure log on CI
https://travis-ci.org/apache/thrift/jobs/81515399#L6011

PR #620, opened 2015-09-22
THRIFT-3329: fix test incompatibility with boost-1.59

jeking3:


lib/cpp/test/SecurityTest.cpp
```diff
@@ -75,13 +75,17 @@ struct GlobalFixture
             if (!boost::filesystem::exists(certFile("server.crt")))
             {
                 keyDir = boost::filesystem::path(master_test_suite().argv[master_test_suite().argc - 1]);
-                BOOST_REQUIRE_MESSAGE(boost::filesystem::exists(certFile("server.crt")),
-                                      "The last argument to this test must be the directory containing the test certificate(s).");
+                if (!boost::filesystem::exists(certFile("server.crt")))
+                {
+                	std::cerr << "The last argument to this test must be the directory containing the test certificate(s)." << std::endl;
```
hcorg: there must be a better way to do it... BOOST_TEST_MESSAGE & BOOST_FAIL?

jeking3: Starting with boost-1.59 you cannot use assertion test macros outside a test scope.  This means you cannot use them in the global fixture.  It's a weird choice but that's why I made this change.

hcorg: damn, I was using "let's do some checks in fixture constructor" method quite offten... ok, another idea (cleaner than exit(1)) - just throw std::runtime_error with message. Boost has to deal with exceptions, this could make better error info etc.

bufferoverflow: I agree with @hcorg, there must be a better alternative from boost test to do this.
the other questions is backward compatibility, miminum version is 1.53.0 right now.

jeking3: Yeah, I was rushing through this yesterday, I'll change to exceptions.

lib/cpp/test/TransportTest.cpp
```diff
@@ -1016,7 +1019,7 @@ struct global_fixture {
   }
 };
 
-BOOST_GLOBAL_FIXTURE(global_fixture)
+BOOST_GLOBAL_FIXTURE(global_fixture);
```
hcorg: is this change necessary for boost 1.59? in previous versions it causes warning "extra ;"

jeking3: boost-1.59 rewrote boost::unit_test and they left the semicolon to the test implementer.  With clang-3.8 it won't compile without this.

hcorg: blah.... so another ifdef :) (I just hate warnings...)

PR #625, opened 2015-09-24
Fix word-extraction substr bug in initialism code

prashantv:
Initialisms were not handled in the second word or later. E.g. for the given Thrift file,
```thrift
service S {
  void get_id()
}
```

The generated code was using `GetIdArgs` instead of `GetIDArgs`. This was happening because fix_common_initialism was passing the position of the underscore to substr when extracting the word, when substr expects a length.


Jens-G:
Tracked in [THRIFT-3354](http://thrift.apache.org/docs/HowToContribute) - good catch.

PR #626, opened 2015-09-25
THRIFT-3336: eliminate extraneous ostream operators in generated code…

jeking3:
…; provide an overridable ostream implementation

### DIFF
--- a/compiler/cpp/src/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/generate/t_cpp_generator.cc
@@ -115,7 +115,8 @@ public:
                                    bool pointers = false,
                                    bool read = true,
                                    bool write = true,
-                                   bool swap = false);
+                                   bool swap = false,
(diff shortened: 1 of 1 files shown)


compiler/cpp/src/generate/t_cpp_generator.cc
```diff
@@ -1506,24 +1523,16 @@ void generate_fields(std::ofstream& out,
...
   out << " {" << endl;
 
   indent_up();
 
   out << indent() << "using apache::thrift::to_string;" << endl;
-
-  // eliminate compiler unused warning
-  const vector<t_field*>& fields = tstruct->get_members();
-  if (fields.empty())
-    out << indent() << "(void) obj;" << endl;
```
hcorg: why this line is removed? I'm pretty sure struct/exception without fields will generate warning "unused function parameter" for operator<</print method

PR #627, opened 2015-09-26
THRIFT-3357: enum_containers option to generate EnumSet/EnumMap

afds:
Additional java compiler option to generate EnumSet/EnumMap instead
of HashSet/HashMap where elements/keys are enums.


afds:
https://issues.apache.org/jira/browse/THRIFT-3357

compiler/cpp/src/generate/t_java_generator.cc
```diff
@@ -5139,8 +5215,11 @@ THRIFT_REGISTER_GENERATOR(
     "    reuse-objects:   Data objects will not be allocated, but existing instances will be used "
     "(read and write).\n"
     "    sorted_containers:\n"
-    "                     Use TreeSet/TreeMap instead of HashSet/HashMap as a implementation of "
+    "                     Use TreeSet/TreeMap instead of HashSet/HashMap as an implementation of "
     "set/map.\n"
+    "    enum_containers:\n"
```
Jens-G: Not so sure if we need a new option at all. The default assumptions for Thrift containers are described at https://thrift.apache.org/docs/types . If the patch does not break compatibility and on top is more efficient, why not make it the default? $0,02

afds:
I'm perfectly fine to make it default.

Theoretically possible issues is that someone heavily relies on HashMap/HashSet or uses _null_ keys/elements (EnumMap/EnumSet does not allow _null_ keys/elements). But in that case, we indeed, need to make it default to punish them.

afds:
Created a new pull request here: https://github.com/apache/thrift/pull/1253

PR #628, opened 2015-09-27
THRIFT-3437 - Fixed MinGW-w64 build && boost/functional conflict

becrux:
Fixed MinGW-w64 build
Fixed Boost - TR1 conflict

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -1004,6 +1004,10 @@ void TNonblockingServer::handleEvent(THRIFT_SOCKET fd, short which) {
  * Creates a socket to listen on and binds it to the local port.
  */
 void TNonblockingServer::createAndListenOnSocket() {
+#ifdef _WIN32
+  TWinsockSingleton::create();
+#endif // _WIN32
+
   THRIFT_SOCKET s;
 
   struct addrinfo hints, *res, *res0;
@@ -1109,10 +1113,17 @@ void TNonblockingServer::listenSocket(THRIFT_SOCKET s) {
   serverSocket_ = s;
 
   if (!port_) {
-    sockaddr_in addr;
-    socklen_t size = sizeof(addr);
-    if (!getsockname(serverSocket_, reinterpret_cast<sockaddr*>(&addr), &size)) {
-      listenPort_ = ntohs(addr.sin_port);
+    #ifdef _WIN32
+      struct sockaddr_storage addr;
+      struct sockaddr_in *pAddr = (sockaddr_in *) &addr;
+      socklen_t size = sizeof(addr);
+    #else
+      struct sockaddr_in addr;
+      struct sockaddr_in *pAddr = &addr;
+      socklen_t size = sizeof(addr);
+    #endif
+    if (!getsockname(serverSocket_, reinterpret_cast<sockaddr*>(pAddr), &size)) {
+      listenPort_ = ntohs(pAddr->sin_port);
     } else {
       GlobalOutput.perror("TNonblocking: failed to get listen port: ", THRIFT_GET_SOCKET_ERROR);
     }
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -69,6 +69,10 @@
 #define PRIu64 "I64u"
 #endif
 
+#if (_WIN32_WINNT < 0x0600)
+  #define AI_ADDRCONFIG 0x0400
+#endif
+
 namespace apache {
 namespace thrift {
 namespace server {
--- a/lib/cpp/src/thrift/windows/GetTimeOfDay.cpp
+++ b/lib/cpp/src/thrift/windows/GetTimeOfDay.cpp
@@ -23,17 +23,28 @@
 // win32
 #include <time.h>
 
+#if defined(__MINGW32__)
+  #include <sys/time.h>
+#endif
+
 #if defined(_MSC_VER) || defined(_MSC_EXTENSIONS)
 #define DELTA_EPOCH_IN_MICROSECS 11644473600000000Ui64
 #else
 #define DELTA_EPOCH_IN_MICROSECS 11644473600000000ULL
 #endif
 
+#if !defined(__MINGW32__)
 struct timezone {
   int tz_minuteswest; /* minutes W of Greenwich */
   int tz_dsttime;     /* type of dst correction */
 };
+#endif
 
+#if defined(__MINGW32__)
+int thrift_gettimeofday(struct timeval* tv, struct timezone* tz) {
+  return gettimeofday(tv,tz);
+}
+#else
 int thrift_gettimeofday(struct timeval* tv, struct timezone* tz) {
   FILETIME ft;
   unsigned __int64 tmpres(0);
@@ -79,6 +90,7 @@ int thrift_gettimeofday(struct timeval* tv, struct timezone* tz) {
 
   return 0;
 }
+#endif
 
 int thrift_sleep(unsigned int seconds) {
   ::Sleep(seconds * 1000);
--- a/lib/cpp/test/CMakeLists.txt
+++ b/lib/cpp/test/CMakeLists.txt
@@ -25,6 +25,10 @@ include_directories(SYSTEM "${Boost_INCLUDE_DIRS}")
 #Make sure gen-cpp files can be included
 include_directories("${CMAKE_CURRENT_BINARY_DIR}")
 
+if (MINGW)
+  add_definitions(-DBOOST_TEST_DYN_LINK)
+endif()
+
 # Create the thrift C++ test library
 set(testgencpp_SOURCES
     gen-cpp/DebugProtoTest_types.cpp
@@ -101,7 +105,7 @@ target_link_libraries(TInterruptTest
     ${Boost_LIBRARIES}
 )
 LINK_AGAINST_THRIFT_LIBRARY(TInterruptTest thrift)
-if (NOT MSVC AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+if (NOT MSVC AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT MINGW)
 target_link_libraries(TInterruptTest -lrt)
 endif ()
 add_test(NAME TInterruptTest COMMAND TInterruptTest)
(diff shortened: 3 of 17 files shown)


nsuke:
@becrux, thank you for your patch.
Do we have a JIRA ticket for this ?
If not please file one: https://github.com/apache/thrift/blob/master/CONTRIBUTING.md

becrux:
Thanks for the suggestion, @nsuke.

I've created the following issue on JIRA:

https://issues.apache.org/jira/browse/THRIFT-3437

Unfortunately, I was not able to "link" it to the pull request. Should I do something more?

Thanks,
A.

lib/cpp/test/qt/TQTcpServerTest.cpp
```diff
@@ -1,18 +1,23 @@
 #define BOOST_TEST_MODULE TQTcpServerTest
 #include <QTest>
-#include <boost/smart_ptr.hpp>
+
+#ifndef Q_MOC_RUN
+  #include <boost/smart_ptr.hpp>
```
nsuke: Please put it to #ifdef below as reordering seems OK here.

lib/cpp/test/CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ include_directories(SYSTEM "${Boost_INCLUDE_DIRS}")
 #Make sure gen-cpp files can be included
 include_directories("${CMAKE_CURRENT_BINARY_DIR}")
 
+if (MINGW)
+  add_definitions(-DBOOST_TEST_DYN_LINK)
+endif()
+
```
nsuke: I don't think `BOOST_TEST_DYN_LINK` is related to mingw.
You may want to pass it to cmake by `-DCMAKE_CXX_FLAGS="-D..."` option instead.

lib/cpp/test/SecurityTest.cpp
```diff
@@ -108,8 +108,13 @@ struct SecurityFixture : public TestPortFixture
...
             pServerSocketFactory.reset(new TSSLSocketFactory(static_cast<apache::thrift::transport::SSLProtocol>(protocol)));
             pServerSocketFactory->ciphers("ALL:!ADH:!LOW:!EXP:!MD5:@STRENGTH");
+          #ifdef BOOST_WINDOWS_API
+            pServerSocketFactory->loadCertificate(certFile("server.crt").string().c_str());
+            pServerSocketFactory->loadPrivateKey(certFile("server.key").string().c_str());
+          #else
             pServerSocketFactory->loadCertificate(certFile("server.crt").native().c_str());
             pServerSocketFactory->loadPrivateKey(certFile("server.key").native().c_str());
+          #endif
             pServerSocketFactory->server(true);
```
nsuke: Please  remove original lines and #ifdef here and below, as `string()` should not do any harm on Unix too.

lib/cpp/src/thrift/windows/GetTimeOfDay.cpp
```diff
@@ -23,17 +23,28 @@
...
 #else
 #define DELTA_EPOCH_IN_MICROSECS 11644473600000000ULL
 #endif
 
+#if !defined(__MINGW32__)
 struct timezone {
   int tz_minuteswest; /* minutes W of Greenwich */
   int tz_dsttime;     /* type of dst correction */
 };
+#endif
```
nsuke: Would you merge  this `#if !` to `#else` below ? It seems integral  part.

lib/cpp/test/OpenSSLManualInitTest.cpp
```diff
@@ -67,6 +67,22 @@ void test_openssl_availability() {
   openssl_cleanup();
 }
```
nsuke: Just an idea but BOOST_TEST_MAIN or (even better) BOOST_TEST_MODULE work instead of these `main`s ?

I'm fine with the current approach, though.
We have a bug report for boost test dynamic linking and this certainly solves it and it's great.

lib/cpp/test/TFileTransportTest.cpp
```diff
@@ -91,19 +99,35 @@ class FsyncLog {
...
+    snprintf(path_, path_len, "%sXXXXXX", prefix);
+    if (_mktemp_s(path_,path_len) == 0) {
+      fd_ = open(path_,O_CREAT | O_RDWR | O_BINARY,S_IREAD | S_IWRITE);
+      if (fd_ < 0) {
+        throw apache::thrift::TException("_mktemp_s() failed");
+      }
+    } else {
+      throw apache::thrift::TException("_mktemp_s() failed");
+    }
+  #else
```
nsuke: What is the purpose here ? mkstemp does not exist ?
`directory` arg value being "/tmp" seems problematic though...

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1105,10 +1113,17 @@ void TNonblockingServer::listenSocket(THRIFT_SOCKET s) {
...
-      listenPort_ = ntohs(addr.sin_port);
+    #ifdef _WIN32
+      struct sockaddr_storage addr;
+      struct sockaddr_in *pAddr = (sockaddr_in *) &addr;
+      socklen_t size = sizeof(addr);
+    #else
+      struct sockaddr_in addr;
+      struct sockaddr_in *pAddr = &addr;
+      socklen_t size = sizeof(addr);
+    #endif
```
nsuke: Maybe `getsockname` does not truncate the result but instead fails when struct is too small ?
Is it whole Win32 API spec or mingw only bug ?

lib/cpp/test/qt/TQTcpServerTest.cpp
becrux: Sorry, I didn't get what you mean. In my build, moc conflicts with boost headers, that's why I added it. Do you have a different method?

lib/cpp/test/CMakeLists.txt
becrux: You're right, it's not related to mingw strictly, it's related to the way you build Boost under MinGW. Your idea is fine, I was finding a way to detect if Boost requires dynamic linking automatically, but I didn't find any method. We could ignore this diff, if that's not possible.

lib/cpp/test/qt/TQTcpServerTest.cpp
nsuke: Oh I just meant to put it to another `#ifndef Q_MOC_RUN` below.

lib/cpp/test/SecurityTest.cpp
becrux: But if fails with MinGW, due to a conflict between string / wstring, when using "native" or "string" method.

Here too, if you have a more portable method, please modify it at will.

lib/cpp/test/TFileTransportTest.cpp
becrux: mkstemp is not available under MinGW-w64. _mktemp_s is the natural replacement. Unfortunately, MinGW-w64 headers do not provide an automatic define to replace _mktemp_s with mkstemp.

lib/cpp/src/thrift/windows/GetTimeOfDay.cpp
becrux: Yes, I can merge it, you're right.

lib/cpp/test/SecurityTest.cpp
nsuke: I don't get it. You are using "string" not "wstring" method.
"string" will be fine for Unix too, so I'd like to keep your version only, removing original "native".

lib/cpp/test/OpenSSLManualInitTest.cpp
becrux: Frankly speaking, I'm not a Boost.Test expert. I had to Google a lot to find a way that could build and run tests successfully. Please apply any change that could clean the test code, and make it work in any case. My patch could not be the correct way to fix it (even if I can ensure you that it works fine, and all tests are passed).

lib/cpp/src/thrift/server/TNonblockingServer.cpp
becrux: getsockname fails cause the buffer is not big enough. If you compare the size of "sockaddr_in" and "sockaddr_storage", you'll find out that "sockaddr_storage" has many more fields (due to IPv6 and many other stuff).

Some hints here:

http://stackoverflow.com/questions/14624427/wsaefault-error-when-use-getsockname-function

nsuke: Sure, and on Unix, the result shall be truncated rather than fail if the size is not enough.
What I was wondering is whether the failure is official Windows API spec or MinGW only behavior.

becrux: Well, I do not have an answer, but it seems an official Windows API, cause all "stackoverflow" posts I found are not related to MinGW usage. Even if you search through different blogs & forums, you'll find the same answer, and MinGW is not mentioned.

Moreover, sample code reported in MSDN documentation comments uses getsockname in the same way:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738543(v=vs.85).aspx

nsuke: OK, let's do it, then.
Here again, I prefer keeping your version only rather than `#ifdef` for the sake of readability over saving transitive ~100 bytes allocation, unless others disagree.
What do you think ? Do you have any concern ?

becrux: Yeah, that's fine for me.

becrux:
@nsuke, thanks for all your reviews.

Shall I perform a new patch, according to your suggestion?

Thanks,
A.

nsuke:
@becrux, yes, it would be great !
Thank you for contributing, it will help others using the platform a lot.

becrux:
Hi @nsuke,

really sorry for the delay, I was very busy during last month.

I've prepared a new patch, according to your suggestions. How can I clean this pull request, and perform a new commit??

Thanks,
A.

nsuke:
@becrux Thanks for that !
You can push your changes to the original branch of this pull request (in your case it seems master branch of your fork) and the changes will be immediately visible here.
If you have conflict or something, `git push -f `... should do.

becrux:
Well, maybe it's better to start from the actual master revision ... do you agree?

nsuke:
yes, that would indeed be better.

becrux:
Great. So, I could clean this pull request? Should I create a new one?

nsuke:
It may not be necessary because it will be updated automatically as you update your branch.
But it's also totally fine to close this one and create new one.

PR #630, opened 2015-09-28
THRIFT-3361 Improve C# library

nsuke:


Jens-G:
Mono refreshes me with this (absolutely correct) warning:

       src/Transport/TTLSSocket.cs(322,47): 
       warning CS0219: The variable `certs' is assigned but its value is never used

Could you have a look at that one?

nsuke:
Thank you for catching this !
I added a commit with this fix and releated test client setup, please have look at it too.

PR #634, opened 2015-10-05
Fix bad links to coding_standards.md

jimmynavio:
I noticed the other bad links to coding_standards.md had been fixed, and somehow CONTRIBUTING.md was missed. This is bad news bears since CONTRIBUTING.md is directly referenced by https://thrift.apache.org/docs/HowToContribute - which means these are currently broken on the live site.


Jens-G:
Too bad you didn't really read the document ;-)  
=> tracked in THRIFT-3367

PR #636, opened 2015-10-08
THRIFT-3253

phongphan:
Hi

These changesets would fixed the deprecation warnings when compiling with latest D compiler. I have tested with 2.067.1.

I have to change the `thriftToHashImpl` a bit in [379bac8] since I cannot get the `FieldNames` to work when using TMethodMeta.


Jens-G:
Committed per 855cf7f089730a7df657f45e25afa0b6043a72c8
Please close the PR by hand, thanks.

phongphan:
Thank you.

PR #639, opened 2015-10-09
Implements ssl sockets on c_glib

gadLinux:
Adds dependencies to gobject at compile time because it needs it.
Server side is not complete so some tests are failing. Looking to fix them all.

I moved openssl check outside of the cpp compilation because now it's needed for at least two subprojects at lib.

-avoid-version has to be reviewed as this is only for android, and must be enabled only when compiling with android-arm or android-*

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,7 +130,6 @@ if test "$enable_libs" = "no"; then
   with_lua="no"
 fi
 
-
 AX_THRIFT_LIB(cpp, [C++], yes)
 have_cpp=no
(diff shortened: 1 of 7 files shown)


jeking3:
@jfarrell this was done in another set of PRs so this can be closed.

gadLinux:
Yes, quite old sofar.

jeking3:
@jfarrell you probably meant to close this, but it was already closed, and got reopened.

PR #647, opened 2015-10-10
THRIFT-2411 C++: UTF-8 sent by PHP as JSON is not understood by TJsonProtocol

phongphan:
Hi

This patch allow consuming JSON encoded by older version of PHP (< 5.4) which doesn't support JSON_UNESCAPED_UNICODE flag.

It also supported surrogate pair which allow non-BMP character.


phongphan:
Looks like clang in travis-ci doesn't use libc++.

phongphan:
The branch is quite messy right now and not ready to merge.
I will clean it up and create the new PR.

Sorry the the noise.

PR #649, opened 2015-10-10
THRIFT-2794 Suppress warnings in autogenerated Java files

aaudiber:
Client: Java

Most importantly, this will ignore "unused" warnings.

### DIFF
--- a/compiler/cpp/src/generate/t_java_generator.cc
+++ b/compiler/cpp/src/generate/t_java_generator.cc
@@ -441,7 +441,7 @@ string t_java_generator::java_type_imports() {
 }
 
 string t_java_generator::java_suppressions() {
-  return "@SuppressWarnings({\"cast\", \"rawtypes\", \"serial\", \"unchecked\"})\n";
+  return "@SuppressWarnings({\"cast\", \"rawtypes\", \"serial\", \"unchecked\", \"unused\"})\n";
 }
 
 /**


compiler/cpp/src/generate/t_java_generator.cc
```diff
@@ -441,7 +441,7 @@ string t_java_generator::java_type_imports() {
 }
 
 string t_java_generator::java_suppressions() {
-  return "@SuppressWarnings({\"cast\", \"rawtypes\", \"serial\", \"unchecked\"})\n";
+  return "@SuppressWarnings(\"all\")\n";
```
bufferoverflow: Is there no other option to fix this?

aaudiber: It makes sense for people writing the code generator to see the warnings for the code, but for external users who have no control over the generator anyway, the autogen code warnings hide actually useful warnings.

Other options could be changing the code generator to avoid doing things which generate warnings or adding more types of warnings to this list until it only covers expected warnings. The first is really high-difficulty and low-reward, and the second is harder to maintain.

It's hard to imagine users being helped by the warnings in their autogenerated code, so we should just suppress them.

bufferoverflow: We fixed the code generator for cpp to generate code without warnings.
Suppress some specific warnings instead of all is preferred as well.

aaudiber: Ok, I've changed the commit to just add "unused" warnings to the suppressions list.

PR #653, opened 2015-10-14
THRIFT-3382 TBase class for C++ Library

beberg:
Add a TBase that all Struct's use as their base class.
Contributed by Sentient Technologies - http://www.sentient.ai/

### DIFF
--- a/compiler/cpp/src/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/generate/t_cpp_generator.cc
@@ -116,7 +116,8 @@ public:
                                    bool read = true,
                                    bool write = true,
                                    bool swap = false,
-                                   bool stream = false);
+                                   bool stream = false,
+                                   bool is_struct = false);
   void generate_struct_definition(std::ofstream& out,
                                   std::ofstream& force_cpp_out,
                                   t_struct* tstruct,
@@ -368,6 +369,7 @@ void t_cpp_generator::init_generator() {
            << endl
            << "#include <thrift/Thrift.h>" << endl
            << "#include <thrift/TApplicationException.h>" << endl
+           << "#include <thrift/TBase.h>" << endl
            << "#include <thrift/protocol/TProtocol.h>" << endl
            << "#include <thrift/transport/TTransport.h>" << endl
            << endl;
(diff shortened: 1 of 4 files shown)


compiler/cpp/src/generate/t_cpp_generator.cc
```diff
@@ -116,7 +116,8 @@ class t_cpp_generator : public t_oop_generator {
                                    bool read = true,
                                    bool write = true,
                                    bool swap = false,
-                                   bool stream = false);
+                                   bool stream = false,
+                                   bool is_struct = false);
```
hcorg: is this necessary? it's either struct or exception, and there is already bool for that. or am I missing something? IMHO there is too many of those bools already :)

lib/cpp/src/thrift/TBase.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef _THRIFT_TBASE_H_
+#define _THRIFT_TBASE_H_ 1
+
+namespace apache {
+namespace thrift {
+
+class TBase {
+public:
+  virtual ~TBase(){};
+  virtual uint32_t read(::apache::thrift::protocol::TProtocol* iprot) = 0;
```
hcorg: this header is not self contained - it needs uint32_t and TProtocol definitions (for second thing - at least forward decl)

compiler/cpp/src/generate/t_cpp_generator.cc
hcorg: couple of solutions:
 - convert some of those bools into enum (especially that is_exception=true and is_struct=true can be set but it's a nonsense) [imho preferred]
- rename `bool stream` to `is_not_internal` or something like that (it was also added to distinguish internal structs and normal ones)
- do not bother that internal structs inherit from TBase, they're structs anyway

lib/cpp/src/thrift/TBase.h
```diff
@@ -0,0 +1,39 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef _THRIFT_TBASE_H_
+#define _THRIFT_TBASE_H_ 1
+
+#include <sys/types.h>
```
hcorg: uint32_t in Thrift comes from Thrift.h - not all platforms have sys/types.h etc, so Thrift.h contains some ifdefs etc

PR #656, opened 2015-10-18
THRIFT-3393 Introducing i8 to provide consistent set of Thrift integers

Jens-G:
THRIFT-3393 Introducing i8 to provide consistent set of Thrift integers
Client: Compiler (general)
Patch: Jens Geyer

### DIFF
--- a/compiler/cpp/src/thriftl.ll
+++ b/compiler/cpp/src/thriftl.ll
@@ -155,7 +155,8 @@ literal_begin (['\"])
 "include"            { return tok_include;              }
 "void"               { return tok_void;                 }
 "bool"               { return tok_bool;                 }
-"byte"               { return tok_byte;                 }
+"byte"               { return tok_i8;                   }
+"i8"                 { return tok_i8;                   }
 "i16"                { return tok_i16;                  }
 "i32"                { return tok_i32;                  }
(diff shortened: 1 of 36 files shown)


compiler/cpp/src/thriftl.ll
```diff
@@ -155,7 +155,8 @@ literal_begin (['\"])
 "include"            { return tok_include;              }
 "void"               { return tok_void;                 }
 "bool"               { return tok_bool;                 }
-"byte"               { return tok_byte;                 }
+"byte"               { return tok_i8;                   }
```
hcorg: could we somehow generate warning here? "using byte type is deprecated as it might lead to misunderstanding in signess" or something

test/ThriftTest.thrift
```diff
@@ -86,7 +86,7 @@ struct Xtruct
 
 struct Xtruct2
 {
-  1: byte   byte_thing,
+  1: i8     byte_thing,  // used to be byte, hence the name
```
hcorg: I'd go with full rename ;) (just my normal hate for comments :D)

Jens-G: Agree, but one of my goals was to break it down into small changes to prevent us from ending up with a monster patch. This could be one of the next steps.

PR #658, opened 2015-10-20
DART: Allow decode JSON escaped unicode string.

phongphan:
Hi

This patch allow Dart's json protocol to be able to comsume escaped unicode string (\uXXXX). To make the `readString` test runnable, I have to initialize `_context` variable by calling `_resetContext();` in ctor. Not sure it's safe or not.

### DIFF
--- a/lib/dart/lib/src/protocol/t_json_protocol.dart
+++ b/lib/dart/lib/src/protocol/t_json_protocol.dart
@@ -41,6 +41,7 @@ class TJsonProtocol extends TProtocol {
   TJsonProtocol(TTransport transport) : super(transport) {
     _rootContext = new _BaseContext(this);
     _reader = new _LookaheadReader(this);
+    _resetContext();
   }
 
   void _pushContext(_BaseContext c) {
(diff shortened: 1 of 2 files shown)


nsuke:
Hello @phongphan , thanks for the patch, would you mind creating a JIRA ticket for this ?

phongphan:
Hi @nsuke,
Thanks, created at [THRIFT-3396](https://issues.apache.org/jira/browse/THRIFT-3396).

lib/dart/lib/src/protocol/t_json_protocol.dart
```diff
@@ -321,12 +327,36 @@ class TJsonProtocol extends TProtocol {
...
-      byte = _hexVal(_tempBuffer[0]) << 4 + _hexVal(_tempBuffer[1]);
-      bytes.add(byte);
+      // it's \uXXXX
+      transport.readAll(_tempBuffer, 0, 4);
+      byte = (_hexVal(_tempBuffer[0]) << 12)
+        + (_hexVal(_tempBuffer[1]) << 8)
+        + (_hexVal(_tempBuffer[2]) << 4)
+        + _hexVal(_tempBuffer[3]);
+      if (_isHighSurrogate(byte)) {
+        if (codeunits.isNotEmpty) {
```
Jens-G: Shouldn't that be `if( codeunits.isEmpty) { throws ... }`, similar to line 344?

phongphan:
Thanks :)

PR #663, opened 2015-10-27
Thrift 3397

JonathanH-UK:
2 Commits for consideration:
5cf0334 - Changes to C# Library Code (lib/csharp/src)
024c204 - Supporting changes to C# Test Code (lib/csharp/test)

### DIFF
--- a/lib/csharp/Makefile.am
+++ b/lib/csharp/Makefile.am
@@ -66,6 +66,10 @@ THRIFTCODE= \
             src/Transport/TTLSSocket.cs \
             src/Transport/TTLSServerSocket.cs \
             src/TProcessor.cs \
+            src/TProcessorFactory.cs \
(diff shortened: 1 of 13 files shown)


JonathanH-UK:
New commit for review - I've rebased it as per the GitHub guidelines so I hope it's okay. The commit message contains a list of all the changes.
It's still failing to build in the Travis CI environment but I'm still not convinced it's anything I've broken.

For some reason I can't get 'make -k check' to work properly on my dev environment  - I get fatal g++ errors of "no such file or directory" for some boost libraries and for '#include <event.h>' - It's odd because the files it can't find are definitely installed.

'make cross' is more hopeful although all charp-java tests which involve ssl in the transport are failing consistently (e.g. csharp-java        compact      buffered-ip-ssl          failure(64)). Oddly no other ssl tests fail, and java-csharp is fine.

PR #673, opened 2015-10-31
THRIFT-2413: UTF-8 sent by PHP as JSON is not understood by TJsonProtocol

phongphan:
Hi

This should fixed THRIFT-2413.

### DIFF
--- a/lib/py/test/thrift_json.py
+++ b/lib/py/test/thrift_json.py
@@ -2,6 +2,7 @@ from thrift import Thrift
 from thrift.protocol.TJSONProtocol import TJSONProtocol
 from thrift.transport import TTransport
 
+import sys
 import unittest
 
 #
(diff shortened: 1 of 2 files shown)


lib/py/test/thrift_json.py
```diff
@@ -0,0 +1,28 @@
...
+
+  def test_escaped_unicode_string(self):
+    unicode_json = '"hello \\u0e01\\u0e02\\u0e03\\ud835\\udcab unicode"'
+    unicode_text = u'hello \u0e01\u0e02\u0e03\U0001D4AB unicode'
+
+    buf = TTransport.TMemoryBuffer(unicode_json)
+    transport = TTransport.TBufferedTransportFactory().getTransport(buf)
+    protocol = TJSONProtocol(transport)
+
+    self.assertEqual(protocol.readString(), unicode_text)
```
nsuke: Since the other protocols' `readString` returned string, we cannot make TJSONProtocol alone return unicode.
(FYI unicode in Thrift python2 is handled by "utf8string" generator option.)

Now that we have Python3, it is valid as is for Python3 (since unicode and string is equivalent in py3) but still needs the modification for Python2.
To do this you may want to look at `testString` in test/py/TestClient.py for Python2/3 switch.

nsuke:
Hi @phongphan, thank you for series of improvements !
Unfortunately, I think I just broke your patch by Python3 commit.
Could you rebase this and also take a look at my code comment ?

phongphan:
Hi @nsuke, thanks for the comment. I just rebase and fixed the `readString` as suggested.

PR #674, opened 2015-11-01
Free getaddrinfo memory for domain sockets

theopolis:
Previously, UNIX domain sockets were not calling `freeaddrinfo` within `listen`.


theopolis:
This has since been fixed with the AddressInfo wrapper.

PR #680, opened 2015-11-03
Thrift 1857 redux 2

nsuke:
Rational:
* No TType.UTF8 (don't break cross-lang compatibility)
* No six.BufferIO (don't break fastbinary)
* Don't make py2 read/writeString UTF8 (don't break backward API compatibility)


bufferoverflow:
:+1: 
tested with autoconf running **make cross** and cmake **make** on Debian Jessie amd64 with Python 2.7.9
-roger

nsuke:
@bufferoverflow thanks ! Not breaking py2 has been hard part for us.
I'll commit this after CI (+ preferably real people) verify for python 3.x too :)

PR #683, opened 2015-11-04
Deprecate this Apache Thrift Fork

breerly:
We're finally to a place where we can deprecate this repo.

* Python & Node use [thriftrw](https://github.com/search?utf8=%E2%9C%93&q=thriftrw).
* Go uses packaged binaries that are separate from this fork.
* Java uses Apache Thrift directly.

This will not break existing functionality - instead, it will halt new development and reduce confusion.


Raynos:
#legit

kriskowal:
:ok:

Raynos:
#oops

PR #688, opened 2015-11-12
THRIFT-3385 warning: format ‘%lu’ expects ‘long unsigned int’

henrique:
THRIFT-3385  warning: format ‘%lu’ expects ‘long unsigned int’

### DIFF
--- a/test/cpp/src/TestClient.cpp
+++ b/test/cpp/src/TestClient.cpp
@@ -488,7 +488,7 @@ int main(int argc, char** argv) {
       testClient.testBinary(bin_result, string(bin_data, 256));
       if (bin_result.size() != 256) {
         printf("}\n*** FAILED ***\n");
-        printf("invalid length: %lu\n", bin_result.size());
+        printf("invalid length: %zu\n", bin_result.size());
         return_code |= ERR_BASETYPES;
       } else {
         bool first = true;


henrique:
Does it work on VS, anyone?

test/cpp/src/TestClient.cpp
```diff
@@ -488,7 +488,7 @@ int main(int argc, char** argv) {
       testClient.testBinary(bin_result, string(bin_data, 256));
       if (bin_result.size() != 256) {
         printf("}\n*** FAILED ***\n");
-        printf("invalid length: %lu\n", bin_result.size());
+        printf("invalid length: %zu\n", bin_result.size());
```
jeking3: Given this is in test code, why not use something that is more flexible:
```printf(boost::str(boost::format("invalid length: %1%\n") % bin_result.size()));```
instead?

hcorg: maybe simple `std::cout` would be enough?

henrique: Yeah +1, we could get rid of the printf's all together... 
The boost format is cool, but I thought we're looking into getting rid of it too? hehe

jeking3: Why avoid boost?  It is already required to run the unit tests.

hcorg: Why constructing such complicated boost format line, when simple `cout <<` is enough? :) simpler code == better code

Also - it is true - long term development should remove boost, especially from compiler, to allow maximum ease of thrift portability for 'non-C++' users

jeking3:
@jfarrell this issue has been closed/fixed and the pull request can be closed.

PR #690, opened 2015-11-12
C++ transport: implement interruptable TSSLSocket

MartinHaimberger:
- implement interruptable TSSLSocket using non-blocking
sockets and non-blocking bios
- introduce getter getInterruptSocket() in TSocket class
- modified TSSLSocketFactory to be able to create
interruptable TSSLSockets.

See issue THRIFT-3420

### DIFF
--- a/lib/cpp/src/thrift/transport/TSSLServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLServerSocket.cpp
@@ -48,9 +48,9 @@ TSSLServerSocket::TSSLServerSocket(int port,
 }
 
 boost::shared_ptr<TSocket> TSSLServerSocket::createSocket(THRIFT_SOCKET client) {
-  boost::shared_ptr<THRIFT_SOCKET> interruptSocketReader = getInterruptSocket();
-  if (interruptSocketReader) {
-      return factory_->createSocket(client, interruptSocketReader);
+  if (interruptableChildren_) {
+      return factory_->createSocket(client, pChildInterruptSockReader_);
+
   } else {
       return factory_->createSocket(client);
   }
--- a/lib/cpp/src/thrift/transport/TSSLServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLServerSocket.cpp
@@ -48,7 +48,12 @@ TSSLServerSocket::TSSLServerSocket(int port,
 }
 
 boost::shared_ptr<TSocket> TSSLServerSocket::createSocket(THRIFT_SOCKET client) {
-  return factory_->createSocket(client);
+  boost::shared_ptr<THRIFT_SOCKET> interruptSocketReader = getInterruptSocket();
+  if (interruptSocketReader) {
+      return factory_->createSocket(client, interruptSocketReader);
+  } else {
+      return factory_->createSocket(client);
+  }
 }
 }
 }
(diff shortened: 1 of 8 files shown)


MartinHaimberger:
Build is failing because cabal could not be downloaded and installed ............


/home/travis/build.sh: line 45: cabal: command not found
The command "if [ "$ALL_DEPS" != "no" ] ; then cabal update ; fi" failed and exited with 127 during .

Can anyone rerun the tests ?

MartinHaimberger:
the pull request builder seems to have a problem:

depbase=`echo src/TestClient.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
	clang++ -DHAVE_CONFIG_H -I. -I../.. -I../../lib/cpp/src/thrift  -I/usr/include  -I../../lib/cpp/src -Igen-cpp  -Wall -Wextra -pedantic -g -O2 -std=c++11 -MT src/TestClient.o -MD -MP -MF $depbase.Tpo -c -o src/TestClient.o src/TestClient.cpp &&\
	mv -f $depbase.Tpo $depbase.Po
clang: error: unable to execute command: Killed
clang: error: clang frontend command failed due to signal (use -v to see invocation)
clang version 3.4 (tags/RELEASE_34/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
clang: note: diagnostic msg: PLEASE submit a bug report to http://llvm.org/bugs/ and include the crash backtrace, preprocessed source, and associated run script.
clang: note: diagnostic msg:

lib/cpp/src/thrift/transport/TSSLServerSocket.cpp
```diff
@@ -48,7 +48,12 @@ TSSLServerSocket::TSSLServerSocket(int port,
 }
 
 boost::shared_ptr<TSocket> TSSLServerSocket::createSocket(THRIFT_SOCKET client) {
-  return factory_->createSocket(client);
+  boost::shared_ptr<THRIFT_SOCKET> interruptSocketReader = getInterruptSocket();
```
jeking3: The way I implemented this in TSocket was to provide two constructors - the second one takes the interrupt socket as the second argument.  If the same logic is applied here, there would be no need to expose it via the method added to TServerSocket called getInterruptSocket.

lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -659,6 +659,10 @@ shared_ptr<TSocket> TServerSocket::createSocket(THRIFT_SOCKET clientSocket) {
   }
 }
 
+boost::shared_ptr<THRIFT_SOCKET> TServerSocket::getInterruptSocket() {
```
jeking3: Perhaps the interrupt socket and flag should be protected so they are accessible to subclasses of TSocket instead of this?  It's more of a style comment however, I wouldn't gate check-in on changing it.

jeking3:
I think the only concern I would have relative to this is that I don't see a test that exercises the new code.  There are tests that exercise all of the interrupt code paths in TSocketInterruptTest, would like to see the same thing for TSSLSocket, especially with the different wait behavior and BIO nonblocking changes.

lib/cpp/src/thrift/transport/TServerSocket.cpp
MartinHaimberger: @jeking3 i can change the baseclass so the socket and the flag will be protected.

MartinHaimberger:
@jeking3 i will also implement the TSSLSocketInterruptTest ....

MartinHaimberger:
@jeking3 tests and nonblocking accept connect and close are implemented now

jeking3:
Nice - having a test like this ensures it is working and more importantly stays working in the future.  It's too bad our build infrastructure seems to be so fragile.  I'd like to see this pass all builds and test before approving for merge.  I'll look at the build failure and see if the new test passed in any case.

PR #695, opened 2015-11-14
Run test for THRIFT-2413

nsuke:
Slightly simplify _toChar method too.


bufferoverflow:
+1

phongphan:
This is break the `official` win32 build which is `narrow build`. The `unichr` isn't support codepoint larger than 0x10000.

nsuke:
@phongphan thanks for catching this ! I'll revert relevant part.

PR #698, opened 2015-11-14
THRIFT-3200 JS and nodejs do not encode JSON protocol binary fields a…

nsuke:
…s base64

### DIFF
--- a/lib/js/Gruntfile.js
+++ b/lib/js/Gruntfile.js
@@ -41,7 +41,7 @@ module.exports = function(grunt) {
         command: 'mkdir test/build; mkdir test/build/js; cp src/thrift.js test/build/js/thrift.js'
       },
       InstallThriftNodeJSDep: {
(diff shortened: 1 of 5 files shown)


bufferoverflow:
:+1:  great!

beberg:
Great testing improvements, but it's still not base64

beberg:
Sorry, to clarify btoa() was only added to IE in v10, so this isn't a fully portable implementation.

bufferoverflow:
I think it is ok to not support IE < v10
This change needs to be committed base64 can be another step.
test integration done here is awesome!

beberg:
Agree it should be commited, but leave behind a ticket for universal support.

nsuke:
It seems that IE10 has btoa/atob and it was the first version.
http://caniuse.com/#feat=atob-btoa
https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/btoa

PR #707, opened 2015-11-20
Fix generated Node JS code

bracki:
As reported in https://issues.apache.org/jira/browse/THRIFT-1840 and https://issues.apache.org/jira/browse/THRIFT-2527 the generated JS code pollutes the global namespace because it lacks `var` keywords.

This PR adds the `var` keyword where appropriate.
It also appends missing `;`, thus making tools like [jshint](http://jshint.com/) et. al. happier.

### DIFF
--- a/compiler/cpp/src/generate/t_js_generator.cc
+++ b/compiler/cpp/src/generate/t_js_generator.cc
@@ -639,11 +639,12 @@ void t_js_generator::generate_js_struct_definition(ofstream& out,
   vector<t_field*>::const_iterator m_iter;
 
   if (gen_node_) {
+    string prefix = has_js_namespace(tstruct->get_program()) ? js_namespace(tstruct->get_program()) : "var ";
     if (is_exported) {
-      out << "var " << js_namespace(tstruct->get_program()) << tstruct->get_name() << " = "
+      out << prefix << tstruct->get_name() << " = "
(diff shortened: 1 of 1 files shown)


cjthompson:
Looking at this change, I believe this will break namespacing because the output will be invalid JavaScript:

````javascript
var Namespace.Object.Type = {}
````

For namespacing to work, the namespace object would have to be declared at the top of the file and the `var` keyword not used when adding to the namespace object.

bracki:
@cjthompson Good catch! I haven't tried out using `js` namespaces.

What are they good for anyways? Wouldn't you define your namespace when requiring the files?

cjthompson:
I don't use the javascript namespaces either.  With the globals fixed, I just do a require of the file where it's needed.  So you might need to change the code to only output the var if `js_namespace()` return an empty value.

timse:
any news on this? would that change do the job @cjthompson?

cjthompson:
I'm not a C developer, but at first glance it looks like it might work. Let's see if anyone from the apache foundation agrees to merge it.

nsuke:
Would you fix this error ?
https://travis-ci.org/apache/thrift/jobs/92427337#L1307

timse:
there were a couple of syntax errors so i allowed myself to force push fix them
should be fixed now @nsuke

nsuke:
Committed, thank you folks !

PR #715, opened 2015-11-23
THRIFT-3441 Stabilize Travis-CI builds

nsuke:


nsuke:
This fixes old "all (automake)" failures where the compiler got killed.
Fix for `#8` is #709.

PR #716, opened 2015-11-23
[THRIFT-3445] Generates Throwable#getMessage override

d2r:


d2r:
Looks like one of the tests defines the missing getMessage method, and now there is a collision on the declaration.  Closing for now until I can fix it.

PR #722, opened 2015-11-30
fixed the import header

jriskin:
Import headers weren’t being prefixed if you had multiple thrift files.


Jens-G:
Hi @jriskin , 

thanks for the patch! 

Could you just have a look at our [contrib guidelines](http://thrift.apache.org/docs/HowToContribute) and file a matching JIRA ticket for it? If you want to make it perfect, you also add some test case that shows the problem. 

Have fun,
JensG

jriskin:
added a jira ticket, the fix is one line...

PR #723, opened 2015-12-01
dub.sdl for integration into Dlang package registry

nrTQgc:


jfarrell:
hey @nrTQgc, what is can you please outline what is required to register the package or link to documentation

nrTQgc:
You need create account at http://code.dlang.org
It is necessary register package in http://code.dlang.org/my_packages

jeking3:
THRIFT-3458

jfarrell:
Any objections to switching this to dub.json as follows ?

```
{
  "name": "apache-thrift",
  "description": "Apache Thrift D library",
  "authors": [
    "Apache Thrift Developers <dev@thrift.apache.org>"
  ],
  "homepage": "http://thrift.apache.org",
  "license": "Apache-2.0",
  "dependencies": {
    "libevent": {
      "version": "*",
      "optional": "true"
    },
    "openssl": {
      "version": "*",
      "optional": "true"
    }
  },
  "targetType": "library",
  "sourcePaths": [
    "src"
  ],
  "importPaths": [
    "src"
  ],
  "excludedSourceFiles": [
    "src/thrift/index.d"
  ]
}
```

nrTQgc:
dub.json is fine too

jeking3:
All parties here, please move this forward or it will eventually be closed due to inactivity.  It looks like we're fairly close to completing this; however I have to point out that the project's use of deimos has shacked it in a bad way, and we don't support current versions of openssl or libevent any more because of it...

nrTQgc:
I replaced dub.sdl with dub.json and fixed deprecation warnings. Can you explain issues with current versions of openssl and libevent? What options we have?

jeking3:
See https://issues.apache.org/jira/browse/THRIFT-4308

Essentially the upstream deimos modules for openssl and libevent no longer work, and we need to move to something else for secure and for non-blocking modes.  I think there's probably a better D crypto lib we can use, but I'm not sure about libevent.  In any case, there hasn't been much of an uproar from folks about these things not working.

jeking3:
Well, I was wrong; it's a link ordering issue.  I could not fix/force the link order coming out of dmd, so I switched to the ld.gold linker for the docker images and that one is smart enough to get it right (also it's way faster than the bfd linker).

PR #731, opened 2015-12-04
Support for float type (part 1: cpp)

andrewcox:
Ported from fbthrift


bufferoverflow:
Thanks Andrew! I will review this during the weekend...

Could you please add THRIFT-2457 as prefix within your commits as mentioned within CONTRIBUTING.md ? This will simplify the automation with apache policies and automation made by @jfarrell and infra team.

best!
-roger

bufferoverflow:
Great patch Andrew!
I would really like to get FLOAT up and running here. The blocker for not merging now is breaking other languages with this at the moment, I had to exclude a lot:
```
sh bootstrap.sh
./configure --without-go --without-c_glib --without-haskell --without-haxe --without-java --without-erlang --without-ruby --without-perl --without-php --without-nodejs --without-python --without-csharp
make
make check
make cross
python test/test.py -s --server cpp --client cpp
```
our first line of defense was hit heavily ;-) https://travis-ci.org/apache/thrift/builds/94772477

Could you fix these issues?
```
In member function ‘std::string t_csharp_generator::declare_field(t_field*, bool, std::string)’:
src/generate/t_csharp_generator.cc:2713:14: warning: enumeration value ‘TYPE_FLOAT’ not handled in switch [-Wswitch]
       switch (tbase) {
```
I would say providing a warning such as here  and add some little stubs with a TODO should be sufficient:
https://github.com/apache/thrift/blob/master/compiler/cpp/src/generate/t_as3_generator.cc#L2358

thanks again!
-roger

nsuke:
@bufferoverflow Just FYI our first line of defense is hit from behind not by this patch.
At that time Travis people was migrating their "legacy" env to Google cloud and that somehow broke most of our builds.
(It doesn't matter if your local build is failing too)

andrewcox:
Updated the commit name to have the correct prefix (the current commit is otherwise unchanged)

Yes, I was just checking whether this looks like what you want, I'll go about making this work for more languages.

When I update the pull request will it re-run the CI tests so I can keep up with what I need to fix?

bufferoverflow:
Yes, CI will re-run.

jeking3:

PR #733, opened 2015-12-06
THRIFT-3474 Docker: thrift-compiler

bufferoverflow:
Would be nice to provide support of 0.9.3 compiler as container.

This closes #732 

```
$ docker pull thrift/thrift-compiler:0.9.3
Pulling repository thrift/thrift-compiler
...
Status: Downloaded newer image for thrift/thrift-compiler:0.9.3
$ sudo docker run thrift/thrift-compiler:0.9.3 --version
Thrift version 0.9.3
```


bufferoverflow:
@nsuke @Jens-G @jfarrell 

What do you think about this?

nsuke:
#732 is already in master so the point is to push this to 0.9.3 branch ?
IMO it's reasonable in this case to have one more commit in 0.9.3 **branch** than 0.9.3 **tag**.

As to implementation, `apt-get update`, `apt-get install` and the last cleaning step should not be separate `RUN`s  to avoid fat intermediate layers and also we typically want to avoid `apt-get upgrade` and leave it to debian image maintainers.
See https://docs.docker.com/engine/articles/dockerfile_best-practices/ and [current "offical" thrift dockerhub image](https://github.com/ahawkins/docker-thrift/blob/61c3478ab828d3e610f192b442ac2a7221749c47/0.9/Dockerfile) for more detail.

BTW we may want to take ownership of the docker hub [official image](https://hub.docker.com/_/thrift/).
The procedure is described here:
https://github.com/docker-library/official-images#maintainership

bufferoverflow:
@nsuke You are right is about pushing an additional commit on top of 0.9.3 branch. 

Thank's for the links, I did some rework [here](https://github.com/bufferoverflow/thrift/commit/4309bd9c17015fa4763878b56fd911fe1a886870) the biggest problem from a size perspective are all the gcc dependencies... What do you think about this? 

Yes, we should take ownership of the official docker hub image as soon as we have the Dockerfile in a perfect shape.

nsuke:
A common technique is to call apt-get purge -y --auto-remove ... in one step.
It is often seen with a temporary variable like this:

    RUN buildDeps=" \
            flex \
            ... \
            "; \
        && apt-get install -y --no-install-recommends $buildDeps
        && ...build and other cleanups
        && apt-get purge -y $buildDeps

@JensG We also need to install Go to use `-gen go` (because it uses `gofmt`), right ?

jfarrell:
@nsuke correct, gofmt is called in the compiler

bufferoverflow:
rework done ready to merge to 0.9.3 branch

jeking3:
@jfarrell this is marked Closed/Fixed in jira; this pull request can be closed.

PR #734, opened 2015-12-07
THRIFT-3328 A callback is not called when a server is down.

HIRANO-Satoshi:
https://issues.apache.org/jira/browse/THRIFT-3328


bufferoverflow:
Thanks for the fix!
Could you close the ticket, I forgot to mention the PR within the commit.

HIRANO-Satoshi:
Thanks.

PR #736, opened 2015-12-08
THRIFT-3433 Doubles aren't interpreted correctly

nsuke:


bufferoverflow:
:+1: tested on Debian Jessie amd64

PR #739, opened 2015-12-10
fix compile error(Windows7 x64) : there is no lib\cpp\src\thrift\Thri…

lizhenghn123:
* fix compile error(Windows7 x64) : there is no lib\cpp\src\thrift\Thrift.cpp file.


kcris:
seems incomplete, until more files are added to the vcxproj, see pull request #745

lizhenghn123:
add TProtocol.cpp and TOutput.cpp in libthrift.vcxproj

kcris:
I did. I have both removed Thrift.cpp from the project file, and also added TProtocol.cpp and TOutput.cpp (see pull request #745, was merged already).
So this pr (#739) should be closed I think.

lizhenghn123:
Thanks!

PR #741, opened 2015-12-11
THRIFT-3485 don't publish extra files to npm

azylman:
The thrift npm package is as big as all of our other node dependencies combined. It currently publishes the entire thrift repository, which is ~30mb right now. As far as I can tell, it only requires lib/nodejs, which is 376k.

I propose adding an .npmignore file. You can more info in the npm documentation: https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

If the package only requires lib/nodejs, we should be able to add an entry to the npmignore file for every folder (including lib), and then a negated entry for lib/nodejs. We'll also want to include everything from gitignore, because npm will no longer use that once an npmignore file is added.

I've tested this using the method recommended in the npmjs documented, basically installing the package globally and making sure it still requires successfully:

```
$ npm install -g .
$ node
> var thrift = require('thrift');
undefined
> thrift
{ Thrift:
   { Type:
  ...etc...
$ du -h -d 1 .
236K	./lib
812K	./node_modules
1.2M	.
```


nsuke:
It seems it's already fixed by f264884901d0e2a1197ae1511ee0fa0b8ca0d2aa.
@azylman could you confirm ?

azylman:
@nsuke Is that commit published at all? I'm still seeing this behavior in the latest version published to npm (0.9.3)

azylman:
@nsuke Looks like that's not published in 0.9.3. I just confirmed that master does _not_ have this problem, making this PR unnecessary. Hopefully that change can get published soon :/

PR #743, opened 2015-12-13
THRIFT-3486 - Java generated `getFieldValue` is incompatible with `setFieldValue` for binary values.

bgould:
For binary fields, the compiler generates Java getter/setter methods for primitive byte arrays (`byte[]`) as well as `java.nio.ByteBuffer` objects.  However, `setFieldValue` only accepts a `ByteBuffer`, and throws a `ClassCastException` if a `byte[]` is passed as an argument. To further confuse things, `getFieldValue` returns a `byte[]`. By allowing either a `ByteBuffer` or a `byte[]` as input, `setFieldValue` could exhibit less confusing behavior for end users.

These changes add an `instanceof` check for binary fields in the generated `setFieldValue()` method so that it accepts either a `byte[]` or a `ByteBuffer`.  There is also an update to the Java library's test suite to test for this functionality.

### DIFF
--- a/lib/java/test/org/apache/thrift/TestStruct.java
+++ b/lib/java/test/org/apache/thrift/TestStruct.java
@@ -309,11 +309,33 @@ public class TestStruct extends TestCase {
   }
 
   public void testBytesBufferFeatures() throws Exception {
-    JavaTestHelper o = new JavaTestHelper();
+    
+    final String testString = "testBytesBufferFeatures";
+    final JavaTestHelper o = new JavaTestHelper();
+
     o.setReq_bin((ByteBuffer)null);
     assertNull(o.getReq_bin());
+
+    o.setReq_bin(ByteBuffer.wrap(testString.getBytes()));
+    assertNotNull(o.getReq_bin());
+
     o.setReq_bin((byte[])null);
     assertNull(o.getReq_bin());
+
+    o.setReq_bin(testString.getBytes());
+    assertNotNull(o.getReq_bin());
+
+    o.setFieldValue(JavaTestHelper._Fields.REQ_BIN, null);
+    assertNull(o.getReq_bin());
+
+    o.setFieldValue(JavaTestHelper._Fields.REQ_BIN, testString.getBytes());
+    assertNotNull(o.getReq_bin());
+
+    o.setFieldValue(JavaTestHelper._Fields.REQ_BIN, null);
+    assertNull(o.getReq_bin());
+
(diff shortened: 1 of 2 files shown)


lib/java/test/org/apache/thrift/TestStruct.java
```diff
@@ -309,11 +309,33 @@ public void testToString() throws Exception {
...
+    o.setReq_bin(testString.getBytes());
+    assertNotNull(o.getReq_bin());
+
+    o.setFieldValue(JavaTestHelper._Fields.REQ_BIN, null);
+    assertNull(o.getReq_bin());
+
+    o.setFieldValue(JavaTestHelper._Fields.REQ_BIN, testString.getBytes());
+    assertNotNull(o.getReq_bin());
+
+    o.setFieldValue(JavaTestHelper._Fields.REQ_BIN, null);
```
Gospodzin: Does it make sense to duplicate this piece of code?

lib/java/test/org/apache/thrift/TestStruct.java
```diff
@@ -309,11 +309,33 @@ public void testToString() throws Exception {
...
-    JavaTestHelper o = new JavaTestHelper();
+    
+    final String testString = "testBytesBufferFeatures";
+    final JavaTestHelper o = new JavaTestHelper();
+
     o.setReq_bin((ByteBuffer)null);
     assertNull(o.getReq_bin());
+
+    o.setReq_bin(ByteBuffer.wrap(testString.getBytes()));
+    assertNotNull(o.getReq_bin());
```
Gospodzin: Wouldn't it make more sense to use `assertArrayEquals(testString.getBytes(), o.getReq_bin())` (same at line 326)?

lib/java/test/org/apache/thrift/TestStruct.java
bgould: Just wanted to make sure the field was null before proceeding before testing the next "set" operation

lib/java/test/org/apache/thrift/TestStruct.java
bgould: Yes, couldn't hurt.  Really I was just trying to test to make sure it doesn't trigger a ClassCastException; but your suggestion seems logical.

lib/java/test/org/apache/thrift/TestStruct.java
Gospodzin: seems logical ;)

PR #745, opened 2015-12-16
[fix] libthrift.vcxproj was not compilable with Visual Studio

kcris:
libthrift.vcxproj did not properly include TOutput.cpp and TProtocol.cpp


kcris:
THRIFT-3493
https://issues.apache.org/jira/browse/THRIFT-3493

PR #753, opened 2015-12-21
THRIFT-3497 Build fails with "invalid use of incomplete type"

simonsouth:
This change fixes an "invalid use of incomplete type" compiler error referring to `struct timeval` when building on Haiku.

- `Monitor.cpp`: Include `thrift-config.h` for definition of `HAVE_SYS_TIME_H`.
- `Monitor.h`: Explicitly include `sys/time.h` on platforms that require this.


Jens-G:
Hi Simon, is that invalid or why did you close it?

PR #755, opened 2015-12-21
THRIFT-3499: Add package_prefix to python generator

ericklaus-wf:
Permit prefixing import paths with a certain string to support devs who will place the generated code below the top level of their python hierarchy.

This is similar to the go generator's package_prefix generator option.

### DIFF
--- a/compiler/cpp/src/generate/t_py_generator.cc
+++ b/compiler/cpp/src/generate/t_py_generator.cc
@@ -59,6 +59,12 @@ public:
     iter = parsed_options.find("slots");
     gen_slots_ = (iter != parsed_options.end());
 
+    iter = parsed_options.find("package_prefix");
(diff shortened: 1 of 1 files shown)


ericklaus-wf:
The CI build has failed when running the cpp-cpp test twice now. I'm not sure how my python changes could affect this. Am I kicking off the CI incorrectly?

Relevant CI failure line is below:
```
cpp-cpp            compact      http-ip                  timeout(128)
```

nsuke:
Not at all. Some cross tests are flaky and should be excluded, if not fixed immediately.
In your case the failure is clearly unrelated so no need to re-invoke any more.

ericklaus-wf:
@Jens-G Do I need to do anything else to get this PR moving?

PR #756, opened 2015-12-23
THRIFT-3500: Strings unicode py2

johnlockwood-wf:
As the test shows, this makes Thrift protocols treat Strings as unicode in Python 2 which is essentially equivalent as the str in Python 3. Indeed Python 3.5 now treats u"hello" as a str literal, just as Python 2 treats it as a unicode literal.

The python 2 standard lib contains the json package, that, when loading JSON, deserializes strings into unicode objects, where the TJSONProtocol deserializes them into utf-8 encoded bytes without this change.

https://issues.apache.org/jira/browse/THRIFT-3500

### DIFF
--- a/lib/py/src/compat.py
+++ b/lib/py/src/compat.py
@@ -1,14 +1,21 @@
 import sys
+import logging
 
 if sys.version_info[0] == 2:
 
   from cStringIO import StringIO as BufferIO
 
   def binary_to_str(bin_val):
-    return bin_val
+    try:
+      return bin_val.decode('utf8')
+    except:
+      return bin_val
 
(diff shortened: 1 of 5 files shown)


nsuke:
In the current design, unicode in py2 is handled by -gen py:utf8strings compiler switch.
Does it cover your use case ?

johnlockwood-wf:
I don't know, is there any documentation on what that option does? I'll just try it.

johnlockwood-wf:
@nsuke, that option does appear to do the job when using the JSONProtocol.

jeking3:

PR #762, opened 2015-12-26
THRIFT-3510 Add HttpTaskAsyncHandler implementation

adamconnelly:
The THttpHandler doesn't support an async pipeline. This means that it's difficult for service implementations to make async calls. I've added an implementation of HttpTaskAsyncHandler, so you can write services using async calls.

Additionally, if you generate the C# classes with the current async support, you get a single interface with both sync and async methods. This doesn't really make sense on the server side since if you implement a service you end up leaving all the async method unimplemented. I've added a new option to the compiler that allows you to generate separate async and sync interfaces: separateasync.

This generates code like the following:

```csharp
public interface ISync
{
    void DoSomething();
}

public interface IAsync
{
    Task DoSomethingAsync();
}

public interface Iface : ISync, IAsync
{
}
```

I tried to write the code for the generator in such a way that I didn't alter the existing functionality. I've probably created a load of duplication as part of that, but that was just because I'm not familiar with the code.


Jens-G:
1. The project files are 2013 now. What minimum version of Mono is required?
2. There are several files lacking the ASF license header. Could you add that?

adamconnelly:
Cheers for the feedback - I'll try to sort all tha . I'm not quite sure what the minimum version of mono is, but I'll find out. 

I added the new project to test the web stuf . Do you think there's actually any point to it? Just asking because i'm not mega familiar with how you guys do your testin . 

Thanks for the feedbac .

nsuke:
> This doesn't really make sense on the server side since if you implement a service you end up leaving all the async method unimplemented.

You can implement both by using Interface.Method syntax for at least one of the two.
Generating XxxAsync as your patch does feels more straightforward and natural, though.

adam-resdiary:
@nsuke Sorry if I'm misunderstanding you, but I don't think I explained myself very well with

> This doesn't really make sense on the server side since if you implement a service you end up leaving all the async method unimplemented.

Here's an example scenario just so it's clear what I'm talking about:

Say I have two servers, Server 1 and Server 2. I also have two thrift services, Thrift Service A and Thrift Service B. Server 1 implements Thrift Service A, and calls Thrift Service B on Server 2. Server 2 implements Thrift Service B. So something like this:

| Server 1 | Server 2 |
|------------|-------------|
| Implements Thrift Service A | Implements Thrift Service B |
| Calls Thrift Service B | |

Here's example thrift definitions:

```
service ThriftServiceA
{
    void Method1();
}

service ThriftServiceB
{
    string GetSomeInformation();
}
```

I then generate the code using the thrift compiler, and use the async:true option. Because of the way we generate the code for our solutions, this causes the async interfaces to be generated for both Thrift Service A, and Thrift Service B.

On Server 1, I then have some kind of API call that calls Thrift Service B:

```csharp
[Route("api/Information")]
public async Task<string> GetSomeInformation()
{
    return await _thriftServiceB.GetSomeInformationAsync();
}
```

This makes perfect sense because I'm making an async call to a thrift service via the thrift client.

I then go to implement Thrift Service A, and what I end up with is something like:

```csharp
public class ThriftServiceAImpl : ThriftServiceA.Iface
{
    public void Method1()
    {
        // Do something
    }

    public Task Method1Async()
    {
        throw new NotImplementedException("There's no point implementing this");
    }
}
```

So in the example above, there's no point implementing Method1Async, because the current HttpHandler implementation doesn't ever call it - it only calls the synchronous implementation. So it's just noise, but you have to implement it because it's part of the interface.

So my patch doesn't actually change the method names at all, it just splits the two sets of methods into two separate interfaces so that if you're only implementing the synchronous set you only need to implement the synchronous methods, and the same for the async ones.

Now that I think about it, I think my change is actually backwards compatible, because Iface implements ISync and IAsync, so there's probably no need for a separate compiler flag. I was just playing it safe. I'll leave that decision up to you guys though.

adam-resdiary:
@Jens-G 

> The project files are 2013 now. What minimum version of Mono is required?

I've actually just reverted the changed VS made to the solution file, because I don't have access to my mono environment to test this right now. Is that good enough for now? If not I'll figure something out.

> There are several files lacking the ASF license header. Could you add that?
> Same for the new AssemblyInfo.cs file(s), the fields are incomplete.

That should be these problems sorted out now. I've just added fixup commits for now so it was easy for you to see the changes in the PR. When you're happy with them, I can rebase them so it makes more sense.

Jens-G:
Whatever works is good. Rebase & would be fine, yes.

I am going to do some tests before I commit it to master. Thanks for the work so far.

adam-resdiary:
Great - thanks. That's it rebased against the current version of master - hopefully that should fix the failing test in the last build as well.

nsuke:
@adamconnelly thanks for the clear and concise explanation and yes that's how I misread somehow.
About the backward compatibility, I couldn't come up with any sensible potential issue either, so it makes sense to me to remove the old path.

@JENS-G do you find it safe too to replace the behavior rather than to add a flag, or any concern ?

For the Mono version: `async` is added in Mono 3.0.0 so it's likely that it's the minimum working version for this addition (as already noted in another ticket, it requires mcs compiler not gmcs).

Jens-G:
If we can avoid another flag (we already have too many of them) and if it does not break compatibility, I'm for it.

adamconnelly:
Well let me know what you want me to do - obviously the lazy part of me doesn't mind if you want to make the change yourself to just use the existing async flag :smile: but given I created the original patch, I'm more than happy to change that if you want me to.

Jens-G:
*confused* Are you talking about adding the new flag or removing the existing ```async```? Either way, as I said above: If we don't need it anymore and it does not break anything, remove it. 

@nsuke ?

adamconnelly:
Not quite sure if you're asking about my last comment, but what I was asking was whether you wanted me to change what I'd done to just use the existing ```async``` flag rather than the new ```separateasync``` flag I added. I'm happy to do that if you think it makes sense.

Jens-G:
Ok, do it.

adam-resdiary:
@Jens-G @nsuke That's the changes added to get rid of the separateasync flag and just use the async flag instead.


One thing I noticed is we've got #if preprocessor directives that only include the Begin_, End_ methods if you're building for Silverlight. However, these directives aren't included if you generate with ```async:true```, but they do if you generate with ```asyncctp:true```. This means the Begin_, End_ methods are always included for async, but only included for asyncctp if you're building for Silverlight.

I didn't want to change this behaviour since it's how the code currently is, but it seems a bit inconsistent.

adam-resdiary:
Sorry about all the whitespace changes - I've got SourceTree setup to ignore whitespace because it can't seem to handle reverting changes properly unless that setting's turned on.

I'll revert all those changes and push up another commit.

adam-resdiary:
Ok - that should be it sorted now. I added the fixup in because for some reason VS decided to prefix a couple of the method prototypes with ```t_csharp_generator::```. The code formatting / intellisense stuff for C++ seems pretty terrible in VS compared to the C# support :(

Jens-G:
When editing the Thrift compiler C++ files please watch out and keep the existing formatting. IIRC by default VS indents with tabs. We don't want tab indents in that files (and we want LInux EOLs - git should take care about the latter). 

The async CTP stuff is only used with WP7 - which I removed from trunk just a few minutes ago. It will not be used with WP8 or higher, as the only purpose was to deal with some incompatible changes made between the async CTP and the version that found its way into later frameworks. 

With regard to this pull request and JIRA ticket I'd say we leave it as it is for now. The flag is better removed either in a separate ticket or as follow-up patch to the WP7 rermoval ticket.

adamconnelly:
Yeah, I already changed my settings to spaces for indentation, and two spaces for indents, which seemed to be what the existing code was using. It's just that there's a load of other things (like newlines before else statements for example), and for some reason VS seemed to reformat the entire file. 

Anyhow, I reverted all those whitespace changes, so it shouldn't be an issue.

adamconnelly:
@Jens-G Thanks for accepting this.

PR #765, opened 2016-01-02
THRIFT-3514: Add PHP 7 version of php_thrift_protocol

dsp:
This is an initial port of php_thrift_protocol to PHP7 and an RFC how to approach it.
However as we deal with zval's all over the place, we opt for separating
the C files completely leading to some overhead. However this
is a good start to see the differences in the implementation. From
there we should follow up with a more unified approach by refactoring
parts of the zval handling to be more generic so we can plug it
into PHP 7 and PHP 5 extensions.

Tested this by running with TestClient.php against a CPP server
and using TBinaryProtocolAccelerated.


ovr:
Awesome!! :cake: :beers:

dsp:
The contbuild error seem to be in java-csharp and perl-py3 which are unrelated.

stigsb:
Great work :8ball:

PR #768, opened 2016-01-03
THRIFT-3474 Docker: thrift-compiler

bufferoverflow:
some tuning suggested by Aki Sukegawa


nsuke:
Thanks for incorporating these :)
Just tested this and it works fine including `-gen go`.
The size is 151 MB which is even smaller than plain ubuntu or centos image and only 26 MB larger than jessie image.

PR #770, opened 2016-01-04
THRIFT-3520 Dart TSocket onError type as Object

markerickson-wf:
Dart TSocket exposes `onError` as a `Stream<String>`. However, the payload may be either a String or an Error, and should be typed as `Stream<Object>`.  

This is a follow up to https://issues.apache.org/jira/browse/THRIFT-3468, and really should have been done with that change.

https://issues.apache.org/jira/browse/THRIFT-3520

@evanweible-wf
@tylertreat-wf
@stevenosborne-wf


evanweible-wf:
+1

stevenosborne-wf:
+1

PR #772, opened 2016-01-05
THRIFT-3521 : TSimpleJSONProtocolTest is not deterministic (fails when run on JDK 8)

bgould:
This PR removes fields that cause TSimpleJSONProtocolTest to fail when run using different version of the Java standard library.

### DIFF
--- a/lib/java/test/org/apache/thrift/protocol/TestTSimpleJSONProtocol.java
+++ b/lib/java/test/org/apache/thrift/protocol/TestTSimpleJSONProtocol.java
@@ -27,6 +27,7 @@ import org.apache.thrift.TException;
 import org.apache.thrift.transport.TMemoryBuffer;
 
 import thrift.test.CompactProtoTestStruct;
+import thrift.test.HolyMoley;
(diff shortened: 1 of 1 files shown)


nsuke:
@bgould, thanks for looking into this.
Would it be tedious and/or complicate the code if we match against regex instead of removing them ?

bgould:
I thought about that too but didn't have time to do it this morning and just wanted to get a working build.  I think that perhaps better than a regex might be to actually parse the JSON string and check the values.  I'll see if I can get that working.

nsuke:
I was mainly concerned about completely losing map/set tests.
Of course preserving all tests for those element types is great, but I think it's fine to unset most of them, as long as some/one of tests for both map and set is retained.

bgould:
I see... in that I think that the tests should be ok as they are because there still exist maps and sets in the `testSanePartsOfCompactProtoTestStruct` test method.

nsuke:
You're right, you were already keeping those which still work for JDK8.

PR #774, opened 2016-01-06
THRIFT-3523 : XML Generator

bgould:
I've put together an XML model for Thrift IDL that is very similar in concept and structure to the model created by the JSON generator of the Thrift compiler.  The code is based on the JSON generator, and inherits the 'merge' option from that module for including the model of all includes in a single file if desired.

This generator is useful if you wish to extend the Thrift compiler with some XML based templating engine such as XSLT.  For example I have prototyped transforms that can create valid XML schema and WSDL files from Thrift IDL, as well as transform SOAP requests and responses to and from Thrift messages.  Those transforms are not included because they may be outside the scope of Thrift, but I can provide them if interested.  Regardless, I feel the XML model is generally useful and might be a good addition to the Thrift compiler.

Also included is an XML schema for the model file format and tests that generate XML from existing IDL in the Thrift test suite to validate against that schema.  These tests are Ant based and so only run when --with-java is enabled in ./configure

### DIFF
--- a/lib/xml/thrift-idl.xsd
+++ b/lib/xml/thrift-idl.xsd
@@ -1,0 +1,265 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<schema xmlns="http://www.w3.org/2001/XMLSchema"
+        targetNamespace="http://thrift.apache.org/xml/idl" 
+        xmlns:tns="http://thrift.apache.org/xml/idl" 
+        elementFormDefault="qualified">
+
+  <element name="idl" type="tns:IDL" />
+  <element name="document" type="tns:Document" />
+
+  <complexType name="IDL">
+    <sequence>
+      <element ref="tns:document" minOccurs="1" maxOccurs="unbounded"/>
+    </sequence>
+  </complexType>
+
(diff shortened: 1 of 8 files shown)


lib/xml/thrift-idl.xsd
```diff
@@ -0,0 +1,265 @@
...
+        <attribute name="doc" type="string" />
+      </extension>
+    </complexContent>
+  </complexType>
+
+  <simpleType name="Requiredness">
+    <restriction base="string">
+      <enumeration value="required" />
+      <enumeration value="optional" />
+    </restriction>
```
Jens-G: If I read this correctly, the ```<attribute name="required" type="tns:Requiredness" />``` must be specified. But there are only two possible values for it. We have three ... if neither ```required``` nor ```optional``` are specified, the member is of the so-called "default" requiredness (which basically says: it may not be in the data you read, but please always write it).

Jens-G: Ok, I did not read it correctly. The attribute is of course optional by default, and that's perfectly fine.

PR #778, opened 2016-01-06
THRIFT-3528 Fix warnings in thrift.ll

nsuke:


Jens-G:
Could you explain what is going on here and why these entries have to be changed/removed? I just don't get it, sorry.

nsuke:
Sure, that's the whole point of putting this as a PR.
In a nutshell I believe it does not change any behavior except that it fixes 3 warnings pasted to JIRA.
Those 3 rules are exact duplicates that never match because precedent ones (Line 182, 267 and 409) always match over these.
You may want to also check `git blame` output to see which patches caused these. It seems simple mistakes to me.

hcorg:
thrift_reserved_keyword(yytext); is used to report "Cannot use reserved language keyword: " (reserved in Thrift target languages, not Thrift itself). As "union" is already used by Thrift, it does not need to be additionally reserved. "Public" is just repeated (looks like typo).
So for those changes: +1

But last one seems more complicated - at least looking at comment (I hate comments, but still, sometimes it's good to read them). I'm not sure which 'default' block should we keep. Why someone needed '*' in parser? Is it still needed? If we stay with "unexpected_token" (I'd prefer that) I think that comment should be killed too.

nsuke:
235f8b5899b2c5b75cb83975695da1377f0cac0c is the commit that added current (preceding) one which is THRIFT-1274.

According to the JIRA ticket

> Currently, if the thrift lexer encounters a token it does not expect, it prints the token it to stdout and continues. (This is the default behavior of flex when a token is unmatched.) This updates thriftl.ll to fail with an error message when it sees an unexpected character.

So according to this, older one does nothing and we wanted Thrift compiler to fail on unexpected token, which seems perfectly reasonable to me.

Removing the comment makes sense. `unexpected_token` call is already more informative than the comment.

hcorg:
Comment is from older version, which for some reason passed all unexpected tokens to bison.. strange :)

So with comment removal this patch has my +1 :+1:

nsuke:
@hcorg oh I actually got your point, thanks for catching this. Comments are harder to read than C++ :p

PR #784, opened 2016-01-10
THRIFT-3539 Use self.process_foo instead of Processor.process_foo

tbartelmess:
https://issues.apache.org/jira/browse/THRIFT-3539


bufferoverflow:
I have some issues using this, did you ran **make check** or **make cross**

jeking3:
@jfarrell this PR can be closed - the fix causes many build failures.

PR #785, opened 2016-01-11
THRIFT-3251 Add a Http Transport for go lib.

dimiro1:
The protocol was implemented as a HandlerFunc using the standard net/http package.

See: https://github.com/apache/thrift/pull/557
See: https://issues.apache.org/jira/browse/THRIFT-3251


dimiro1:
@Jens-G  please take a look at this first iteration. Thanks

PR #790, opened 2016-01-13
THRIFT-3545 Cocoa: Container type literals do not compile

ChristopherRogers:


kdubb:
Have these changes been tested with nested containers?  I don't think they will work.

My apologies for being late to the party.  I am only scanning my thrift email intermittently lately.

kdubb:
@ChristopherRogers Can you provide an example of what does not compile for you?

kdubb:
These do work with nested types so my hunch was wrong on that but with the current code (including your PR) it now has awkward extra semi-colons; so my munch was somewhat correct that they appear unnecessary

The NSSet missing a single brace was correct. I don't know why you switched it to the longer syntax when you could have just added a brace.

It seems to me the issue was only to do with set literals and could have been solved with a single character change (adding a brace). Is there some other place I am missing where semi-colons are not added? If so I cannot seem to find it. 

Currently as generated...
```objc
  NSDictionary<NSString *, NSString *> * tmp0 = @{@"new": @"world"};
  MapConst = @{@"hello": tmp0};
;

  NSSet<NSString *> * tmp1 = [[NSSet alloc] initWithArray:@[@"hello"]];
  SetConst = [[NSSet alloc] initWithArray:@[tmp1]];
;

  NSArray<NSString *> * tmp2 = @[@"hello"];
  ArrayConst = @[tmp2];
;
```

kdubb:
@ChristopherRogers Please check #797 and see if it produces working code for you as well.

ChristopherRogers:
Ah, my apologies for the extra semicolons. It was missing semicolons in the default init methods where a collection type had a default value.

I believe this semicolon problem is due to discrepancies in how the caller of this method assumes code will be generated--with semicolons and line breaks or not. I think if the call sites were to be cleaned up the ugliness would go away. I believe I checked them but I might've overlooked them. I'll give it another look again later too.

The NSSet change should've been in another PR, I admit, but it was for performance reasons (since one of Thrift's main goals is performance). I believe that optimized ARC code cannot cancel out the autorelease that setWithArray makes because Foundation isn't using ARC itself (I think, but it's worth testing.) Even if it could I think it's still faster to not have to have it back out of autoreleasing. I considered changing it to the initWithObjects variant but it seemed a little too drastic and needs performance testing.

ChristopherRogers:
Also, I just remembered from an old version of Xcode that even the @[]/@{} syntax used autorelease (probably to easily codegen in both ARC and MRC) but that might've changed since then, and I might've been looking at a non-optimized build.

kdubb:
I am fairly sure that #797 fixes the line breaks and semi-colons in the proper way. So, just check it with your code to make sure.

As far as performance of ARC I will just say that you attempt to be micro optimizing code that is about to, or just has, sent data over a network connection. Any thought to the nanosecond of time ARC *may* (and most likely doesn't) add is not something we should be concerned about. Finally, this code initializes static constants so discussing performance at all is unnecessary.

PR #792, opened 2016-01-13
THRIFT-3439 Run make cross using Python3 when available

nsuke:
Add Makefile.am change that was missing in the original patch.


bufferoverflow:
+1

PR #794, opened 2016-01-14
Update NodeClient.js

cleverchao:
NodeClient.js has some mistakes, 
When running this file, 
"var ThriftTransports = require('thrift/transport');
var ThriftProtocols = require('thrift/protocol');" 
will raise an error because the path passed in require is not correct.

### DIFF
--- a/tutorial/nodejs/NodeClient.js
+++ b/tutorial/nodejs/NodeClient.js
@@ -18,14 +18,12 @@
  */
 
 var thrift = require('thrift');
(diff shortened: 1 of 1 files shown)


nsuke:
@cleverchao thanks for the fix ! Could you create a JIRA issue for this ?

jeking3:
@jfarrell THRIFT-3904 and other, earlier commits already did the same thing, so this PR can be closed.

PR #797, opened 2016-01-17
THRIFT-3559 - Fix awkward sem-colons & newlines from THRIFT-3545

kdubb:
This essentially reverts THRIFT 3545 and uses simpler changes to fix the problem. These changes do not introduce the extra newlines and semi-colons.


kdubb:
@Jens-G ??

Jens-G:
kdubb!!

PR #804, opened 2016-01-18
THRIFT-3566: fixed TQTcpServerTest - it was never executed and working

szenker:


nsuke:
@szenker  thanks for the patches.
Please close this one and #802 as they are merged.

PR #809, opened 2016-01-19
THRIFT-3571 Make feature test result browsable

nsuke:
https://issues.apache.org/jira/browse/THRIFT-3571


bufferoverflow:
nice!
+1

PR #812, opened 2016-01-22
THRIFT-3574 Cocoa: use the right namespace for included generated .h files

stigsb:
The Cocoa generator currently uses the namespace of the including file in imports statements, instead of the included file.


kdubb:
This fix looks correct. Out of curiosity does it work with Pods frameworks option? I assume it will since you are just including another project file.

stigsb:
@kdubb yes, we are using the pods option

PR #817, opened 2016-01-23
THRIFT-3578 Make THeaderTransport detect TCompact framed and unframed

nsuke:
https://issues.apache.org/jira/browse/THRIFT-3578


lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -39,9 +39,11 @@
 
 enum CLIENT_TYPE {
   THRIFT_HEADER_CLIENT_TYPE = 0,
-  THRIFT_FRAMED_DEPRECATED = 1,
-  THRIFT_UNFRAMED_DEPRECATED = 2,
-  THRIFT_UNKNOWN_CLIENT_TYPE = 4,
+  THRIFT_FRAMED_BINARY = 1,
```
nsuke: @andrewcox @djwatson
Here I presume `_DEPRECATED` is only true for fbthrift.
Or is there any reason we should keep them and add the suffix to `COMPACT` ?

lib/cpp/src/thrift/transport/THeaderTransport.h
```diff
@@ -39,9 +39,11 @@
...
+  THRIFT_UNFRAMED_BINARY = 2,
+  THRIFT_FRAMED_COMPACT = 3,
+  THRIFT_UNFRAMED_COMPACT = 4,
```
nsuke: (Moving the comment to right position as it was misleading)
@andrewcox @djwatson
Here I presume _DEPRECATED is only true for fbthrift.
Or is there any reason we should keep them and add the suffix to COMPACT ?

PR #818, opened 2016-01-23
THRIFT-3579 Introduce retry to make cross

nsuke:
https://issues.apache.org/jira/browse/THRIFT-3579


nsuke:
oops, 5e432ec should have closed this, not #817

PR #819, opened 2016-01-23
THRIFT-3578 Make THeaderTransport detect TCompact framed and unframed

nsuke:
Closed previous one by mistake: #817


jsirois:
@nsuke - could you kill the 2 hung shards in https://travis-ci.org/apache/thrift/builds/106198835 ?
They >24hrs old and are blocking the Apache TravisCI pipeline afaict.

PR #820, opened 2016-01-23
THRIFT-3580 THeader for Haskell

nsuke:
Current Protocol type class had to be modified for reasons such as THeader needing two transports or protocol type dynamically changing on IO results.
I tried to make it similar to existing interface and most of the surface API should still work but signatures that involves Protocol will need to be updated.


jeking3:
Want to rebase on master and see if CI passes?

PR #821, opened 2016-01-24
THRIFT-2980 Accept external buffer in thrift_memory_buffer constructor

cjmay:
Here's a pass at letting thrift_memory_buffer take its buffer as a constructor argument:

Add buf (GByteArray pointer) and owner (gboolean) to constructor-only properties.  Create buf (object member) during construction if buf (constructor-only property) is NULL, the default.  Unref buf during finalize if buf is non-NULL and owner is TRUE.

Also change semantics of buf_size slightly: default is now zero, max is G_MAXUINT32, and zero is interpreted as no limit.  (My thrift objects are often very large; setting no limit as the default seems to simplify the usage in my opinion, and I believe raising the maximum is harmless.)

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/transport/thrift_memory_buffer.c
+++ b/lib/c_glib/src/thrift/c_glib/transport/thrift_memory_buffer.c
@@ -32,7 +32,9 @@
 enum _ThriftMemoryBufferProperties
 {
   PROP_0,
-  PROP_THRIFT_MEMORY_BUFFER_BUFFER_SIZE
+  PROP_THRIFT_MEMORY_BUFFER_BUFFER_SIZE,
+  PROP_THRIFT_MEMORY_BUFFER_BUFFER,
+  PROP_THRIFT_MEMORY_BUFFER_OWNER
 };
 
 G_DEFINE_TYPE(ThriftMemoryBuffer, thrift_memory_buffer, THRIFT_TYPE_TRANSPORT)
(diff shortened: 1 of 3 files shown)


lib/c_glib/src/thrift/c_glib/transport/thrift_memory_buffer.c
```diff
@@ -209,16 +242,39 @@ thrift_memory_buffer_class_init (ThriftMemoryBufferClass *cls)
 
   param_spec = g_param_spec_uint ("buf_size",
                                   "buffer size (construct)",
-                                  "Set the read buffer size",
+                                  "Set the read/write buffer size limit"
+                                  " (0 for unlimited)",
                                   0, /* min */
-                                  1048576, /* max, 1024*1024 */
-                                  512, /* default value */
+                                  G_MAXUINT32, /* max, 1024*1024 */
```
nsuke: nit: could you update or remove outdated part of the comment ?

lib/c_glib/src/thrift/c_glib/transport/thrift_memory_buffer.c
```diff
@@ -209,16 +242,39 @@ thrift_memory_buffer_class_init (ThriftMemoryBufferClass *cls)
 
   param_spec = g_param_spec_uint ("buf_size",
                                   "buffer size (construct)",
-                                  "Set the read buffer size",
+                                  "Set the read/write buffer size limit"
+                                  " (0 for unlimited)",
                                   0, /* min */
```
nsuke: ` " (0 for unlimited)",  0, /* min */` is confusing. Is the minimum actually unlimited ?

cjmay: That was my original intent, but now that you mention it I realize there's not actually an improvement over the size type (currently uint32) max.  Fixed.

lib/c_glib/src/thrift/c_glib/transport/thrift_memory_buffer.c
cjmay: Sorry about that, fixed.

PR #822, opened 2016-01-26
Fix THRIFT-2157 (generated code would cause ClassCastException)

mbreslow:
Ran into this issue as well and implemented the suggestions by Iker Jimenez and dvdeepankar on https://issues.apache.org/jira/browse/THRIFT-2157?focusedCommentId=14380813&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14380813


lib/java/src/org/apache/thrift/TSerializable.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.Serializable;
+
+import org.apache.thrift.protocol.TProtocol;
+
+/**
+ * Generic base interface for generated Thrift objects.
+ *
+ */
+public interface TSerializable extends Serializable {
```
bgould: I don't think this interface should extend java.io.Serializable.

Being able to serialize and deserialize with Thrift suggests nothing about an object's suitability to participate in standard Java serialization.

lib/java/src/org/apache/thrift/TApplicationException.java
```diff
@@ -73,7 +74,18 @@ public int getType() {
     return type_;
   }
 
-  public static TApplicationException read(TProtocol iprot) throws TException {
```
bgould: Technically this is a breaking change.  However this is not a method that would typically be used by applications I think, so the impact would probably be very low... and even for affected applications, it should be easy to fix (that is, create a new `TApplicationException` and call the read method).  It might be nice to add a static convenience method as a replacement (named `TApplicationException.readFrom(protocol)` or something like that)

mbreslow:
@bgould , thanks for the feedback. I implemented your suggestions.

lib/java/src/org/apache/thrift/TApplicationException.java
```diff
@@ -86,29 +98,29 @@ public static TApplicationException read(TProtocol iprot) throws TException {
...
+          }
+          break;
+        default:
           TProtocolUtil.skip(iprot, field.type);
-        }
-        break;
-      default:
-        TProtocolUtil.skip(iprot, field.type);
-        break;
+          break;
```
bgould: nitpick:  this whole section looks like it is just an indentation change

bgould:
Overall this looks like a useful change IMO.

PR #824, opened 2016-01-28
THRIFT-3584 To check element of self whether nil or not,avoid the value(false) of BOOL lost.

mygityf:
… BOOL is lost.

eg.
Before:
  if self.argBool then
    oprot:writeFieldBegin('argBool', TType.BOOL, 13)
    oprot:writeBool(self.argBool)
    oprot:writeFieldEnd()
  end

After:
  if self.argBool ~= nil then
    oprot:writeFieldBegin('argBool', TType.BOOL, 13)
    oprot:writeBool(self.argBool)
    oprot:writeFieldEnd()
  end

Why:
  the value(false) for paramater self.argBool will be lost when condition is 'if self.argBool then' as before.


mygityf:
Thank you.

nsuke:
@mygityf thanks for great contribution !
Could you create two issues in our JIRA for these changes ?

PR #825, opened 2016-01-29
THRIFT-3589 Pass references to field names in constructor through get_field_name

markerickson-wf:
Currently we are incorrectly handling fields that have default values and uppercase names, like `ID = -1`

@stevenosborne-wf
@ericklaus-wf
@tylertreat-wf

https://issues.apache.org/jira/browse/THRIFT-3589


stevenosborne-wf:
+1

PR #826, opened 2016-02-01
THRIFT-3584 Lua: To check element of self whether nil or not,avoid the value(false) of BOOL is lost.

mygityf:
…e value(false) of BOOL is lost.

eg.
    Before:
      if self.argBool then
        oprot:writeFieldBegin('argBool', TType.BOOL, 13)
        oprot:writeBool(self.argBool)
        oprot:writeFieldEnd()
      end
    
    After:
      if self.argBool ~= nil then
        oprot:writeFieldBegin('argBool', TType.BOOL, 13)
        oprot:writeBool(self.argBool)
        oprot:writeFieldEnd()
      end
    
    Why:
      the value(false) for paramater self.argBool will be lost when condition is 'if self.argBool then' as before.

### DIFF
--- a/compiler/cpp/src/generate/t_lua_generator.cc
+++ b/compiler/cpp/src/generate/t_lua_generator.cc
@@ -463,7 +463,9 @@ void t_lua_generator::generate_lua_struct_writer(ofstream& out, t_struct* tstruc
 
   indent(out) << "oprot:writeStructBegin('" << tstruct->get_name() << "')" << endl;
   for (f_iter = fields.begin(); f_iter != fields.end(); ++f_iter) {
-    indent(out) << "if self." << (*f_iter)->get_name() << " then" << endl;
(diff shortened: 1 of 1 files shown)


nsuke:
While it might fix a problem in sending `false` from Lua,
combined send + receive test does not seem to work even with this fix.
Do you want to fix reading side too ?

mygityf:
To read a field by field id instead of argument of 'self ' in reading side.
The problem can not be reproduced in reading side.

nsuke:
OK, "read" was not a precise word.
The problem I was trying to describe is about receiving boolean return value (not field) that happened to be `false`.
At a quick glance, it seems that `if result.success then` in generated code can be fixed in the same way as this.

If you're running the test mentioned in the other patch, you can uncomment the line `assertEqual(client:testBool(false), ...` and verify locally.

PR #828, opened 2016-02-01
THRIFT-3585 Lua: compact protocol implement of lua for thrift.

mygityf:


nsuke:
@mygityf thanks for tidying things up, will look into it soon.

lib/lua/src/luabpack.c
```diff
@@ -103,7 +104,7 @@ static int l_bunpack(lua_State *L) {
   const char *code = luaL_checkstring(L, 1);
   luaL_argcheck(L, code[1] == '\0', 0, "Format code must be one character.");
   const char *data = luaL_checkstring(L, 2);
-  size_t len = lua_rawlen(L, 2);
+  size_t len = lua_objlen(L, 2);
```
nsuke: This breaks Lua 5.2+. We need for example #ifdef against Lua version macro.

nsuke:
Tested with `test/lua/test_basic_client.lua` and found several problems.
Examples of problematic valus were negative i8 values (-128 ~ -1), some i32 values (0x40000000, -1 ...), i64 values.
For i64 values, we need it to work with `liblualongnumber.new`.

I recommend you to run the test yourself. Here's how:

1. pull master branch
2. `make precross`
3. in `test/go/bin` run `./testserver --protocol=compact --transport=framed` (leave it blocked)
4. uncommend compact entry around line 75 in `test/lua/test_basic_client.lua`. also add require to compact
5. in `test/lua` run `lua test_basic_client.lua --transport=framed --protocol=compact` with appropriate LUA_PATH and LUA_CPATH

mygityf:
Tested with test/lua/test_basic_client.lua and test_basic_server.lua, result is expected.
new comment(df693bf) added for the bug.

PR #831, opened 2016-02-02
THRIFT-3595 Perl: Unify implementation of set as a hashref

adamu:
https://issues.apache.org/jira/browse/THRIFT-3595

### DIFF
--- a/test/perl/TestClient.pl
+++ b/test/perl/TestClient.pl
@@ -250,16 +250,16 @@ print("}\n");
 #
 # SET TEST
 #
-my $setout = [];
+my $setout = {};
 for (my $i = -2; $i < 3; ++$i) {
-    push(@$setout, $i);
+    $setout->{$i} = 1;
 }
 
-print("testSet({".join(",",@$setout)."})");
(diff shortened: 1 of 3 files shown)


test/perl/TestClient.pl
```diff
@@ -250,16 +250,16 @@ sub usage {
 #
 # SET TEST
 #
-my $setout = [];
+my $setout = {};
 for (my $i = -2; $i < 3; ++$i) {
-    push(@$setout, $i);
+    $setout->{$i} = 1;
```
amillerchip: It's just been pointed out to me this this will not work if `$i` is an object, because it will get serialised to its string representation when used as the hash key like this.

jeking3: Do you have a follow-up to this?

jeking3:
I have an open backlog item to use Set::Scalar for perl sets.  https://issues.apache.org/jira/browse/THRIFT-3190

jeking3:
Closing based on 2 years on inactivity; superceded by THRIFT-3190

PR #834, opened 2016-02-04
THRIFT-2103 [python] Support for SSL certificates with Subject Altern…

nsuke:
…ative Names

https://issues.apache.org/jira/browse/THRIFT-2103


nsuke:
25536ad83a85cfda6d5388278e4e378f2d4df73e

PR #835, opened 2016-02-04
THRIFT-3599 Validate client IP address against cert's SubjectAltName

nsuke:
https://issues.apache.org/jira/browse/THRIFT-3599


nsuke:
f39f7dbd26fe090f0fc6566c100ca7adc9ace714

PR #840, opened 2016-02-08
THRIFT-3112 [Java] AsyncMethodCallback should be typed in generated AsyncIface

jsirois:
The parametrization brings the existing actual parametrization with
client call implementation objects to the fore and so this change
also fixes that parametrization to be a simple parametrization over
the return type as is done in the server-side AsyncProcessor code.

NB: This is a breaking change in both generated code and the client
libs.

### DIFF
--- a/lib/java/src/org/apache/thrift/async/TAsyncMethodCall.java
+++ b/lib/java/src/org/apache/thrift/async/TAsyncMethodCall.java
@@ -33,11 +33,15 @@ import org.apache.thrift.transport.TNonblockingTransport;
 import org.apache.thrift.transport.TTransportException;
 
 /**
- * Encapsulates an async method call
+ * Encapsulates an async method call.
+ * <p>
  * Need to generate:
- *   - private void write_args(TProtocol protocol)
- *   - public T getResult() throws <Exception_1>, <Exception_2>, ...
- * @param <T>
+ * <ul>
+ *   <li>protected abstract void write_args(TProtocol protocol)</li>
+ *   <li>protected abstract T getResult() throws <Exception_1>, <Exception_2>, ...</li>
+ * </ul>
+ *
+ * @param <T> The return type of the encapsulated method call.
  */
 public abstract class TAsyncMethodCall<T> {
 
@@ -112,6 +116,8 @@ public abstract class TAsyncMethodCall<T> {
   }
 
   protected abstract void write_args(TProtocol protocol) throws TException;
+
+  protected abstract T getResult() throws Exception;
 
   /**
    * Initialize buffers.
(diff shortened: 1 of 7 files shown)


jsirois:
NB: This is a straw-man at this point that picks up the work started in #500 by @mreve and whose merits is being discussed here: http://markmail.org/message/k77l6helvkh4igrv

Please do not merge pending results of that discussion - this is a breaking change for async client users.

lib/java/src/org/apache/thrift/async/TAsyncMethodCall.java
```diff
@@ -225,8 +231,13 @@ private void cleanUpAndFireCallback(SelectionKey key) {
     key.interestOps(0);
     // this ensures that the TAsyncMethod instance doesn't hang around
     key.attach(null);
```
nsuke: AFAIK, key is not removed from `selector.keys()` unless `key.cancel()` is called, so `interestOps(0)` and `attach(null)` is not enough (maybe wrong assumption ?).
Do you happen to have any idea why the original code does not call `cancel()` for the success path here ?

As your intention here seems to be to prevent `key.cancel()` in `transition()` for the error path, I thought you might know about it.

nsuke: maybe it's just a bug in the original code ?
Then I think we can simply add cancel() and stop key set from growing indefinitely without doing any harm at all.
What do you think ?

jsirois: Adding cancel won't (doesn't) work.  The key is associated with the single underlying channel (wrapped in a `TNonblockingTransport`) from the `TAsyncClient` which is intended for use with multiple concurrent calls afaict.  The `org.apache.thrift.async.TestTAsyncClientManager#testParallelCalls` test uses the client in this way - 1 client, multiple concurrent calls - and demonstrates that cancelling the key cancels the underlying channel and, for example, you get 1 successful call and 4999 failed calls in the 5000 call test case as a result.

jsirois: NB: I still don;t fully understand this code, but you've got me reading it more closely.  I'll report back when I fully grok what's going on here in the pre-existing code.

nsuke: It turned out that it's one key per connection.
So indeed canceling here does not work at all, it spoils the connection after the first call.
Instead, cancel() should be called when the connection is closed but is not unless aborted by exceptions, so key objects are certainly leaking.

As it now seems a completely separate problem so I'll stop pulling it here and focus on this patch for now.
Can you explain as to why you are introducing try catch anyway ?
It's the last bit that stops me from going forward.

nsuke: Another implication of the current (pre-existing) design is that the client cannot handle unordered or even queued calls.
IMO it should but that would need major rewrite.
(To be clear, testParallelCalls has bunches of concurrent clients each of which does sequential calls)

nsuke: I started working on another async issue (THRIFT-3680) on top of your change.
As I actually ran your code, I think it immediately became evident that the try-catch is simply for new checked exception, period.
In that case, I'd like to put cancel() to the **catch** part to restore the original behavior.
Any concern for that ?

PR #841, opened 2016-02-10
lib/cpp/test/SecurityTest is flaky in jenkins Thrift-precommit build.

jsirois:
Previously, the server socket might not be fully closed from a prior
test, leading to flaky port bind failures in CI.  Avoid the issue
altogether by binding an ephemeral port.

Add a TODO to TestPortFixture.h to warn folks away from its usage
and link THRIFT-3609 to track the work of removing or replacing it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,7 @@ erl_crash.dump
 /lib/c_glib/test/testbufferedtransport
 /lib/c_glib/test/testcontainertest
 /lib/c_glib/test/testdebugproto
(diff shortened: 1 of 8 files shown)


jsirois:
Hrm, although it looks like Travis-CI will go green with this change, Jenkins still flaked out.  More investigation needed.  This PR will get a bit noisy with debug attempts and I may close it after debugging the issue and start a fresh PR + patch with the ultimate fix.

PR #848, opened 2016-02-11
lib/cpp/test/SecurityTest is flaky in jenkins Thrift-precommit build.

jsirois:
Previously, the deprecated `linux` symbol was tested, but under both g++
and clang++ using `-std=c++11` this deprecated symbol is not defined.
Test the accepted modern standard of `__linux__` instead to ensure the
SIGPIPE ignore is in-effect and the test runs stably.

PR #850, opened 2016-02-12
lib/py/test/test_sslsocket.py is flaky

jsirois:
Previously a combination of fixed ports, fixed paths and delays was used
in all TSSLSocketTest tests that involved making a client-server
connection; now ephemeral ports, unique tmp files and no delays for
successful connection tests are all implemented. A delay still remains
for the failed connection tests to allow for SSL handshake initiation
but not wait too long.

### DIFF
--- a/lib/py/test/test_sslsocket.py
+++ b/lib/py/test/test_sslsocket.py
@@ -155,6 +155,7 @@ class TSSLSocketTest(unittest.TestCase):
     def test_deprecation(self):
         if not six.PY3:
             # The checks below currently only work for python3.
+            # See: https://issues.apache.org/jira/browse/THRIFT-3618
             print('skiping test_deprecation')
             return
 
--- a/lib/py/test/test_sslsocket.py
+++ b/lib/py/test/test_sslsocket.py
@@ -78,7 +78,7 @@ class ServerAcceptor(threading.Thread):
             self._client_accepted.set()
 
     def await_listening(self):
-        self._port_bound.wait()
+        self._listening.wait()
 
     @property
     def port(self):
--- a/lib/py/test/test_sslsocket.py
+++ b/lib/py/test/test_sslsocket.py
@@ -77,17 +77,17 @@ class ServerAcceptor(threading.Thread):
         finally:
             self._client_accepted.set()
 
-    def await_listening(self, timeout=None):
-        self._port_bound.wait(timeout=timeout)
+    def await_listening(self):
+        self._port_bound.wait()
 
     @property
-    def port(self, timeout=None):
-        self._port_bound.wait(timeout=timeout)
+    def port(self):
+        self._port_bound.wait()
         return self._port
 
     @property
-    def client(self, timeout=None):
-        self._client_accepted.wait(timeout=timeout)
+    def client(self):
+        self._client_accepted.wait()
         return self._client
 
 
--- a/lib/py/test/test_sslsocket.py
+++ b/lib/py/test/test_sslsocket.py
@@ -28,6 +28,8 @@ import unittest
 import warnings
 from contextlib import contextmanager
 
+import six
+
 import _import_local_thrift
 from thrift.transport.TSSLSocket import TSSLSocket, TSSLServerSocket
 from thrift.transport.TTransport import TTransportException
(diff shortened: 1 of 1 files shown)


jsirois:
@nsuke - hoping you have time to review this one

jsirois:
Noting that Jenkins CI failed on lib/cpp/test/SecurityTest.cpp before getting to test_sslsocket.py.
I have a fix for the SecurityTest in #848 that still needs review attention so I'll mark THRIFT-3616 as blocked on THRIFT-3608

lib/py/test/test_sslsocket.py
```diff
@@ -75,185 +104,176 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-            if acc.client:
-                acc.client.close()
-            server.close()
+            try:
+                self.assertTrue(acc.client is not None)
+            finally:
+                client.close()
 
     # deprecated feature
     def test_deprecation(self):
```
nsuke: For some reason, deprecation count now differs between Python 2 and Python 3
https://travis-ci.org/apache/thrift/jobs/108805112#L2383

>    File "test/test_sslsocket.py", line 164, in test_deprecation
>        self.assertEqual(len(w), 7)
>    AssertionError: 6 != 7

I've found another issue on deprecate messages so I'll need to modify this part shortly after, after all.
So you can just skip this with if + print for now.

lib/py/test/test_sslsocket.py
```diff
@@ -75,185 +104,176 @@ def __exit__(self, exc_type, exc_value, traceback):
...
             acc.start()
-            time.sleep(CONNECT_DELAY / 2)
-            client.setTimeout(CONNECT_TIMEOUT / 2)
-            with self._assert_raises(Exception):
-                logging.disable(logging.CRITICAL)
-                client.open()
-                select.select([], [client.handle], [], CONNECT_TIMEOUT / 2)
-            # self.assertIsNone(acc.client)
-            self.assertTrue(acc.client is None)
+            acc.await_listening()
```
nsuke: You've merged `_port_bound.wait` to `port` property so this method is no longer needed ?

lib/py/test/test_sslsocket.py
```diff
@@ -42,22 +43,50 @@
...
+        try:
+            self._client = self._server.accept()
+        finally:
+            self._client_accepted.set()
+
+    def await_listening(self, timeout=None):
+        self._port_bound.wait(timeout=timeout)
+
+    @property
+    def port(self, timeout=None):
```
nsuke: property argument is unusable, but fortunately not needed either.
Could you remove this ?

lib/py/test/test_sslsocket.py
```diff
@@ -42,22 +43,50 @@
...
+        self._port_bound.wait(timeout=timeout)
+        return self._port
+
+    @property
+    def client(self, timeout=None):
+        self._client_accepted.wait(timeout=timeout)
```
nsuke: same as above

lib/py/test/test_sslsocket.py
jsirois: This is used below on line 117

lib/py/test/test_sslsocket.py
jsirois: Used on lines 134 and 148 below.

lib/py/test/test_sslsocket.py
jsirois: Its needed for the unix domain socket case which binds no port, but must wait for the server to be up and listening.

lib/py/test/test_sslsocket.py
jsirois: will do - diff coming presently

nsuke: Just FIY, this is the issue https://issues.apache.org/jira/browse/THRIFT-3618

lib/py/test/test_sslsocket.py
nsuke: Oops, edit was too late, please check the above out.

jsirois: Aha - gotcha, will do.  One more diff coming.

lib/py/test/test_sslsocket.py
jsirois: Added a link to the issue in the comment - thanks!

lib/py/test/test_sslsocket.py
nsuke: Ah, I think I finally got it.
Maybe you mean `self._listening.wait(timeout=timeout)` rather than `_port_bound` here ?
I missed your intention of waiting for `listen()` rather than `getsockname()`
Could you change it to _listening or remove the _listening that is unused ?

jsirois: Aha - that was copypasta - fixed to make `await_listening` use the `_listening` `Event` which was the intention all along.

PR #854, opened 2016-02-13
THRIFT-3623 Fix cpp/lib/test/TSSLSocketInterruptTest.cpp to use ephem…

jsirois:
…eral ports.

Instead of using a fixed port, use an ephemeral port to improve
robustness and make way for parallelizability.

This change has an undesirable amount of repeated code but this fact is
not different from the prior so being a bit expedient here.  Ideally
setup of a `TSSLServerSocket` listening on an ephemeral port would be
DRYed up into a helper of some sort.

### DIFF
--- a/lib/cpp/test/TSSLSocketInterruptTest.cpp
+++ b/lib/cpp/test/TSSLSocketInterruptTest.cpp
@@ -38,6 +38,8 @@ using apache::thrift::transport::TTransport;
 using apache::thrift::transport::TTransportException;
 using apache::thrift::transport::TSSLSocketFactory;
 
+BOOST_AUTO_TEST_SUITE(TSSLSocketInterruptTest)
+
 boost::filesystem::path keyDir;
 boost::filesystem::path certFile(const std::string& filename)
 {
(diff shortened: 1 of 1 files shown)


lib/cpp/test/TSSLSocketInterruptTest.cpp
```diff
@@ -91,8 +90,6 @@ BOOST_GLOBAL_FIXTURE(GlobalFixtureSSL);
 BOOST_GLOBAL_FIXTURE(GlobalFixtureSSL)
 #endif
 
-BOOST_FIXTURE_TEST_SUITE(TSSLSocketInterruptTest, TestPortFixture)
```
nsuke: Can you keep 3 of these as BOOST_**AUTO**_TEST_SUITE ?

This would remove a namespace and put everything into global namespace.
In this regard, it seems that this one should have been placed before the `keyDir` symbol already...

jsirois: Thanks @nsuke - all fixed up here and in #855 and #857

PR #860, opened 2016-02-13
THRIFT-3622 Fix deprecated uses of std::auto_ptr

jsirois:


nsuke:
Really unlucky for us that Travis builds aren't invoked for this PR.
We still need C++98 so cannot use `unique_ptr` yet, so we're using `boost::scoped_ptr` in such cases.

nsuke:
It seems that our Autools build does not use -isystem to include boost (i.e. using -I), so I'm afraid that we cannot remove the warnings by using scoped_ptr.
Because last time I checked, scoped_ptr was using auto_ptr inside their code and there was no workaround.

But it would still remove them from CMake build.

jsirois:
Thanks for taking a look @nsuke, I'm going to abandon this change until I understand the c/c++ build landscape better.

PR #863, opened 2016-02-14
THRIFT-3631 Lua-This is JSON protocol implement for thrift-lua library.

mygityf:
TJsonProtocol.lua is new source of lua for thrift JSON protocol.
Test result is expected of test_basic_server.lua, test_basic_client.lua.
Test cases:
client            <-------------->            server
json_protocol_client.lua                 json_server.go (golang thrift server)
json_client.go (golang thrift client) json_protocol_server.lua
json_protocol_client.lua                 json_protocol_server.lua
test_basic_server.lua                    test_basic_client.lua


nsuke:
@mygityf thanks for another contribution !
But it really needs cleanup.
Does it make sense to you if I say "rebase to the current master and keep only the THRIFT-3631  commit" ?
If you have really no idea, just paste the output of `git remote -v` and I'll tell you exactly what commands to run.

After that, could you do the same as ebd71cedc3f7c641c76abc2ec13119465c5b02a6 to enable TJSONProtocol test ?

Jens-G:
An extra test is not bad. But integrating into the standard ThriftTest suite similar to all other implementations would be awesome to let it become part of "make cross".

nsuke:
@Jens-G yeah, that's what is happening here.
Lua is recently added to make cross, although in a very limited client only state.
So json protocol client cross tests can be enabled by adding 3 lines of code.

mygityf:
pull request is moved to https://github.com/apache/thrift/pull/864 #864

PR #864, opened 2016-02-14
THRIFT-3631 Lua- JSON protocol implement for thrift-lua library.

mygityf:
THRIFT-3631 Lua-This is JSON protocol implement for thrift-lua library.
TJsonProtocol.lua is new source of lua for thrift JSON protocol.

### DIFF
--- a/lib/lua/TJsonProtocol.lua
+++ b/lib/lua/TJsonProtocol.lua
@@ -1,0 +1,725 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements. See the NOTICE file
+-- distributed with this work for additional information
+-- regarding copyright ownership. The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"), you may not use this file except in compliance
+-- with the License. You may obtain a copy of the License at
+--
+--   http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied. See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+require 'TProtocol'
+require 'libluabpack'
+require 'libluabitwise'
+
+TJSONProtocol = __TObject.new(TProtocolBase, {
+  __type = 'TJSONProtocol',
+  THRIFT_JSON_PROTOCOL_VERSION = 1,
+  jsonContext = {},
+  jsonContextVal = {first = true, colon = true, ttype = 2, null = true},
+  jsonContextIndex = 1,
+  hasReadByte = ""
+})
+
+TTypeToString = {}
+TTypeToString[TType.BOOL]   = "tf"
+TTypeToString[TType.BYTE]   = "i8"
+TTypeToString[TType.I16]    = "i16"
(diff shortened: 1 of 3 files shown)


mygityf:
@nsuke 
I move THRIFT-3631 to new branch and keep only the THRIFT-3631 commit in this pull request.

nsuke:
@mygityf the tests passed against all but C++ server.
https://travis-ci.org/apache/thrift/jobs/109138968#L8071

Could you look into the failures ?
You can run the specific part by

    test/test.py --regex "cpp.*lua.*json"

and can find log in `test/log/unexpected_failures.log`

mygityf:
The precision is 15 bits after dot of format casting from double to string in Thrift-cpp-library- json-protocol.
But the precision is 16 bits after dot in Thrift-lua-library-json-protocol.
So, test with C++ server is break.
@nsuke can we modify the precision to 16 bits for C++ server?
The solution:
To Change code 'str.precision(std::numeric_limits<double>::digits10 + 1);' to 'str.precision(std::numeric_limits<double>::digits10 + 2);'
in function doubeToString file TJSONProtocol.cpp at line 524,
e.g:
before:
In C++ TJsonProtocol.cpp:
double a = 1.12345678906666663;
string astr = doubleToString(a);
double b = stringToDouble(astr);
the result as below:
a = 1.1234567890666667
astr = "1.123456789066667"
b = 1.1234567890666669
after changing:
the result as below:
a = 1.1234567890666667
astr = "1.1234567890666667"
b = 1.1234567890666667
This result is expected.

mygityf:
The solution to the test failure:
pull request for C++ thrift https://github.com/apache/thrift/pull/870.

nsuke:
I've tested this on my local env but i8(byte) and i32 values are encoded like 1.0 instead of 1, and naturally the receiver side fails to handle this.
I'm using Lua 5.3.2.
Would be nice if you add a workaround to force them to be integer strings.

lib/lua/TJsonProtocol.lua
```diff
@@ -0,0 +1,725 @@
...
+    else
+      self:writeJSONEscapeChar(ch)
+    end
+  end
+end
+
+function TJSONProtocol:writeJSONString(str)
+  self:writeElemSeparator()
+  self.trans:write(JSONNode.StringDelimiter)
+  -- TODO
```
nsuke: Can you elaborate on this ?

mygityf: It can be escaped by being preceded with a backslash,
like " to \", \t to \\t, \r to \\r, \f to \\f, \n to \\n , \b to \\b.
Otherwise, character " will break JSON format.
e.g:
To be sent string is "Begin:This is Send\"String\b\t\n\r\f:End."

mygityf:
@nsuke issue  of i8(byte) and i32 values are encoded like 1.0 instead of 1.
It is reproduced in receiver side or sender side or both side?

nsuke:
sender side.
used lua client against cpp / py server and extracted the json in the packet with tcpdump.

lib/lua/TJsonProtocol.lua
nsuke: thanks. will include this info to the comment.

mygityf:
@nsuke Integer is convert to string without dot. can you try again in your env to check whether it work well or not, thanks.

nsuke:
Nice, the last commit fixed the problem. I guess it's a Lua 5.3 issue.

nsuke:
please close this PR as it's merged.

PR #869, opened 2016-02-14
THRIFT-3609 Remove or replace TestPortFixture.h.

jsirois:
Removed since it is no longer used.


nsuke:
Great, you've killed them all. Will definitely merge it soon.

PR #870, opened 2016-02-15
THRIFT-3636 The precision is 15 bits after dot of casting from double to string in Thrift-cpp-library- json-protocol.

mygityf:
Cpp json protocol double precision should be 16 bits


nsuke:
thanks for uncovering it !
The "2" is for "0." ?
I think the comment can be improved.
@hcorg maybe you have better alternative than code comment here ?

mygityf:
The "1" is for "0.123456789012345".
The "2" is for "0.1234567890123456".

nsuke:
ok, so 2 is the difference between max_digits10 and digits10.
since we cannot use c++11 max_digits10 yet, I suggest using 17 directly with a comment or a clearly named local variabe.

mygityf:
using 17 directly with a comment.

mygityf:
It looks no error detected for Jenkins checking.

nsuke:
Actually test is failing quite badly.
Your fix is fine but the test needs fix too.
A possible fix is to replace the PI and golden ratio with float literals and use the same digits in the assertions.

lib/cpp/src/thrift/protocol/TJSONProtocol.cpp
```diff
@@ -524,7 +524,8 @@ namespace {
 std::string doubleToString(double d) {
   std::ostringstream str;
   str.imbue(std::locale::classic());
-  str.precision(std::numeric_limits<double>::digits10 + 1);
+  // new precision is 17 equal to max_digits10(c++11) or digits10+2
+  str.precision(17);
```
hcorg: IMHO
`const double max_digits10 = 2 + std::numeric_limits<double>::digits10;`
`str.precision(max_digits10);`
looks better than comment ;)

I'd even consider naming variable `max_digits10_cpp11`

mygityf: It looks better than before.
Thanks.

mygityf:
@nsuke I replace the PI and golden ratio with float literals and use the same digits(17) in file lib/cpp/test/JSONProtoTest.cpp.
The result looks OK in Jenkins checking.
why CI build failed.

nsuke:
@mygityf the failure is unrelated to your patch so no need to worry about that any more.

Would you mind if I remove the comment as per @hcorg 's suggestion, now that we have nicely self explanatory variable ?

mygityf:
@nsuke just do it to make it better.

PR #871, opened 2016-02-17
THRIFT-3639 C# Thrift library forces TLS 1.0, thwarting TLS 1.2 usage

nsuke:


nsuke:
This enables passing which TLS version to use as a constructor argument.
Note that the default value is still SslProtocols.Tls because we're still stuck at .NET 3.5 and equivalent Mono right now.

PR #881, opened 2016-02-21
THRIFT-3650 union (and optionals) serialization

Jens-G:
Contains all fixes for THRIFT-3650, 3652, 3654 and 3656 (4 patches)


markerickson-wf:
+1 on the Dart fix, thanks

PR #884, opened 2016-02-21
THRIFT-3657 D TFileWriterTransport close should use non-priority send

nsuke:


matyapiro31:
I think the change on Dockerfile must be irrelevant to version  updating of debian/ubuntu package.

jeking3:
@nsuke a reminder to rebase and resubmit

jeking3:
@nsuke a reminder to rebase and submit, and I would limit it to just the lib/d/src/thrift/transport/file.d file.

PR #895, opened 2016-02-26
[THRIFT-3689] thrift_reconnecting_client start failed when server is not available

laozhp:
When first time reconnect without a success call, line 214: Backoff = 2 * R, will cause exception, because R is undefined. Set reconn_time's default value to 0 resolved this.

### DIFF
--- a/lib/erl/src/thrift_reconnecting_client.erl
+++ b/lib/erl/src/thrift_reconnecting_client.erl
@@ -43,7 +43,7 @@
                   thrift_opts,
                   reconn_min,
                   reconn_max,
-                  reconn_time,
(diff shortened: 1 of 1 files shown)


nsuke:
@laozhp thanks for the fix ! Could you create an issue in our JIRA ?
https://github.com/apache/thrift/blob/master/CONTRIBUTING.md

laozhp:
@nsuke OK, I have created an issue just now, [THRIFT-3689] thrift_reconnecting_client start failed when server is not available

nsuke:
@laozhp, it indeed is a problem and the fix looks good.
Merging shortly after, thanks again !

PR #896, opened 2016-02-26
THRIFT-3676

ngrewe:
This PR adds proper error handling to the generated Cocoa server implementations. To find which err property in the result object we need to set the error to, we check have to check the type of the error at runtime, which is a consequence of using NSError*\* arguments to pass things around.


jeking3:
Hi, would you be able to rebase on the current master and squash the commits, and force push?  This will restart the CI build with our current codebase so we know the change is still applicable and passes all tests.  Thanks.

jeking3:
I think this should be closed and resubmitted, the history got messed up.

PR #904, opened 2016-02-27
THRIFT-3686 Java processor should report internal error on uncaught exception

nsuke:


jeking3:
This should be rebased against upstream/master since there are conflicts, and resubmitted; also see THRIFT-1805 as there is some discussion on what is okay and what is not in this context.

jeking3:
@nsuke This should be rebased against master since there are conflicts, and resubmitted; also see THRIFT-1805 as there is some discussion on what is okay and what is not in this context.

PR #916, opened 2016-02-29
THRIFT-3637: Dart compact protocol

brianshannan-wf:
https://issues.apache.org/jira/browse/THRIFT-3637


nsuke:
@brianshannan-wf @markerickson-wf  could you look into 3 make cross failures ?
It may be http servers and haskell's faults so in that case we can put these three to the test/known_failures.

markerickson-wf:
@nsuke Thanks, it looks like it is just these three combinations.  Other combinations involving dart + compact and the same languages / transports are passing.  We'll look into it.

https://issues.apache.org/jira/browse/THRIFT-3702

PR #917, opened 2016-03-01
THRIFT-3702 Fix cross tests for Dart compact protocol (3 failing)

markerickson-wf:
I looked at Go-Dart locally, and the response body from the Go server appears to be empty.  Go server - Java client also fails for me for http.  So without much more time to investigate, I think the best course of action is to add these to known_failures_Linux.json to allow CI to pass.  There are already other combinations excluded for those server - transport pairs.

I did some cleanup in the Dart test client while I was in there.

go-dart                 compact      http-ip                  failure(13)
hs-dart                 compact      framed-ip                failure(13)
cpp-dart                compact      http-ip                  failure(13)

https://issues.apache.org/jira/browse/THRIFT-3702

@brianshannan-wf
@stevenosborne-wf
@tylertreat-wf


brianshannan-wf:
+1

tylertreat-wf:
+1

PR #919, opened 2016-03-04
THRIFT-3465 Go Code With Complex Const Initializer Compilation Depends On Struct Order

tylertreat:
https://issues.apache.org/jira/browse/THRIFT-3465
https://issues.apache.org/jira/browse/THRIFT-3553

@markerickson-wf 
@tomdeering-wf
@mattsanders-wf


tylertreat:
This appears to be an issue with forward usages of types, but I haven't been able to figure out a proper solution. Closing for now...

PR #927, opened 2016-03-06
THRIFT-3706 : Implement protocol multiplexor in c_glib [Rebased Master]

gadLinux:
This patch implements protocol multiplexor for the client. For the server there's a need for a processor that currently we don't have. So a test of this can only be tested against other language that already has this in place. We did testing against Java.

Example of use:

/*

```
We will use binary encoding
*/
ThriftBinaryProtocol *binary_protocol= g_object_new (THRIFT_TYPE_BINARY_PROTOCOL,
"transport", manager->transport,
NULL);
```

/*

```
FIXME Transport in the multiplexed should be got from binary_protocol
*/
ThriftMultiplexedProtocol *multiplexed_protocol= g_object_new (THRIFT_TYPE_MULTIPLEXED_PROTOCOL,
"transport", manager->transport,
"protocol", binary_protocol,
"service-name", "MyService",
NULL);
```

### DIFF
--- a/lib/c_glib/Makefile.am
+++ b/lib/c_glib/Makefile.am
@@ -35,6 +35,7 @@ libthrift_c_glib_la_SOURCES = src/thrift/c_glib/thrift.c \
                               src/thrift/c_glib/processor/thrift_processor.c \
                               src/thrift/c_glib/processor/thrift_dispatch_processor.c \
                               src/thrift/c_glib/protocol/thrift_protocol.c \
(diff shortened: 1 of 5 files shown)


gadLinux:
Sorry for the previous attempt I did so horrible wrong I had to overwrite my own history.

gadLinux:
@nsuke I rebased the patch here. You asked for. Do you see any problems to integrate?

gadLinux:
I don't know if the ad0bf54must be here or not.

gadLinux:
I will redo it

PR #929, opened 2016-03-07
Fix problem on #THRIFT-3370

gadLinux:
This is a rebased master of the thrift-3370.


nsuke:
@gadLinux the patch is already [merged](b343feee0d048e214c93bd78560af9185531c094) but without the comment, as noted in the JIRA issue.

gadLinux:
Hi @nsuke, 

I see I just rebased everything in case it was not. I close this then.

PR #930, opened 2016-03-07
THRIFT-3369 : Implement SSL/TLS support on C with c_glib

gadLinux:
This is a proper pull request for this open issue. The patch that's on the issue is no longer valid and migrated everything to this pull request. 

Again the server part is missing so some tests are done over published SSL server and are incomplete.

Requires ssl initialization:

```
    thrift_ssl_socket_initialize_openssl();
```

and finalization
        thrift_ssl_socket_finalize_openssl();

And can set pinning manager. Thought one is provided. 

```
        thrift_ssl_socket_set_manager(manager.socket, my_access_manager);// Install pinning manager

        read_from_file(cert_buffer, 65534, "./owncloud.level2crm.pem");
        if(!thrift_ssl_load_cert_from_buffer(manager.socket, cert_buffer)){
           g_warning("Certificates cannot be loaded!");
        }
```

Good example of client creation can be:

```
  if(ssl_proto>=0 && ssl_proto<=LATEST){
    /*
     * We will communicate using sockets
     */
    int cypher=ssl_proto;
    do{
        manager->socket = thrift_ssl_socket_new_with_host(cypher, host, port, error);
        if((*error!=NULL && (*error)->code==THRIFT_SSL_SOCKET_ERROR_CIPHER_NOT_AVAILABLE)){
            fprintf (stderr, "Cypher not available: %s\n", (*error)->message);
            g_error_free (*error);
            (*error)=NULL;
            cypher++;
        }else{
            cypher = LATEST;
        }
    }while(manager->socket==NULL && cypher<=LATEST);
    if(manager->socket==NULL && cypher>=LATEST){
        if((*error)==NULL){
            g_set_error (error,
                    THRIFT_TRANSPORT_ERROR,
                    THRIFT_SSL_SOCKET_ERROR_CIPHER_NOT_AVAILABLE,
                    "no cypher overlap");
        }
    }
}
```


nsuke:
OpenSSL link seems to be failing on CI. You may want to take a look at lib/cpp/Makefile.am for a working setup.
Also, could you add these files and link settings to `CMakeLists.txt` too ?

gadLinux:
I'm sorry but don't know how to do it for CMakeLists.txt

gadLinux:
@nsuke Can you take a look to this patch? I think it's important that Thrift has SSL/TLS in version 1.0

lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
```diff
@@ -0,0 +1,700 @@
...
+	gint ret = 0;
+	guint sent = 0;
+	if(THRIFT_SSL_SOCKET_GET_CLASS(ssl_socket)->handle_handshake(transport, error)){
+
+		BIO* bio = SSL_get_wbio(ssl_socket->ssl);
+		if (bio == NULL) {
+			g_set_error (error, THRIFT_TRANSPORT_ERROR,
+					THRIFT_TRANSPORT_ERROR_SEND,
+					"failed to flush, wbio returned null");
+			//			    throw TSSLException("SSL_get_wbio returns NULL");
```
nsuke: Shall we return here ?

lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
```diff
@@ -0,0 +1,700 @@
...
+		}
+		if (BIO_flush(bio) != 1) {
+			g_set_error (error, THRIFT_TRANSPORT_ERROR,
+					THRIFT_TRANSPORT_ERROR_SEND,
+					"failed to flush it returned error");
+			//			    int errno_copy = THRIFT_GET_SOCKET_ERROR;
+			//			    string errors;
+			//			    buildErrors(errors, errno_copy);
+			//			    throw TSSLException("BIO_flush: " + errors);
```
nsuke: Please remove comments that are not meant for documentation etc.

lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
```diff
@@ -0,0 +1,700 @@
...
+
+static void thrift_ssl_socket_static_locking_callback(int mode, int n, const char* unk, int id) {
+	if (mode & CRYPTO_LOCK) {
+		//		g_printf("We should lock thread %d\n", n);
+		//mutexes[n].lock();
+	} else {
+		//		g_printf("We should unlock thread %d\n", n);
+		//mutexes[n].unlock();
+	}
+}
```
nsuke: Why you commented out entire method body ?

lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
```diff
@@ -0,0 +1,700 @@
...
+
+static void* thrift_ssl_socket_dyn_lock_create_callback(const char* unk, int id) {
+	g_print("We should create a lock\n");
+	return NULL;
```
nsuke: AFAIK, dynamic locking is optional. If we're not doing it, maybe we shouldn't register callbacks at all.
We need static locking above before merging, though.

lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
```diff
@@ -0,0 +1,700 @@
...
+		thrift_ssl_socket_global_context = NULL;
+	}
+
+	if (!thrift_ssl_socket_openssl_initialized) {
+		return;
+	}
+	thrift_ssl_socket_openssl_initialized = FALSE;
+#if (OPENSSL_VERSION_NUMBER < OPENSSL_VERSION_NO_THREAD_ID)
+	CRYPTO_set_id_callback(NULL);
+#endif
```
nsuke: It's original cpp lib TSSLSocket problem but we're supposed to call CRYPTO_THREADID_set_callback for `#else` path.
It does not seem to cause any problem on most platforms, according to the doc though.

lib/c_glib/test/owncloud.level2crm.pem
```diff
@@ -0,0 +1,59 @@
...
+IkxldmVsMiBTb2x1Y2lvbmVzIE1hcmtldGluZyAzNjAgU0wxETAPBgNVBAsTCFNp
+c3RlbWFzMQ8wDQYDVQQDEwZMZXZlbDIxJTAjBgkqhkiG9w0BCQEWFmdhZ3VpbGFy
+QGxldmVsMmNybS5jb22CCQCowoJ4Kn/b7DAMBgNVHRMEBTADAQH/MAsGA1UdDwQE
+AwIBBjANBgkqhkiG9w0BAQsFAAOCAQEAh2kPzv8jFNQbCrAD770m8T4W/KMhxMOv
+Uhm/wSv7/g3Ir1pZ9LgXAWJx1qLQdCcOikBparkUpHIrqDU+wBL6o+CFNWp7WR6g
+bft627bL9R3z3fTQHIchjPHQ7S4ID/STWfTqWDJQESBnTp1KTuJ8sMlvACMWZSUc
+jZOAQIodNiUUNPkfreurPb9KH/nArwV/yTDoZMhiqoqNwOIY/0sxUJaNUAX3pgDd
+FsGMBn4MiyGAw4XaK9LJBh+6DLH5D2Y7h0hXv+AUP3SSC+w5XA/D+SHcAbjtPzMW
+cdq9jr1Gu0KO3PSGIAUpI1o+XQeLqgyieFcVF/FlD31MDC152DifNA==
+-----END CERTIFICATE-----
```
nsuke: We could make use of cert files under `test/keys` directory instead of adding new one.
That way we can easily add c_glib ssl to the cross test too.

lib/c_glib/test/testtransportsslsocket.c
```diff
@@ -0,0 +1,579 @@
...
+//                                   "hostname", "localhost",
+//                                   "port",     port,
+//                                   NULL);
+//
+//  /* thrift_transport_peek returns FALSE when the socket is closed */
+//  g_assert (thrift_transport_is_open (client_transport) == FALSE);
+//  g_assert (thrift_transport_peek (client_transport, &error) == FALSE);
+//  g_assert (error == NULL);
+//
+//  pid = fork ();
```
nsuke: Haven't looked into detail yet but ca we have this test enabled on our CI, rather than commented out ?

nsuke:
Thanks for your intense work, I believe it's almost there but needs some cleanup and a bit more safer initialization (static mutex at least).
Also could you check the coding style of c_glib ? I think we're using 2 spaces rather than tabs and also some spacing before `{` are different.

gadLinux:
Hi @nsuke I will take a look to the style guide.
About the initialization. I will also take a look to it but I think OpenSSL already takes care of it. Let me see. 
If I have a little time I want to implement server side.

lib/c_glib/test/owncloud.level2crm.pem
gadLinux: Great. Didn't see it. Will remove that because this is my own. :D

lib/c_glib/test/testtransportsslsocket.c
gadLinux: This is because Server part was not ready. If I remember well. I will see also to this.

gadLinux:
@nsuke Sorry for delay. At the end I had to put this in a hurry in one of my projects and had no time to update. I did some efforts to clear the code a little bit. 
I cannot remove the certificate of the tests because still I'm unable to run the server, so I cannot put local certificate. I'm still using the server of my owncloud server that's ssl to test that SSL handshake and pinning is working. I will try to clear it at latter step. 

Is it possible to merge without it? I can comment the test if you want. Then I can remove the certificate.

gadLinux:
@nsuke It's giving errors in Travis about things I think didn't touched. Maybe a rebase is required?

nsuke:
@gadLinux locking, indent and cmake changes look good !
Only concern is the test dependency on external service.
One way to avoid it is to use cpp for server side as in `lib/c_glib/test/testthrifttestclient.cpp`.

Also could you enable `test/c_glib/src/test_client.c` to initialize your SSL implementation with `--ssl` option ?
We'll be able to add this to our cross language tests then.

gadLinux:
@nsuke But this would mean you will have to build c++ library to be able to test c. I prefer to add the server on c side. But for this I will need more time. I will miss 1.0 I think. Prefer to disable this test until correctly done, if you agree. 

About adding the --ssl I will check how it's done in c++ and do same way there if we leave test enabled.
Tell me what you think.

Best regards,

nsuke:
@gadLinux  I guess we can remove not-yet-possible part of the test for now, then.
I agree that if we eventually have server-side too, it would be cleaner to have c_glib-only test.

For --ssl, you can check out the behaviour as follows, if you’ve configured to build cpp:
```shell
$ make precross
$ test/cpp/TestServer # with or without --ssl
(on another shell)
$ test/c_glib/test_client # the same for --ssl above
```
otherwise, test/tests.json has full list of languages that has ssl (search for “ip-ssl”) and its command.

gadLinux:
Maybe it can be finally merged.

gadLinux:
I want to work on server after this is merged. So maybe I will open a new incident to track it in a different task. I want to see also if I can introduce timeouts to current implementation and see if there's any valgrind leakage. But this is another history.

gadLinux:
The two tests failing on travis are because missing build dependencies. They are not because our patch.

gadLinux:
@nsuke Do you think we can merge it finally?

jeking3:
We cannot merge it until it passes all CI builds.
Recommend you rebase against the current master, squash to a single commit, and force push to get another build going.

gadLinux:
As I told, there problem on CI is not because something we did. But it seem that it was already in the branch we forked from, cause it gave in a place we didn't touch. Anyway I rebased. Hope did it right!

gadLinux:
This is unmanageable. Now  it fails because something in Java is not properly working. Would I be able to merge?

jeking3:
It looks like you merged instead of rebasing.

If memory serves you had 11 commits originally.
What I would recommend is that you:

1. git checkout master
2. git fetch upstream master (you need to have upstream configured per github's instructions)
3. git merge FETCH_HEAD

Now your fork's master will match the master in github.

4. git checkout -b redo/THRIFT-3369
5. git cherry-pick <each of your 11 commits in order from earliest ot latest>
6. git push ... to github
7. open another pull request
8. close this pull request

gadLinux:
Hi, 
I know that it's not a good idea to rewrite history but since this is not merged anyplace I just overwrote the commit. This time I think it's correct. Le'ts see what the tests are saying...

gadLinux:
Again a problem with code outside the domain area. Can someone help me to fix this? Because it's failing on some places where I never touched. Should I continue doing rebase until everyone fixes their problems?

What should I do?

gadLinux:
I'm rebasing again...

lib/c_glib/test/owncloud.level2crm.pem
jeking3: This needs to be removed in favor of existing test keys.

configure.ac
```diff
@@ -199,6 +196,12 @@ if test "$with_c_glib" = "yes"; then
 fi
 AM_CONDITIONAL(WITH_C_GLIB, [test "$have_glib2" = "yes" -a "$have_gobject2" = "yes"])
 
+echo "OpenSSL check"
```
jeking3: There may be other libraries with native builds that need OpenSSL.  Why are we making it conditional on cpp and c_glib library support?

lib/c_glib/Makefile.am
```diff
@@ -45,6 +45,7 @@ libthrift_c_glib_la_SOURCES = src/thrift/c_glib/thrift.c \
                               src/thrift/c_glib/transport/thrift_buffered_transport_factory.c \
                               src/thrift/c_glib/transport/thrift_framed_transport_factory.c \
                               src/thrift/c_glib/transport/thrift_socket.c \
+                              src/thrift/c_glib/transport/thrift_ssl_socket.c \
```
jeking3: Does this need to be conditional on openssl being found?

lib/c_glib/src/thrift/c_glib/transport/thrift_platform_socket.h
```diff
@@ -0,0 +1,120 @@
+/*
```
jeking3: This looks a lot like the platform socket code that already exists in the C++ library.
Given their similarity should they be combined and moved into a common location?

lib/c_glib/test/CMakeLists.txt
```diff
@@ -108,13 +108,6 @@ add_test(NAME testoptionalrequired COMMAND testoptionalrequired)
 
 include_directories("${PROJECT_SOURCE_DIR}/test/c_glib/src" "${CMAKE_CURRENT_BINARY_DIR}/gen-c_glib")
 
-add_executable(testthrifttest testthrifttest.c
```
jeking3: Why was the testthrifttest executable removed?

lib/c_glib/test/Makefile.am
```diff
@@ -37,16 +37,17 @@ BUILT_SOURCES = \
...
 AM_CXXFLAGS = $(AM_CFLAGS)
-AM_LDFLAGS = $(GLIB_LIBS) $(GOBJECT_LIBS) @GCOV_LDFLAGS@
+AM_LDFLAGS = $(GLIB_LIBS) $(GOBJECT_LIBS) $(OPENSSL_LIBS) @GCOV_LDFLAGS@
 
 check_PROGRAMS = \
   testserialization \
   testapplicationexception \
   testcontainertest \
   testtransportsocket \
+  testtransportsslsocket \
```
jeking3: Does this need to be conditional on openssl being found?

configure.ac
gadLinux: But this is the latest I fixed... Maybe I lost the changes when doing the rebase. OpenSSL was not used everyplace when I did the patch. In fact it was not even included for c_glib. C++ had this optionally included. I followed the C++ directions.

lib/c_glib/src/thrift/c_glib/transport/thrift_platform_socket.h
gadLinux: I tottally agree. It's a reimplementation of the library for c_glib to not mix things. But this can be seen as an upgrade. And start making room for an common library after merge. Don't you think?

lib/c_glib/test/CMakeLists.txt
gadLinux: Thrift test should be there. Iif I removed it it was because error. I will readd it.

lib/c_glib/test/Makefile.am
gadLinux: I think so. But since I don't have yet detection code for ssl there's no opt-out code. I think it will not work without the openssl library. So we have to choices. Require OpenSSL as you recommended above. Or make the code conditional, something is not ready yet. What do you think?

gadLinux:
Re-rebased

gadLinux:
I will reopen the merge

PR #942, opened 2016-03-11
THRIFT-3734 To compare two string as lowercase.

mygityf:
it looks unexpected result got when parameter a is uppercase and b is lowercase.
split[0] is string of 'Content-Length', it contains character with uppercase.

### DIFF
--- a/lib/d/src/thrift/transport/http.d
+++ b/lib/d/src/thrift/transport/http.d
@@ -134,7 +134,7 @@ protected:
     }
 
     static bool compToLower(ubyte a, ubyte b) {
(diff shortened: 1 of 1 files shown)


mygityf:
@nsuke 
CI error is 'Failed to install HUnit-1.3.1.1', how to fix it?

nsuke:
@mygityf it's a haskell dependency build failure but doesn't seem to be reproducible. It's obviously not related to your patch.
And thanks for fixing the other side of the cross lang again. D http implementation is indeed weak and patches are much appreciated.

PR #943, opened 2016-03-11
THRIFT-3735 JSON protocol left in incorrect state when an exception is thrown during read or write operations

tylertreat-wf:
The JSON context stack may be left in an incorrect state when an
exception is thrown during read or write operations. This leads to
further errors while writing/reading the NEXT message, because incorrect
characters may be written or expected.

This is related to THRIFT-1473, but there is an additional issue in that
the bufio.Writer needs to be reset if an error occurs.

@markerickson-wf 
@stevenosborne-wf


matthinrichsen-wf:
+1

stevenosborne-wf:
+1

PR #947, opened 2016-03-14
THRIFT-3743 Java JSON protocol left in incorrect state when an exception is thrown during read or write operations

tylertreat-wf:
This is the same issue as THRIFT-1473 (Delphi) and THRIFT-3735 (Go) but for the Java library.

The JSON context stack may be left in an incorrect state when an exception is thrown during read or write operations. This leads to further errors while writing/reading the NEXT message, because incorrect characters may be written or expected.

@stevenosborne-wf
@markerickson-wf

### DIFF
--- a/lib/java/src/org/apache/thrift/protocol/TJSONProtocol.java
+++ b/lib/java/src/org/apache/thrift/protocol/TJSONProtocol.java
@@ -308,6 +308,13 @@ public class TJSONProtocol extends TProtocol {
     context_ = contextStack_.pop();
   }
 
+  // Reset the context stack to its initial state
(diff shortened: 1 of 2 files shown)


tylertreat-wf:
The Travis build failure appears to be an unrelated npm error...

markerickson-wf:
+1

We do something similar in Dart, for what it's worth - https://github.com/apache/thrift/blob/master/lib/dart/lib/src/protocol/t_json_protocol.dart#L187

stevenosborne-wf:
+1

jeking3:
@jfarrell this was merged, and the PR can be closed.

PR #938, opened 2016-03-09
THRIFT-3728 http transport for thrift-lua

mygityf:
THttpTransport.lua is http transport implement for thrift-lua.

PR #948, opened 2016-03-15
THRIFT-3744 The precision should be 17 (16 bits need after dot) after dot for double type.

mygityf:
The precision is lost when converting double to string.
E.g:
double PI = 3.1415926535897931;
string value = format("%.16g", PI);
The value will be '3.141592653589793' and last 1 is lost after format operation.
But expected value should be '3.1415926535897931'.
Solution:
string value = format("%.17g", PI);


mygityf:
@nsuke can you merge the pull into master,
Then i can try to rebase #938

PR #955, opened 2016-03-18
THRIFT-3746 JSON protocol left in incorrect state on read errors

tylertreat-wf:
This is related to THRIFT-3735, but the problem was not fully addressed
there. The JSON protocol's read buffer needs to be reset to avoid being
put in an invalid state on read errors.

This also changes how the writer is reset to match what we do with the
reader for consistency.


nsuke:
LGTM
@Jens-G any concern for this ?

nsuke:
It seems that this change makes corss tests for Go server with JSON protocol + Buffered transport flaky.
A normal RPC call right after oneway call sometimes fails to respond.

@tylertreat-wf do you have any idea how this happens ?

tylertreat-wf:
@nsuke Interesting, I'm not entirely sure why this would be happening. I would expect the `writer` reset to be a safe operation to perform on `WriteMessageBegin`.

PR #956, opened 2016-03-18
Sargis/rdr 812/t socket patch

SargisYonan:


Jens-G:
Please rebase against current master. That PR will never get accepted, as it is.

PR #958, opened 2016-03-20
THRIFT-3750 Check isSet in copyWithZone to ensure a true copy

zachhowe:
For each property being copied to the new object, ensure isSet is checked before setting the new object's property. Otherwise isSet will always be YES on new object even if it was not set to YES on the original object. This ensures the new copied object is a true copy.


chefnobody:
Yes please. :+1:

PR #959, opened 2016-03-23
THRIFT-3753 Fixed a race condition in TServerFramework::stop that prevents clean server shutdown

tpcwang:
This is a sequence that exposes the race condition:
(1) Thread-1: in serve(), blocked on accept()
(2) Thread-2: in stop(), calls interrupt()
(3) Thread-2: in stop(), in interruptChildren(), checked that socket is valid
(4) Thread-1: in serve(), unblocked because of interrupt(), calls releaseOneDescriptor
(5) Thread-2: in interruptChildren(), send fails so clients are not interrupted

Closing a socket while it is being polled/selected may or may not work on all platforms, so I think changing the order is better.

### DIFF
--- a/lib/cpp/src/thrift/server/TServerFramework.cpp
+++ b/lib/cpp/src/thrift/server/TServerFramework.cpp
@@ -214,8 +214,10 @@ void TServerFramework::setConcurrentClientLimit(int64_t newLimit) {
 }
 
 void TServerFramework::stop() {
-  serverTransport_->interrupt();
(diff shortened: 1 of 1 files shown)


jeking3:
I recommend you rebase this against origin/master because the list of commits above leads me to believe it wasn't on origin/master to begin with, and the fix should be small.  Plus this will re-run the build so it passes, and we can look at merging it in.

tpcwang:
@jeking3 Thanks for catching it. It is rebased correctly now.

PR #961, opened 2016-03-23
THRIFT-3756 Improve requiredness documentation

Jens-G:
Client: Website
Patch: Jens Geyer

### DIFF
--- a/doc/specs/idl.md
+++ b/doc/specs/idl.md
@@ -98,7 +98,7 @@ N.B.: The `xsd_all` keyword has some purpose internal to Facebook but serves no 
 
 ### Union
 
-Unions are similar to structs, except that they provide a means to transport exactly one field of a possible set of fields, just like union {} in C++. Consequently, union members cannot be required fields.
+Unions are similar to structs, except that they provide a means to transport exactly one field of a possible set of fields, just like union {} in C++. Consequently, union members are implicitly considered optional (see requiredness).
 
     [13] Union          ::=  'union' Identifier 'xsd_all'? '{' Field* '}'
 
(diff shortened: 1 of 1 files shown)


doc/specs/idl.md
```diff
@@ -126,8 +126,39 @@ A service provides the interface for a set of functionality provided by a Thrift
...
+- Read: Optional fields may, or may not be part of the input stream. 
+- Default values: are written when the isset flag is set
+
+Most language implementations use the recommended pratice of so-called "isset" flags to indicate whether a particular optional field is set or not. Only fields with this flag set are written, and conversely the flag is only set when a field value has been read from the input stream. 
+	
+#### default requiredness (implicit)
+
+- Write: Like required, the fields are always written.
+- Read: Like optional, the field may, or may not be part of the input stream. 
+- Default values: may not be written
```
jsirois: This 'may' is confusing.  Id this 'may' in the [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt) sense, or is this 'will never be written' or, in RFC-2119-speak 'must not be written'.

jsirois:
LGTM @Jens-G - just one question/comment above.  Thanks for surfacing this very important part of the spec in the docs!

doc/specs/idl.md
```diff
@@ -126,8 +126,39 @@ A service provides the interface for a set of functionality provided by a Thrift
...
+
+- Write: Optional fields are only written when they are set
+- Read: Optional fields may, or may not be part of the input stream. 
+- Default values: are written when the isset flag is set
+
+Most language implementations use the recommended pratice of so-called "isset" flags to indicate whether a particular optional field is set or not. Only fields with this flag set are written, and conversely the flag is only set when a field value has been read from the input stream. 
+	
+#### default requiredness (implicit)
+
+- Write: Like required, the fields are always written.
```
jsirois: This also seems misleading.  I added commentary on https://issues.apache.org/jira/browse/THRIFT-3752, but the java generator, for example, only emits default requiredness collections when they are non-null; which seems like the only thing to do in order for the read-side of the wire to correctly interpret the field as not set (unless there were a special byte-sequence in the protocol for indicating unset or null).  As such, this language seems misleading to me.  Perhaps something like: "Write: Fields that are set are always written" ?

doc/specs/idl.md
Jens-G: Actually it is a "may", but you hit an interesting point. There are some lengthy discussions in JIRA about the semantics of optional fields, so I tried to reflect the current status quo here. There are arguments for both, whether an default value should be written or not, and at the very end this depends on how "default value" in the Thrift context should be defined. If it means that the value is assigned by default, then the only difference would be how the field is set. But if "default" implies that any value that equals the default should be omitted during serialization, that's something completely different. From my understanding, the latter is not really implemented anywhere (think of a default value consisting of a struct that contains a list of structs which ...) so for the sake of having one clear definition, I assume the former, much easier semantics is the one intended by the inventor. I admit that I took a bit of freedom here. I will add a note to at least point out that this topic is still not finally settled.

More information here:
- https://issues.apache.org/jira/browse/THRIFT-2429
- https://issues.apache.org/jira/browse/THRIFT-1528
- http://lionet.info/asn1c/blog/2010/07/18/thrift-semantics/

PR #964, opened 2016-03-24
THRIFT-3751 Compiler allows field ids that are too large for generate…

Jens-G:
…d code

Client: Compiler (General)
Patch: Jens Geyer


nsuke:
+1

npm fails too often...

PR #970, opened 2016-03-28
THRIFT-3765 fix memory leak in python compact protocol extension

cjmay:
This fixes a significant memory leak in the new Python extension module for the compact protocol.

### DIFF
--- a/lib/py/src/ext/protocol.tcc
+++ b/lib/py/src/ext/protocol.tcc
@@ -418,19 +418,24 @@ bool ProtocolBase<Impl>::encodeValue(PyObject* value, TType type, PyObject* type
   }
 
   case T_STRING: {
+    ScopedPyObject nval;
+
     if (PyUnicode_Check(value)) {
-      value = PyUnicode_AsUTF8String(value);
-      if (!value) {
+      nval.reset(PyUnicode_AsUTF8String(value));
+      if (!nval) {
         return false;
       }
+    } else {
+      Py_INCREF(value);
+      nval.reset(value);
     }
 
-    Py_ssize_t len = PyBytes_Size(value);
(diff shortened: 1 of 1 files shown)


nsuke:
@cjmay thanks for catching this before the release.
[C extension + UTF8 fix](https://github.com/apache/thrift/pull/769/files#diff-cf2b2a6a8c345e5979b22a2ccc932abfR439) seems to be the true culprit here (so it's even much more worse in terms of the scope of this regression).

lib/py/src/ext/protocol.tcc
```diff
@@ -431,6 +433,7 @@ bool ProtocolBase<Impl>::encodeValue(PyObject* value, TType type, PyObject* type
     }
```
nsuke: I think we need DECREF before the "return false" above too.
Feel free to use ScopedPyObject (or not).

cjmay: I thought so too, originally, but isn't value null there?

cjmay:
Thank you!  We were excited to see this extension get written.

lib/py/src/ext/protocol.tcc
nsuke: It shouldn't be...
The value is either null-checked or passed to `Py_INCREF` that does not accept null.
The intention here is to simply reject too long string values.
If you experienced null value here, we'll need to investigate on that matter too.

cjmay: Oh, I apologize, I was looking at the wrong "return false."

cjmay: Fixed, and thanks for the tip.

nsuke:
Committed, thanks again for the fix.

PR #971, opened 2016-03-28
Update php_thrift_protocol7.cpp

visoeclipse:
bug fix


jeking3:
Is there a Apache Thrift Jira ticket for this?  The check-in comment does not refer to it.

PR #972, opened 2016-04-01
THRIFT-3770 Implement Python 3.4+ asyncio support

vmarkovtsev:
I decided to take an approach which is different from Tornado implementation and do not nail supported protocols to the Framed one. My effort enables BinaryProtocol.

### DIFF
--- a/test/py/TestAsyncioClient.py
+++ b/test/py/TestAsyncioClient.py
@@ -1,0 +1,339 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+import asyncio
+import os
+import ssl
+import sys
+import time
+import unittest
+from optparse import OptionParser
+
+from util import local_libpath
+
+SCRIPT_DIR = os.path.abspath(os.path.dirname(__file__))
+
+
+def async_test(f):
+    def wrapper(*args, **kwargs):
+        coro = asyncio.coroutine(f)
+        future = coro(*args, **kwargs)
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(future)
+    return wrapper
+
+
+class AbstractTest(unittest.TestCase):
+    @async_test
+    def setUp(self):
+        if options.ssl:
+            ssl_ctx = ssl.create_default_context()
+            ssl_ctx.check_hostname = False
+            ssl_ctx.verify_mode = ssl.CERT_NONE
+        else:
+            ssl_ctx = None
+        self.transport = yield from TAsyncio.TAsyncioTransport.connect(
+            options.host, options.port, ssl=ssl_ctx,
+            framed=(options.trans == 'framed'))
+
+        self.transport.open()
+        protocol = self.get_protocol(self.transport)
+        self.client = ThriftTest.Client(protocol)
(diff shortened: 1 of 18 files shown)


vmarkovtsev:
``flake8 --ignore=E501 lib/py`` in Travis failed because the syntax "yield from" is Python 3.3+ and flake [does not support it](https://gitlab.com/pycqa/flake8/issues/53).

nsuke:
@vmarkovtsev thanks for the contribution !
A few points:
* Is there any way to create a generic protocol wrapper ?
* gen_asyncio_ seems uninitialized.
* we need something similar to test/py.tornado and tutorial/py.tornado

As to the test, even much better way is to port test/py/TestClient.py and TestServer.py so that we can test asyncio against other languages.

vmarkovtsev:
@nsuke, about the generic wrapper, I feel bad about these dupes too but cannot see the possibility because asyncio coroutine's "viral" nature with "yield from" is going to infect every piece of reading code. What's even worse, "yield from" result in SyntaxError on Python 2.7. In order to support all the protocols, we will have to adapt each existing Protocol similar to what's done to Binary. Framed's already been ported, so I don't think it is too hard anyway.

gen_asyncio_ - overseen, fixed.

I will add the tests in a separate commit in this PR, thanks.

vmarkovtsev:
I've added the tests. These only belong to Client, since I do not plan to port Server.

```
test/test.py --server cpp --client py3.asyncio
Apache Thrift - Integration Test Suite
Mon Apr 04 14:50:11 2016
==========================================================================
server-client:          protocol:    transport:               result:
cpp-py3.asyncio         binary       buffered-ip              success
cpp-py3.asyncio         compact      framed-ip                success
cpp-py3.asyncio         compact      buffered-ip              success
cpp-py3.asyncio         binary       buffered-ip-ssl          success
cpp-py3.asyncio         compact      framed-ip-ssl            success
cpp-py3.asyncio         compact      buffered-ip-ssl          success
cpp-py3.asyncio         binary       framed-ip                success
cpp-py3.asyncio         binary       framed-ip-ssl            success
==========================================================================
No unexpected failures.
```

nsuke:
@vmarkovtsev great, thanks for the tests and flake8 fix.
Could you also provide an example in either tutorial/py or tutorial/py.asyncio directory ?

test/tests.json
```diff
@@ -268,6 +268,41 @@
...
+    "name": "py3.asyncio",
+    "server": {
+      "extra_args": ["TSimpleServer"],
+      "command": [
+        "python3",
+        "TestAsyncioServer.py",
+        "--verbose",
+        "--genpydir=gen-py.asyncio"
+      ]
+    },
```
nsuke: If you're not providing it, server: {...} should be omitted.

test/tests.json
```diff
@@ -268,6 +268,41 @@
...
+    "client": {
+      "timeout": 10,
+      "command": [
+        "python3",
+        "TestAsyncioClient.py",
+        "--host=localhost",
+        "--genpydir=gen-py.asyncio"
```
nsuke: Can you update test/py/Makefile.am and test/py/generate.cmake so that we can easily/automatically run your test with the generated code ?

test/py/TestAsyncioClient.py
```diff
@@ -0,0 +1,350 @@
...
+        self.assertEqual(y.string_thing, 'foobar')
+
+    @async_test
+    def testOneway(self):
+        print('testOneway')
+        start = time.time()
+        yield from self.client.testOneway(1)  # type is int, not float
+        end = time.time()
+        self.assertTrue(end - start < 3,
+                        "oneway sleep took %f sec" % (end - start))
```
nsuke: It seems rather an issue of original test code, but 3 seconds are way too large for this test to make any sense.
Something like 200 ms should be enough.

vmarkovtsev:
Added the tutorial and some tests, fixed the comments

AndreLouisCaron:
I'm interested in using this.  I see that [facebook/fbthrift](https://github.com/facebook/fbthrift) has support for it, but it's not in PyPI.

Any chance work on this PR will resume?

vmarkovtsev:
I will resume this as soon as the maintainers show up themselves :) Since the last time there appeared conflicts and I don't want to resolve them multiple times - not working with Thrift anymore.

jeking3:
This needs to be rebased on the current upstream/master, conflicts resolved, and pushed to kick a new build.

jeking3:
This needs to be rebased on the current master, conflicts resolved, and pushed to kick a new build.

jeking3:
@jfarrell please close this; the author indicated no interest in finishing the work due to the team's past inaction on pull request management.

PR #973, opened 2016-04-01
THRIFT-3771 TBufferedTransport gets in invalid state on read/write errors

tylertreat-wf:
Go's TBufferedTransport can enter an invalid state after an error occurs
while calling read, write, or flush. This is because TBufferedTransport
uses a bufio.ReadWriter, which "caches" the error returned by a call to
read or write such that subsequent calls return the same error. This can
be problematic if you wish to reuse the transport after a failed read or
write. The solution is to reset the reader/writer on failed calls.

@stevenosborne-wf @markerickson-wf


stevenosborne-wf:
+1

PR #974, opened 2016-04-04
THRIFT-3776 Go code from multiple thrift files with the same namespace

creker:
Currently compiler puts all generated Go code into two files - ttypes.go and constants.go. Because of that you can't compile multiple thrift files with the same namespace. They will just overwrite each other.

This PR fixes two things that closely related to each other:
1) Allow compiler to put generated Go code into the files with the same names as original thrift files. That mimics behaviour of other language generators. Types and services will be put into [thrift file name].go and constants will be put into [thrift file name]-consts.go. As a bonus, if you put everything into one file result will be basically the same as before.
The only difference is I put GoUnusedProtection__ variable into it's own .go file to prevent redefinition.

2) Prevent importing self package. Before this compiler would put self package into import section which Go doesn't support. Now compiler will recognize that types are from the same namespace. Without this you can't generate working Go code from multiple thrift files.

This PR doesn't add new compiler flag for this.

### DIFF
--- a/compiler/cpp/src/generate/t_go_generator.cc
+++ b/compiler/cpp/src/generate/t_go_generator.cc
@@ -762,7 +762,7 @@ void t_go_generator::init_generator() {
   f_const_values_ << endl << "func init() {" << endl;
 
   // Create file for the GoUnusedProtection__ variable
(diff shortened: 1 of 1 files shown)


creker:
Looks like failed tests are not relevant to this PR

PR #976, opened 2016-04-04
THRIFT-3467 Go Maps for Thrift Sets Should Have Values of Type struct{}

creker:


test/go/src/bin/stress/main.go
```diff
@@ -201,7 +201,27 @@ func client(protocolFactory thrift.TProtocolFactory) {
...
+			0: struct{}{},
+			1: struct{}{},
+			2: struct{}{},
+			3: struct{}{},
+			4: struct{}{},
+			5: struct{}{},
+			6: struct{}{},
+			7: struct{}{},
+			8: struct{}{},
+		}
```
Jens-G: That's my favourite part of this change.

creker: And probably the most important one

PR #977, opened 2016-04-04
THRIFT-3768 fix TThreadedServer refactoring issues with client lifetime guarantees

jeking3:


lib/cpp/src/thrift/server/TThreadedServer.h
```diff
@@ -20,19 +20,26 @@
...
 
 /**
- * Manage clients using a thread pool.
+ * Manage clients using threads.  Once the refactoring for THRIFT-3083 took place it became
+ * obvious that the differences between the two threaded server types was becoming insignificant.
+ * Therefore to satisfy THRIFT-3096 and fix THRIFT-3768, TThreadedServer is simply a wrapper
+ * around TThreadedPoolServer now.  If backwards compatibility was not a concern, it would have
+ * been removed.
+ *
+ * The default thread pool size is
```
tpcwang: Unfinished comment.

lib/cpp/src/thrift/server/TThreadedServer.h
```diff
@@ -20,19 +20,26 @@
...
  */
-class TThreadedServer : public TServerFramework {
+
+/* [[deprecated]] */
+class TThreadedServer : public TThreadPoolServer {
```
tpcwang: I'd suggest using composition and the thread manager internally rather than inheriting from TThreadPoolServer seeing that TThreadPoolServer exposes other methods that would defeat the original purpose of the TThreadedServer.

lib/cpp/src/thrift/server/TThreadedServer.cpp
```diff
@@ -77,44 +83,29 @@ TThreadedServer::TThreadedServer(const shared_ptr<TProcessor>& processor,
...
-  try {
-    Synchronized s(clientsMonitor_);
-    while (getConcurrentClientCount() > 0) {
-      clientsMonitor_.wait();
-    }
-  } catch (TException& tx) {
-    string errStr = string("TThreadedServer: Exception joining workers: ") + tx.what();
-    GlobalOutput(errStr.c_str());
+void TThreadedServer::onClientConnected(const shared_ptr<TConnectedClient>& pClient) {
+  if (!threadManager_->idleWorkerCount())
```
tpcwang: I would personally like to see how the threads would be reaped before we go down this path, but that's mostly because I am unfamiliar with the thread manager. I'll defer to you and the maintainers on whether to do this without the accompanying thread manager change.

jeking3: As submitted the number of threads will grow to the maximum number of concurrent connections and will not shrink.  Additional work is needed to allow ThreadManager::removeWorker() to do the right thing when many threads arrive at it at the same time.  In gdb I found that workerCount_ was 10 but workerMaxCount_ was 0; the stop behavior for ThreadManager calls removeWorker(workerCount_) and gets an exception, which eventually causes a core.  This is why, for this submission, I left it as stated.

jeking3:
Declining this pull request - I have a far simpler one to submit.

lib/cpp/src/thrift/server/TThreadedServer.h
Jens-G: Still the case.

PR #979, opened 2016-04-06
THRIFT-3779 Qt Client Android and Windows QTcpSocket

Gjergj:
libthriftQt.pro creates .dll and .lib for Windows but also .so and .a libraries for Android.
To generate .lib or .a library uncomment CONFIG += staticlib inside the .pro file.
The port is based on TQtcpSocket class. Currently it only supports TBinaryProtocol and TCompactProtocol.
The TQtcpSocket is not tested for server or SSL usage.
libthriftQt/thrift/config.h is just a hack to have config.h in our include path and #define STRERROR_R_CHAR_P 0


lib/cpp/libthriftQt.pro
```diff
@@ -0,0 +1,161 @@
...
+        DESTDIR = ../cpp/DebugQtWin
+        OBJECTS_DIR = ../cpp/DebugQtWin/.obj
+    } else {
+        DESTDIR = ../cpp/releaseQtWin
+        OBJECTS_DIR = ../cpp/releaseQtWin/.obj
+    }
+}
+android {
+    #SOURCES +=
+    HEADERS += libthriftQt/thrift/config.h \
```
bufferoverflow: please use `#include <thrift/thrift-config.h>`
`config.h` is generated and should not be part of the source tree

Gjergj: At first I did not want to change thrift-config.h
Fixed now.

nsuke:
Why not enhance lib/cpp/src/thrift/qt/TQIODeviceTransport.cpp instead ?

As to the qmake file, I don't think we want to introduce yet another build system.
As autotools and cmake handles config.h generation for android, the pre-generated config.h won't be needed either.

Gjergj:
I'll have a look at TQIODeviceTransport and see how it works as a client.
As per config.h the idea is the same as in Windows port I guess.
My cmake exprience is limited and I do not know how cmake can generate the config.h on Windows.
Also I do not know how the cmake can help to build/generate using Qt for Android without QtCreator.
The idea is that it should support Windows as well as Linux development using Qt.

Gjergj:
lib/cpp/src/thrift/qt/TQIODeviceTransport.cpp  seems to be written for  lib/cpp/src/thrift/qt/TQTcpServer.cpp  which is an async server. The lib/cpp/src/thrift/transport/TQtcpSocket I wrote is for a sync client.
Qt sync TcpSocket communication requires that you call waitForBytesWritten , waitForReadyRead and waitForConnected so I am not quite sure how to modify TQIODeviceTransport.
I am open for any suggestion about this and cmake/config.h issue.

nsuke:
@Gjergj to me, as Qt sockets are inherently async, synchronous-only support seems a bit strange.
I think the whole point of existing TQTcpServer is to provide async support using Qt event loop, as an alternative to libevent based one.
What is your motivation for this ?

Gjergj:
@nsuke I was just looking for something like TSocket based on Qt as I needed a synchronous client. And I do not see how async Qt sockets would fit in synchronous client.

nsuke:
> I was just looking for something like TSocket based on Qt 

Why you needed one ?
In other words, how does it work better for you compared to plain TSocket ?

Gjergj:
Since I use Qt for Android development an I use Windows as environment I had a difficult time compiling TSocket for Android on Windows. And i figured that QTcpSocket might do the same and compile for all platforms that Qt runs and an all development environments that Qt runs.

nsuke:
FWIW I use Thrift C++ lib on Windows and Android.
We're building it for android only on Linux but I'm sure it's possible on Windows too.
I guess it's a general cross compiling problem, not something specific to Thrift.
You need to use CMake with a toolchain file (or maybe Autotools on msys2/cygwin) to setup proper cross compiling for Android NDK.

Gjergj:
I tried to build for Android under Linux but failed.
cmake -DCMAKE_TOOLCHAIN_FILE=../build/cmake/android-toolchain.cmake ..
Failed on serveral errors.
Do you mind sharing how you did it?

nsuke:
The toolchain file is rudimentary and full of hard coded values.
You can manually edit the file to match your environemnt or instead pick more full fledged one from the web, for example https://github.com/Itseez/opencv/blob/master/platforms/android/android.toolchain.cmake .

Gjergj:
I finally succeeded to build for Android under Windows and Linux.
I took me one day and a half though.
Using Qt it would be much easier if one used Qt as client.
Is there anything I can change for the Qt Client based on QTcpSocket patch to be accepted?
Thanks for the help.

jeking3:
Folks who originally commented on this, please follow up.

nsuke:
This PR does the following
* use `qmake` in addition to autotools and cmake to build C++ lib
* introduce blocking variant of Qt based transport

The motivation was to use blocking transport on android. The existing solution for this is to build TSocket for Android using CMake. While I agree that qmake is much more easier to build for android, it's only because CMake is not that easy to use for android cross compilation.

Gjergj:
Indeed, compiling 0.9.3 on Android was a pain. I had to get the cross compile cmake file from another project.
Since I was using Qt fro Android development, using QSocket was an easier to cross compile thrift for Android.

However I am not interested anymore to take this any further.

PR #981, opened 2016-04-06
THRIFT-3038: fix up some volatiles in cpp

jeking3:
I addressed the concerns in TFileTransport (forceFlush_) and in TThreadPoolServer, so items #2, #4, #5 from the original issue description.

### DIFF
--- a/lib/cpp/src/thrift/transport/TFileTransport.h
+++ b/lib/cpp/src/thrift/transport/TFileTransport.h
@@ -27,6 +27,7 @@
 #include <string>
 #include <stdio.h>
 
+#include <boost/atomic.hpp>
 #include <boost/scoped_ptr.hpp>
 #include <boost/shared_ptr.hpp>
 
@@ -352,7 +353,7 @@ private:
 
   // To keep track of whether the buffer has been flushed
   Monitor flushed_;
-  volatile bool forceFlush_;
+  boost::atomic<bool> forceFlush_;
 
   // Mutex that is grabbed when enqueueing and swapping the read/write buffers
   Mutex mutex_;
@@ -438,3 +439,4 @@ private:
 } // apache::thrift::transport
 
 #endif // _THRIFT_TRANSPORT_TFILETRANSPORT_H_
+
--- a/lib/cpp/src/thrift/server/TThreadPoolServer.cpp
+++ b/lib/cpp/src/thrift/server/TThreadPoolServer.cpp
@@ -69,7 +69,6 @@ TThreadPoolServer::TThreadPoolServer(const shared_ptr<TProcessorFactory>& proces
                      inputProtocolFactory,
                      outputProtocolFactory),
     threadManager_(threadManager),
-    stop_(false),
     timeout_(0),
     taskExpiration_(0) {
 }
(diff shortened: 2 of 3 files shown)


jeking3:
Quite interesting, the job that failed is related to the code changes but the error message comes from a version of boost before 1.53.  It looks like that build job isn't making sure boost is at the minimum required version.

ben-craig:
Unless you have extremely compelling profiling results, the only atomic memory order you should be using is memory_order_seq_cst.  Even better, you get seq_cst for free as the default parameter of load and store.
memory_order_consume is almost entirely broken in C++11 and 14.  Basically all compilers promote it to an acquire instead of a consume.
memory_order_relaxed is extremely difficult to reason about, and should be avoided.

jeking3:
I am okay letting things default to something else here.  I followed the boost examples and it passed the build with those settings.

nsuke:
Seeing the [commit log](https://github.com/apache/thrift/commit/05b19b57cf2e224ab49e5cf983677361acb18311#diff-487498268b6e5a491d25456332fd44a1R107), it looks like more about not to use `acquire` for `store` and `release` for `load` (it's swapped).

lib/cpp/src/thrift/transport/TFileTransport.h
```diff
@@ -352,7 +352,7 @@ class TFileTransport : public TFileReaderTransport, public TFileWriterTransport
 
   // To keep track of whether the buffer has been flushed
   Monitor flushed_;
-  volatile bool forceFlush_;
+  bool forceFlush_;
```
nsuke: We're doing `while (forceFlush_) {`, so we need atomic here.

jeking3: Thanks for catching that - I pushed an update changing this to a boost::atomic<bool> type.

nsuke:
@jeking3 the atomics for TThreadPoolServer fields are for 64-bit field read-write on 32bit-x86 and other platforms ? If so, we only need relaxed.

That said, as @ben-craig suggested, leaviing it as default strict order (+without `load` and `store`) would even be better for readability.

jeking3:
Are you suggesting I get rid of the load and store calls and instead just use the provided "operator=" from the atomic?  I'm okay with that... it is much more readable.

ben-craig:
The default op=, and even the default conversion operators from atomic<T> to T are fine.

@nsuke I would prefer we keep the access to TThreadPoolServer timeout variables as seq_cst.  The setter operations shouldn't ever be in performance critical code, and the 'get' side of things is happening in an already slow area.  On x86, the getters will still turn into regular 'mov' operations.  On ARM, there will be a couple of memory barriers generated, but those shouldn't be a huge deal.

jeking3:
@ben-craig are you saying additional code changes are needed?

ben-craig:
Your latest revision look fine, I was mainly trying to address nsuke's post.

jeking3:
The latest build failure was environmental and passed all other tests:

```
make[3]: Leaving directory `/thrift/src/lib/lua'
make[3]: write error
make[2]: *** [check-recursive] Error 1
```

PR #982, opened 2016-04-07
THRIFT-3783: python code generator dose not handle struct dependent

collinmsn:


compiler/cpp/src/generate/t_py_generator.cc
```diff
@@ -21,12 +21,19 @@
...
 
 #include <stdlib.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sstream>
 #include <algorithm>
+#include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/topological_sort.hpp>
+#include <boost/graph/depth_first_search.hpp>
+#include <boost/graph/visitors.hpp>
```
Jens-G: We don't want to add boost as a dependency when building the Thrift compiler. Sorry.

compiler/cpp/src/generate/t_py_generator.cc
```diff
@@ -384,6 +406,55 @@ void t_py_generator::init_generator() {
...
     py_imports() << endl <<
     "from .ttypes import *" << endl;
+
+  construct_object_dependent_graph();
+}
+
+/**
+ * Construct object dependent graph to determine object generate order
+ * if cycle dependent is detected, generate process will be aborted
+ * object dependent example
```
Jens-G: But your test case is perfectly valid Thrift IDL:
```
struct A {
  1: B b;
}

struct B {
  1: A a;
}
```

One can do this with Thrift. Is there no way to support that in Python? Something like forward-declaration of types maybe?

compiler/cpp/src/generate/t_py_generator.cc
```diff
@@ -384,6 +406,55 @@ void t_py_generator::init_generator() {
...
+    }
+  }
+
+  DependentGraph dependent_graph(edges.begin(), edges.end(), objects.size());
+  // check cycle dependent
+  bool has_cycle = false;
+  cycle_detector detector(has_cycle);
+  boost::depth_first_search(dependent_graph, boost::visitor(detector));
+  if (has_cycle) {
+    throw "type error: cycel dependent found in " + program_->get_path();
```
Jens-G: Would be nice to know which type(s) are affected.

test/py/object_gen_order/cycle.thrift
```diff
@@ -0,0 +1,29 @@
...
+
+namespace py cycle
+
+struct A {
+  1: B b;
+}
+
+struct B {
+  1: A a;
+}
```
Jens-G: As I said, that's legal code.

jeking3:
Does this need further work or should it be closed?

jeking3:
Closing due to 2 years of inactivity.

KarDik465:
This problem is actual.
I have problems because of it

PR #983, opened 2016-04-07
THRIFT-3780 Use fixnum Int64 to write/read binary encoded i64

stevenosborne-wf:
Reading/writing `i64` is broken using the binary protocol when compiled to js. Use the [`fixnum`](https://pub.dartlang.org/packages/fixnum) library to address this. Note: fixnum `Int64` is already used in the [compact protocol](https://github.com/apache/thrift/blob/master/lib/dart/lib/src/protocol/t_compact_protocol.dart#L246) to correct the same issue.

@markerickson-wf


markerickson-wf:
+1 - the CI failures are in unrelated language - pairs so I don't have concerns about those.

PR #984, opened 2016-04-07
THRIFT-3784: fix and tests

csosborn:
This patch fixes [THRIFT-3784](https://issues.apache.org/jira/browse/THRIFT-3784), a bug in the thrift-maven-plugin.

### DIFF
--- a/contrib/thrift-maven-plugin/src/main/java/org/apache/thrift/maven/AbstractThriftMojo.java
+++ b/contrib/thrift-maven-plugin/src/main/java/org/apache/thrift/maven/AbstractThriftMojo.java
@@ -117,7 +117,7 @@ abstract class AbstractThriftMojo extends AbstractMojo {
      * @parameter default-value="${localRepository}"
      * @required
      */
-    private ArtifactRepository localRepository;
+    protected ArtifactRepository localRepository;
 
     /**
      * Set this to {@code false} to disable hashing of dependent jar paths.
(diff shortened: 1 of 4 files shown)


contrib/thrift-maven-plugin/src/main/java/org/apache/thrift/maven/AbstractThriftMojo.java
```diff
@@ -319,15 +331,6 @@ public boolean accept(File dir, String name) {
...
     }
 
-    ImmutableSet<File> findThriftFilesInDirectories(Iterable<File> directories) throws IOException {
-        checkNotNull(directories);
-        Set<File> thriftFiles = newHashSet();
-        for (File directory : directories) {
-            thriftFiles.addAll(findThriftFilesInDirectory(directory));
-        }
-        return ImmutableSet.copyOf(thriftFiles);
-    }
```
csosborn: This code was unused.

csosborn:
What's the protocol here? It does not look like the build failures have anything to do with my changes.

csosborn:
Just checking in here.. This request has been open for about two months with no response. 

Is the project still accepting patches? It doesn't look like anything has been merged in a few months, and the CI system appears unhealthy.

jeking3:
The project is still accepting patches, however as you can see we have over 100 pull requests (I have 3 in the queue) but we haven't been able to stabilize the apache build job for quite some time, leading to a backlog.  Only a few people have access to the apache build job and until we can get builds to pass the project lead decided (correctly, in my opinion) not to merge any more code.

PR #988, opened 2016-04-15
Fix command option error.

matyapiro31:
There's no --dbg for dh_strip, maybe someone has mistaken this for --dbg-package.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -199,8 +199,8 @@ binary-common:
 	dh_installexamples
 	dh_installman
 	dh_link
-	dh_strip -ppython-thrift --dbg=python-thrift-dbg
-	dh_strip -ppython3-thrift --dbg=python3-thrift-dbg
+	dh_strip -ppython-thrift --dbg-package=python-thrift-dbg
+	dh_strip -ppython3-thrift --dbg-package=python3-thrift-dbg
 	dh_strip -pthrift-compiler -plibthrift0
 	dh_compress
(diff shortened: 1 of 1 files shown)


matyapiro31:
These build failures don't come from my changes.

debian/rules
```diff
@@ -199,8 +199,8 @@ binary-common:
 	dh_installexamples
 	dh_installman
 	dh_link
-	dh_strip -ppython-thrift --dbg=python-thrift-dbg
-	dh_strip -ppython3-thrift --dbg=python3-thrift-dbg
+	dh_strip -ppython-thrift --dbg-package=python-thrift-dbg
+	dh_strip -ppython3-thrift --dbg-package=python3-thrift-dbg
```
Jens-G: @nsuke: any comments from your side? 
this is from f43d0ca6e57c4c30ea742e5f80e086288e999ecb

nsuke: Actually, -dbg= was from original line from THRIFT-71.
I don't know much about this but seeing the doc it seems this fix seems correct.

Jens-G:
@matyapiro31, may I ask you to file a JIRA ticket for this, according to [our contrib guidelines](http://thrift.apache.org/docs/HowToContribute)? Or do we have one already maybe?

matyapiro31:
@Jens-G What do you mean?

Jens-G:
tracked in THRIFT-3796

PR #991, opened 2016-04-21
dependency version constraints for haskell

SmartHoneybee:
I was unable to use `cabal install thrift` to install the thrift haskell package from [hackage](https://hackage.haskell.org/package/thrift). Apparently the implementation is not compatible with `unordered-containers` newer than 0.2.5.1 and `QuickCheck` newer than 2.8.1.

This change enabled me to build the package.

(I hope this is the correct file/way to submit my suggestion.)

### DIFF
--- a/lib/hs/Thrift.cabal
+++ b/lib/hs/Thrift.cabal
@@ -40,7 +40,7 @@ Library
   Hs-Source-Dirs:
     src
   Build-Depends:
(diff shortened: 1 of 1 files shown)


jeking3:
@jfarrell I checked the current haskell build file and we've already modified it to work the dependencies.  This item doesn't have a Jira ticket I can comment on, so this can be closed.

PR #992, opened 2016-04-22
THRIFT-3233: fix accounting of workers in thread manager

jeking3:
Per patch submitted by Paweł Janicki.


jeking3:
Looking for an advocate who would consider merging this fix.

jeking3:
Looking for an advocate who would consider merging this fix despite the busted Apache Jenkins environment result.

PR #995, opened 2016-04-27
THRIFT-3799 Make constants to use iota

astromahi:
https://issues.apache.org/jira/browse/THRIFT-3799

Converted constant declaration to use `iota` construct rather than manually 
specifying value.


jeking3:
@jfarrell please close this pull request as it was rejected last year.  Thanks.

PR #1002, opened 2016-05-02
THRIFT-3812 Add option to prefix namespace Swift type names

zhaakhi:
Swift does not appear to have a way to namespace things within a single pod.
This adds an option swift:prefix_namespace that will add Cocoa-style prefixes to the generated code if the file has a defined Swift namespace.

E.g. if a file has
"namespace swift ABC"
then the struct "MyStruct" will be renamed to "ABCMyStruct"

### DIFF
--- a/compiler/cpp/src/generate/t_swift_generator.cc
+++ b/compiler/cpp/src/generate/t_swift_generator.cc
@@ -1920,7 +1920,7 @@ void t_swift_generator::render_const_value(ostream& out,
       throw "compiler error: no const of base type " + t_base_type::t_base_name(tbase);
     }
   } else if (type->is_enum()) {
-    out << definition_name(value->get_identifier());
+    out << definition_name(value->get_identifier_with_parent());
   } else if (type->is_struct() || type->is_xception()) {
     
     out << type_name(type) << "(";
--- a/compiler/cpp/src/generate/t_swift_generator.cc
+++ b/compiler/cpp/src/generate/t_swift_generator.cc
@@ -1920,7 +1920,14 @@ void t_swift_generator::render_const_value(ostream& out,
       throw "compiler error: no const of base type " + t_base_type::t_base_name(tbase);
     }
   } else if (type->is_enum()) {
-    out << definition_name(value->get_identifier_with_parent());
+    string enum_value = value->get_identifier_with_parent();
+    if (prefix_namespace_) {
+      t_program* program = type->get_program();
+      if (program) {
+        enum_value = program->get_namespace("swift") + enum_value;
+      }
+    }
+    out << enum_value;
   } else if (type->is_struct() || type->is_xception()) {
     
     out << type_name(type) << "(";
--- a/compiler/cpp/src/generate/t_swift_generator.cc
+++ b/compiler/cpp/src/generate/t_swift_generator.cc
@@ -58,6 +58,7 @@ public:
     async_clients_ = false;
     promise_kit_ = false;
     debug_descriptions_ = false;
+    prefix_namespace_ = false;
 
     for( iter = parsed_options.begin(); iter != parsed_options.end(); ++iter) {
       if( iter->first.compare("log_unexpected") == 0) {
@@ -68,6 +69,8 @@ public:
         promise_kit_ = true;
       } else if( iter->first.compare("debug_descriptions") == 0) {
         debug_descriptions_ = true;
+      } else if( iter->first.compare("prefix_namespace") == 0) {
+        prefix_namespace_ = true;
       } else {
         throw "unknown option swift:" + iter->first; 
       }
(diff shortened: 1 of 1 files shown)


mfrawley:
This is a must-have for Swift support. Any eta on merging?

apocolipse:
In my opinion this is very Un-Swifty.  Namespace prefixes are an Objective-C Cocoa convention, Swift tries to get away from this convention (in fact with Swift 3.0, NS prefixes on many Foundation types have been dropped*)
It would be more proper to Nest them instead of rename them, and not that difficult either, simply install everything in a struct:
```
namespace swift MyNamespace
struct MyStruct { ... }
service MyService { ... } 
```
```swift
struct MyNamespace {
  struct MyStruct { ... }
  protocol MyService { ... }
}
```
Alternatively, Swift uses Module scoping for namespacing, so having the generator spit out an Xcodeproj with a Framework scheme, or SPM formatted file hierarchy, would be ideal, this would give you import semantics and direct access to your types when there's no conflicts, 
```swift
import MyNamespace
var aStruct = MyStruct()
```
as well as nested access when conflicts exist
```swift
var aStruct = MyNamespace.MyStruct()
```

Either module scoping or nesting are the ideal Swifty ways to leverage namespaces, Name prefixing shouldn't be used here.

*Dropped where types have been added as value types for more Swift-like handling, mutability, etc.

mfrawley:
@apocolipse Like the ideas. Protocols can't be nested under a struct. You get:
```
error: declaration is only valid at file scope protocol MyService {}
```
I like the Swift packages suggestion (which unfortunately are not in a stable Swift release yet). Frameworks influence app startup time and Cocoapods/Carthage already provide this functionality with podspecs (we use Cocoapods atm). 
  A bigger issue we've run into when updating (since we generate code for 6+ client langs) is that we have property naming collisions .e.g. .code in subtypes of NSError had to be renamed to errorCode, or .description. Being able to have some way of dealing with that would be nice. One possible solution might be a language-specific alias e.g. generate 'errorCode' for Swift, but leave it as 'code' for other langs. That would avoid having to uglify all client code with manually prefixed names. It's a different issue I know but it's also name-collision related, and has cost me some considerable effort.

apocolipse:
So for Cocoapods/Carthage, you're also using Frameworks (startup time is negligible).  I'd almost recommend manually handling separate Podspec's to keep things Module scope namespaced, it doesn't seem right at this point to have Thrift generate Carthage/Cocoapods configurations.  SPM should be good enough when Swift 3.0 is stable to where it can point to specific files for modules, additionally Foundation is ready for server side use then as well so perhaps SPM is the way to go.

Either way I wouldn't recommend this as a 1st-class feature for Swift code generation from Thrift.  Perhaps as an option "Use Obj-C Cocoa Namespacing", so that its explicitly clear that this isn't a Swift convention, but an old Obj-C/Cocoa convention.

zhaakhi:
Proper Swift modules would be great. The Cocoa-style prefixes have worked well for us as a stop-gap solution, though.

I don't mind changing the option name or description if that would clarify things.

apocolipse:
Yeah I think that would be the most appropriate way, making sure that people that use it /want/ to use it and/or know that its not a Swift convention.
SPM integration should come with Swift 3.0 support, which while in beta now is still not complete.  

Sidenote: If anyone's interested I'm working on building a Pure-Swift Thrift Library, basing it off of Swift 3.0 now (Started with Swift 2.0, but it was limited to Mac/iOS environments due to Foundation use, now Swift Foundation is part of Swift Standard Library 3.0 can be pushed for a public lib that works on Linux + Mac)  It's going to be in the works until Swift dev's finish up Foundation (still a bit incomplete as of writing this) but it will be much more "Swifty" (Following Swift 3.0 API Guidelines quite strictly), and may require an updated Code generator as well.  I'm not too familiar with the Code generators so should someone like to work with me for a Swift 3.0 Library + Code Generator that'd be great :D (Ideally i'm shooting to make it still compatible with the existing Swift generator, but should any changes be needed we should add a 3.0 generator option)

jeking3:
This hasn't been touched in over a year - what needs to be done?

zhaakhi:
Hello, sorry that I haven't had time to work more on this.
In principle it could be merged with just a rebase and a change to the option name.
In practice https://github.com/apache/thrift/pull/1084 should take priority, and may reduce the need for this by using module namespacing.
If anyone still wants this after https://github.com/apache/thrift/pull/1084 has been merged, I could do a rebase and fix it up.

jeking3:
It looks like module namespacing in #1084 is the preferred mechanism for doing this.  If you agree, we can close this.

jeking3:
THRIFT-3773 was merged.

jeking3:
module namespacing in #1084 is the preferred mechanism for doing this - I am closing this.  Feel free to reopen if additional insights arise.

PR #1005, opened 2016-05-05
THRIFT-3814 Fix contention in TNonblockingServerTest

nsuke:


lib/cpp/test/TNonblockingServerTest.cpp
```diff
@@ -47,14 +47,32 @@ struct Handler : public test::ParentServiceIf {
...
+        }
         server->serve();
-      } catch (const TException&) {
-        error = true;
+      } catch (const transport::TTransportException&) {
+        if (retry_count > 0) {
+          ++port;
+          startServer(retry_count - 1);
+        } else {
+          throw;
```
tpcwang: Seems bad to throw an unhandled exception from a thread, but maybe this is catastrophic and rare enough that this is fine? The rest of the change looks good to me.

nsuke: It's a test failure so my intention here was "no need to catch" and let it die with std::terminate with exception log.
The original `error` flag was needed for retrying from the main thread.

PR #1010, opened 2016-05-15
Thrift 3046: Allow PSR4 class loading for generated classes (PHP)

sinceat:
Original patch provided by @fduch, just integrated it into the current version and added a test.
(Original author didn't respond for over a year)
Hopefully this PR will drain more attention than the issue (as hoped by @creker )

### DIFF
--- a/compiler/cpp/src/generate/t_php_generator.cc
+++ b/compiler/cpp/src/generate/t_php_generator.cc
@@ -62,6 +62,7 @@ public:
     validate_ = false;
     json_serializable_ = false;
     nsglobal_ = ""; // by default global namespace is empty
+    psr4_ = false;
     for( iter = parsed_options.begin(); iter != parsed_options.end(); ++iter) {
       if( iter->first.compare("inlined") == 0) {
         binary_inline_ = true;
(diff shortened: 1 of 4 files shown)


fduch:
Hi @sinceat, thank you for sibmitting my patch as a PR.

My response time is not so big as you mention since the question to me was submitted [about month ago](https://issues.apache.org/jira/browse/THRIFT-3046?focusedCommentId=15246649&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15246649) :)

Anyway it would be great to save original commit history.
In order to do this we need to apply original patch to the version that was used to create this patch against: https://github.com/apache/thrift/commit/624781c1c4b89dce9f48fa80b841b5cd3d93aebb
After that result HEAD should be rebased against current master with minimal conflict (you have been resolving it on the patch level during submitting this PR) and populated with your additional commit. If you need any help with this, please let me know.

sinceat:
Hi @fduch, did the rebase. Would have done it that way from the start up but that's hard without parent commit id ;-)

fduch:
@sinceat yeah, great that now its allowed for THRIFT to make PR's instead old-style JIRA-issues with patches submitted)

Now hoping that this could be reviewed..

appkr:
👍 You guys are so cool~ I and my colleague will test the merged master today. While we are testing, can you tag a release?

PR #1013, opened 2016-05-19
THRIFT-3833 haxe http php server implementation (embeding into php web server)

oprudkyi:
- support of TSimpleServer.runOnce (run till input EOF)
- TWrappingServerTransport wraps TTransport into TServerTransport
- README.md - notes about integration into PHP web server
- tests
- update tutorial with http support
  - make tutorialserver_php_http
  - make tutorialclient_http


tutorial/haxe/src/Main.hx
```diff
@@ -51,14 +51,44 @@ class Main {
...
+        #if ! (flash || js || phpwebserver)
         try {
               ParseArgs();
         } catch (e : String) {
             trace(e);
             trace(GetHelp());
             return;
         }
+
+        #elseif  phpwebserver
```
Jens-G: I would like to have all those "//remap trace to error log" sections reafctored into a dedicated method. As it is now, it bloats the main() method with stuff that definitely belongs onto a different level of abstraction. The main method serves two purposes: parsing args (or similar) to set up the system, then start the right program mode. Aside from some added root level error handling, that's it and that should be it. Anything else belongs into either the first part, or the second.

Considereing the fact that we seem to need the same code over and over again for PHP (see your own README changes), why don't we put the stuff it into a library utility function, preventing the need for further copy+paste programming?

lib/haxe/README.md
```diff
@@ -117,3 +117,48 @@ may only support parts of the library, or not be compatible at all.
...
+========================
+
+- you have to import PHP files generated by haxe into PHP
+```php
+require_once  dirname(__FILE__) . '/bin/php-web-server/Main-debug.php';
+```
+
+- trace() by default outputs into stdout (http response), so you have to redirect it to stderr or you own logs, something like
+```haxe
+//remap trace to error log
```
Jens-G: see my other comment

test/haxe/router.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
Jens-G: Both router.php files lack the ASF header.

test/haxe/src/Main.hx
```diff
@@ -31,6 +31,32 @@ import thrift.test.*;  // generated code
 class Main
 {
     static function main() {
+        #if phpwebserver
+        //remap trace to error log
+        haxe.Log.trace = function(v:Dynamic, ?infos:haxe.PosInfos)
```
Jens-G: see my other comment

Jens-G:
Looks nice, good work! Only a few things ... ;-)

tutorial/haxe/src/Main.hx
oprudkyi: Hi, 
There is a problem with php - there is no standard way for logs, while stderr(error_log) is ok for tests/samples the real applications/frameworks use  own logs , so for laravel it will be \Log::info(...) , for Kohana - Kohana::log('info', .. ) etc 
also python/nodejs/lua etc will require own trace handler as well 
i.e. trace/logs issue belongs to haxe only and can't be fixed at thrift side .
Of course it is possible to create thrift oriented trace() but then user level code will be required to follow such scheme  too, and it will be even more unusable and inconvenient (like using thrift provided trace function over all haxe libs)
so I decided to leave this issue out of thrift codebase

test/haxe/router.php
oprudkyi: Sorry, 
ASF headers has been added to both router.php

oprudkyi:
I've filed issue to the haxe https://github.com/HaxeFoundation/haxe/issues/5264

tutorial/haxe/src/Main.hx
Jens-G: But we still can put everything between ```#elseif  phpwebserver``` and ```#endif``` into a separate method, couldn't we? In my opintion, we should not sacrifice code clarity to some obscure PHP edge case.

oprudkyi: code has been updated

oprudkyi: also, just sample how it looks in a real server
php side:
```php
<?php

namespace App\Libs\Tsp;

function tsp_trace_function($message) 
{
	\Log::info($message);
}	

require_once  dirname(__FILE__) . '/../../../tsp_handler_web/tsp_web_handler.php';
```

haxe side:
```haxe
	static public function main() 
	{
		//remap trace
		haxe.Log.trace = function(v:Dynamic, ?infos:haxe.PosInfos) 
		{ 
			....
			untyped __call__('App\\Libs\\Tsp\\tsp_trace_function', '${msg}${newValue}');
		}
	}
```

PR #1015, opened 2016-05-24
Update connection.js

ldjking:
bug when more than 1 package,
eg: the second package   service_name is null but the client is similar with the first


jeking3:
This needs an Apache Jira ticket in the Thrift project, and needs to be rebased against master so that it gets a clean CI build.

jeking3:
@jfarrell this PR can be closed; the line of code removed here was already done by another commit.

jeking3:
@jfarrell this PR can be closed; the line of code removed here was already done by another commit.

PR #1017, opened 2016-05-24
thrift-3841: dart compact protocol incorrectly serializes doubles

brianshannan-wf:
The compact protocol expects doubles to be serialized/deserialized with a little endian byte order instead of with a network (big endian) byte order. This fixes that. Tested locally by running a go server and dart client with a service containing a method that sent and received a double and making sure what was received on both ends was correct.

@markerickson-wf @tylertreat-wf @stevenosborne-wf


stevenosborne-wf:
+1

markerickson-wf:
+1

markerickson-wf:
@brianshannan-wf Can you close this PR?  I did not construct the commit message correctly for it to be closed automatically.

PR #1021, opened 2016-05-31
Fix THRIFT-3844

RobberPhex:
update `php_thrift_protocol7.cpp`
let `php_thrift_protocol` can compiled with php7.0.7


jeking3:
@RobberPhex could you rebase this against master and re-submit to get a build that passes CI tests?

PR #1022, opened 2016-05-31
THRIFT-3845

RobberPhex:
use `isBinaryAccelerated` to indicate use extension `thrift_protocol`.

When `TBinaryProtocolAccelerated` warped in `TMultiplexedProtocol`,extension `thrift_protocol` can also be used for `client serialization`, `client deserialization`, `server serialization`.

notice: `server deserialization` can't use extension `thrift_protocol`.

### DIFF
--- a/compiler/cpp/src/generate/t_php_generator.cc
+++ b/compiler/cpp/src/generate/t_php_generator.cc
@@ -78,7 +78,7 @@ public:
       } else if( iter->first.compare("nsglobal") == 0) {
         nsglobal_ = iter->second;
       } else {
-        throw "unknown option php:" + iter->first; 
+        throw "unknown option php:" + iter->first;
       }
     }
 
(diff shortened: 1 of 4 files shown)


jeking3:
@RobberPhex could you rebase this against master so we can get a build that passes all tests?

RobberPhex:
@jeking3 rebased

compiler/cpp/src/generate/t_php_generator.cc
```diff
@@ -1341,9 +1341,13 @@ void t_php_generator::generate_process_function(t_service* tservice, t_function*
     return;
   }
 
-  f_service_ << indent() << "$bin_accel = ($output instanceof "
-             << "TBinaryProtocolAccelerated) && function_exists('thrift_protocol_write_binary');"
+  // for compatibility, add method_exists
+  f_service_ << indent() << "$bin_accel = method_exists($output, 'isBinaryAccelerated')" << endl;
+  indent_up();
+  f_service_ << indent() << " && $output->isBinaryAccelerated()" << endl
+             << indent() << " && function_exists('thrift_protocol_write_binary');"
```
nsuke: The real problem may be that existing code generation here is too specific to one concrete type.
What if we move these thrift_protocol_..._binary calls to TBinaryProtocolAccelerated.php ?
That way we can move isStrict... checks there too so that we don't need them in TProtocolDecorator either.

lib/php/lib/Thrift/Protocol/TBinaryProtocolAccelerated.php
```diff
@@ -62,4 +62,9 @@ public function isStrictWrite()
   {
     return $this->strictWrite_;
   }
+
+  public function isBinaryAccelerated()
+  {
+    return true;
+  }
```
nsuke: I agree we need something like this but the name can be more general.
For example, `isAccelerated` or even `isDynamic`.
(Thinks about the day when we add TCompactProtocolAccelerated or anything without generated read/write calls...)

PR #1023, opened 2016-06-06
THRIFT-3849

keatoncarter-wf:
Created a Serializer/Deserializer in dart that mirrors the go version.

@markerickson-wf 
@tylertreat-wf 
@stevenosborne-wf 
@brianshannan-wf

ref: https://issues.apache.org/jira/browse/THRIFT-3849

### DIFF
--- a/lib/dart/lib/src/serializer/t_deserializer.dart
+++ b/lib/dart/lib/src/serializer/t_deserializer.dart
@@ -1,0 +1,51 @@
+/// Licensed to the Apache Software Foundation (ASF) under one
+/// or more contributor license agreements. See the NOTICE file
+/// distributed with this work for additional information
+/// regarding copyright ownership. The ASF licenses this file
+/// to you under the Apache License, Version 2.0 (the
+/// "License"); you may not use this file except in compliance
+/// with the License. You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing,
+/// software distributed under the License is distributed on an
+/// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+/// KIND, either express or implied. See the License for the
+/// specific language governing permissions and limitations
+/// under the License.
+
+part of thrift;
+
+class TDeserializer {
+  final message = new TMessage('Deserializer', TMessageType.ONEWAY, 1);
+  TTransport transport;
+  TProtocol protocol;
+
+  TDeserializer() {
+    this.transport = new TBufferedTransport();
+    this.protocol = new TBinaryProtocolFactory().getProtocol(transport);
+  }
+
+  void read(TBase msg, Uint8List data) {
+    protocol.writeMessageBegin(message);
+    
+    protocol.writeBinary(data);
+    
+    protocol.writeMessageEnd();
+
+    msg.read(protocol);
+  }
+
+  void readString(TBase msg, String data) {
+    protocol.writeMessageBegin(message);
+    
+    protocol.writeString(data);
+
+    protocol.writeMessageEnd();
+    
+    transport.flush();
+
+    msg.read(protocol);
+  }
+}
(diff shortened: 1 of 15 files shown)


lib/dart/lib/src/serializer/t_serializer.dart
```diff
@@ -0,0 +1,43 @@
...
+
+part of thrift;
+
+class TSerializer {
+  final message = new TMessage('Serializer', TMessageType.ONEWAY, 1);
+  TBufferedTransport transport;
+  TProtocol protocol;
+
+  TSerializer() {
+    this.transport = new TBufferedTransport();
```
brianshannan-wf: Would it be better to put these as optional arguments in the constructor with the buffered transport/binary protocol as the defaults?

keatoncarter-wf: you need to use the transport to instantiate the protocol. That way the protocol will use the correct transport when writing/reading.

brianshannan-wf:
+1

tylertreat-wf:
lgtm

lib/dart/lib/src/serializer/t_deserializer.dart
```diff
@@ -0,0 +1,41 @@
...
+/// under the License.
+
+part of thrift;
+
+class TDeserializer {
+  final message = new TMessage('Deserializer', TMessageType.ONEWAY, 1);
+  TBufferedTransport transport;
+  TProtocol protocol;
+
+  TDeserializer() {
```
markerickson-wf: This could take an optional TProtocolFactory

lib/dart/lib/src/serializer/t_serializer.dart
markerickson-wf: I think `TSerializer` could take an optional `TProtocolFactory`, e.g. the Java versions of the serializer / deserializer take arguments.

lib/dart/lib/src/serializer/t_deserializer.dart
```diff
@@ -0,0 +1,41 @@
...
+    this.transport = new TBufferedTransport();
+    this.protocol = new TBinaryProtocolFactory().getProtocol(this.transport);
+  }
+
+  void read(TBase msg, Uint8List data) {
```
markerickson-wf: I think there might be a better name for `msg` in these four locations, since msg makes me think of a `TMessage` not a `TBase`.  Java uses `base`

lib/dart/pubspec.yaml
```diff
@@ -23,7 +23,7 @@ author: Apache Thrift Developers <dev@thrift.apache.org>
 homepage: http://thrift.apache.org
 documentation: http://thrift.apache.org
 environment:
-  sdk: ">=1.12.0 <2.0.0"
+  sdk: ">=1.13.0 <2.0.0"
 dependencies:
   crypto: "^0.9.0"
```
markerickson-wf: I think you can remove the dependency on `crypto` now?

lib/dart/lib/src/transport/t_buffered_transport.dart
```diff
@@ -22,13 +22,13 @@ class TBufferedTransport extends TTransport {
...
 
-  Uint8List _consumeWriteBuffer() {
+  Uint8List consumeWriteBuffer() {
     Uint8List buffer = new Uint8List.fromList(_writeBuffer);
     _writeBuffer.clear();
     return buffer;
   }
 
-  void _setReadBuffer(Uint8List readBuffer) {
+  void setReadBuffer(Uint8List readBuffer) {
```
markerickson-wf: I don't think you need to change the visibility of this method here and in all of the subclasses.

lib/dart/lib/src/serializer/t_deserializer.dart
```diff
@@ -0,0 +1,41 @@
...
+    transport.setReadBuffer(data);
```
markerickson-wf: Instead of changing the visibility of `setReadBuffer`, I think you can use the existing public method:
```
transport.writeAll(data);
```

markerickson-wf:
Just a few minor things, but looks good - thanks for adding this and updating the BASE64 encoder.

markerickson-wf:
+1 thanks!

brianshannan-wf:
+1

PR #1024, opened 2016-06-12
THRIFT-3850: All apache builds are failing when initiated from a github

jfarrell:
THRIFT-3850: All apache builds are failing when initiated from a github pull request to Jenkins

Disabled jenkins jobs, added blank line to .travis.yml and triggering builds to get baseline for fixing travis.

PR #1025, opened 2016-06-12
THRIFT-3852 A Travis-CI job fails with "write error"

nsuke:


nsuke:
As steted in #1024 (THRIFT-3850) , it's a workaround against hypothetical resources starvation, by re-balancing workload between jobs.

PR #1026, opened 2016-06-12
THRIFT-3851

connorgorman:


jeking3:
We can ignore the failed apache jenkins build - as such this one looks okay from a build perspective.  I would like someone with some go experience to code review it however.

PR #1028, opened 2016-06-16
MOD: fix the blocked bug when call Stop() twice or more

huaiyun:
When Stop() is called twice or more, and no new connection accepted from AcceptLoop(), the Stop() will be blocked because the quit channel is full.
Use a flag to ensure the Stop() executed only once, and no block.

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -22,11 +22,13 @@ package thrift
 import (
 	"log"
 	"runtime/debug"
+	"sync/atomic"
 )
 
 // Simple, non-concurrent server for testing.
 type TSimpleServer struct {
-	quit chan struct{}
+	quit    chan struct{}
+	stopped int64
 
 	processorFactory       TProcessorFactory
 	serverTransport        TServerTransport
@@ -149,8 +151,10 @@ func (p *TSimpleServer) Serve() error {
 }
 
 func (p *TSimpleServer) Stop() error {
-	p.quit <- struct{}{}
-	p.serverTransport.Interrupt()
+	if atomic.CompareAndSwapInt64(&p.stopped, 0, 1) {
+		p.quit <- struct{}{}
+		p.serverTransport.Interrupt()
+	}
 	return nil
 }
 


jeking3:
Is there an Apache Jira ticket for this issue?

lib/go/thrift/simple_server.go
```diff
@@ -22,11 +22,13 @@ package thrift
...
 	"log"
 	"runtime/debug"
+	"sync/atomic"
 )
 
 // Simple, non-concurrent server for testing.
 type TSimpleServer struct {
-	quit chan struct{}
+	quit    chan struct{}
+	stopped int64
```
jeking3: Pardon my ignorance here - in go do int64 automatically initialize to zero?

lib/go/thrift/simple_server.go
```diff
@@ -149,8 +151,10 @@ func (p *TSimpleServer) Serve() error {
 }
 
 func (p *TSimpleServer) Stop() error {
-	p.quit <- struct{}{}
-	p.serverTransport.Interrupt()
+	if atomic.CompareAndSwapInt64(&p.stopped, 0, 1) {
```
jeking3: Do you want to be checking for a return value of 0 here instead of non-zero?

huaiyun: Actually no ticket, I am sorry.

Itn64 is initialized to 0 automatically. It will be setted to 1 and return true when 0, return false with no change otherwise.

I want to make the Stop() content actually executed only at the first calling, when stopped equals to 0, and repeated callings will be skipped.






------------------------------------------------------------------------------------------------------------
姓名：尹智宇
电话：13312111858
邮箱：yinzhiyu@xiaomi.com
地址：北京市海淀区毛纺路58号小米总参
发件人: James E. King, III <notifications@github.com>
发送时间: 2016年6月17日 8:21
收件人: apache/thrift
抄送: 尹智宇; Author
主题: Re: [apache/thrift] MOD: fix the blocked bug when call Stop() twice or more (#1028)


In lib/go/thrift/simple_server.go<https://github.com/apache/thrift/pull/1028#discussion_r67445983>:

> @@ -149,8 +151,10 @@ func (p *TSimpleServer) Serve() error {
>  }
>
>  func (p *TSimpleServer) Stop() error {
> -     p.quit <- struct{}{}
> -     p.serverTransport.Interrupt()
> +     if atomic.CompareAndSwapInt64(&p.stopped, 0, 1) {


Do you want to be checking for a return value of 0 here instead of non-zero?

―

jeking3: When I looked at the go documentation for CompareAndSwapInt64 the documentation states that it returns the old value, so I assume you want the quit and interrupt code to run when this method returns 0, meaning it is the first time it has run.

jeking3: I created a Jira item for your issue:

https://issues.apache.org/jira/browse/THRIFT-3855

jeking3: Hi, still looking for a reply to my last comment about 0 and 1.  I think your code relies on the fact that stop is called twice, as it will skip the first bump of the counter from 0 to 1, and then on the second call to stop it will actually stop, so I cannot approve the request.  If I misread it please let me know.

jeking3: @huaiyun need you to look at this again.

huaiyun: @jeking3 
The CompareAndSwapInt64 will return true of false to mark if the counter has been swapped.

the golang doc:
// The compare-and-swap operation, implemented by the CompareAndSwapT
// functions, is the atomic equivalent of:
//
//      if *addr == old {
//              *addr = new
//              return true
//      }
//      return false

At the first call, it returns true. Because the counter is 0 as initialized, and will be swapped to 1.
At the second call or more, it will return false. Because the counter is already 1, not 0.
So this make the Stop() only works at the first call. The more will be skipped.

a simple function is here:
https://play.golang.org/p/s-yHhQHG_O

jeking3: Okay the example makes it clear.  I thought I found documentation that said it returns the previous value... I guess not!

PR #1030, opened 2016-06-18
THRIFT-3856 Update Debian Dependencies

matyapiro31:
Issue is https://issues.apache.org/jira/browse/THRIFT-3856 .

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,14 +1,14 @@
 Source: thrift
 Section: devel
 Priority: extra
-Build-Depends: debhelper (>= 9), build-essential, mono-gmcs, python-dev, ant,
-    mono-devel,  libmono-system-web2.0-cil, erlang-base, ruby-dev | ruby1.9.1-dev, autoconf, automake,
+Build-Depends: debhelper (>= 9), build-essential, mono-mcs, python-dev, ant,
+    mono-devel,  libmono-system-web4.0-cil, erlang-base, ruby-dev | ruby1.9.1-dev, autoconf, automake,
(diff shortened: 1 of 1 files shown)


matyapiro31:
@nsuke , what makes it be failed?
I could not find its cause.

nsuke:
@matyapiro31 sorry for the delay, as the two "debian package" jobs for ubuntu and debian passed, I don't think we need any further action for CI.

debian/control
```diff
@@ -1,14 +1,14 @@
...
 Priority: extra
-Build-Depends: debhelper (>= 9), build-essential, mono-gmcs, python-dev, ant,
-    mono-devel,  libmono-system-web2.0-cil, erlang-base, ruby-dev | ruby1.9.1-dev, autoconf, automake,
+Build-Depends: debhelper (>= 9), build-essential, mono-mcs, python-dev, ant,
+    mono-devel,  libmono-system-web4.0-cil, erlang-base, ruby-dev | ruby1.9.1-dev, autoconf, automake,
     pkg-config, libtool, bison, flex, libboost-dev | libboost1.53-dev,
     python-all, python-setuptools, python-all-dev, python-all-dbg,
     python3-all, python3-setuptools, python3-all-dev, python3-all-dbg,
-    openjdk-6-jdk | java-sdk,
+    openjdk-7-jdk | openjdk-8-jdk,
```
nsuke: It might make sense to add [default-jdk](https://packages.debian.org/jessie/default-jdk) here.
What do you think ?

matyapiro31: No problem, I think.

PR #1031, opened 2016-06-22
THRIFT-3859: Unix Domain Socket Support in Objective-C

clindsay:
Existing behavior for port-based socket transport should be unchanged by this commit. TSocketServer and TSocketTransport have been refactored to support sockets created using either a port or a path.

### DIFF
--- a/lib/cocoa/src/server/TSocketServer.h
+++ b/lib/cocoa/src/server/TSocketServer.h
@@ -41,6 +41,10 @@ extern NSString *const TSockerServerTransportKey;
              protocolFactory:(id <TProtocolFactory>)protocolFactory
             processorFactory:(id <TProcessorFactory>)processorFactory;
 
+- (instancetype) initWithPath: (NSString *) path
+              protocolFactory: (id <TProtocolFactory>) protocolFactory
+             processorFactory: (id <TProcessorFactory>) processorFactory;
+
 @end
 
 
--- a/lib/cocoa/src/server/TSocketServer.m
+++ b/lib/cocoa/src/server/TSocketServer.m
@@ -25,7 +25,7 @@
 
 #import <sys/socket.h>
 #include <netinet/in.h>
-
+#include <sys/un.h>
 
 
 NSString *const TSocketServerClientConnectionFinished = @"TSocketServerClientConnectionFinished";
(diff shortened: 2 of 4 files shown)


clindsay:
I have written unit tests for this change, but am not sure where to contribute them. I see no "cocoa" directory under the "test" directory. Is there another place where I can contribute my unit tests?

clindsay:
The diff generated by GitHub makes the change look a little bigger than it actually is. In TSocketServer.m, the `initWithSocket` method is effectively the same as the old `initWithPort` method, but with the socket creation code factored out into `createSocketWithPort`. Additional code has been added to create a socket using a path instead of a port.

jeking3:
Would it be possible to rebase this on master so we can get a clean CI build?  Also enabling unix socket transport in the "make cross" test would be really nice to validate this code.

clindsay:
No problem. Just rebased on master. I will work on adding to the "make cross" test.

jeking3:
To enable "make cross" support you can follow the pattern in test/tests.json to add the "domain" transport to your cocoa test.  Your cocoa test_client needs to understand that transport and create the correct kind of client, etc.

clindsay:
Thanks @jeking3. Sorry for the slow response, a bit busy these days!

Just to clarify, as far as I can tell there isn't a cocoa test client today. Is this correct? So, in order to add these tests to "make cross", we would first need to add a "cocoa" folder to `thrift/test`, and put together a test server, and test client executable that could participate in these tests?

jeking3:
If there isn't one there yet I would recommend following the pattern for java in there.  Specifically you would need a new test/ subdirectory for cocoa with test_client and test_server scripts, and you would need a stanza in tests.json that the test.py script could run.

You can test your scripts against a single client or server like this, after running a "make" following configure in test/:

./test.py --client cocoa --server java
./test.py --client java --server cocoa

You could substitute cpp for java as well.  Then when those pass, try a "make cross" to see if everything passes.

jeking3:
Hi @clindsay just checking in to see if you had made any progress on adding cross test support for cocoa.   If not that's okay... it would be really nice to have it though.

clindsay:
Hey @jeking3, sorry for going silent there. I've had a lot of other things on my plate, and haven't gotten to this yet, but I still intend to add this support when I have a chance!

jeking3:
Have you used this in a project and know that it works?  Given this is an improvement I'd like to get it in, but if you don't have time right now to add make cross support, I can add a story in Jira to complete the work.

clindsay:
I've been using it in a project since last summer, communicating between a python process and a Swift/Objective-C process. The project has been in beta with a few hundred users, and no reports of any issues attributable to this yet. I've also written a set of unit tests, testing the behavior of a server and client in the same process both using this transport, that are all passing.

So yes, it's definitely at least working in these two scenarios.

jeking3:
Okay, well I'm going to approve and merge, and open a Jira item to add crosstest support for cocoa (if not there, I'll check) and for unix sockets in cocoa.

gaoshanyu:
Who can tell me why the cocoa lib has Swift2.3 codes? Where I can download full Objective-C lib? Or is there a way to make the lib to support Swift 3 +?

PR #1032, opened 2016-06-22
THRIFT-3476: fix missing header

jeking3:
stdexcept was missing, but std::runtime_error is used


tpcwang:
:+1: makes sense

jeking3:
Waiting for master to rev so I can rebase and kick this build again.

nsuke:
@jeking3 can you rebase this one too ? It seems quite safe and I could just test it on my travis account but it's always better to have history here.

nsuke:
@jeking3 thanks, merging as failures cannot possibly be related (go and dart).
Filed a [ticket](https://issues.apache.org/jira/browse/THRIFT-3883) for non-cross-test one.

PR #1033, opened 2016-06-22
THRIFT-2156: fix errno handling in server socket

jeking3:
I took the information from the patch submitted in 2013 and applied it to the code where appropriate to fix the errno handling.

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -284,7 +284,7 @@ void TServerSocket::listen() {
   hints.ai_family = PF_UNSPEC;
   hints.ai_socktype = SOCK_STREAM;
   hints.ai_flags = AI_PASSIVE | AI_ADDRCONFIG;
-  sprintf(port, "%d", port_);
+  snprintf(port, sizeof("65535"), "%d", port_);
 
   // If address is not specified use wildcard address (NULL)
   TGetAddrInfoWrapper info(address_.empty() ? NULL : &address_[0], port, &hints);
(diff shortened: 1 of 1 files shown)


lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -284,7 +284,7 @@ void TServerSocket::listen() {
   hints.ai_family = PF_UNSPEC;
   hints.ai_socktype = SOCK_STREAM;
   hints.ai_flags = AI_PASSIVE | AI_ADDRCONFIG;
-  sprintf(port, "%d", port_);
+  snprintf(port, sizeof("65535"), "%d", port_);
```
tpcwang: sizeof(port) seems ever-so-slightly clearer.

lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -520,23 +523,23 @@ void TServerSocket::listen() {
   if (retries > retryLimit_) {
     char errbuf[1024];
     if (!path_.empty()) {
-      sprintf(errbuf, "TServerSocket::listen() PATH %s", path_.c_str());
+      snprintf(errbuf, 1024, "TServerSocket::listen() PATH %s", path_.c_str());
```
tpcwang: ditto here with sizeof(errbuf)

tpcwang:
:+1: I don't really mind if you want to leave the two noted lines as-is.

lib/cpp/src/thrift/transport/TServerSocket.cpp
jeking3: I think the allocation of the buffer for port is just as bizarre, but I wanted to remain consistent.  Let's see if the build passes; if it fails and I need to rebase on master to kick another because of a transient issue, I can make the changes you suggest.

jeking3:
Rebased now that the correct fix for THRIFT-3768 is in which was causing build failures... using this to prove that is resolved so I can re-mark THRIFT-3768 closed and THRIFT-3861 fixed.

PR #1034, opened 2016-06-27
THRIFT-3857

leaves4j:
thrift js:node complier support an object as parameter not an instance of struct


leaves4j:
What should I do with the fail checks

leaves4j:
I have rebased on upstream/master and add those commits into latest branch @jeking3

jeking3:
The only build failure can be attributed to THRIFT-4071.  Otherwise, this is just waiting for someone with node.js experience to review.  Specifically I need to know if this change is backwards compatible (i.e. will existing handler code work unchanged, or is this a breaking change?).

leaves4j:
@jeking3 it's not a breaking change

jeking3:
I'll keep monitoring.  I would prefer a code review from someone else who uses node.js; it looks like a positive change to me.

PR #1036, opened 2016-06-29
THRIFT-3867 Specify BinaryProtocol and CompactProtocol

erikvanoosten:
First version of a specification of the binary protocol and compact protocol. Please review and comment.


doc/specs/thrift-binary-protocol-encoding.md
```diff
@@ -0,0 +1,467 @@
...
+
+Binary protocol (old, 9+ bytes):
++--------+--------+--------+--------+--------+...+--------+--------+--------+--------+--------+--------+
+| name length                       | name                |00000mmm| seq id                            |
++--------+--------+--------+--------+--------+...+--------+--------+--------+--------+--------+--------+
+```
+
+Where:
+
+* `vvvvvvvvvvvvvvv` is the version, an unsigned 15 bit number fixed to `1`. The leading bit must be `1`.
```
Jens-G: That's a bit too short. Please compoare against
https://github.com/apache/thrift/blob/master/lib/cpp/src/thrift/protocol/TBinaryProtocol.h#L40

doc/specs/thrift-binary-protocol-encoding.md
```diff
@@ -0,0 +1,467 @@
...
+* `unused` is an ignored byte.
+* `mmm` is the message type, an unsigned 3 bit integer. The 5 leading bits must be `0` as some clients (checked for java in 0.9.1) take the whole byte.
+* `name length` is the byte length of the name field, a signed 32 bit integer encoded in network (big endian) order (must be >= 0).
+* `name` is the method name to invoke, a UTF-8 encoded string.
+* `seq id` is the sequence id, a signed 32 bit integer encoded in network (big endian) order.
+
+Because name length must be positive (therefor the first bit is always `0`), the first bit allows the receiver to see
+whether the strict format or the old format is used. Therefore a server and client using the different variants of the
```
Jens-G: thereforE

Jens-G: Ah, I see. I still think that it is a bit confusing. First we say the value must always be 1, but later we add the footnotes that it actually may not be /that/ easy? Mmhh.

doc/specs/thrift-binary-protocol-encoding.md
```diff
@@ -0,0 +1,467 @@
...
+* _Oneway_: 4
+
+For a method name of 32 bytes, the binary protocol (strict) needs 44 bytes and the compact protocol needs 36 to 40 bytes.
+
+## Method arguments, return types and exceptions
+
+TODO: method arguments are encoded as a struct
+
+TODO: return value are encoded as a ?
+
```
Jens-G: ... struct. They also may contain exceptions which are technically just a special kind of structs.

doc/specs/thrift-binary-protocol-encoding.md
```diff
@@ -0,0 +1,467 @@
...
+```
+
+Because each field header contains the field-id (as defined by the IDL), the fields can be encoded in any order.
+Thrift's type system is not extensible, you can only encode the primitive types and structs. Therefore is also possible
+to handle unknown fields while decoding; these are simply ignored. While decoding the field type can be used to
+determine how to decode the field value.
+
+In the binary protocol the field header is always 3 bytes long. In the compact protocol the field header packs a lot
+more cleverness. In most cases it is 1 byte long. In special cases it can grow to 2 or 3 bytes. When your field-ids are
+very large it can even grow to 4 bytes.
```
Jens-G: I'd just refer to zigzag above. That's enough cleverness.

Jens-G:
Overall pretty neat. I still have a few complaints.

1. The document repeats a lot of stuff that can be found elsewhere, e.g. in the Thrift Whitepaper. There is nothing per se wrong with repeating stuff, but from my feeling it makes the spec a bit bloaty. 

2. I think the structure should be revised grossly. No need to panic, I'm going to explain what I have in mind. By looking at the bits and bytes the reader may miss an important point: The order of data is always the same and entitrely Independent from the wire format, because that's the way Thrift Protocols work. 

 * At the first level we have messages. They have a certain structure.

 * at the next level, a particular data item has a certain structure: a  map always consists of two types and a count, plus the elements. That does not change between protocols. What changes is the way how a particular protocol organizes the data onto the wire.

 * and one level deeper we are at what I mean by revising the structure: I would not constantly jump back and forth between protocols to explain structs, lists, sets, integers, etc. Instead, there should be one chapter focusing on binary, and another one for compact. That makes it much easier 
  a) for the reader who is usually interested in only one protocol at a time, and 
  b) for future maintainers to add more protocols, such as JSON.

doc/specs/thrift-binary-protocol-encoding.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Oneway
+
+Type `Oneway` is only used starting from Apache Thrift 0.9.3. Earlier versions do _not_ send TMessages of type `Oneway`,
+even for service methods defined with the `oneway` modifier.
+
+When client sends a request with type `Oneway`, the server must _not_ send a response (steps 3 and 4 are skipped).
+Strangely enough (in the Java code generated by Apache Thrift 0.9.1 up to 0.9.3), only responses of type `Response` are
+skipped. Responses of type `Exception` are always send. There is no correct way to handle this situation from the client
+perspective; you either wait for a response or not, you can't do both. Luckily this has been fixed _after_ Apache Thrift
```
Jens-G: The Thrift compiler prevents the combination of ```oneway``` plus ```throws``` right from the start [since March 2003](https://github.com/apache/thrift/commit/6d94390375e865e0c774df1dc072ea1774eba7b1), any types other than ```void``` are disallowed since [THRIFT-2030](https://issues.apache.org/jira/browse/THRIFT-2030). So that does not hold.

erikvanoosten:
Thanks @Jens-G . That's all useful feedback. Next week I have time to work on it further.

Regarding structure: it took me weeks to collect all this information and create the document. At some points I no longer could keep the entire thing in my head. Your structure proposal makes sense and clears it for me also :)

doc/specs/thrift-binary-protocol-encoding.md
erikvanoosten: I don't understand. https://github.com/apache/thrift/blob/master/lib/cpp/src/thrift/protocol/TBinaryProtocol.h#L40 indicates that the version mask takes the first 2 bytes = 16 bits. The first bit must be `1` is used to distinguish the format and is therefore not part of the version number.

Would it be clearer if I say that the version is fixed to `000 0000 0000 0001`?

doc/specs/thrift-binary-protocol-encoding.md
erikvanoosten: Yeah, funny stuff. I'll work on better wording.

doc/specs/thrift-binary-protocol-encoding.md
erikvanoosten: Field headers are more clever then that: they have a short and long form. In the short form they contain a field-id delta.

erikvanoosten:
@Jens-G wrote:
> The Thrift compiler prevents the combination of oneway plus throws right from the start since March 2003, any types other than void are disallowed since THRIFT-2030. So that does not hold.

I looked at it again. `ProcessFunction` in thrift 0.9.3 contains the following:

```
    try {
      result = getResult(iface, args);
    } catch(TException tex) {
      LOGGER.error("Internal error processing " + getMethodName(), tex);
      TApplicationException x = new TApplicationException(TApplicationException.INTERNAL_ERROR, 
        "Internal error processing " + getMethodName());
      oprot.writeMessageBegin(new TMessage(getMethodName(), TMessageType.EXCEPTION, seqid));
      x.write(oprot);
      oprot.writeMessageEnd();
      oprot.getTransport().flush();
      return;
    }
```

As you can see when `getResult` throws it is encoded regardless of oneway. You may ask when does `getResult` throw? Looking at the generated code, for example:

```
      public getById_result getResult(I iface, getById_args args) throws org.apache.thrift.TException {
        getById_result result = new getById_result();
        try {
          result.success = iface.getById(args.userId);
        } catch (UserNotFoundException exc1) {
          result.exc1 = exc1;
        }
        return result;
      }
```

it throws when it can not decode the request, or when the service implementation throws any unchecked exception.

erikvanoosten:
Note: last commit is still work in progress.

erikvanoosten:
All, @Jens-G, this documentation is ready for another review. Thanks!

erikvanoosten:
Closing this due to lack of interest of repo owners.

Jens-G:
Why?

Jens-G:
Thanks for the reminder. Actually, I like it. Too bad it's closed now.

erikvanoosten:
Feel free to take it. I won't support it anymore though.

erikvanoosten:
The feedback cycle took more then a month. This is too long for me. I do 
not have the mental capacity to keep work in my head for that long.

Kind regards,
     Erik.



Op 20-09-16 om 21:49 schreef Jens Geyer:

erikvanoosten:
The specification documents are up at https://erikvanoosten.github.io/thrift-missing-specification/.

PR #1037, opened 2016-06-30
nodejs client fails to import 'assert'

edenhochbaum:
So when the connection fails, we just get an error about assert not being defined (rather than details of the connection failure.)


jeking3:
Please open an Apache Thrift Jira ticket for this issue and rebase the pull request against master so we get a clean CI build.

jeking3:
THRIFT-4172

PR #1038, opened 2016-07-01
replace 'binary' on the base types

DontPanicBaby:
json generator creates file with "binary" type instead thrift base types. As a result, it is impossible to identify the type.


Jens-G:
But ```binary```is a basis type?

DontPanicBaby:
I need to know exactly type of variable. The current implementation does not allow it. I write serializer based on json generator's result and i can't serialize binary, becose i don't know how many bytes needs for variable.

jeking3:
This needs a Jira item opened for it, and needs to be rebased and merge conflicts fixed.
See; https://thrift.apache.org/docs/HowToContribute
Thanks.

jeking3:
This needs a Jira item opened for it, and needs to be rebased and merge conflicts fixed.
See; https://thrift.apache.org/docs/HowToContribute
Thanks.

PR #1039, opened 2016-07-06
THRIFT-2835 Add possibility to distribute generators separately

dtmuller:
I rebased the original pull request #368 from @nsuke and added a few fixes on top of that.


nsuke:
@dtmuller thanks ! I'll look into the weekend after next.

lib/c_glib/test/CMakeLists.txt
```diff
@@ -18,12 +18,15 @@
...
 set(TEST_PREFIX "c_glib")
 
-include_directories(${Boost_INCLUDE_DIRS})
+# Find required packages
+set(Boost_USE_STATIC_LIBS ON) # Force the use of static boost test framework
+find_package(Boost 1.53.0 REQUIRED COMPONENTS unit_test_framework)
+include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
+
+#Make sure gen-cpp and gen-c_glib files can be included
+include_directories("${CMAKE_CURRENT_BINARY_DIR}")
```
nsuke: I couldn't find any boost test inside lib/c_glib. Is this needed ?

compiler/cpp/test/CMakeLists.txt
```diff
@@ -18,15 +18,15 @@
...
+    set(Boost_USE_STATIC_LIBS ON) # Force the use of static boost test framework
+    find_package(Boost 1.53.0 REQUIRED COMPONENTS unit_test_framework)
+    include_directories(SYSTEM "${Boost_INCLUDE_DIRS}")
 
-#Make sure gen-cpp files can be included
-include_directories("${CMAKE_CURRENT_BINARY_DIR}")
+    #Make sure gen-cpp files can be included
+    include_directories("${CMAKE_CURRENT_BINARY_DIR}")
 
-if(${WITH_PLUGIN})
```
nsuke: Shared boost test support is added after this patch.
So we now need to fix the test.

compiler/cpp/test/plugin/conversion_test.cc
```diff
@@ -269,6 +268,8 @@ void test_const_value(t_const_value* sut) {
 #undef T_CONST_VALUE_CASE
   case t_const_value::CV_MAP:
     BOOST_CHECK_EQUAL(sut->get_map().size(), sut2->get_map().size());
+    // NOTE This test only works with map with a single entry because we can't
+    // rely on map ordering with pointers as key.
     BOOST_CHECK_EQUAL(sut->get_map().begin()->first->get_type(),
                       sut2->get_map().begin()->first->get_type());
     BOOST_CHECK_EQUAL(sut->get_map().begin()->second->get_type(),
```
nsuke: I think we still want to have multiple values inside map, e.g. by using `operator[]`/`find` on sut2.

compiler/cpp/src/audit/t_audit.cpp
```diff
@@ -41,12 +41,12 @@ void thrift_audit_warning(int level, const char* fmt, ...) {
...
   va_list args;
-  fprintf(stderr, "[Thrift Audit Failure:%s] ", g_curpath.c_str());
+  //fprintf(stderr, "[Thrift Audit Failure:%s] ", g_curpath.c_str());
   va_start(args, fmt);
   vfprintf(stderr, fmt, args);
   va_end(args);
   fprintf(stderr, "\n");
-  g_return_failure = true;
+  //g_return_failure = true;
 }
```
nsuke: We cannot just undo others' warnings and exit value manipulation...
I imagine there are some difficulties to keep these ?

compiler/cpp/src/plugin/plugin.thrift
```diff
@@ -0,0 +1,198 @@
...
+namespace delphi Thrift.Plugin
+namespace go thrift
+namespace haxe org.apache.thrift.plugin
+namespace hs Thrift.Plugin
+namespace java org.apache.thrift.plugin
+namespace ocaml Thrift
+namespace perl Thrift.Plugin
+namespace php thrift.plugin
+namespace py thrift.plugin
+namespace rb Thrift
```
nsuke: `namespace erl` is added after this patch. Would be nice to have it from the day 1 of this feature.

nsuke:
This looks mostly good except for a few issues I commented.
The t_audit part seems to be the biggest challenge.
Not sure you're going to work through them, as some of them are from original code and others' newer patches.
Let me know if you have any plan on it.

lib/c_glib/test/CMakeLists.txt
dtmuller: I removed the unit_test_framework dependency.

compiler/cpp/test/plugin/conversion_test.cc
dtmuller: Fixed this by extracting to a map of dereferenced types for comparison.

compiler/cpp/src/plugin/plugin.thrift
dtmuller: Added plugin namespace for erlang.

compiler/cpp/src/plugin/plugin.cc
```diff
@@ -402,6 +402,13 @@ THRIFT_CONVERSION(t_program, from.path, from.name) {
   boost::for_each(from.services | boost::adaptors::transformed(&resolve_service),
                   boost::bind(&::t_program::add_service, to, _1));
 
+  for (std::vector<t_program>::const_iterator it = from.includes.begin();
+       it != from.includes.end();
+       it++) {
+    ::t_program* tmp = new ::t_program((*it).path, (*it).name);
+    convert<t_program, ::t_program>((*it), tmp);
+    to->add_include(tmp);
+  }
```
dtmuller: Just stumbled over another thing: the `t_program::includes_` were empty on the plugin side...

compiler/cpp/test/CMakeLists.txt
dtmuller: I'm not sure what you mean by that?!

dtmuller:
I fixed a few of your findings. On the remaining two issues concerning `t_audit/t_logging` and the shared boost test support I'm not sure what to do...

nsuke:
@dtmuller thanks, the new changes look good.
As you uncovered another defect on includes, I've devised simple integration test using `diff` and resolved a few issues too.
I've pushed some commits on top of yours into the PR #368 .
(moved original patch to https://github.com/nsuke/thrift/tree/THRIFT-2835-orig)
Would be awesome if you can review/try it.

dtmuller:
I added the deployment of headers needed to build plugins. I changed the directory structure to look the same as under /lib (added an extra thrift folder). The reason: for cmake I batch-copied the headers with `install(DIRECTORY ...)` which put them into `/usr/local/include` instead of `/usr/local/include/thrift`. But I changed this back and now copy the files as needed (to exclude unneeded generated headers). Still, I think it's nice to have the dir structure matching the namespaces...

@nsuke If you agree with the header deployment you can pull over my last 5 commits to your original pull-request. If not I can change it back to deploy the headers using the old structure.

The rest looks good to me, I also tested it with my plugin. Only I couldn't get CI test #4 green - though it works locally?!

nsuke:
The `thrift` directory makes sense.
The test # 4 was relying on `sleep` in test script that is inherently flaky.
I've removed a double pclose in parent process code so we'll see if it fixes the problem in #368 result.

Another problem is that Appveyor CI is failing due to include error.
Not sure why it's only happening on Windows.

dtmuller:
So, it looks like this could make it into the master soon - once CI is green. :smiley: 
Unfortunately I'm afraid that I can't help with the windows fix.

dtmuller:
@nsuke Do you have a rough idea when this PR can be merged? If there is more work to do I can assist, please let me know...

nsuke:
@dtmuller as you may have noticed I've squashed commits for merging in #368.
If there's no concern I'll push them within a week or so.

dtmuller:
@nsuke That's great, thanks. I see, the two PRs are in sync, so I will close this one...

nsuke:
@dtmuller just committed, thanks for all the help and improvements !
It might have not happened at all if you didn't pick it up and revive it ...

dtmuller:
@nsuke Thanks a lot for your effort - I really enjoyed working with you.

PR #1040, opened 2016-07-08
THRIFT-3848: Perl ForkingServer must ignore SIGCHLD to avoid infinite loop

jeking3:


jeking3:
Rebased on master to fix the TServerIntegrationTest hang.

PR #1044, opened 2016-07-09
THRIFT-3873: fix compiler warnings on windows with VS2010

jeking3:
Also fixed: concurrency tests were using assert() in a bad way limiting them to debug builds only
Also fixed: changed config.h VERSION to THRIFT_VERSION to avoid conflicts with windows or third party headers

With this pull request in a CMake build on Windows VS2010 or on Linux (Ubuntu 14.04) the C++ code compiles clean with /W3 level warnings enabled.

### DIFF
--- a/build/cmake/ConfigureChecks.cmake
+++ b/build/cmake/ConfigureChecks.cmake
@@ -23,11 +23,11 @@ include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckFunctionExists)
 
-# If AI_ADDRCONFIG is not defined we define it as 0
-check_symbol_exists(AI_ADDRCONFIG "sys/types.h;sys/socket.h;netdb.h" HAVE_AI_ADDRCONFIG)
-if(NOT HAVE_AI_ADDRCONFIG)
-set(AI_ADDRCONFIG 1)
-endif(NOT HAVE_AI_ADDRCONFIG)
+if(MSVC)
+  check_symbol_exists(AI_ADDRCONFIG "winsock2.h" HAVE_AI_ADDRCONFIG)
+else(MSVC)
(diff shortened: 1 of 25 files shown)


jeking3:
Found that the automake environment wasn't building properly - will push a fix.

tpcwang:
I didn't review the changes in detail, but I agree with the change and like the fact that we can compile cleanly with /W3 :+1:

jeking3:
I rebased on THRIFT-3936 so that it would compile and resolved merge conflicts that occurred since July.

nsuke:
@jeking3 hi, sorry for the delay, what is your thought on "#define __STDC_LIMIT_MACROS + ..._MAX" v.s. numeric_limists<...>::max() ?
If it were my own code, I would always use numeric_limists but I had once very strange compilation error on VC++14 with this too.
Do you think define + _MAX is better or as good for our code base ?

jeking3:
I have tended to use the C99 limit macros in the past, however either one should be acceptable.  I don't believe I have experienced issues with numeric_limits before.  Either solution should be portable going forward as well so I don't see the need to favor one over the other.

jeking3:
Going to rebase one more time on THRIFT-3936 which now includes the fix for appveyor.

jeking3:
Rebasing again on THRIFT-3936 so it compiles on VS2010 after THRIFT-2836 was introduced... so we'll want THRIFT-3936 to be merged before this one.

jeking3:
I'll rebase on master once THRIFT-3936 is merged, since something odd happened with the build.

jeking3:
I'm closing this to re-submit it without THRIFT-3936 in it since it was merged to master, but when I rebased it's still here.

PR #1045, opened 2016-07-10
THRIFT-3876 haxe js/nodejs client

oprudkyi:
- only http transport is supported 
- testing of browser js made via phantomjs 
- tutorial of browser js uses haxe-php web server for files delivery 
- default implementation of haxe's HttpClient reimplemented/patched:
    - browser js implemented to support binary encoding (avoid any utf8 conversions)
    - for nodejs implemented non-async client (via ChildProcess.spawnSync and it's slow) 
- makefile for tests updated to avoid sleeps (it kills server when client stops)

### DIFF
--- a/lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
+++ b/lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
@@ -125,7 +125,7 @@ class TFileStream implements TStream {
         var copyJsBuf = new js.node.Buffer(count);
         bytesRead = Fs.readSync(Input, copyJsBuf, 0, count, lastInputOffset);
         if(bytesRead > 0) {
-			lastInputOffset += bytesRead;
+            lastInputOffset += bytesRead;
             var readBuf = copyJsBuf.hxToBytes();
             buf.blit(offset, readBuf, 0, bytesRead);
         }
--- a/lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
+++ b/lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
@@ -24,6 +24,9 @@ import haxe.io.BytesBuffer;
 import haxe.io.Input;
 import haxe.io.Output;
 
+#if js
+import js.node.Fs;
+#end
 
 enum TFileMode {
     CreateNew;
@@ -36,8 +39,14 @@ class TFileStream implements TStream {
 
     public var FileName(default,null) : String;
 
+    #if !js
     private var Input  : sys.io.FileInput;
     private var Output : sys.io.FileOutput;
+    #else
+    private var Input  : Null<Int>;
+    private var Output  : Null<Int>;
+    private var lastInputOffset : Int = 0;
+    #end
 
 
     public function new( fname : String, mode : TFileMode) {
@@ -45,13 +54,25 @@ class TFileStream implements TStream {
         switch ( mode)
         {
             case TFileMode.CreateNew:
+                #if !js
                 Output = sys.io.File.write( fname, true);
+                #else
+                Output = Fs.openSync(fname, WriteCreate);
+                #end
 
             case TFileMode.Append:
+                #if !js
                 Output = sys.io.File.append( fname, true);
+                #else
+                Output = Fs.openSync(fname, AppendCreate);
+                #end
 
             case TFileMode.Read:
+                #if !js
                 Input = sys.io.File.read( fname, true);
+                #else
+                Input = Fs.openSync(fname, js.node.FsOpenFlag.Read);
+                #end
 
             default:
                 throw new TTransportException( TTransportException.UNKNOWN,
@@ -62,11 +83,19 @@ class TFileStream implements TStream {
 
     public function Close() : Void {
         if( Input != null) {
+            #if !js
             Input.close();
+            #else
+            Fs.closeSync(Input);
+            #end
             Input = null;
         }
         if( Output != null) {
+            #if !js
             Output.close();
+            #else
+            Fs.closeSync(Output);
+            #end
             Output = null;
         }
     }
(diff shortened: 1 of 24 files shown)


lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
```diff
@@ -24,6 +24,9 @@ import haxe.io.BytesBuffer;
 import haxe.io.Input;
 import haxe.io.Output;
 
+#if js
+import js.node.Fs;
+#end
```
Jens-G: We have basically two dource files in one. Maybe we should refactor this .later on.

test/haxe/src/Main.hx
```diff
@@ -46,11 +46,14 @@ class Main
             if( args.printHelpOnly)
                 return;
 
+            #if js
+                TestClient.Execute(args);
+            #else
             if (args.server)
```
Jens-G: Why do we need the ifdef here? No other way?

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
```diff
@@ -39,12 +46,25 @@ class THttpClient extends TTransport {
...
-          request_ = new Http(requestUrl);
-        request_.addHeader( "contentType", "application/x-thrift");
+        #if !js
+        request_ = new Http(requestUrl);
+        #else
+        request_ = new JsHttp(requestUrl);
+        request_.async = false;
+        #end
+        request_.addHeader("Content-Type", "application/x-thrift");
+        request_.addHeader("Accept", "application/x-thrift");
```
Jens-G: BTW, would you mind to integrate the new Content-Type(s) from THRIFT-2756 as well? New ticket, of course, and only a question.

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
```diff
@@ -94,10 +114,241 @@ class THttpClient extends TTransport {
...
+        try {
+            if( post )
+                r.open("POST",url,async);
+            else if( uri != null ) {
+                var question = url.split("?").length <= 1;
+                r.open("GET",url+(if( question ) "?" else "&")+uri,async);
+                uri = null;
+            } else
+                r.open("GET",url,async);
+        } catch( e : Dynamic ) {
```
Jens-G: Did you test that against which other target language? If it works only against Haxe, that's not sufficient :-)

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
```diff
@@ -94,10 +114,241 @@ class THttpClient extends TTransport {
...
+          + "});"
+          + "response.on('error', function(error) {"
+          + "process.stdout.write(JSON.stringify({err: error, errorMessage : error.message}));"
+          + "});"
+          + "}).on('error', function(error) {"
+          + "process.stdout.write(JSON.stringify({err: error, errorMessage : error.message}));"
+          + "});"
+          + headersCode
+          + sendCode 
+          + "req.end();";
```
Jens-G: Did you think about possible code injection issues? How do we prevent that? Tested?

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
```diff
@@ -94,10 +114,241 @@ class THttpClient extends TTransport {
...
+            "path":request,
+            "method":post ? "POST" : "GET",
+            "agent":false,
+            "withCredentials":#if (haxe_ver >= 3.3) me.withCredentials #else false #end
+        };
+
+        var headersCode = "";
+        for( h in headers )
+            headersCode += 'req.setHeader("${h.header}","${h.value}");';
+
```
Jens-G: Same here. What happens if some bad guy passes some "handcrafted special effects" header values?

Jens-G:
Good work!

Jens-G:
Some of the added files lack the ASF Header. Not sure about package.json, but phantom-js-client.test.js should have one.

jeking3:
It does seem like this could be refactored into a base implementation holding the common code and then an implementation per #ifdef branch that was introduced to clean it up.

lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
oprudkyi: Well, the haxe by self is such
anyway this file is mostly used for tests

test/haxe/src/Main.hx
oprudkyi: without def it tries to compile server side which isn't implemented

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: this will need redesign of api as well, because transport doesn't have access to protocol type information

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: as for me haxe has never worked with other langs

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: all params are wrapped into json or generated from private data (like headers) 
so it's seems safe for me

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: access to request_ and headers is 'private' for THttpClient. so there no way to pass extra headers into code 

of course jsHttp type isn't safe , but it isn't part of public api

oprudkyi:
@Jens-G 
>phantom-js-client.test.js should have one.
ASF header has been added

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: fix added - headers are passed as JSON

lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
Jens-G: > Well, the haxe by self is such

This is a code smell, no matter what "the haxe" thinks. But not blocking issue for the commit.

> anyway this file is mostly used for tests 

That's why it says "TTestFileStream"? No, wait ... it doesn't. Joking aside, why do you want to limit the usage artifically by saying "oh, this is only test stuff? I see a slass that may be used for other purposes.

test/haxe/src/Main.hx
Jens-G: Implement and throw if called?

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
Jens-G: But we /are/ already adding the Content Type here. Why do we need an API redesign for switching a string? *confused*

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
Jens-G: Then we need to fix that first, and I can't commit it. Thrift is about talking cross-language and cross-platform, not about one system/machine/language talking to itself. I tested the original implementation against C#, Delphi and Go, crossing Windows and Linux borders, and it worked.

lib/haxe/src/org/apache/thrift/transport/TFileStream.hx
oprudkyi: as for me current implementation is better over scattering the same functionality over different files 
i.e. always visible 'original' implementation as well js part which just reimplements original one in target depended way

test/haxe/src/Main.hx
oprudkyi: done

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: how code in the transport will get info what is content type used (binary/json/compact) ?
right now I see only analysis of data stream might help (ie check first byte, etc) but this is very ugly 
real implementation will include some TTransport.setContentType() function which should be called by every  TProtocol implementation , in all languages

lib/haxe/src/org/apache/thrift/transport/THttpClient.hx
oprudkyi: That incompatibility bug isn't related to haxe-js. but to all thrift-haxe , because when I run haxe-cpp client against non-haxe cpp server I have got errors

Jens-G: File JIRA ticket(s) for what you think is broken.

jeking3:
Was this set of changes tested with "make cross" as part of the build?

jeking3:
@Jens-G any thoughts on what needs to happen next?

Jens-G:
After I read that the cross tests doesn't need to be made because "it doesn't work anyway" my enthusiasm for this patch was basically send to zero . First, this is a plain lie. Second, when I read such statements it gives great insight into the attitude of the contributor. I'm not willing to even look at it anymore under that premise.

oprudkyi:
I don't care either , I had the particular issue and shared fix for the issue 
I do not have time to create JIRAs for unrelated stuff nor to implement cross-platform things . 
of course even asking me todo so I'm seeing for the first time in my life and I have never seen in any other foss projects 
though, any maintainer have all the rights todo anything he wants

jeking3:
So I agree that changes like this need representation in the cross test.  It would be nice if folks could patch up any ill will and move this forward.

PR #1048, opened 2016-07-13
Ensuring that HTTP failures will clear the http transport outbuf var

thomas07vt:
With the current implementation, any Net HTTP failure will raise from the #flush() method without resetting the @outbuf variable.

I think that resetting the @outbuf on these failures is more "expected" behaviour. Especially if there is a malformed request that the downstream server does not want to/can't handle. As far as I can tell, there is not way to clear the @outbuf var apart from the #flush() method, so if that fails, then you will just keep appending requests to the out buffer.

### DIFF
--- a/lib/rb/lib/thrift/transport/http_client_transport.rb
+++ b/lib/rb/lib/thrift/transport/http_client_transport.rb
@@ -50,6 +50,7 @@ module Thrift
       data = resp.body
       data = Bytes.force_binary_encoding(data)
       @inbuf = StringIO.new data
+    ensure
       @outbuf = Bytes.empty_byte_buffer
     end
   end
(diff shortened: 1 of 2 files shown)


thomas07vt:
Travic CI seemed to fail on something not related to my change. Can someone rerun the Travis CI build?

nsuke:
@thomas07vt hi thanks for the patch. The change makes sense to me.
Could you file an issue on our JIRA, if you haven't done so ?
https://github.com/apache/thrift/blob/master/CONTRIBUTING.md

thomas07vt:
@nsuke Just filed an issue:
https://issues.apache.org/jira/browse/THRIFT-3920

Please feel free to edit the issue(or tell me what should be edited) if I have left out any necessary info since this is the first time I have opened an issue on the Thrift Jira board.

PR #1049, opened 2016-07-14
THRIFT-3880 Fix erlang compact protocol boolean representation

binaryseed:
I've noticed that when using the compact protocol in `erlang`, the boolean flags are inverted.

It appears the we've simply defined the constants wrong. For reference, here are a few other languages which appear to constantly define `true` as `0x01` and `false` as `0x02`:

* https://github.com/apache/thrift/blob/master/lib/rb/lib/thrift/protocol/compact_protocol.rb#L36-L37
* https://github.com/apache/thrift/blob/master/lib/py/src/protocol/TCompactProtocol.py#L84-L85
* https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/protocol/TCompactProtocol.java#L101-L102

JIRA tracking issue: https://issues.apache.org/jira/browse/THRIFT-3880


jeking3:
In order to get this into the project we need an Apache Jira ticket to be opened against the THRIFT project and the pull request to ideally identify this ticket.  I agree with your assessment and your fix.  I would also recommend that as part of this fix we also attempt to re-enable the erlang compact protocol "make cross" tests which appear to be disabled:

jking@dvm-4:~/thrift/test$ grep 'erl' known_failures_Linux.json | grep compact
  "csharp-erl_compact_buffered-ip-ssl",
  "csharp-erl_compact_framed-ip-ssl",
  "erl-cpp_compact_buffered-ip",
  "erl-cpp_compact_buffered-ip-ssl",
  "erl-cpp_compact_framed-ip",
  "erl-cpp_compact_framed-ip-ssl",
  "erl-nodejs_compact_buffered-ip",
  "erl-rb_compact_buffered-ip",
  "erl-rb_compact_framed-ip",

Jens-G:
Good catch!

What about other languages? Did you check all of them? 
=> If yes, what was the result? 
=> If not, why not? ;-) Could you do that?

binaryseed:
Unfortunately, adding back in the cross language tests still causes failures :( Seems like there are other bugs lurking beyond the boolean one here

```
*** Following 9 failures were unexpected ***:
csharp-erl              compact      buffered-ip-ssl          failure(1)
csharp-erl              compact      framed-ip-ssl            failure(1)
erl-nodejs              compact      buffered-ip              failure(timeout)
erl-cpp                 compact      buffered-ip              failure(1)
erl-cpp                 compact      buffered-ip-ssl          failure(1)
erl-cpp                 compact      framed-ip                failure(1)
erl-cpp                 compact      framed-ip-ssl            failure(1)
erl-rb                  compact      buffered-ip              failure(1)
erl-rb                  compact      framed-ip                failure(1)
```

binaryseed:
I didn't check out all the languages, just a handful to verify that the spec is: true = 1, false = 2

nsuke:
@binaryseed thanks, I've verified your patch locally and it does fix the problem. It'll be merged soon.

I'm also working on extending cross tests so that we can assert true/false and endianess which cannot be verified by current round trip tests.

jeking3:
Aren't we supposed to have a clean build before merging?

nsuke:
The failures are from tentative re-activation submitted later as a separate commit
I've excluded it from the merge.

jeking3:
Would I be able to tell that happened from any log or history that is public?  From here it looks like the tests were re-enabled because you merged the hash with them enabled (d150f8b).  I think it is important to have a clean and visible history so everyone knows what's going on.  If in effect you simply merged 49cefb9 by reverting the changes in d150f8b before pushing, that makes sense.

nsuke:
In my environment d150f8b41ef3a3f63fadcc595a99bc38628cd636 looks exactly same as 49cefb9563a18bc1cc7ac70ad0f1830019a3cd65 ...

PR #1050, opened 2016-07-15
THRIFT-3879 Undefined evaluation order

dtmuller:
The evaluation order of f1 and f2 in `x = f1() + f2()` is undefined...


tpcwang:
Good catch!

PR #1051, opened 2016-07-18
THRIFT-3884 Fix Erlang compact protocol double endianess and boolean …

nsuke:
…list


lib/erl/src/thrift_compact_protocol.erl
```diff
@@ -355,7 +356,7 @@ read(This0, i64) ->
 read(This0, double) ->
   {This1, Bytes} = read_data(This0, 8),
   case Bytes of
-    {ok, <<Val:64/float-signed-big, _/binary>>} -> {This1, {ok, Val}};
+    {ok, <<Val:64/float-signed-little, _/binary>>} -> {This1, {ok, Val}};
```
jeking3: Why is "float" used here instead of "Double"?  I don't know erlang; just looking at line 222 above and expecting it to be similar.

nsuke: @jeking3 thanks, good poinit. I've checked the [doc](http://erlang.org/doc/programming_examples/bit_syntax.html) but couldn't find any `double`.

> The type can be integer, float, or binary.

So I'll replace the other part with float and also unify the order.

nsuke: Ooops, the `Double` part is just a variable, the same as `Val`.
Anyway I'll unify the order.

PR #1008, opened 2016-05-10
THRIFT-3827:php CompactProtocol readI64 function has bug, when value has 32bit ~64bit, Example：value=1461563457000

lnn1123:
php CompactProtocol readI64 function
when value = 1461563457000, old readI64 function get wrong value 1463710940648

PR #1052, opened 2016-07-20
THRIFT-3885 PHP: Error when readI64 in TCompactProtocol

czm1989:
fix bug when readI64 in TCompactProtocol of php.
see https://issues.apache.org/jira/browse/THRIFT-3885

### DIFF
--- a/lib/php/lib/Thrift/Protocol/TCompactProtocol.php
+++ b/lib/php/lib/Thrift/Protocol/TCompactProtocol.php
@@ -615,7 +615,7 @@ class TCompactProtocol extends TProtocol
       if ($shift >= 32) {
         $hi |= (($byte & 0x7f) << ($shift - 32));
       } elseif ($shift > 25) {
-        $hi |= (($byte & 0x7f) >> ($shift - 25));
+        $hi |= (($byte & 0x7f) >> (7 - ($shift - 25)));
       }
       if (($byte >> 7) === 0) {
         break;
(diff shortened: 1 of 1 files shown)


jeking3:
Please create an Apache Jira ticket in the THRIFT project to cover the defect, and put a link to the pull request in there.  In the future if you open a ticket first and put the name of the Jira item (like THRIFT-3855) into the pull request comments, it will get linked to the Jira ticket automatically.

jeking3:
I'm also wondering if this was causing any "make cross" errors that have been ignored and whether we can re-enable some of those tests.

nsuke:
I believe this is duplicate of THRIFT-3827 (#1008) and the correct fix is there too.
As PHP cross test has been useless, I'm verifying the fix in THRIFT-3886 (#1054).

czm1989:
@nsuke Yes, I agree.
It also can simply be `$hi |= (($byte & 0x7f) >> 4);` because of the shift is only 28 that satisfy the condition.
But according to the algorithm of writeVarint, I think 
`elseif ($shift > 25) {
  $hi |= (($byte & 0x7f) >> (32 - $shift));
}`
 to be the best. 
(When the shift is 25, the 7 bit only fill in the low 32bits. so the condition should be `$shift > 25`.)
Thanks for reading.

nsuke:
I'm inclined to make it stupid simple...
``` php
// Shift hi and lo together.
if ($shift < 28) {
  $lo |= (($byte & 0x7f) << $shift);
} elseif ($shift == 28) {
  $lo |= (($byte & 0x0f) << 28);
  $hi |= (($byte & 0x70) >> 4);
} else {
  $hi |= (($byte & 0x7f) << ($shift - 32));
}
```
@czm1989 @lnn1123 Thought ?
Otherwise I'd apply @lnn1123 's patch but keeping `> 25`.

nsuke:
Also I noticed that `if (true)` part below inside the existing function seems completely redundant.
`else` code even uses different set of variables.
We can clean it up without losing any value, right ?

PR #1056, opened 2016-07-25
THRIFT-3887: adding factories to the generated Java classes

korovkin:
adding factories to the generated Java classes, allowing the user to control the type of classes to be allocated when de-serializing the incoming message.

example of generated code:

```
public class Team implements ... {

  ...
  public static class TeamFactory {
    public Team allocate() { return new Team(); }
    public Team allocate(Team other) { return new Team(other); }
    public Team allocate(String teamId, String name, String domain, String email_domain, String plan) { return new Team(teamId, name, domain, email_domain, plan); }
  }

  public static volatile TeamFactory factory = new TeamFactory();
  public static void setFactory(TeamFactory f) { factory = f;}
  public static TeamFactory getFactory() { return factory;}

  ...
}
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,442 @@
 Apache Thrift Changelog
 
+Thrift 0.9.3
+--------------------------------------------------------------------------------
+## Bug
+    * [THRIFT-2441] - Cannot shutdown TThreadedServer when clients are still connected
+    * [THRIFT-2465] - TBinaryProtocolT breaks if copied/moved
+    * [THRIFT-2474] - thrift.h causes a compile failure
+    * [THRIFT-2540] - Running configure from outside the source directory fails
+    * [THRIFT-2598] - Add check for minimum Go version to configure.ac
+    * [THRIFT-2647] - compiler-hs: don't decapitalize field names, do decapitalize argument bindings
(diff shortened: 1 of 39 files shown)


korovkin:
contributing back from slackhq

korovkin:
CC: @smourachov

Jens-G:
Hi @korovkin, thanks! Could you please create a matching JIRA ticket for it, according to our [contrib guidelines](https://thrift.apache.or/docs/HowToContribute)?

korovkin:
sure. 
added: https://issues.apache.org/jira/browse/THRIFT-3887

jeking3:
This needs to be rebased on the current master and resubmitted; we need a clean CI build before we can proceed.

jeking3:
This needs to be rebased on the current master and resubmitted; we need a clean CI build before we can proceed.

jeking3:
Please change the destination to master, and rebase on master so we can get a fresh CI build in order for this to proceed.  Thanks.

jeking3:
Would also prefer to see a test exercising the new code.

korovkin:
we are no longer using this approach, safe to close

PR #1057, opened 2016-07-30
Update TSSLSocket.cpp

udit043:
Older version of Open SSL may create problem because they do not support 'SSL_set_tlsext_host_name()'.

### DIFF
--- a/lib/cpp/src/thrift/transport/TSSLSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLSocket.cpp
@@ -471,8 +471,10 @@ void TSSLSocket::checkHandshake() {
       }
     } while (rc == 2);
   } else {
-    // set the SNI hostname
-    SSL_set_tlsext_host_name(ssl_, getHost().c_str());
+    /* OpenSSL < 1.0.0 does not have SSL_set_tlsext_host_name() */
+    #if defined(SSL_set_tlsext_host_name) // set the SNI hostname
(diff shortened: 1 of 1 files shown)


jeking3:
Hi, could you please create a ticket for this in Apache Jira following the instructions at: https://thrift.apache.org/docs/HowToContribute

Thanks!

udit043:
Okay

lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -471,8 +471,10 @@ void TSSLSocket::checkHandshake() {
       }
     } while (rc == 2);
   } else {
-    // set the SNI hostname
-    SSL_set_tlsext_host_name(ssl_, getHost().c_str());
+    /* OpenSSL < 0.9.8f does not have SSL_set_tlsext_host_name() */
+    #if defined(SSL_set_tlsext_host_name) // set the SNI hostname
```
jeking3: Typically we would add something to the build system environment to differentiate this; also is there an alternative that can be used with older OpenSSL?  Other folks who are committers will need to decide if it is worth supporting an older and likely quite vulnerable (to hacks) OpenSSL library.

udit043: I have created a issue regarding this: https://issues.apache.org/jira/browse/THRIFT-3892
No, there is not any alternative for older version. There are very less possibilities of someone using very old OpenSSL library. This issue is negligible, it is just for improving the project.

PR #1058, opened 2016-08-02
Node.js: Set/unset client seqid for json_protocol and compact_protocol

vh:
Like binary_protocol does.


jeking3:
This change needs a Jira ticket; please see https://thrift.apache.org/docs/HowToContribute.  Also if you can rebase on master and push that will kick a new CI build; CI needs to pass (or failures easily explained) to merge.

jeking3:
This change needs a Jira ticket; please see https://thrift.apache.org/docs/HowToContribute. Also if you can rebase on master and push that will kick a new CI build; CI needs to pass (or failures easily explained) to merge.

jeking3:
@bananer is this worth pulling forward if it passes a build?

bananer:
@jeking3 To be honest, I don't know why this is useful/necessary. I think there is no need for a seqid because the connection between request and response is handled through callbacks automatically.

jeking3:
In general I want to make sure all clients send a seqid and all servers put that seqid in the result.  Without this we won't be able to have clients that can have multiple outstanding requests in the future.

bforbis:
This should have updates added to it that were done in https://issues.apache.org/jira/browse/THRIFT-4480, otherwise the server will spam logs for every requests if they come from clients that send seqid = 0 (perl does this)

jeking3:
Original author is not responsive and deleted his upstream branch/repo.  I am closing this.

PR #1059, opened 2016-08-04
Grift

chedeti:


jeking3:
Please open an Apache Jira ticket per the [contributing guidelines](https://thrift.apache.org/docs/HowToContribute) so we know more about what this is.

PR #1060, opened 2016-08-08
[THRIFT-3895] Fix java 1.8 with ant < 1.9 compatibility

dieend:


jeking3:
Would you be able to rebase this against master and force push an update?  This will kick off a new CI build.  I would like to see the build pass before merging into the project.

jeking3:
If the only build failure turns out to be the unrelated hspec-core issue in job #17 I will ignore that and merge it.

PR #1075, opened 2016-09-02
THRIFT-3916 Throw proper errors from JS, not strings

SimenB:
I'd love to add a test for this, but I can't find where to do that.

EDIT: I'm unable to get bison to register correctly in order to run the tests locally (Mac).

Is there some Dockerfile I could use that installs dependencies and an elegant way for me to run the tests?

### DIFF
--- a/compiler/cpp/src/generate/t_js_generator.cc
+++ b/compiler/cpp/src/generate/t_js_generator.cc
@@ -66,7 +66,7 @@ public:
       } else if( iter->first.compare("ts") == 0) {
         gen_ts_ = true;
       } else {
-        throw "unknown option js:" + iter->first; 
+        throw "unknown option js:" + iter->first;
       }
     }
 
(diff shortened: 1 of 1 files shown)


nsuke:
@SimenB we have `Dokerfile` s for build env inside build/docker/ subdirectories.
You can also do `docker pull thrift/ubuntu` or thrift/debian.

jeking3:
@SimenB if you are interested in resurrecting this let me know - rebase on current master and push to CI will run all our tests.

jeking3:
@SimenB if you are interested in resurrecting this let me know - rebase on current master and push to CI will run all our tests.

jeking3:
@bananer is this worth pulling forward if it passes a build?  Is it a breaking change?

bananer:
@jeking3 This is a well-intended change but I think it is not implemented correctly and will break error handling for users.

In some places in the compiler the input to `render_recv_throw` already is a proper `Error` (the JS exception class), e.g. https://github.com/apache/thrift/blob/2b09dfed9c6b858571e7d8829a2b4a4bcda18d6a/compiler/cpp/src/thrift/generate/t_js_generator.cc#L1648-L1652

This is not the case for other places:
https://github.com/apache/thrift/blob/2b09dfed9c6b858571e7d8829a2b4a4bcda18d6a/compiler/cpp/src/thrift/generate/t_js_generator.cc#L1673

The proper way of doing this would be to inspect all usages of `render_recv_throw` to see if the argument needs to be wrapped in `new Error(…)`. Still then, this could break error handling that relies on these cases being strings.

jeking3:
Original author has not responded to any comments and PR needs more work.  Closing the PR.

PR #1080, opened 2016-09-04
THRIFT-3891 TNonblockingServer configured with more than one IO threa…

nsuke:
…ds does not always return from serve() upon stop()

imported from JIRA issue.

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -126,6 +126,9 @@ private:
   /// Libevent object
   struct event event_;
 
+  /// Libevent object initialized
+  bool isEventInitialized_;
+
   /// Libevent flags
   short eventFlags_;
 
@@ -215,7 +218,8 @@ public:
   TConnection(THRIFT_SOCKET socket,
               TNonblockingIOThread* ioThread,
               const sockaddr* addr,
-              socklen_t addrLen) {
+              socklen_t addrLen)
+    : isEventInitialized_(false) {
     readBuffer_ = NULL;
     readBufferSize_ = 0;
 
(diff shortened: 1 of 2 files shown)


lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1524,7 +1532,7 @@ void TNonblockingIOThread::breakLoop(bool error) {
   }
 
   // sets a flag so that the loop exits on the next event
-  event_base_loopbreak(eventBase_);
+  event_base_loopexit(eventBase_, 0);
```
jeking3: I am curious to understand why this was changed.  event_base_loopexit will cause the lop to continue processing until drained.  This function is called breakLoop and I assume that means stop as soon as possible, which is what it was doing.

bgedik: There are some subtle differences between the two. The first difference is what you already mentioned. The second difference is that, event_base_loopbreak will have no effect when there is no loop running: http://www.wangafu.net/~nickm/libevent-book/Ref3_eventloop.html. In contrast, event_base_loopexit is not `lost` if it is called without having a loop running.

In the TNonblockingServer code, event_base_loopbreak is called without checking if the loop started running.

An alternative fix would be to keep the event_base_loopbreak, but somehow introduce additional synchronization to make sure it is not lost.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1524,7 +1532,7 @@ void TNonblockingIOThread::breakLoop(bool error) {
   }
 
   // sets a flag so that the loop exits on the next event
```
bgedik: I guess the comment is stale now. The loop now exits once the current iteration is complete.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
jeking3: Your commends correspond to the discussion in the Jira entry for this item from @nsuke.  Is that something you would have time to do?

bgedik: I created the JIRA entry, with a patch that works for us (in production and testing). I know it is not the ideal fix, but I was hoping that someone who is more versed in the ``TNonblockingServer`` implementation can work it. I don't have the time to dig into the design of ``TNonblockingServer``.

PR #1081, opened 2016-09-04
THRIFT-3854 add a way in java to clear TFramedTransport read buffers

nsuke:
imported from JIRA issue.

### DIFF
--- a/lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
+++ b/lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
@@ -65,7 +65,8 @@ public class TFastFramedTransport extends TTransport {
 
   private final TTransport underlying;
   private final AutoExpandingBufferWriteTransport writeBuffer;
-  private final AutoExpandingBufferReadTransport readBuffer;
+  private AutoExpandingBufferReadTransport readBuffer;
+  private final int initialBufferCapacity;
   private final byte[] i32buf = new byte[4];
   private final int maxLength;
 
@@ -104,6 +105,7 @@ public class TFastFramedTransport extends TTransport {
   public TFastFramedTransport(TTransport underlying, int initialBufferCapacity, int maxLength) {
     this.underlying = underlying;
     this.maxLength = maxLength;
+    this.initialBufferCapacity = initialBufferCapacity;
     writeBuffer = new AutoExpandingBufferWriteTransport(initialBufferCapacity, 1.5);
     readBuffer = new AutoExpandingBufferReadTransport(initialBufferCapacity, 1.5);
   }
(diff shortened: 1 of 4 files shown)


lib/java/test/org/apache/thrift/transport/TestTFastFramedTransport.java
```diff
@@ -19,6 +19,8 @@
 package org.apache.thrift.transport;
 
 public class TestTFastFramedTransport extends TestTFramedTransport {
+  protected final static int INITIAL_CAPACITY = 50;
+
   @Override
   protected TTransport getTransport(TTransport underlying) {
     return new TFastFramedTransport(underlying, 50, 10 * 1024 * 1024);
```
jeking3: I think 50 was supposed to be changed to ``INITIAL_CAPACITY`` here, something I could do as part of the merge.

lib/java/src/org/apache/thrift/transport/TFramedTransport.java
```diff
@@ -123,6 +123,12 @@ public void consumeBuffer(int len) {
     readBuffer_.consumeBuffer(len);
   }
 
+  public void clear() {
+    if (readBuffer_ != null) {
```
jsirois: `readBuffer_` is declared `private TMemoryInputTransport readBuffer_ = new TMemoryInputTransport(new byte[0]);` and never re-assigned in this class. Unless there is known code re-assigning via reflection, this should just be promoted to final and all the null checks - like the one added here - dispensed with.

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
```diff
@@ -164,6 +166,10 @@ public void consumeBuffer(int len) {
     readBuffer.consumeBuffer(len);
   }
 
+  public void clear() {
+    readBuffer = new AutoExpandingBufferReadTransport(initialBufferCapacity, 1.5);
```
jsirois: It would be nice to share the initialization code on line 110 above in the constructor - consider factoring out a factory or calling reset from the constructor to initialize `readBuffer` for the 1st time.

lib/java/test/org/apache/thrift/transport/TestTFastFramedTransport.java
clockfort: Yes.

lib/java/src/org/apache/thrift/transport/TFramedTransport.java
clockfort: (I'm the person who originally wrote this patch / feature request)
I agree, and can't find any reflected use cases in at least just the thrift codebase.

jeking3:
I made the requested changes locally but I don't know if I can push on top of this, so I will likely open a new pull request to obsolete this one.

PR #1082, opened 2016-09-07
THRIFT-3828 In cmake avoid use of both quoted paths and SYSTEM with include_directories()

jmesmon:
This allows us to avoid issues where there are no paths to be added to
the include path (include_directories() errors when given an empty
string).

Specifically, gcc-6 requires that libraries stop passing paths like
'/usr/include' (or they will get libstdc++ build errors), so these paths
will be empty more often in the future.

### DIFF
--- a/lib/cpp/CMakeLists.txt
+++ b/lib/cpp/CMakeLists.txt
@@ -24,7 +24,7 @@ else()
   find_package(Boost 1.53.0 REQUIRED)
 endif()
 
-include_directories(SYSTEM "${Boost_INCLUDE_DIRS}")
+include_directories(${Boost_INCLUDE_DIRS})
 include_directories(src)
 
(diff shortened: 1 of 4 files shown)


nsuke:
I cannot reproduce this using GCC 6 and boost headers in /usr/include.

jmesmon:
Issue was observed in yocto (aka open-embedded) when building for target. It likely requires cross compilation (and thus /usr/include is actually some other path) to trigger.

jeking3:
@jfarrell the Jira ticket was resolved as "Won't Fix" here; can you close this PR?  Thanks.

jeking3:
@jfarrell the Jira ticket was resolved as "Won't Fix" here; can you close this PR? Thanks.

PR #1086, opened 2016-09-14
THRIFT-3926 Emit an error for bad http status code

lifei:
Sometimes the thrift servers is not ready to serve requests from the client. The status code of response from the services is 40x or 50x. But the client is never telled in this version of http_connections. There is no error at all even when the response protocol is bad.

### DIFF
--- a/lib/nodejs/lib/thrift/http_connection.js
+++ b/lib/nodejs/lib/thrift/http_connection.js
@@ -169,6 +169,10 @@ var HttpConnection = exports.HttpConnection = function(host, port, options) {
     var data = [];
     var dataLen = 0;
 
+    if (response.statusCode !== 200) {
+      this.emit("error", new THTTPException(statusCode, response));
+    }
+
     response.on('error', function (e) {
       self.emit("error", e);
     });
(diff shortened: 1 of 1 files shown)


jfarrell:
ping @RandyAbernethy for thoughts on this patch

lifei:
https://issues.apache.org/jira/browse/THRIFT-3926

RobberPhex:
status code not equal 200 doesn't means it's a `TApplicationException`.

* When status code is 301/3xx, should thrift client follow the redirect?
* In some server implements, server return Exception with 4xx, Exception contains the message "Access Denied"

Httpcode is expression of server status, it can be used for debugging, for logging, not just a part of rpc.

If someone want client process this error, return Exception when http code is not right, and client can parse and raise this exception.

lifei:
> I think that putting the statusCode into the text description would be useful here

added

jeking3:
As-is the patch would require string parsing to get the statusCode which is brittle.  I like the overall direction of handling unexpected status code with an exception; consider using a derived exception that encapsulates the information so one could call getStatusCode().

lifei:
@jeking3 modified

jeking3:
Looks good to me.  I guess there is no TTransportException for the js runtime?  I'd like one other person more familiar with using the language to approve then we can merge it in, like @RandyAbernethy.

NameFILIP:
Looks like a critical error was introduced in this PR. Here is the fix for it: https://github.com/apache/thrift/pull/1531

PR #1089, opened 2016-09-15
THRIFT-3929 php namespace remove tail "\\"

RobberPhex:
when namespace is empty, and global namespace is not empty, don't add tail "\\"


RobberPhex:
@jfarrell Can you review this pr?

bmeynell:
FWIW I tested to use case described in `THRIFT-3929` locally and it worked as expected.

PR #1091, opened 2016-09-19
THRIFT-1108: Adding SSL support for ruby

mansinahar:


jfarrell:
Thanks for the patch,Committed, 04e6f62

PR #1093, opened 2016-09-23
Add persistence patch

StachOverflow:


compiler/cpp/src/generate/t_php_generator.cc
```diff
@@ -1720,6 +1720,7 @@ void t_php_generator::generate_service_client(t_service* tservice) {
                << "TBinaryProtocolAccelerated) && function_exists('thrift_protocol_write_binary');"
                << endl;
 
+    f_service_client << indent() << "$this->seqid_ = rand();" << endl;
     f_service_client << indent() << "if ($bin_accel)" << endl;
```
Jens-G: I'm not quite sure what that ```rand()``` should do here? The sequential number is expected to be sequential. I would be a bit surprised when a function called ```rand()``` starts to return sequential numbers.

lib/php/lib/Thrift/Protocol/TBinaryProtocol.php
```diff
@@ -246,6 +257,10 @@ public function readMessageBegin(&$name, &$type, &$seqid)
       }
     }
 
+    if ($seqid != $this->seqid_) {
+      throw new TApplicationException("TBinaryProtocol::ReadMessageBegin received SequenceID: $seqid not matches requested ID: $this->seqid_ " . TApplicationException::BAD_SEQUENCE_ID);
+    }
```
Jens-G: Ah, got it. Sorry, but you are doing it wrong.

sstach: Thanks for your comment but I'am very interested in what you believe would be appropriate in this case and how you would solve it, because I think this is a *major issue* in the php thrift implementation.

Here is mho:
I understand why you complain about the `rand()`. 
But a random number is much more useful than an sequential number that start with 0 and only increment itself. Before `seqid_` wasn't respected at all and it's used to identify whether the response belongs to my request or not, and not to determine the order of requests and responses. And lastly, the php client initiates the `seqid_` with 0 and so it could happen that two clients communicate with the server at the same time and client A receives the response of client B and doesn't track that issue because accidentally the `seqid_` is that one that client A expect.

compiler/cpp/src/generate/t_php_generator.cc
jeking3: I thign seqId is sequence ID and not sequential ID, however I agree it should be sequential.  Random is less efficient and less effective.

lib/php/lib/Thrift/Protocol/TBinaryProtocol.php
jeking3: The sequence ID was originally put in place for future expansion of the servers to allow for multiple outstanding requests from a client.  Most servers do not implement this.  For example on the C++ threaded servers there is one request per client at any given time.  The C# server is the same way.  The servers could reply with a seqId that matches the request the client sent, allowing a client to submit multiple requests, if the servers were written differently.  Most client implementations ignore seqId because they only issue one request at a time and not all servers fill in the field properly.  The seqId should be handled on the server as a connection-specific unique request id.

My recommendation would be to follow what other clients do and atomically increment a number starting at 1.  This will be safer than random.  It probably isn't safe to check that the server's response seqId equals the last request sent.  Some servers may not fill in the field.  This also limits your client to one concurrent request and hampers future expansion.

lib/php/lib/Thrift/Transport/TSocket.php
```diff
@@ -255,6 +255,16 @@ public function close()
   }
 
   /**
+   * Close the socket and try to
+   */
+  public function reconnect()
```
jeking3: I don't see what this has to do with seqId changes and it is incorrect.  You would want to call the TSocket::close() and not fclose here.

lib/php/lib/Thrift/Protocol/TBinaryProtocol.php
sstach: Okay the seqid_ itself is not the problem. I agree to your statement that it depends on the implementations of the client or the server.

But here is our problem, we introduced Thrift as a performance boost by using persistent connections. Now that we use it in production we identified that the PHP implementation is not robust enough for us. It occurs that the client Socket gets closed by timeout and the client got an Exception. If another client requested something he got the response that was meant for the client which got the timeout.

With the seqid_ we have a robust functionality to check that our java server returns the correct answer to our question, irrelevant whether its a rand() or a simple increment.
And if we have a miss match we can react on that and our client can handle the error by opening a new Socket and request again.

We spend a lot of time to get to the bottom of this problem and the seqid_ seems to be the missing piece we are looking for. We were surprised to find out that it seems that no one else is running into the same problem. Only a 5 year old  [abandoned issue](https://issues.apache.org/jira/browse/THRIFT-1310) seems to reference this topic.

Do we miss something about the persistence? Is there another way to determine if the response matches the request?

lib/php/lib/Thrift/Transport/TSocket.php
sstach: Here we need a functionality to close a socket even if it`s persistent and open it again. This way we can react to the situation mentioned in the above comment.

lib/php/lib/Thrift/Protocol/TBinaryProtocol.php
jeking3: I would suggest that the issue lies in the server implementation in PHP based on your description.  If a client can disconnect, then a new client connects and it receives the response from the server originally intended for the client that disconnected then the server is misbehaving and a massive security hole.  Perhaps it keeps a list of responses based on IP address of something (I haven't looked).  In any case, if you look at the C++ server, each connection is handled by a single thread, so there can only be one outstanding request at a time.  If the client disconnects, the thread completes the request and then fails to send the reply, and ends.  There is no possibility that another client would receive the reply.  I would recommend instead of trying to solve the issue on the client side that fixing the root cause on the server side would be better as it would be much more secure.  I don't like the notion that someone can connect to a thrift PHP server and possibly receive confidential information originally destined for another connection.

StachOverflow:
Still an issue and we fixed it that way. We see that this is not the best way and there are still security issues when a server sends the wrong message to a client. But this prevents the client to accept a message which is meant for another client.

Are there any new ideas how to solve this problem?

jeking3:
So overall, proper sequence ID handling is important.  Most languages don't need it for proper operation, as I suggested with C++ which processes one request per connection at a time.  In some work I did for Dell EqualLogic a number of years ago, I leveraged a sequence number to rewrite the C# thrift engine so it could support multiple outstanding requests, and effectively would allow multiple threads to share the same connection and submit multiple concurrent requests.  So I do like to see proper sequence IDs in the request/response for each language.  It would be nice to have better tests proving these work in two ways - one is to make sure that a seqId a client passes in for each server language and protocol returns the seqId in the response.  The other would be to make sure every client actually sends a unique seqId.  Both are possible through crosstest with some work.

In the meantime though, having proper support for sequence IDs probably starts in the protocol implementation.  The protocol carries the sequence.  Generating a sequence ID in the compiler, or adding code that does seems incorrect to me.  It should be up to the constructor for a new binary protocol header to generate a new sequence ID, and the server implementation should honor it by carrying it through the response.

jeking3:
@StachOverflow will you be moving this forward or should it be closed?

jeking3:
@StachOverflow will you be moving this forward or should it be closed?

StachOverflow:
@jeking3 I put this topic back on my schedule and try to find a proper solution.

jeking3:
It looks like there's a deeper issue here, and I don't quite understand the use case.  Are two clients being used on the same transport?  As far as I know, this isn't supported by any of the implementations today.  Recommend opening multiple connections to the server - one for each client, even if they are in the same process.  While the change proposed perhaps moves in the direction stated, I'm not convinced proper locking would be done to avoid interleaving message contents and corrupting the stream.

This issue has no Jira ticket and has been idle for a long time.  I'm going to close it.  It can be reopened in the future by opening a Jira ticket.

PR #1094, opened 2016-09-24
Fixes THRIFT-3855

asm0dey:
Replaced complex CompareAndSwapInt64 with syncOnce.

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -185,7 +185,10 @@ func (p *TSimpleServer) processRequests(client TTransport) error {
 			log.Printf("error processing request: %s", err)
 			return err
 		}
-		if !ok {
+		if err, ok := err.(TApplicationException); ok && err.TypeId() == UNKNOWN_METHOD {
+			continue
+		}
+ 		if !ok {
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


lib/go/thrift/simple_server.go
```diff
@@ -183,9 +186,6 @@ func (p *TSimpleServer) processRequests(client TTransport) error {
 			log.Printf("error processing request: %s", err)
 			return err
 		}
-		if err, ok := err.(TApplicationException); ok && err.TypeId() == UNKNOWN_METHOD {
-			continue
-		}
```
Jens-G: I can't see the connection. Please explain.

Jens-G: How is that related? Please explain.

asm0dey: It's removed by gofmt. I thought that gofmt knows better tan me, so I didn't revert that change. should I?

asm0dey: @Jens-G --^

Jens-G: I would doubt that. If that is true that would probably qualify for a bug report. I would not expect a formatting tool to arbitrarily delete valid code lines.

asm0dey: @Jens-G  Reverted back that change. Thank you for review!

asm0dey: And again I dunno why one of builds has failed, but it looks like it's nothing to do with Fo code.

PR #1096, opened 2016-09-26
fix snprintf error in TServerSocket.cpp under vs2013

CompileMaster:
fix snprintf error in vs2013

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -16,7 +16,9 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
+#if _MSC_VER
+#define snprintf _snprintf
+#endif
 #include <thrift/thrift-config.h>
 
 #include <cstring>
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-#if _MSC_VER
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
 #define snprintf _snprintf
 #endif
 #include <thrift/thrift-config.h>


Jens-G:
Hi @CompileMaster, a bit more information about the problem would be nice. Aside from that, we have a nice [contribution guide](http://thrift.apache.org/docs/HowToContribute) that we would like you to follow.

CompileMaster:
Compiling under VS2013 
error C3861: “snprintf”:  Can't find the identifier  , line 282, 527, 528 in TServerSocket.cpp

CompileMaster:
@Jens-G windows+msvc has no "snprintf" funtion, but has "_snprintf" funtion. so should use "_snprintf"  function replace the "snprintf" function.

CompileMaster:
Second commit
change "#if _MSC_VER" to "#if defined(_MSC_VER) && (_MSC_VER < 1900)" , compiliable with vs2015

lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -16,7 +16,9 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+#define snprintf _snprintf
```
jeking3: This is already done in PlatformSocket.h - you can replace snprintf in this file with THRIFT_SNPRINTF and remove this code block instead.

CompileMaster: Thank you first, vs2013 has no  snprintf function,but vs2015 has  snprintf function, i see the code of THRIFT_SNPRINTF in PlatformSocket.h,  use the  THRIFT_SNPRINTF replace  snprintf in TServerSocket.cpp will occurr link errors when compiling under vs2015. use my patch code will compile sucessfully under vs2013 and vs2015 at the same time.

jeking3: Then I would suggest updating the code that makes THRIFT_SNPRINTF in PlatformSocket.h to be correct for VS2013 and VS2015, and then changing the code that uses snprintf directly to use THRIFT_SNPRINTF.  We should keep this type of platform abstraction in one place, in PlatformSocket.h.

jfarrell: One of our goals early on was to maintain latest version and not try and support everything available as it would become too difficult to maintain. What is the minimum set here that is in LTS, I would say we try and cover latest and possible one or two back (if that)

CompileMaster:
@jeking3 I'm glad you help me to submit a PR ,please do it, thanks a lot.

jeking3:
Looks like I introduced the issue:

https://github.com/apache/thrift/commit/e5176241c325837967a07d6d20e0e9a789fbd107

Fortunately it isn't in a release yet.

jeking3:
@CompileMaster see https://github.com/apache/thrift/pull/1099

PR #1099, opened 2016-09-28
THRIFT-3936: fix compile error on VS2013 and earlier

jeking3:
Submitting to replace https://github.com/apache/thrift/pull/1096.

### DIFF
--- a/compiler/cpp/src/thrift/windows/config.h
+++ b/compiler/cpp/src/thrift/windows/config.h
@@ -42,4 +42,25 @@
 // squelch bool conversion performance warning
 #pragma warning(disable : 4800)
 
+// MSVC10 (2010) or later has stdint.h
+#if _MSC_VER >= 1600
+#define HAVE_STDINT_H 1
+#endif
+
+// Must be using VS2010 or later, or boost, so that C99 types are defined in the global namespace
+#ifdef HAVE_STDINT_H
+#include <stdint.h>
+#else
+#include <boost/cstdint.hpp>
+
+typedef boost::int64_t int64_t;
+typedef boost::uint64_t uint64_t;
+typedef boost::int32_t int32_t;
+typedef boost::uint32_t uint32_t;
+typedef boost::int16_t int16_t;
+typedef boost::uint16_t uint16_t;
+typedef boost::int8_t int8_t;
+typedef boost::uint8_t uint8_t;
+#endif
+
 #endif // _THRIFT_WINDOWS_CONFIG_H_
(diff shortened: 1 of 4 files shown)


jeking3:
@CompilerMaster this contains the fix I described.

lib/cpp/src/thrift/TApplicationException.h
```diff
@@ -21,6 +21,7 @@
 #define _THRIFT_TAPPLICATIONEXCEPTION_H_ 1
 
 #include <thrift/Thrift.h>
+#include <stdint.h>
```
nsuke: We have `HAVE_STDINT_H` for this.
For older VC++, uint32_t is still available via windows/config.h, switching between boost and std, which is included via Thrift.h.
It's problematic for any non-windows without stdint.h, though.

jeking3: Okay then what happens with the "uint32_t" lines at the bottom of this header if someone does not have STDINT_H?

nsuke: As I said, for windows it's OK, for non-windows without stdint.h, it will fail.
We could improve config.hin and build/cmake/config.h.in to do the same thing as Windows when boost is avaiable.

jeking3: I see this work was done in windows/config.h so we just need to have TApplicationException.h include the correct header to pull that in on Windows.

jeking3: This is a little more complicated than I anticipated.  It looks like when compiling plugin_types.cpp we end up including TApplicationException through some code path that either undefines HAVE_STDINT_H (which is defined in windows/config.h, from thrift/Thrift.h) or something else is happening.  Still working on it.

nsuke: Oh... it sounds pretty much caused by THRIFT-2835... I'll look into it within a few days or so too.

jeking3: Still perplexing me... it looks like _STDINT is already defined by the time we get there but only stdint.h seems to define that... so I'm still looking...

jeking3: FInally... the compiler has a windows/config.h that's different than the library one...

jeking3: Figured it out; wasn't able to build with VS2010 due to changes made in THRIFT-2835.  I made the necessary changes and ensured I could build on windows and on linux (to a point... I am still running into THRIFT-3937 on my linux system).

PR #1101, opened 2016-09-29
THRIFT-3937: if the compiler is too old (gcc before 4.8) disable new compiler plugin during configuration

jeking3:
This will disable building the new compiler plug-in code if the compiler is too old (gcc earlier than 4.8).


jfarrell:
License on this autoconf macro is GPLWithACException, we can not include it with Apache Thrift. see http://www.apache.org/legal/resolved.html#category-x

jeking3:
I'll back out the autoconf changes then and we'll leave the CMake ones... we should probably document on the thrift web site that although gcc-4.2 or later is required, gcc-4.8 or later is required for compiler plug-in support.

jeking3:
I added a couple changes to properly report in the options summary why plugin was disabled, either because of an old compiler or because WITH_PLUGIN was set to OFF.

build/cmake/DefineOptions.cmake
```diff
@@ -164,31 +164,36 @@ message(STATUS "Thrift package version:                       ${PACKAGE_VERSION}
 message(STATUS "Build configuration Summary")
 message(STATUS "  Build Thrift compiler:                      ${BUILD_COMPILER}")
 message(STATUS "  Build compiler plugin support:              ${WITH_PLUGIN}")
+if (WITH_PLUGIN_COMPILER_TOO_OLD)
+  MESSAGE_DEP(WITH_PLUGIN "Disabled due to older compiler")
+else()
+  MESSAGE_DEP(WITH_PLUGIN "Disabled by WITH_PLUGIN=OFF")
```
jeking3: I found that in the CMake environment, WITH_PLUGIN is honored properly.  It's in the autoconf environment where it seems to be messed up.

build/cmake/DefineOptions.cmake
```diff
@@ -164,31 +164,36 @@ message(STATUS "Thrift package version:                       ${PACKAGE_VERSION}
 message(STATUS "Build configuration Summary")
 message(STATUS "  Build Thrift compiler:                      ${BUILD_COMPILER}")
 message(STATUS "  Build compiler plugin support:              ${WITH_PLUGIN}")
+if (WITH_PLUGIN_COMPILER_TOO_OLD)
+  MESSAGE_DEP(WITH_PLUGIN "Disabled due to older compiler")
```
jeking3: If I do this and leave the MESSAGE_DEP(WITH_PLUGIN below, you get two messages.  It should be fine the way it is?

build/cmake/DefineOptions.cmake
nsuke: I mean, for example, `WITH_CPP` is a flag that users can directly set ON/OFF while `BUILD_CPP` is an internal flag that is based on `WITH_CPP` but can be disabled by `Boost-FOUND=OFF`.

The problem for `WITH_PLUGIN` is that, we're directly modifying this flag without internal equivalent, based on `BUILD_CPP` and `BUILD_COMPILER` values.
So we cannot tell what ON/OFF value the user has put to it.

About autotools flag, try `--enable-plugin=no` instead.
Internally inside configure.ac, with/without flags seemed to be tied to libraries so I used `enable` flag instead.

build/cmake/DefineOptions.cmake
nsuke: Yes, the current policy seems to be to provide all causes which disabled a feature at once.

jeking3:
@nsuke I simplified this one as you suggested, please review.

PR #1106, opened 2016-10-02
THRIFT-3868 Java struct equals should do identity check before field comparison

dustinpho:
https://issues.apache.org/jira/browse/THRIFT-3868


jeking3:
Looks good to me.  :+1:

PR #1107, opened 2016-10-04
THRIFT-3941 WinXP version of thrift_poll() relies on undefined behavior by passing a destructed variable to select()

tpcwang:


jeking3:
Looks good to me.  :+1:

PR #1108, opened 2016-10-05
THRIFT-3942 Make TSSLSocket honor send and receive timeouts

tpcwang:
The test no longer uses signals because we can just wait for read/peek to time out, which is nice because it simplifies the test and gets rid of the error messages.

### DIFF
--- a/lib/cpp/test/TSSLSocketInterruptTest.cpp
+++ b/lib/cpp/test/TSSLSocketInterruptTest.cpp
@@ -59,8 +59,8 @@ struct GlobalFixtureSSL
 
 #ifdef __linux__
       // OpenSSL calls send() without MSG_NOSIGPIPE so writing to a socket that has
-		// disconnected can cause a SIGPIPE signal...
-		signal(SIGPIPE, SIG_IGN);
+      // disconnected can cause a SIGPIPE signal...
+      signal(SIGPIPE, SIG_IGN);
 #endif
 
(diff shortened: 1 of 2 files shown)


lib/cpp/test/TSSLSocketInterruptTest.cpp
```diff
@@ -57,12 +54,6 @@ struct GlobalFixtureSSL
         BOOST_TEST_MESSAGE(boost::format("argv[%1%] = \"%2%\"") % i % master_test_suite().argv[i]);
       }
 
-#ifdef __linux__
-      // OpenSSL calls send() without MSG_NOSIGPIPE so writing to a socket that has
-		// disconnected can cause a SIGPIPE signal...
-		signal(SIGPIPE, SIG_IGN);
```
tpcwang: On second thought, I think I'll just leave this in, in case we need to force-close for other tests in the future.

jeking3: Agree, I would leave that in otherwise applications built on thrift with SSL that do not do this will exit (without a core if I remember correctly, which was really hard to figure out).

tpcwang: This is part of the test code for setup/teardown, so it should not affect applications, but I've reverted my change here.

jeking3:
Looks good to me. :+1:

PR #1111, opened 2016-10-06
THRIFT-3546: Remove global namespace objects from nodejs generated code

bgould:
Modified the node.js code generator to not generate namespace object by default, as using require() is the idiomatic way to reference classes contained in other modules for node, instead of relying on a object heirarchy existing in the global namespace.

This change makes the generated code work under Javascript's "use strict" mode.

In case others are relying on the existing behavior, I added a ":with_ns" flag to the compiler that enables the current behavior for backwards compatibility.


jfarrell:
at quick run through patch looks ok to me, @Jens-G @RandyAbernethy can either of you please review also

PR #1114, opened 2016-10-14
Added Nonblocking SSL java client

dthaluru:


nsuke:
@dthaluru I think we can avoid duplicating the entire client stack.
As far as I can see the only thing that really matters is `TNonblockingSSLSocket`.
If we make it compatible with `TNonblockingTransport`, almost anything else would not be needed any longer.

The most significant incompatibility is `int read(ByteBuffer)` v.s. `ByteBuffer read(int)`.
Would it be possible for `TNonblockingSSLSocket` to accept and reuse `ByteBuffer` so that the method signature can be the same ?

dthaluru:
Thanks for your replies. Sorry, I  will send update pull request soon. I can close pull request for now and will send again later.

PR #1116, opened 2016-10-21
java: cause of TTransportException added to TNonblockingServerSocker constructor

Eternity-Yarr:


lib/java/src/org/apache/thrift/transport/TNonblockingServerSocket.java
```diff
@@ -93,7 +93,7 @@ public TNonblockingServerSocket(NonblockingAbstractServerSocketArgs args) throws
       serverSocket_.bind(args.bindAddr, args.backlog);
     } catch (IOException ioe) {
       serverSocket_ = null;
-      throw new TTransportException("Could not create ServerSocket on address " + args.bindAddr.toString() + ".");
+      throw new TTransportException("Could not create ServerSocket on address " + args.bindAddr.toString() + ".", ioe);
```
jeking3: I like the use of inner exception handling here (that's what C# calls it, in Java it's a Throwable Cause).

jeking3:
I added THRIFT-4040 for this issue.
For future changes please see the Contributing to Thrift guide at:

PR #1117, opened 2016-10-25
Fixed TThreadedServer Memory Leak

tzongw:
There is memory leak in `TThreadedServer` because of reference loop between `TConnectedClientRunner` and `Thread`. A `Runnable` should hold a weak reference to `Thread`. PR should fix this issue.


tzongw:
Added a ticket
https://issues.apache.org/jira/browse/THRIFT-3955

jeking3:
I resolved the conflicts and committed it.

PR #1119, opened 2016-10-29
Java generator: Added "java.lang." to all generated instances of type…

ne0h:
… java.lang.Object to enable the usage of "Object" as struct name in thrift.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -893,7 +893,7 @@ void t_java_generator::generate_union_constructor(ofstream& out, t_struct* tstru
   indent_down();
   indent(out) << "}" << endl << endl;
 
-  indent(out) << "public " << type_name(tstruct) << "(_Fields setField, Object value) {" << endl;
+  indent(out) << "public " << type_name(tstruct) << "(_Fields setField, java.lang.Object value) {" << endl;
   indent(out) << "  super(setField, value);" << endl;
   indent(out) << "}" << endl << endl;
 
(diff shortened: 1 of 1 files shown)


Jens-G:
Hi @ne0h,

could you just [create a JIRA ticket for this](http://thrift.apache.org/docs/HowToContribute)?

@Java devs: Comments anyone?

ne0h:
Hi Jens-G,

of course: https://issues.apache.org/jira/browse/THRIFT-3954

bgould:
@Jens-G 

IMO this is a valid use case but it is not necessarily a great idea because in general its not very user friendly to design an API that clashes names in java.lang.* ... it is even more complicated in Thrift because other languages may also have constructs called "Object" (javascript for example).  There is not a Java-specific technical reason why this shouldn't be allowed however.

That said, there is a related JIRA, not exactly the same problem because java.lang classes do not require import statements, but the idea is the same:

https://issues.apache.org/jira/browse/THRIFT-3301

jeking3:
@bgould Generated code should always be explicitly fully pathed or namespaced and never make any assumptions.  Thrift already uses namespaces so I would have no problem with org.king.Object as a thrift structure so long as it derives from java.lang.Object.  If the code generator uses "Object" without qualifications that could be ambiguous, and that is a valid defect .. I like the idea of eliminating that.  With regards to using a structure named Object, well that's really up to the folks using thrift, it isn't up to the thrift project to tell them how to name their structures and we should stay out of their way where we can.

bgould:
@jeking3 

I agree completely.  Didn't mean to give the impression that it shouldn't be supported.

PR #1120, opened 2016-11-01
THRIFT-3301: Naming collisions possible in generated Java code

bgould:
Switched to fully-qualified names for all class name references in generated code.  Added test IDL to verify generated code compiles successfully.

### DIFF
--- a/lib/java/test/org/apache/thrift/test/TestClient.java
+++ b/lib/java/test/org/apache/thrift/test/TestClient.java
@@ -19,33 +19,38 @@
 
 package org.apache.thrift.test;
 
-// Generated code
-import thrift.test.*;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import org.apache.thrift.TApplicationException;
 import org.apache.thrift.TException;
 import org.apache.thrift.TSerializer;
-import org.apache.thrift.transport.TTransport;
-import org.apache.thrift.transport.TSocket;
-import org.apache.thrift.transport.THttpClient;
-import org.apache.thrift.transport.TFramedTransport;
-import org.apache.thrift.transport.TFastFramedTransport;
-import org.apache.thrift.transport.TTransportException;
-import org.apache.thrift.transport.TSSLTransportFactory;
-import org.apache.thrift.transport.TSSLTransportFactory.TSSLTransportParameters;
 import org.apache.thrift.protocol.TBinaryProtocol;
-import org.apache.thrift.protocol.TProtocol;
-import org.apache.thrift.protocol.TJSONProtocol;
 import org.apache.thrift.protocol.TCompactProtocol;
+import org.apache.thrift.protocol.TJSONProtocol;
+import org.apache.thrift.protocol.TProtocol;
 import org.apache.thrift.protocol.TSimpleJSONProtocol;
+import org.apache.thrift.transport.TFastFramedTransport;
+import org.apache.thrift.transport.TFramedTransport;
+import org.apache.thrift.transport.THttpClient;
+import org.apache.thrift.transport.TSSLTransportFactory;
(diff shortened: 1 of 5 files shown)


jeking3:
> The following tests FAILED:
> 	 40 - JavaTest (Failed)
> Errors while running CTest

nsuke:
@bgould the failing CI job is make dist tarball verification.
It's typically caused by missing `EXTRA_DIST` entries for newly added files in `Makefile.am`.

bgould:
Thanks @nsuke I'm pretty sure that was the problem.  I created a new IDL file for validating this change, and I've now added it to tests/Makefile.am

jeking3:
Does the cmake install job need to be updated as well?

jeking3:
I'm not seeing why JavaTypes was introduced - was it simply used to manually check the build for output of every native type?  Is the result from JavaTypes compilation used in some unit test that would fail if it didn't produce the correct output?  If not, does it belong in the distribution?

bgould:
@jeking3 

Yes exactly.  I added it to the build process that compiles the java unit tests... so if some change to compiler introduces a regression, the IDL in that file should fail to compile.  I didn't add a unit test for this as I felt that the compilation itself was the test.

Consider the following IDL:

    struct String {
        1: string val
    }

    struct Object {
        1: map<string, String> somemap
    }

It would be possible to make a modification to the compiler that erroneously generates the map type as `java.util.Map<String, String>` instead of `java.util.Map<java.lang.String, String>` or to add some reference in the generated code to `java.lang.Object` that is not fully-qualified... however with the generated classes above in the `thrift.test` package, `javac` should flag the unqualified reference(s) as ambiguous.

bgould:
Also one thing to note is that the Java code generated from following valid IDL will still fail to compile because the names match Java primitive types... I know of no way to avoid this problem:

```thrift
struct int {
  1: i32 val
}

struct boolean {
  1: bool val
}

struct long {
  1: i64 val
}

struct short {
  1: i16 short
}

struct char {
  1: i16 val
}

struct Primitives {
  1: i32 int,
  2: i64 long,
  3: i16 short,
  4: bool boolean,
  5: i16 char
}
```

jeking3:
I think a number of those things are in the "reserved" keyword list for the thrift compiler, so that's okay.  Nice, I'll mark this one approved.

lib/java/test/org/apache/thrift/test/TestClient.java
```diff
@@ -576,7 +581,7 @@ public static void main(String [] args) {
           goodbye.string_thing = "Goodbye4";
           goodbye.byte_thing = (byte)4;
           goodbye.i32_thing = 4;
-          goodbye.i64_thing = (long)4;
+          goodbye.i64_thing = 4;
```
Jens-G: Probably not really an issue, but that change seems unrelated?

test/JavaTypes.thrift
```diff
@@ -0,0 +1,94 @@
...
+struct Object {
+  1: Integer integer,
+  2: String str,
+  3: Boolean boolean_field,
+  4: Double dbl,
+  5: Byte bite,
+  6: map<i32, Integer> intmap,
+  7: Map somemap,
+}
+
```
Jens-G: Do we need the same for exception?

lib/java/test/org/apache/thrift/test/TestClient.java
bgould: Yes that is unrelated... I think my IDE settings are such that it was removed automatically because it was an unncessary... no problem for me to restore it if you prefer.

test/JavaTypes.thrift
bgould: that is probably a good idea, will get that added.

bgould: As it happens, adding the Exception type did expose a naming collision... I will track it down and push an update when fixed.

bgould:
@jeking3 @Jens-G 
Looks my last commit caused travis-ci to fail... from what I can tell it doesn't look like it was related to my changes tho (it in the test suite during Haskell generation it looks like)

PR #1125, opened 2016-11-08
[THRIFT-3961] TConnectedClient terminates connection when the message…

cmhe:
… could not be processed

Signed-off-by: Claudius Heine <ch@denx.de>


lib/cpp/src/thrift/server/TConnectedClient.cpp
```diff
@@ -86,7 +86,8 @@ void TConnectedClient::run() {
     } catch (const TException& tex) {
```
jeking3: Just wondering, is catching TException enough here or should we catch std::exception here instead?

PR #1126, opened 2016-11-10
Ant fixes.

garydgregory:
Fix mkdir calls to use the mkdir Ant task. Otherwise, this does not build on Windows.


bufferoverflow:
+1

Jens-G:
Tracked in THRIFT-3962

PR #1129, opened 2016-11-13
THRIFT-3038 clean up remaining volatile issues

jeking3:
This takes care of items 3 and 6 in the PR description.

### DIFF
--- a/lib/cpp/src/thrift/concurrency/Mutex.cpp
+++ b/lib/cpp/src/thrift/concurrency/Mutex.cpp
@@ -37,10 +37,9 @@ namespace concurrency {
 
 #ifndef THRIFT_NO_CONTENTION_PROFILING
 
-static sig_atomic_t mutexProfilingSampleRate = 0;
+static int32_t mutexProfilingCounter = 0;
+static int32_t mutexProfilingSampleRate = 0;
 static MutexWaitCallback mutexProfilingCallback = 0;
-
-volatile static sig_atomic_t mutexProfilingCounter = 0;
 
 void enableMutexProfiling(int32_t profilingSampleRate, MutexWaitCallback callback) {
   mutexProfilingSampleRate = profilingSampleRate;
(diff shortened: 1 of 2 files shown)


lib/cpp/src/thrift/concurrency/Mutex.h
```diff
@@ -101,25 +101,6 @@ class ReadWriteMutex {
...
-public:
-  NoStarveReadWriteMutex();
-
-  virtual void acquireRead() const;
-  virtual void acquireWrite() const;
-
-private:
-  Mutex mutex_;
-  mutable volatile bool writerWaiting_;
-};
```
nsuke: It's in a "public" header (kind of).
Are you sure you delete this without deprecation warning etc ?

In this file, ReadWriteMutex and RWGuard are not used inside Thrift either.
How about removing or deprecating all the three classes in a separate issue,
since these are somewhat more involved compared to private volatile removal ?

lib/cpp/src/thrift/concurrency/Mutex.cpp
```diff
@@ -37,11 +37,10 @@ namespace concurrency {
 
 #ifndef THRIFT_NO_CONTENTION_PROFILING
 
+static sig_atomic_t mutexProfilingCounter = 0;
 static sig_atomic_t mutexProfilingSampleRate = 0;
```
jeking3: According to the comment in the implementation file, a race on this variable is acceptable, so I will change the type as you suggested and re-submit.

lib/cpp/src/thrift/concurrency/Mutex.h
jeking3: Item #1 in the original pull request indicates that NoStarveReadWriteMutex may not be correct; given it is not used by the project and is not a core Thrift mechanism I see no need to keep it because it increases the maintenance burden of the project.  I will use a separate pull request to remove them instead.

jeking3:
By moving the issue with NoStargeReadWriteMutex to THRIFT-3972 this became much simpler and it ready for approval.

PR #1131, opened 2016-11-17
THRIFT-3886 PHP cross test client returns 0 even when failing

zhaakhi:
PHP tests are mostly not asserting that the returned result is as
expected, nor do they return an error for unknown protocols.

Add more assertions in the style of the existing code, and check that the
tested protocol is supported.
This requires some fixes so tests will actually run for the accelerated
binary protocol.

Fixing the tests uncovers some brokenness in compact protocol and the extension, will submit fixes as separate patches.

### DIFF
--- a/test/php/TestClient.php
+++ b/test/php/TestClient.php
@@ -44,6 +44,7 @@ $loader->register();
 /** Include the Thrift base */
 /** Include the protocols */
 use Thrift\Protocol\TBinaryProtocol;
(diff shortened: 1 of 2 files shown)


zhaakhi:
I think the tests should pass on current master if re-run... It would be nice to have this merged so future pull requests can include tests.

zhaakhi:
I've tried rebasing, but the integration tests seem to fail on an unrelated issue about Haskell.

PR #1136, opened 2016-11-28
fixed  when the connection is closed if continued to read，the CPU reaches 100%

Janson-Leung:
lib/php/lib/Thrift/Transport/TSocket.php  
line 277:
    elseif ($data == '' && feof($this->handle_))      =>    elseif ($data == '')


Jens-G:
-1 Something is terribly wrong here.

PR #1138, opened 2016-12-11
Add union example to ThriftTest

allengeorge:
This simply adds a union definition to ThriftTest.thrift.

### DIFF
--- a/test/ThriftTest.thrift
+++ b/test/ThriftTest.thrift
@@ -116,11 +116,11 @@ struct CrazyNesting {
 }
 
 union SomeUnion {
- 1: map<Numberz, UserId> map_thing,
- 2: string string_thing,
- 3: i32 i32_thing,
- 4: Xtruct3 xtruct_thing,
- 5: Insanity insanity_thing
+  1: map<Numberz, UserId> map_thing,
+  2: string string_thing,
+  3: i32 i32_thing,
+  4: Xtruct3 xtruct_thing,
+  5: Insanity insanity_thing
 }
 
 exception Xception {
--- a/test/ThriftTest.thrift
+++ b/test/ThriftTest.thrift
@@ -115,6 +115,14 @@ struct CrazyNesting {
   4: binary binary_field
 }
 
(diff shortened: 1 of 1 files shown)


allengeorge:
Do I have to create a ticket number for this?

Also, I decided to separate simply *defining* the union from actually using it in a service call.

mrwiggles:
Thanks @Jens-G

allengeorge:
If I understand correctly - this is merged now? If so - thanks @Jens-G, and I'll get the second half of the patch in later this week. Sorry about the slow turnaround time.

Jens-G:
Correct. I merged it, but left the ticket open. The idea is to have it act as an umbrella ticket to cover all languages after some suitable call has been added to the service and a test definition has been formulated.

Since this will break the cross test until all relevant languages are integrated we should either have  a feature branch. I could do that and coordinate the patches via my personal Thrift fork, then finally merge the whole feature branch back into Apache Thrift master. 

@jfarrell Does that sound like a plan to you?

jfarrell:
I just rebased the release candidate against master and fixed the broken haskel tests so if parts of this are committed all ready then they will be apart of the release candidate. I would advocate for committing  as portions become available and not hiding in a branch or personal fork and try and structure the commits as logical groups which do not break anything as they go in

Jens-G:
But adding a service function will break things?

jfarrell:
could use a stub with a successful return and a TODO comment

allengeorge:
Ok. Should I put notes in the ticket as to what would make a good candidate service call?






On Mon, Dec 12, 2016 at 9:07 PM -0500, "jfarrell" <notifications@github.com> wrote:












could use a stub with a successful return and a TODO comment

Jens-G:
The calls are documented in ThiftTest.thrift. I'd suggest to follow the model we have. Maybe a bit more verbose. What should become claer is the expected inputs and the expected returns. Needed in order to have all server/client pairs run successfully during cross tests or the like.

PR #1139, opened 2016-12-13
THRIFT-3999 Fix debian package set up scripts.

matyapiro31:
debian/control: fixing typo and remove innecessary in dependecies.
debina/rules: changing to use cleanup.sh in "clean" target as config.log and config.status remain after make clean.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,19 +1,19 @@
 Source: thrift
 Section: devel
 Priority: extra
-Build-Depends: debhelper (>= 9), build-essential, mono-mcs, python-dev, ant,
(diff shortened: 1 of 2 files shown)


Jens-G:
Anyone to confirm this?

jfarrell:
would leave in build-essential, otherwise loks good and +1, @matyapiro31 mind adding build-essential back in

jfarrell:
Thanks for the patch, commit 0c27352

matyapiro31:
@jfarrell , **build-essential** isn't necessary to write in as build-essential is necessary to build whole debian package.  
And also if build-essential is written in debian/control, lintian tells a warning.
So **Must not write build-essential in debian/control**

The error detail is here: https://lintian.debian.org/tags/build-depends-on-build-essential.html

PR #1141, opened 2016-12-19
support for timeout in node http connection

guptanishank:


Jens-G:
Thanks! Could you please file a JIRA ticket according to our [contrib guidelines](http://thrift.apache.org/docs/HowToContribute)? 

@all: Anyone to comment on this?

lib/nodejs/lib/thrift/http_connection.js
```diff
@@ -214,6 +214,18 @@ HttpConnection.prototype.write = function(data) {
...
+  //support for timeout
+  var timeout = self.nodeOptions.timeout;
+  if(timeout){
+    req.on('socket', function (socket) {
+        socket.setTimeout(timeout);  
+        socket.on('timeout', function() {
+            req.abort();
+        });
+    });
+  }
```
nsuke: Reading the doc, simply putting 'timeout' key value to `nodeOptions` (without any code change) would activate request timeout.
`socket.setTimeout` is for idle connections which is very different thing.
So reusing the former value to the latter does not seem a good idea.
Do you need idle timeout for your use-case at all ?

guptanishank: @nsuke the node version which i am using doesn't have support for timeout option: http://devdocs.io/node~4_lts/http.

nsuke: Makes sense.

For newer versions of Node, it can become potentially problematic to do redundant timeout handling.
Current Node implementation seems to be ending up with setting socket timeout, so it may not be a problem for now.

Do you have any idea how to check whether the option is supported ?

Another minor point: we can reduce one indent level by using req.setTimeout(timeout, func...) ?

guptanishank:
@Jens-G there is already a JIRA ticket related to this issue  THRIFT-2968

lib/nodejs/lib/thrift/http_connection.js
nsuke: Yes it turned out basically the same thing.
The potential problem was that, as long as we're using the name "timeout", Node will put their own timeout handler on newer versions.
It's especially nasty because a future version of Node can change it's internal behavior as to how exactly the timeout is handled, and it can cause subtle conflict with our own handling later.

But thinking about it, if it's what people have been widely doing, it's unlikely Node will start doing things that conflict with it.

guptanishank: @nsuke  One workaround could be to use a different option like "socketTimeout" and mention in documentation that this option is for node versions not having support for timeout.

Also, could you look patch submitted in this JIRA ticket: THRIFT-2968

nsuke: I cannot see the content of the patch as it has been removed.
It's visible to everyone while it's available.

I slightly prefer the different-name idea.
As to the actual name, I would choose "requestTimeout" especially if we're going to invoke `request.setTimeout`.
Seeing the [code](https://github.com/nodejs/node/blob/master/lib/_http_client.js#L501), the timeout callback is only active during each request life-cycle, and removed after completion so that it does not affect HTTP keep-alive behavior.
In other words, direct `socket.setTimeout` would break socket reuse.

guptanishank: Doesn't setTimeout function handle that?
https://github.com/nodejs/node/blob/master/lib/_http_client.js#L625

jeking3: It looks like additional work is needed here; @guptanishank will you be moving this forward?

jeking3:
@bananer this would be a good candidate to pick up and run with.

bananer:
@jeking3 it looks like a simple change now that support for node 4 can be dropped. However, I'm not sure what the expected behaviour in case of a timeout should be. Just from looking at the code it does not seem like error handling is triggered by the call to `req.abort()`…

lib/nodejs/lib/thrift/http_connection.js
ramukg: I am working on a project which is using node 4 version. I need timeout. Are you guys going to merge this PR ? If not what are possible way to handle timeout thing without touching thrift module?

jeking3: Based on the comments from @nsuke and @bananer, it looks like the solution is not quite right - that's why it hasn't been merged.  Someone needs to step up and finish or make a convincing argument that this is correct.

jeking3:
I'm going to close this if no progress is made soon.

lib/nodejs/lib/thrift/http_connection.js
jeking3: @ramukg are you using this patch?  Does it work?  Have you tested the timeout?

jeking3:
Closed due to inactivity.

PR #1147, opened 2017-01-01
THRIFT-2945 Add Rust support

allengeorge:
This is a PR to add Rust support to Thrift. It is based on today's master, and I've verified that Rust server/client successfully communicates with all cross-platform clients and servers (*). I would be happy to accept and incorporate feedback!

Not implemented:

* Struct/union constants: Honestly, this looks like it's "not possible" (tm)
* Multiplexed processor

I will be continuing to add documentation, comments and clean up the code in both the C++ generator as well as the Rust client library.

(*) With exception of:

1. D: no D compiler installed
2. Lua: need to find out how to use `.dylib` when using Lua libs
3. Dart: weird problem with pubspec requiring 1.0.0 instead of 1.0.0-dev


test/ThriftTest.thrift
```diff
@@ -115,14 +115,6 @@ struct CrazyNesting {
   4: binary binary_field
 }
 
-union SomeUnion {
```
allengeorge: I, uhh...am not sure why this got removed. I'll re-add it. Suffice it to say, the code handles unions.

test/csharp/TestServer.cs
```diff
@@ -273,39 +273,24 @@ public long testTypedef(long thing)
...
+            // { 1 => { 2 => argument,
+            //          3 => argument,
+            //        },
+            //   2 => { 6 => <empty Insanity struct>, },
+            // }
             public Dictionary<long, Dictionary<Numberz, Insanity>> testInsanity(Insanity argument)
             {
                 testLogDelegate.Invoke("testInsanity()");
 
-                Xtruct hello = new Xtruct();
```
allengeorge: I had to fix the C# server because it was not handling `testInsanity` properly.

anatol:
* Non-Rust changes (like C# test fix) should be extracted into separate change.
* Could you please run 'rustfmt' tool over rust sources to bring the sources to common codestyle?

allengeorge:
I thought that there was no agreed-upon rust codestyle... At least, I think
rustfmt's defaults are winding their way through an RFC process right now
no?

Terminal Musings: http://www.allengeorge.com/
Raft in Java: https://github.com/allengeorge/libraft/
Twitter: https://twitter.com/allenageorge/

Jens-G:
The usualities regarding tests are as follows:
* the ThriftTest suite tests should be placed under /test/<language>
* any other tests that do not belong to the Test Suite should be placed under /lib/<language>/test

Don't worry, common mistake. A lot of people did it wrong at least once, myself included.

test/ThriftTest.thrift
```diff
@@ -67,7 +67,7 @@ typedef i64 UserId
 struct Bonk
 {
   1: string message,
-  2: i32 type
+  2: i32 bonkType
 }
```
Jens-G: Please revert that.

lib/rs/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "rift"
+description = "Rust Thrift library"
+version = "0.5.1"
+license = "Apache-2.0"
+authors = ["Allen George <allen.george@gmail.com>"]
+readme = "README.md"
+documentation = "https://docs.rs/rift/"
+exclude = ["Makefile*"]
+keywords = ["thrift"]
+
```
Jens-G: That should be adjusted too.

test/csharp/TestServer.cs
allengeorge: Thank you @Jens-G

test/ThriftTest.thrift
allengeorge: @Jens-G will do, but now the autogenerated rust code won't work :(

lib/rs/Cargo.toml
allengeorge: What should change? And what should I change it to?

test/ThriftTest.thrift
anatol: @Jens-G "type" is a Rust keyword https://doc.rust-lang.org/grammar.html#keywords What is the best way to deal in such situation?

Jens-G: Well, it is a keyword in Delphi too. How did we do that? Magic? No.

Jens-G: Well, first, it is a keyword in Delphi too. Next, you cannot safely assume that anyone offering a Thrift API will test all languages implemented today if they work with the curent names you use in your IDL. And you cannot look into the future and foresee what language Apple or Google or whoever else comes around next week which may use "bonkType" as a keyword. So we have to deal with it.

Jens-G: Well, first, it is a keyword in Delphi too. Next, you cannot safely assume that anyone offering a Thrift API will test all languages implemented today if they work with the curent names you use in your IDL. And you cannot look into the future and foresee what language Apple or Google or whoever else comes around next week which may use "bonkType" as a keyword. So we have to deal with it. One way to do it is like the C# generator does it (look for `csharp_keywords`). The only bad thing about it is that we still have no common mechanism that is used by all generators. Today all generators more or less implement it their own way. I still have it on my list and some day I'll start working on it and finally consolidate that stuff.

Jens-G: BTW, I just did a search for *.thrift files already using `type` across the Thrift repo. Besides ThriftTest.thrift itself, there are quite a few more. So if we put "type" into the global list, we will have to fix a whole lot of existing code.

jsgf:
Awesome! The thing that I got stalled on was non-support for other protocols/framing, which meant it wasn't really useful for real use. The other blocker was support for async, which is pretty much a must-have for practical uses, and the Rust async story is in flux right now with Tokio, etc.

> Not implemented:
> - Struct/union constants: Honestly, this looks like it's "not possible" (tm)
> - Multiplexed processor

Have a look at [my repo](https://github.com/jsgf/thrift/tree/rust). I implemented these with `lazy_static`. I didn't find anything that couldn't be implemented; in general Rust and Thrift seem like a good match.

The main complexity I found was that you have to use `BTreeMap`/`BTreeSet` if you want to use maps/sets as keys for other maps/sets, but otherwise `HashMap`/`HashSet` are preferred.

The other tricky part was using `Result` to represent exceptions and handling some edge cases around void functions which are extended to throw extensions.

I'll review in detail.

lib/rs/Cargo.toml
Jens-G: I was more referring to the legalese section. If its ASF code, then it should say ASF.

test/ThriftTest.thrift
allengeorge: @Jens-G I'm sorry if my comment came across as rude. I'll change the code generator to prefix reserved names with an identifier.

allengeorge:
@jsgf I opted not to do anything with Tokio because I'd like to see it settle down and get some solid real-world usage for a while before coding anything up with it. IMO, it would be great to get Rust into the hands of people (like me) who'd be OK with writing Rust clients against non-Rust servers.

lib/rs/Cargo.toml
allengeorge: No problem.

1. Should the authors be "Thrift developers"?
2. Isn't the license OK?

allengeorge: I have changed it to "Apache Thrift Developers <dev@thrift.apache.org>". This is what other author tags read, so I assume that's OK.

Jens-G: Yep, fine. We also have a template on the bottom of [that page over here]((http://thrift.apache.org/docs/committers/HowToVersion)). The `name` should probably not be "Rift" and `documentation` should point to http://thrift.apache.org. The `version` tag needs to be the same as for all other client libraries.

test/ThriftTest.thrift
Jens-G: It didn't.

markerickson-wf:
@allengeorge re: "Dart: weird problem with pubspec requiring 1.0.0 instead of 1.0.0-dev" - I can help look into that if you give some more detail.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -0,0 +1,2846 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <string>
+#include <fstream>
+#include <iostream>
+#include <boost/algorithm/string/replace.hpp>
```
Jens-G: Oops! The Thrift compiler should not rely on boost being installed. We really need to remove that. Sorry.

allengeorge: OK. I'll code up my own string replace.

allengeorge:
What should the version number of the client library be? Looks like they're currently set to 1.0.0?

allengeorge:
@markerickson-wf Thanks Mark. Here's the error I get when I run `make precross`:

```
Resolving dependencies...
The pubspec for thrift 0.10.0 from path has version 1.0.0-dev.
make[4]: *** [pub-get-client] Error 65
```

I see the `pubspec.yaml` in `lib/dart` is set to `1.0.0-dev`, but what's weird (to me) is that the `pubspec.yaml` in the `test/dart` directory only has a path dependency, not a version dependency. There must be a lock file with the specific version somewhere that I missed.

allengeorge:
@jsgf For structs it's more involved than it appears - perhaps because of the way I've chosen to represent them. Here are some factors that cause issues:

1. I represent `required` fields as bare types (i.e. no `Option` wrapper). This means that if the user doesn't specify them in the IDL const declaration then `rustc` will fail.
2. I'm not sure how to handle a case where a user specifies a single struct field in the IDL const declaration but ignores the others. I *could* conceivably use ".." but (see below)
3. I currently don't implement a `Defaults` trait for structs. Even if I did generate one, I'd only do it for structs with all optional or opt-in-req-out fields

markerickson-wf:
@allengeorge I just did `make clean` followed by `make precross` and I don't see any warnings.

The `pubspec.lock` files are in `.gitignore`, so you can safely delete those and try again.

`make clean` should remove all of the artifacts leftover from `pub get`, but if you want to make sure, you could do this from the root of the thrift repo:
```
find . -type f -name "pubspec.lock" | xargs rm
find . -type d -name ".pub" | xargs rm -r
find . -type d -name "packages" | xargs rm -r
find . -type f -name ".packages" | xargs rm
```

As far as the `version` in `pubspec.yaml`, that should be updated on the next release.  It looks like it's on the checklist here - https://thrift.apache.org/docs/committers/HowToVersion

allengeorge:
Thanks @markerickson-wf - I also saw the commit improving "make clean".

allengeorge:
With my latest commits (fix type reference that broke on another version of clang) and removing boost, it appears like the code compiles on all platforms. It's not clear to me why the debian builds are breaking though. Did I forget something?

allengeorge:
Also, I think the Appveyor build breakage is unrelated?

allengeorge:
I'm continuing to add tests and update documentation, and I think I've addressed all the initial comments on this PR. Any further thoughts?

allengeorge:
Oh. Interesting. I forgot that Debian doesn't have a Rust package (yet) though it's being worked on IIRC.

anatol:
@allengeorge thanks for this great work!! A few comment I had earlier that looks unresolved

- Name of the official crate should be `thrift`, not `rift`.
- Fields with reserved names should be renamed to `type_` rather than to `res_type`.


Also please squash all commits in `thrift-2945-pr` and rebase if needed.

allengeorge:
OK. Will do.

@anatol Do you know if it's possible for someone on the thrift team to "reserve" the thrift crate on crates.io? I originally named the library `rift` because I didn't want to claim the name until it was decided that this code was reasonable.

allengeorge:
@anatol I've made the changes. I'll try and rebase (it's a little tough because I merged from master in the middle of my work to pick up some build fixes)

anatol:
@allengeorge as of your question about `thrift` crate. Someone from Thrift team should own access to it and deploy a new library version every time Thrift makes a release.

For example [thrift rubygem](https://rubygems.org/gems/thrift/) is owned by @jfarrell. @jfarrell maybe you have a suggestion who from Thrift team needs to make Rust library releases? Or maybe [crate.io](crate.io) can be co-owned with @allengeorge who will deploy new Rust library on behalf of Thrift team?

allengeorge:
@anatol Ran `rustfmt` with default settings. Humorously, it has a bug where it transforms `::{SomeTypeName}` to `::::{SomeTypeName}`.

Jens-G:
> Oh. Interesting. I forgot that Debian doesn't have a Rust package (yet) though it's being worked on IIRC.

There is no (suitable) Rust package for Suse either, only an older version. I finally managed [installing through the Rust web site](https://www.rust-lang.org/en-US/install.html), works like a charm & build went through.

allengeorge:
OK. I believe I've addressed all outstanding comments.

@anatol:

* Reserved type names now have a `_` suffix. I've confirmed that this does not trip warnings for either snake-cased or camel-cased identifiers
* Ran `rustfmt` with default settings on the code
* Rebased onto master and squashed all the changes into a single commit

@Jens-G:

* Should the version of the client library be 1.0.0?

allengeorge:
As for crate ownership, I'm happy to publish (and thereby reserve `thrift`), and then transfer ownership to whoever will be in charge of pushing published crates.

The relevant documentation is here: http://doc.crates.io/crates-io.html#cargo-owner

Note: I've not done anything yet.

Jens-G:
Master is (currently) always at 1.0.0, since we haven't released a 1.0 yet.

If you tell me all the relevant files, I can add it to 
https://thrift.apache.org/docs/committers/HowToVersion

allengeorge:
The only file that has to be changed is `lib/rs/Cargo.toml`. It's a simple INI-style file, and it has a `version` field in there. Note that Rust does not allow you to overwrite crates, so if someone accidentally releases a 1.0.0 onto https://crates.io the version is going to have to be bumped upwards.

allengeorge:
@Jens-G Do you know why the Debian steps in Travis are failing with:

```
The command "docker run --net=host -e BUILD_LIBS="$BUILD_LIBS" $BUILD_ENV -v $(pwd):/thrift/src -it thrift-build:$DISTRO build/docker/scripts/$SCRIPT $BUILD_ARG" exited with 2.```

**Comment by allengeorge** [comment]
I've now updated the PR with `TMultiplexedProcessor`, which was the last functional piece missing. I'm now fleshing out the remaining tests as well as documentation and examples.

**Comment by allengeorge** [comment]
@Jens-G Could you add `lib/rs/README.md` to the list of files for which a version has to be updated? I will also look and see if I can automate this somehow.

**Comment by allengeorge** [comment]
Hi guys - any comments on this PR? I think I've addressed all review comments.

File: .gitignore [hunk 6]
```diff
@@ -206,6 +206,7 @@ project.lock.json
 /lib/delphi/test/typeregistry/*.local
 /lib/delphi/test/typeregistry/*.dcu
 /lib/erl/.generated
+/lib/erl/.rebar
```
anatol: it seems unrelated to the rust library

.gitignore
```diff
@@ -281,6 +293,7 @@ project.lock.json
 /test/log/
 /test/test.log
 /test/erl/.generated
+/test/erl/.rebar
```
anatol: ditto

anatol:
This patch looks really good. Thank you @allengeorge for your work! Thrift team, is there anything left that prevents rust library from merging?

Is there a such status as 'thrift library maintainer'. It would be great if @allengeorge keep maintaining rust library and `thrift` cargo.

allengeorge:
Sorry @anatol - I thought I'd changed the package name everywhere! I must have missed that. Let me fix it ASAP. Sorry!






On Sat, Jan 14, 2017 at 3:20 PM -0500, "Anatol Pomozov" <notifications@github.com> wrote:












This patch looks really good. Thank you @allengeorge for your work! Thrift team, is there anything left that prevents rust library from merging?


Is there a such status as 'thrift library maintainer'. It would be great if @allengeorge keep maintaining rust library and thrift cargo.

allengeorge:
[puzzled] I removed the reserved prefix and added an underscore suffix as you requested. Perhaps I didn't commit that. Let me check. Sorry about that!
And, BTW - I would be very happy to maintain the Rust thrift library and cargo. I have additional changes I'd like to make, including a Tokio implementation.






On Sat, Jan 14, 2017 at 3:24 PM -0500, "Allen George" <allen.george@gmail.com> wrote:










Sorry @anatol - I thought I'd changed the package name everywhere! I must have missed that. Let me fix it ASAP. Sorry!






On Sat, Jan 14, 2017 at 3:20 PM -0500, "Anatol Pomozov" <notifications@github.com> wrote:












This patch looks really good. Thank you @allengeorge for your work! Thrift team, is there anything left that prevents rust library from merging?


Is there a such status as 'thrift library maintainer'. It would be great if @allengeorge keep maintaining rust library and thrift cargo.

lib/rs/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "rift"
```
allengeorge: @anatol I've previously changed this - perhaps you accidentally looked at an older version of the PR?

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -2722,7 +2756,30 @@ string t_rs_generator::rust_namespace(t_service* tservice) {
...
+    return "Res" + base_struct_name;
+  } else {
+    return base_struct_name;
+  }
+}
+
+string t_rs_generator::rust_field_name(t_field* tfield) {
+  string base_field_name(rust_snake_case(tfield->get_name()));
+  if (is_reserved(base_field_name)) {
+    return "res_" + base_field_name;
```
allengeorge: I agree. I've changed this as well - perhaps you were looking at an older version of the PR?

.gitignore
allengeorge: I'll remove this.

.gitignore
allengeorge: Will do so.

allengeorge:
@anatol My apologies for the earlier spam. FWIW, the latest version of the PR has the following:

- The "thrift" crate name (I do not use "rift" anywhere)
- I use the "_" suffix instead of the prefixes I used earlier

I am just about to submit another change that removes the ".erl" gitignores.

I would be very happy to be the thrift library maintainer for rust. There are additional server types I want to add, including one for Tokio when it stabilizes.

anatol:
@allengeorge Hi Allen, I was intended to send only `.gitignore` comments related to `erl`. The other two comments were create a long time ago and for some reason github decided to publish it this time. But you've already resolved it. Sorry for the confusion. And thanks one more time for your work.

allengeorge:
@anatol No worries! I was glad to do it :)

Jens-G:
@all: Everyone happy with it?

allengeorge:
@Jens-G Sorry to bother you, but all the Appveyor builds seem stalled. It appears to be holding up the Travis fix #1158 that was approved yesterday as well :/

allengeorge:
I've now added the tutorial.

allengeorge:
Any thoughts on this?

allengeorge:
Updated:

- Auto-generate constructors for all public structs
- Auto-generate `Default` impl for structs with all-optional fields
- Add constructors for protocol/transport factories
- Use constructors/defaults where appropriate in test/tutorial, etc.
- Fix `make clean` bugs in Rust makefiles
- Clean up some generator code

lib/rs/src/errors.rs
```diff
@@ -0,0 +1,678 @@
...
+    /// remote Thrift endpoint into its corresponding rift `ApplicationError`
+    /// representation.
+    ///
+    /// Application code should never call this method directly.
+    pub fn read_application_error_from_in_protocol(i: &mut TInputProtocol)
+                                                   -> ::Result<ApplicationError> {
+        let mut message = "general remote error".to_owned();
+        let mut kind = ApplicationErrorKind::Unknown;
+
+        try!(i.read_struct_begin());
```
anatol: It should probably use shorter and nicer `?` operator instead of `try!` macro. Here and everywhere else. I counted ~50 usages of `try!` macro.

allengeorge: That was a conscious choice. Personally, I find the try! Macro much easier to read in my IDE

anatol: The [doc](https://doc.rust-lang.org/std/macro.try.html) says

```
Prefer using ? syntax to try!. ? is built in to the language and is more succinct than try!. It is the standard method for error propagation.
```

So `?` is more Rust-idiomatic and future-proof way to check the result.

allengeorge: Very well. I'll change it. Please note that I started this in an older version of Rust (1.12) that didn't have the ? operator, and by the time 1.13 came out I'd over a thousand lines of code already written.

anatol: Yes I understand that. Rust is very actively developed language.

anatol:
FYI: run the code through `clippy` checker and [here is its output](http://pastebin.com/icvD3yNA)  Some of the warnings (but not all) look interesting.

lib/rs/src/errors.rs
allengeorge: @anatol Yup. That said - I'll fix up the usage of `try!` and replace it with `?`

allengeorge:
@anatol Thanks for the clippy output! I'll look through all of the items and fix them. I'd no idea that `new()` without `Default` was considered a code smell. Interesting.

allengeorge:
@anatol Code has been updated with the following changes:

- All usage of `try!` converted to `?`
- Fix all but two clippy lint warnings (I thought the existing code was clearer)

allengeorge:
More updates:

- Minor cleanups to Rust code generators
- Confirmed that all Rust code does not trigger any clippy lint warnings

allengeorge:
Fixed makefile errors with `make distdir`

allengeorge:
Yay! Build passed :) Any thoughts on whether this is OK to merge?

PR #1150, opened 2017-01-04
THRIFT-4019 Dart Makefiles should also clean pubspec.lock

markerickson-wf:
@ericklaus-wf 

FYI @allengeorge this addresses the omission that I mentioned here - https://github.com/apache/thrift/pull/1147#issuecomment-270401425


ericklaus-wf:
+1

PR #1151, opened 2017-01-05
fix appveyor builds - ant version change

jeking3:
ANT version changed to 1.9.8


jeking3:
This was merged into master as it fixed appveyor.

PR #1152, opened 2017-01-05
THRIFT-3984 PHP7 extension causes segfault 

zhaakhi:
Fix a double-free when serializing string map keys or the name of called methods, and a memory leak when deserializing maps or sets.

### DIFF
--- a/lib/php/src/ext/thrift_protocol/php_thrift_protocol7.cpp
+++ b/lib/php/src/ext/thrift_protocol/php_thrift_protocol7.cpp
@@ -596,6 +596,7 @@ void binary_deserialize(int8_t thrift_typeID, PHPInputTransport& transport, zval
           if (Z_TYPE(key) != IS_STRING) convert_to_string(&key);
           zend_symtable_update(Z_ARR_P(return_value), Z_STR(key), &value);
         }
+        zval_dtor(&key);
       }
       return; // return_value already populated
     }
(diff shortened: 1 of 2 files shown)


Jens-G:
Any PHPers around to comment/confirm this solution? Will be happy to commit it if there are no objections against it.

jeking3:
Not a php expert but this looks okay to me, however I would like to see it pass CI.  How do we trigger a new build on someone else's pull request?  Otherwise, if the original author could rebase on master, ideally squash the commits to one, and push again that would be helpful and would kick off a new build.

wesm:
In investigating long Travis CI queue times, I discovered that the Thrift project has 24 builds in its Travis CI matrix:

https://travis-ci.org/apache/thrift/builds/199234155 

Since the entire Apache Software Foundation only has 30 concurrent build slaves on Travis CI, I've requested that concurrent builds for Thrift be throttled:

https://issues.apache.org/jira/browse/INFRA-13477

zhaakhi:
Rebased, the failed test seems unrelated.

jeking3:
I agree the failed CI build job does not look like it is related to the php change.

zhaakhi:
Ping. It would be nice to get this fix in since the double free is potentially a data corrupting bug.

jeking3:
Thanks for the reminder.  Things with failures often slip (as you can see from the current backlog).   I will merge this.

PR #1153, opened 2017-01-06
fix a bug in windows

GZShi:
The `SOCKET` type is defined as a unsigned number in windows, the `evutil_socket_t` variables are always large than zero, therefore the condition not work well here.

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -1324,7 +1324,7 @@ TNonblockingIOThread::~TNonblockingIOThread() {
     ownEventBase_ = false;
   }
 
-  if (listenSocket_ >= 0) {
+  if (listenSocket_ >= 0 && listenSocket_ != THRIFT_INVALID_SOCKET) {
     if (0 != ::THRIFT_CLOSESOCKET(listenSocket_)) {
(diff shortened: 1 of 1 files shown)


lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1324,7 +1324,7 @@ TNonblockingIOThread::~TNonblockingIOThread() {
     ownEventBase_ = false;
   }
 
-  if (listenSocket_ >= 0) {
+  if (listenSocket_ >= 0 && listenSocket_ != THRIFT_INVALID_SOCKET) {
```
jeking3: Perhaps the only test needed is to see if the socket is not an invalid socket?  Is the >= 0 test necessary at this point?

Jens-G: @GZShi: Any comments from your side?

GZShi: I agree with @jeking3 .
@Jens-G thanks for reply :)

Jens-G: @GZShi : would you mind updating your PR accordingly?

Jens-G:
Tracked in THRIFT-4038
@GZShi: next time please have a look at our [contrib guidelines](http://thrift.apache.org/docs/HowToContribute)

PR #1156, opened 2017-01-15
THRIFT-4011 Use slices for Thrift sets

dcelasun:
As discussed in [THRIFT-4011](https://issues.apache.org/jira/browse/THRIFT-4011), this commit changes the Go generator to use slices, instead of maps for Thrift sets.

I've specifically didn't touch the Go library since there was no agreement on panicking for duplicates. We have three options:

1. Leave it as is and add documentation stating deduplication is the caller's responsibility.
2. Silently deduplicate before serialization.
3. panic on duplicates.

2 and 3 probably requires [`reflect.DeepEqual`](https://golang.org/pkg/reflect/#DeepEqual), which is not ideal.

@Jens-G Thoughts?

### DIFF
--- a/lib/go/test/tests/thrifttest_driver.go
+++ b/lib/go/test/tests/thrifttest_driver.go
@@ -162,7 +162,7 @@ func (p *ThriftTestDriver) Start() {
 		t.Fatal("TestStringMap failed")
 	}
 
-	setTestInput := map[int32]struct{}{1: {}, 2: {}, 3: {}}
+	setTestInput := []int32{1, 2, 3}
 	if r, err := client.TestSet(setTestInput); !reflect.DeepEqual(r, setTestInput) || err != nil {
 		t.Fatal("TestSet failed")
 	}
--- a/lib/go/test/tests/thrifttest_handler.go
+++ b/lib/go/test/tests/thrifttest_handler.go
@@ -96,7 +96,7 @@ func (p *ThriftTestHandler) TestStringMap(thing map[string]string) (r map[string
 	return thing, nil
 }
 
-func (p *ThriftTestHandler) TestSet(thing map[int32]struct{}) (r map[int32]struct{}, err error) {
+func (p *ThriftTestHandler) TestSet(thing []int32) (r []int32, err error) {
 	return thing, nil
 }
 
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -882,6 +882,7 @@ string t_go_generator::go_imports_begin(bool consts) {
   return string(
       "import (\n"
       "\t\"bytes\"\n"
+      "\t\"reflect\"\n"
       + extra +
       "\t\"fmt\"\n"
       "\t\"" + gen_thrift_import_ + "\"\n");
@@ -890,7 +891,7 @@ string t_go_generator::go_imports_begin(bool consts) {
 /**
  * End the import statement, include undscore-assignments
  *
- * These "_ =" prevent the go compiler complaining about used imports.
+ * These "_ =" prevent the go compiler complaining about unused imports.
  * This will have to do in lieu of more intelligent import statement construction
  */
 string t_go_generator::go_imports_end() {
@@ -899,6 +900,7 @@ string t_go_generator::go_imports_end() {
       "// (needed to ensure safety because of naive import list construction.)\n"
       "var _ = thrift.ZERO\n"
       "var _ = fmt.Printf\n"
+      "var _ = reflect.DeepEqual\n"
       "var _ = bytes.Equal\n\n");
 }
 
@@ -1155,12 +1157,12 @@ string t_go_generator::render_const_value(t_type* type, t_const_value* value, co
   } else if (type->is_set()) {
     t_type* etype = ((t_set*)type)->get_elem_type();
     const vector<t_const_value*>& val = value->get_list();
-    out << "map[" << type_to_go_key_type(etype) << "]struct{}{" << endl;
+    out << "[]" << type_to_go_key_type(etype) << "{" << endl;
     indent_up();
     vector<t_const_value*>::const_iterator v_iter;
 
     for (v_iter = val.begin(); v_iter != val.end(); ++v_iter) {
-      out << indent() << render_const_value(etype, *v_iter, name) << ": struct{}{}," << endl;
+      out << indent() << render_const_value(etype, *v_iter, name) << ", ";
     }
 
     indent_down();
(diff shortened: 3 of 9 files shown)


Jens-G:
http://thrift.apache.org/docs/types states that (as one would expect) a Thrift map is defined as "A map of **strictly unique keys** to values. Translates to an STL map, Java HashMap, PHP associative array, Python/Ruby dictionary, etc.". 

Given that, I would say it could be one option to error, when the user inserts a duplicate. On the other hand, in that case the better option could be to simply replace the current value. 

But **what should not happen is that serialized data come in with duplicated keys**. That would be clearly an error as it is a violation of the rule above. In that case it might be absolutely ok to return some kind of error.

Re `panic()`, from my understanding of Go that's not the idiomatic way to go, since everything form tha above is an easily recoverable error.

Further reading:
 * https://blog.golang.org/error-handling-and-go
 * https://golang.org/doc/effective_go.html#panic
 * http://stackoverflow.com/questions/28472922/when-to-use-os-exit-and-panic-in-golang

dcelasun:
Hey @Jens-G, I'm not sure I follow you, what does this have anything to do with maps? Assuming you meant sets, the docs say:

> An unordered set of unique elements. Translates to an STL set, Java HashSet, set in Python, etc. Note: PHP does not support sets, so it is treated similar to a List

Similar to PHP, Go does not have a native type for sets, so the best thing to do is to treat it similar to a list.

> Given that, I would say it could be one option to error, when the user inserts a duplicate.

This is not possible, because the caller doesn't "insert" anything, they simply return a slice. Consider the following:

```thrift
service Foo {
    set<string> bar() throws (1: Something error)
}
```

This generates an interface called `Foo` with the following method:

```go
type Foo interface {
    Bar() ([]string, error)
}
```

So the user simply returns a string slice, the Thrift library has no control over it. Once `Foo` returns, it's too late for Thrift itself to return an error, only panic. Speaking of which, panicking in case of ***programmer error*** is very common and idiomatic in Go. The standard library is full of such panics (e.g search for "misuse" [here](https://golang.org/src/sync/waitgroup.go)). I would consider returning a non-unique slice for a Thrift set a programming error (and hence deserving a panic), but if you disagree, I can update the PR with deduplication in the library.

Jens-G:
Oopsie. After I read the word "map" above I must have been mentally switched to maps somehow.  My mistake.

Nevertheless what I said about panic and errors in general above still holds true. At this time we have exactly 0 (zero) panics in the code under /lib/go/thrift. So I would argue that introducing a panic at least breaks the usual patterns of that library.

>  Speaking of which, panicking in case of programmer error is very common and idiomatic in Go. The standard library is full of such panics 

Put four Go experts together and you get five opinions, all idiomatic. Especially about the high and lofty Go error handling design principles:

> In Go, error handling is important. **The language's design and conventions encourage you to explicitly check for errors** where they occur (as distinct from the convention in other languages of throwing exceptions and sometimes catching them). 

Or mostly assigning them to `_` as many people do in Go. That's for the "encourage you to explicitly check for errors" part. In contrast, an exception must be caught, not only "sometimes", because there is no other error handling option. 

> The decision to not include exceptions in Go is an example of its **simplicity** and orthogonality. Using multiple return values and a simple convention, **Go solves the problem** of letting programmers know when things have gone wrong and **reserves panic for the truly exceptional**.

I can't see any "simplicity" in replacing exceptions by panics, "encouraging" error handling by leading people into misusing the `_` operator and at the same time claiming "*this time, we get it right*". Go's error handling is FUBAR from the beginning to the end. Good intentions, absolutely, but the implementation .... All personal opinion, of course,

dcelasun:
> Good intentions, absolutely, but the implementation ..

FWIW, I agree. However, we still need a solution and as I explained above, we don't have a way of returning an error here. That leaves us with silent deduplication (during serialization) in lib/go/thrift. Are you OK with that?

Jens-G:
+1

Sent from mobile device, please ignore spelling mistakes.
Von: Duru Can Celasun
Gesendet: 16.01.2017 08:00
An: apache/thrift
Cc: Jens Geyer; Mention
Betreff: Re: [apache/thrift] THRIFT-4011 Use slices for Thrift sets (#1156)


Good intentions, absolutely, but the implementation ..

FWIW, I agree. However, we still need a solution and as I explained above, we don't have a way of returning an error here. That leaves us with silent deduplication (during serialization) in lib/go/thrift. Are you OK with that?

-

lib/go/test/tests/thrifttest_handler.go
```diff
@@ -96,7 +96,7 @@ func (p *ThriftTestHandler) TestStringMap(thing map[string]string) (r map[string
 	return thing, nil
 }
 
-func (p *ThriftTestHandler) TestSet(thing map[int32]struct{}) (r map[int32]struct{}, err error) {
+func (p *ThriftTestHandler) TestSet(thing []int32) (r []int32, err error) {
```
jeking3: This looks like a breaking change; any generated go implementation's handler will need to change.  How will this be communicated?  What happens to existing handlers that implement the older method?

dcelasun: Yes, this is a breaking change which was originally discussed in THRIFT-4011; I only started working on this PR after I was given the go ahead.

> How will this be communicated?

I would imagine posts to the mailing lists and an announcement on the website, several weeks in advance of a new release?

> What happens to existing handlers that implement the older method?

Compiling IDLs with this patch will change the signature of any RPC or struct in the generated Go code, so it's very easy to catch at compile time and make the changes.

jeking3: As long as it is a compile time failure if someone doesn't change it, that's good.  In C++ (C++03, or C++11 without the "override" keyword being used) if you change the signature of a virtual method, anything that overrides it will silently define a new method instead of overriding what you want.  It's pretty messy.

Jens-G:
Hi @dcelasun,any news? In caseyou're donw with it, could you rebase the PR to trigger a new Travis build?

dcelasun:
Hey @Jens-G sorry, I didn't have time to finish this up just yet. The lib part of the PR is still missing, but I'll be doing that in a few days. I'll rebase, push and let you know.

Jens-G:
Ok, all fine. Don't panic ;-)

dcelasun:
I finally hade some time to take a look at this. Duplicate detection is now implemented with `reflect.DeepEqual`. It isn't really ideal, but Go lacks a way of defining equality for arbitrary types, so this is the only possible implementation that supports `set<T>` for any `T`.

lib/go/test/tests/thrifttest_driver.go
```diff
@@ -162,7 +162,7 @@ func (p *ThriftTestDriver) Start() {
 		t.Fatal("TestStringMap failed")
 	}
 
-	setTestInput := map[int32]struct{}{1: {}, 2: {}, 3: {}}
+	setTestInput := []int32{1, 2, 3}
```
jeking3: Syntax looks much better than before.

dcelasun: Yeah, it's also finally serializable to JSON for non-primitive types.

dcelasun:
@Jens-G Squashed commits and rebased from master.

Jens-G:
```
src/common/clientserver_test.go:61: cannot use handler (type *MockThriftTest) as type thrifttest.ThriftTest in argument to GetServerParams:
	*MockThriftTest does not implement thrifttest.ThriftTest (wrong type for TestSet method)
		have TestSet(map[int32]struct {}) (map[int32]struct {}, error)
		want TestSet([]int32) ([]int32, error)
src/common/clientserver_test.go:226: cannot use s (type map[int32]struct {}) as type []int32 in argument to client.TestSet
FAIL	common [build failed]
make[2]: *** [check] Error 2
make[2]: Leaving directory `/thrift/src/test/go'
make[1]: *** [check-recursive] Error 1
make[1]: Leaving directory `/thrift/src/test'
make: *** [check-recursive] Error 1

```

dcelasun:
Whoops, I must have missed that one. Should be fixed now.

dcelasun:
@Jens-G All tests green and rebased from master.

Jens-G:
Ok, thanks.

PR #1157, opened 2017-01-16
THRIFT-4033 disable compiler plug-ins for standard debian build

jeking3:
as the thrift-compiler packaging is not ready for it yet

### DIFF
--- a/build/cmake/DefineOptions.cmake
+++ b/build/cmake/DefineOptions.cmake
@@ -84,7 +84,7 @@ if(WITH_CPP)
 endif()
 CMAKE_DEPENDENT_OPTION(BUILD_CPP "Build C++ library" ON
                        "BUILD_LIBRARIES;WITH_CPP;Boost_FOUND" OFF)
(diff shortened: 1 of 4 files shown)


jeking3:
@jfarrell I'd like one review of this before I merge it.  It did fix the issue I had with the stock 0.10.0 debian package for the compiler not working.

jfarrell:
We have had issues with plugin support being on by default in a couple build instances now, and this does not fix when we pull out autoconf that it wont reappear with cmake. I'd like to see us change the defaults here and make --with-plugin the optional to enable and having WITH_PLUGIN=no so it is disabled by default

jeking3:
I'm okay making the default for plug-in support "OFF", and you have to enable it specifically.  Would it be okay to merge this as it fixes the ticket as described, and open a new one to change the global behavior?

jfarrell:
sure, or modify the ticket description to 'disable compiler plug-ins by default for all builds'

PR #1162, opened 2017-01-20
THRIFT-3973: Provide some tools to make it easier to build thrift C++ on Windows

jeking3:
As a contributor and committer on the project with a focus on C++ and CMake, I needed a way to build thrift C++ on windows with all the bells and whistles.  While cmake is a great tool for generating build environments on different platforms, there is still an ecosystem of third party libraries surrounding thrift that need to be acquired or built.

I was not certain if these files should end up in build/wincpp or in contrib/wincpp.  These files are not used in any of our CI builds.  They simply exist as a guide to make it easier to build thrift C++ on windows, including how to build a redistributable thrift compiler for windows without any dependencies, and how to generate Visual Studio projects for thrift on windows.

As part of this effort, the current C++ MSVC projects that are distributed with the project will be removed once folks are happy with this toolset.

The README.md file contains information that can eventually be posted on:
https://thrift.apache.org/docs/install/windows
and the content there will be split into "older build information" and this would be the current.

With these tools can decide what version of Visual Studio and what platform (x86, x64) you want to work with simply by opening a Visual Studio Command Prompt of the appropriate type, and then the scripts will auto-configure to that.  I have Visual Studio 2010, 2012, 2013, and 2015 installed and I have tested all of the combinations.

The CI build is superfluous because these are not used by anything.  They are for interactive use right now.  Eventually we might be able to have the appveyor build leverage them.  I'm not sure if "wincpp" is the right name but I had to start somewhere.

### DIFF
--- a/build/wincpp/build-thrift-compiler.bat
+++ b/build/wincpp/build-thrift-compiler.bat
@@ -1,0 +1,79 @@
+::
+:: Licensed under the Apache License, Version 2.0 (the "License");
+:: you may not use this file except in compliance with the License.
+:: You may obtain a copy of the License at
+:: 
+::     http://www.apache.org/licenses/LICENSE-2.0
+:: 
+:: Unless required by applicable law or agreed to in writing, software
+:: distributed under the License is distributed on an "AS IS" BASIS,
+:: WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+:: See the License for the specific language governing permissions and
+:: limitations under the License.
+:: 
+
+::
+:: Produces a production thrift compiler suitable for redistribution.
+:: The compiler is linked to runtime statically for maximum portability.
+:: Assumes the thirdparty files for "winflexbison" have been placed
+:: according to the README.md instructions.
+::
+:: Open a Visual Studio Command Prompt of your choosing and then
+:: run this script.  
+
+@ECHO OFF
+SETLOCAL EnableDelayedExpansion
+
+IF NOT DEFINED PACKAGE_NAME    SET PACKAGE_NAME=thrift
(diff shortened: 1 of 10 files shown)


build/wincpp/build-thrift-compiler.bat
```diff
@@ -0,0 +1,80 @@
...
+:: Assumes the thirdparty files for "winflexbison" have been placed
+:: according to the README.md instructions.
+::
+:: Open a Visual Studio Command Prompt of your choosing and then
+:: run this script.  
+
+@ECHO OFF
+SETLOCAL EnableDelayedExpansion
+
+IF NOT DEFINED CMAKE_ROOT      SET CMAKE_ROOT=C:\Program Files\CMake
```
jeking3: I need to remove this, since it can confuse CMAKE.  We'll assume that CMAKE is in the path, like the other top level batch does.  With this here I was getting errors with CMAKE-3.7 so I will fix in the next push.

build/wincpp/build-thrift-compiler.bat
```diff
@@ -0,0 +1,80 @@
...
+  ECHO   Build Directory: %BUILDDIR%
+  ECHO Install Directory: %OUTDIR%
+  ECHO  Source Directory: %SOURCEDIR%
+  ECHO =========================================================================
+  ECHO/
+
+  MKDIR "%BUILDDIR%"
+  CD "%BUILDDIR%" || EXIT /B
+
+  "C:\Program Files\CMake\bin\cmake.exe" %~dp0thrift ^
```
jeking3: This should just be "cmake.exe" and let the path pick it up.

jeking3:
Build failure here cannot be related to the pull request because it is not used in any Travis CI job.

PR #1164, opened 2017-01-27
THRIFT-3854 add a way in java to clear TFramedTransport read buffers

jeking3:
I took https://github.com/apache/thrift/pull/1081 and applied changes to address the code comments.


jeking3:
Job 17 continues to fail on builds.  Something to do with haskell.  I am going to ignore it here.

PR #1170, opened 2017-01-29
THRIFT-3993 fix build error with gcc 6.2 and mingw

jeking3:
We don't have a CI build that covers mingw and gcc 6.2 on windows... can anybody else try this to see if it works?


jeking3:
Closing; there are more MinGW issues that need to be handled holistically.

PR #1171, opened 2017-01-29
THRIFT-4046 fix PlatformSocket.h for mingw64/msys2 and provide instructions

jeking3:
Instructions provided for building with cmake under that toolchain.

See build/cmake/README-MSYS2.md

### DIFF
--- a/build/cmake/README-MSYS2.md
+++ b/build/cmake/README-MSYS2.md
@@ -1,0 +1,63 @@
+<!---
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


build/cmake/README-MSYS2.md
```diff
@@ -0,0 +1,63 @@
...
+After that, close your shell and open a new one.
+
+Use cmake to create a MinGW makefile, out of tree (assumes you are in the top level of the thrift source tree):
+
+    mkdir ../thrift-build
+    cd ../thrift-build
+    cmake -G"MinGW Makefiles" -DCMAKE_MAKE_PROGRAM=/mingw64/bin/mingw32-make \
+       -DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc.exe \
+       -DCMAKE_CXX_COMPILER=x86_64-w64-mingw32-g++.exe \
+       -DWITH_BOOSTTHREADS=ON -DWITH_LIBEVENT=OFF \
```
jeking3: Doesn't work properly without requiring BOOST.  I tried it without this directive and it fails in TFileTransport looking for BoostThreadFactory for some reason.  I will dig into it.

jeking3: BOOST is used in the particular case I have because we have a hardcoded windows/config.h we are using instead of relying on the one generated by cmake.  There's already a backlog item for this, so given there's a bugfix here + some instructions I am going to merge this.

PR #1172, opened 2017-01-30
THRIFT-4060 add better support in the cpp generator for custom ostream operators on structures

jeking3:
This pull request adds support for an annotation that will disable the emit code for operator << and printTo, allowing a custom streaming output format to be provided by the consuming application.


compiler/cpp/src/thrift/generate/t_cpp_generator.cc
```diff
@@ -1117,7 +1129,7 @@ void t_cpp_generator::generate_struct_declaration(ofstream& out,
         << " &b);" << endl << endl;
   }
 
-  if (is_user_struct) {
+  if (is_user_struct && !has_custom_ostream(tstruct)) {
```
jeking3: This needs to change; on ostream declaration must be in the header even if thrift does not emit an implementation for it so that a thrift class that emits output code can contain a thrift class that does not.

jeking3:
Going to add the ability to generate with cpp:no_ostream_operators to omit definitions for ostream operators entirely (when set, declarations will still be made, which means consuming implementations must provide all the implementations just like they did before thrift-0.9.3).

jeking3:
So job #2 failed strangely in c_glib which is unrelated to this change.
I pushed an empty change to kick a rebuild.  I expect it to clear...
Here's the job that failed:
https://travis-ci.org/apache/thrift/jobs/197100291

lib/cpp/test/AnnotationTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+  // Now show that if cpp.customostream is added as an annotation,
+  // the behavior can be overridden in a safe and usable way (THRIFT-4060).
+  //
+
+  ostr_custom cus;
+  cus.__set_bar(10);
+
+  std::stringstream csd;
+  csd << cus;
+  BOOST_CHECK_EQUAL(csd.str(), "{ bar = 10; }");
```
hcorg: IMHO its better to keep "single test-case - single test" (or AAA or GivenWhenThen scheme). This file could include couple of SUITES (for example - one per annotation, and then well-named tests). Anytime I see comment I try to find a way to remade code so it does not need that comment ;)

jeking3: I split it out into two tests without comments.

PR #1173, opened 2017-01-31
Remove debug print from TServiceClient

tdavis:
Client: Java


Jens-G:
Tracked in THRIFT-4062
@tdavis: next time please have a look at our [contrib guidelines](http://thrift.apache.org/docs/HowToContribute). Thank you!

PR #1178, opened 2017-02-08
THRIFT-4072 php: TCurlClient - Add the possibility to send custom hea…

swatkatz:
…ders

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -52,18 +52,22 @@ install:
 - move WIN32-Code\event2\* include\event2\
 - move *.h include\
 - cd ..
-- appveyor-retry cinst -y winflexbison
-- appveyor DownloadFile http://www.us.apache.org/dist/ant/binaries/apache-ant-%ANT_VERSION%-bin.zip
-- 7z x apache-ant-%ANT_VERSION%-bin.zip > nul
+- appveyor-retry cinst -y ant
+- appveyor-retry cinst -y winflexbison3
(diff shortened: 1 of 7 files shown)


jeking3:
Would it be possible to add a unit test for this?

swatkatz:
@jeking3 : I can add the tests as you suggest, but since this is a minor feature improvement and there don't seem to be any tests that exist for this class (and I see that other such patches don't have them either e.g. https://github.com/apache/thrift/commit/1f9717d192137d06927846cc2f2f7e380e5da834), I was just wondering what the standard practice around this was?

Jens-G:
@swatkatz: Having a test is better than not having one. We have way too many patches without tests, so if you provide one, we increase quality standards (and you will be our positive role model 👍) How awesome is that?

jeking3:
If the class has no testing at all then I would not require you to add tests in order to get committed, however as Jens suggested it would improve the quality of the project if tests were put in, so I always ask.

It looks like you are making modifications in your master branch and they are getting sync'd into this pull request.  In the future I would recommend that you create a branch off master first, like "git checkout -b THRIFT-4072" and then submit a pull request from that branch instead so it is isolated from the changes in your master.  There should only be one commit for your changes... it looks like the way you have it set up, every time the apache master is revised your pull request will rebuild here, which is a bit heavy.

If the previous build was clean I can merge this.

swatkatz:
thank you @jeking3 and @Jens-G .. I will give a shot at adding tests for this class in a separate PR. Also, thank you for letting me the correct way of setting up the PRs. Really appreciate the guidance.

PR #1180, opened 2017-02-08
THRIFT-4076 fix appveyor ant issue

jeking3:
This is something of an experiment as I am changing the appveyor build to use chocolatey to install ant instead of grabbing it directly from the apache server.  Apache does not provide a stable link to get the latest stable build so every time ant is revised, the build we download and use is deleted and our builds fail.


jeking3:
This was merged; I wanted to close it to cancel any travis job that hasn't run yet.

PR #1181, opened 2017-02-09
THRIFT-4079 add missing use clauses for included thrift sources in generated perl modules

jeking3:
Sadly the state of perl unit testing appears to be pretty sad... I don't see any tests for perl being run with "make check", nor are they run in the perl package maker (for cpan installs).  There are three test files but I'm not sure how they get run, if at all.  So there's no test for this.  I tested it on the example I submitted and it resolves the issue.


jeking3:
Job 3357.7 failure appears environmental: make[3]: write error

PR #1187, opened 2017-02-12
Review for pull up of RUST extended namespace support, addition of macros to allow application to macroize over defintions ...

przygienda:
Please review initial request here ...

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -102,9 +102,6 @@ private:
 
     // provides rust formatted namespace
     string namespace_preamble();
-    // provides rust formatted preamble of the current namespace & file from
-    // which it came
-    string namespace_origin_preamble();
 
   // Write the common compiler attributes and module includes to the top of the auto-generated file.
   void render_attributes_and_includes();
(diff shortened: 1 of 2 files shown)


przygienda:
OK, finished. The patch does 

* improve the namespace support by allowing to nest Thrift models inside a package now
* moves the reading/writing into a proper trait so models can be differentiated and generics written over them
* adds macro definitions to all generated types which allows the package generating the code to e.g. generate code over model definitions, e.g. to derive custom traits via macrosthat use the internal structure of enum/struct/union such as custom defaults, statistics and so on ...

jeking3:
Is there an Apache Jira ticket associated with this pull request?  I know there have been a number of folks working on RUST support.  Given the size of this, it needs reviews and it also needs to get committed so it doesn't gather conflicts over time.

If there is no Jira ticket for it yet, if you could create one that would be great.

jeking3:
@allengeorge would you agree we should close this pull request and tackle each item individually per the open Jira items?

allengeorge:
Yes please - that would be the best course of action.

jeking3:
If @przygienda could close it, that would be great.

przygienda:
Our MailScanner believes that the attachment to this message sent to you

jeking3:
@jfarrell this can be closed.

przygienda:
Thanks for closing, yes, the work moved to main thrift tracker ...

PR #1188, opened 2017-02-13
Check for joinable before attempting join in thread destructors

cgrebeld:
A spurious exception is thrown when shutting down TThreadedServer via drainDeadClients because threads are joined before being destroyed.

### DIFF
--- a/lib/cpp/src/thrift/concurrency/BoostThreadFactory.cpp
+++ b/lib/cpp/src/thrift/concurrency/BoostThreadFactory.cpp
@@ -61,7 +61,7 @@ public:
   }
 
   ~BoostThread() {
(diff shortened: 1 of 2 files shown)


cgrebeld:
Thanks, Sorry for not being clear.  As you said the exception was being thrown and caught, there was no incorrect behavior, I was just annoyed that the normal control flow entered a block marked "// We're really hosed." when I was debugging an unrelated issue.  I'll be sure to follow the protocol next time, thanks for the advice.

jeking3:
boost has a join_if_joinable wrapper for this, but I don't know what version it showed up in.  It is better to avoid an exception when you can so this is good.

PR #1194, opened 2017-02-16
THRIFT-3921 cpp library- optional operator ostream support for enumerations

jeking3:


compiler/cpp/src/thrift/generate/t_cpp_generator.cc
```diff
@@ -566,6 +567,47 @@ void t_cpp_generator::generate_enum(t_enum* tenum) {
...
+
+    f_types_ << indent() << "std::map<int, const char*>::const_iterator it = _"
+             << tenum->get_name() << "_VALUES_TO_NAMES.find(val);" << endl;
+    f_types_ << indent() << "if (it != _" << tenum->get_name() << "_VALUES_TO_NAMES.end()) {" << endl;
+    indent_up();
+    f_types_ << indent() << "out << it->second;" << endl;
+    indent_down();
+    f_types_ << indent() << "} else {" << endl;
+    indent_up();
+    f_types_ << indent() << "out << val;" << endl;
```
viveksjain: I realized later there's a bug in my implementation - this should be `out << static_cast<int>(val);` otherwise the code will infinite loop in this branch.

jeking3: It isn't an infinite loop as you run our of memory to hold the ostream that gets very large, but it is bad nonetheless.  I fixed it and added a test for it. :)

PR #1196, opened 2017-02-18
THRIFT-3891 TNonblockingServer configured with more than one IO threads does not always return from serve() upon stop()

jeking3:
I took the comments from @nsuke and applied them to the PR.  I was able to reasonably eliminate the boolean you mentioned however I could not go back to loopbreak.  I see no reasonable way to synchronize here.  The run() thread will eventually need to release a mutex and THEN call the event loop.  There's no way inside breakLoop() to know if it is actually IN the event loop.  Given this, the exit variant is the only safe method to break the loop.

The obsoletes PR #1080.

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -510,7 +510,7 @@ void TNonblockingServer::TConnection::workSocket() {
 
     // If there is no data to send, then let us move on
     if (writeBufferPos_ == writeBufferSize_) {
-      GlobalOutput("WARNING: Send state with no data to send\n");
+      GlobalOutput("WARNING: Send state with no data to send");
       transition();
       return;
     }
@@ -765,11 +765,9 @@ void TNonblockingServer::TConnection::setFlags(short eventFlags) {
   }
 
   // Delete a previously existing event
-  if (eventFlags_ != 0) {
-    if (event_del(&event_) == -1) {
-      GlobalOutput("TConnection::setFlags event_del");
-      return;
-    }
+  if (eventFlags_ && event_del(&event_) == -1) {
+    GlobalOutput.perror("TConnection::setFlags() event_del", THRIFT_GET_SOCKET_ERROR);
+    return;
   }
 
   // Update in memory structure
@@ -812,7 +810,7 @@ void TNonblockingServer::TConnection::setFlags(short eventFlags) {
 
   // Add the event
   if (event_add(&event_, 0) == -1) {
-    GlobalOutput("TConnection::setFlags(): could not event_add");
+    GlobalOutput.perror("TConnection::setFlags(): could not event_add", THRIFT_GET_SOCKET_ERROR);
   }
 }
 
@@ -820,9 +818,9 @@ void TNonblockingServer::TConnection::setFlags(short eventFlags) {
  * Closes a connection
  */
 void TNonblockingServer::TConnection::close() {
-  // Delete the registered libevent
-  if (event_del(&event_) == -1) {
+  if (eventFlags_ && event_del(&event_) == -1) {
     GlobalOutput.perror("TConnection::close() event_del", THRIFT_GET_SOCKET_ERROR);
+    return;
   }
 
   if (serverEventHandler_) {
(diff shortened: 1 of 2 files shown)


jeking3:
Looks like I ran the build and tests outside the dockerfile so it didn't do any libevent stuff. :|

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -212,7 +212,8 @@ class TNonblockingServer::TConnection {
   TConnection(THRIFT_SOCKET socket,
               TNonblockingIOThread* ioThread,
               const sockaddr* addr,
-              socklen_t addrLen) {
+              socklen_t addrLen)
+  {
```
nsuke: Enter key by accident ?

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1520,13 +1524,9 @@ void TNonblockingIOThread::breakLoop(bool error) {
     ::abort();
   }
 
-  // sets a flag so that the loop exits on the next event
-  event_base_loopbreak(eventBase_);
+  // cause the loop to stop ASAP - even if it has things to do in it
+  event_base_loopexit(eventBase_, 0);
```
nsuke: +1
According to another look at the header doc, loopbreak is typically to be called from inside the loop.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -820,9 +821,11 @@ void TNonblockingServer::TConnection::setFlags(short eventFlags) {
  * Closes a connection
  */
 void TNonblockingServer::TConnection::close() {
-  // Delete the registered libevent
-  if (event_del(&event_) == -1) {
-    GlobalOutput.perror("TConnection::close() event_del", THRIFT_GET_SOCKET_ERROR);
+  if (eventFlags_) {
+    if (event_del(&event_) == -1) {
```
nsuke: nits: single `if` with `eventFlags_ && event_del...` ?

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -820,9 +821,11 @@ void TNonblockingServer::TConnection::setFlags(short eventFlags) {
...
+      GlobalOutput("TConnection::close event_del");
```
nsuke: Not sure why changing from `GlobalOutput.perror` to `GlobalOutput`.

jeking3: Sorry I copied this from another location and it must have been using GlobalOutput.  I will fix this.

jeking3:
The failure looks like an inability to stop the loop.
I an going to look into changing the nonblocking server's stop() method to inject an event that causes it to call loopbreak from within the loop.

jeking3:
In the next push I have changed the way the nonblocking server test starts up, avoiding a ~100ms delay three times.  Then I changed how the loop stops so that it always stops from inside using loopbreak.

jeking3:
@nsuke to test this I ran TNonblockingServerTest in a tight loop for 30 minutes.  It never made it past 30 seconds before, and that was with 300ms delay built into each test.  The test runs much faster now because I properly synchronized the "wait for the server to be listening" part of the test, and I fixed the shutdown mechanism so that it always occurs from within the event loop.

jeking3:
@nsuke this is ready for review again.

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1566,24 +1562,26 @@ void TNonblockingIOThread::setCurrentThreadHighPriority(bool value) {
...
+      setCurrentThreadHighPriority(false);
+    }
 
-  // cleans up our registered events
-  cleanupEvents();
+    // cleans up our registered events
+    cleanupEvents();
+  }
 
   GlobalOutput.printf("TNonblockingServer: IO thread #%d run() done!", number_);
```
bgedik: Is there any way to control the outputting of these messages? When we run with TNonblockingServer, we get these output. I couldn't find a way to mask them.

jeking3: You can install your own GlobalOutput handler.  The hooks should be in the top level thrift header.

bgedik: Thanks, but for a message like {{"TNonblockingServer: IO thread #0 run() done!"}} is there any kind of metadata that tells me if this is a info message or a warning/error message. It seems to be being output without any information attached to it. That makes it hard to filter the messages that we consider unnecessary.

jeking3: All of these messages were already in the implementation and they do not appear to have any discrimination as to the severity.  My assumption is they are all meant for "printf style debugging".

jeking3:
Once someone has a chance to approve the changes I can merge.

jeking3:
Could someone please do a code review and approve or request changes?  Thanks.

PR #1197, opened 2017-02-19
THRIFT-4084: Add a SSL/TLS negotiation check to crossfeature to verify SSLv3 is not active and that one of TLSv1.0 through 1.2 are accepted

jeking3:
Fixed the following server implementations to properly disable SSLv3 when using default settings:
go, nodejs, perl, py3

### DIFF
--- a/lib/go/thrift/ssl_socket.go
+++ b/lib/go/thrift/ssl_socket.go
@@ -48,6 +48,9 @@ func NewTSSLSocket(hostPort string, cfg *tls.Config) (*TSSLSocket, error) {
 // NewTSSLSocketTimeout creates a net.Conn-backed TTransport, given a host and port
 // it also accepts a tls Configuration and a timeout as a time.Duration
 func NewTSSLSocketTimeout(hostPort string, cfg *tls.Config, timeout time.Duration) (*TSSLSocket, error) {
+	if cfg.MinVersion == 0 {
+		cfg.MinVersion = tls.VersionTLS10
+	}
 	return &TSSLSocket{hostPort: hostPort, timeout: timeout, cfg: cfg}, nil
 }
 
--- a/lib/nodejs/lib/thrift/connection.js
+++ b/lib/nodejs/lib/thrift/connection.js
@@ -18,6 +18,7 @@
  */
 var util = require('util');
 var EventEmitter = require("events").EventEmitter;
+var constants = require('constants');
 var net = require('net');
 var tls = require('tls');
 var thrift = require('./thrift');
@@ -221,7 +222,7 @@ Connection.prototype.connection_gone = function () {
   this.retry_timer = setTimeout(function () {
     if (self._debug) {
        console.log("Retrying connection...");
-	}
+  }
 
     self.retry_totaltime += self.retry_delay;
 
@@ -247,6 +248,11 @@ exports.createConnection = function(host, port, options) {
 };
 
 exports.createSSLConnection = function(host, port, options) {
+  if (!('secureProtocol' in options) && !('secureOptions' in options)) {
+    options.secureProtocol = "SSLv23_method";
+    options.secureOptions = constants.SSL_OP_NO_SSLv2 | constants.SSL_OP_NO_SSLv3;
+  }
+
   var stream = tls.connect(port, host, options);
   var connection = new Connection(stream, options);
   connection.host = host;
--- a/test/tests.json
+++ b/test/tests.json
@@ -121,8 +121,8 @@
       "framed:fastframed"
     ],
     "sockets": [
-      "ip-ssl",
-      "ip"
+      "ip",
+      "ip-ssl"
     ],
     "protocols": [
       "compact",
@@ -415,7 +415,7 @@
         "-I../../lib/perl/lib/",
         "TestClient.pl",
         "--ca=../keys/CA.pem",
-        "--cert=../keys/client.pem",
+        "--cert=../keys/client.crt",
         "--key=../keys/client.key"
       ]
     },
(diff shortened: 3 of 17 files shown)


test/secure/test_secure.bash
```diff
@@ -0,0 +1,69 @@
...
+
+#
+# Negotiation Test Expectations
+#
+
+declare -A EXPECT_NEGOTIATE
+EXPECT_NEGOTIATE[ssl3]=0
+EXPECT_NEGOTIATE[tls1]=1
+EXPECT_NEGOTIATE[tls1_1]=1
+EXPECT_NEGOTIATE[tls1_2]=1
```
nsuke: Maybe split to multiple `tests.json` entries for no-ssl3 (and 2 ?) and tls 1.x tests ?
We wouldn't want any of the former in known_failures.

lib/go/thrift/ssl_socket.go
```diff
@@ -20,152 +20,155 @@
 package thrift
```
nsuke: See https://github.com/golang/go/commit/014f3dcc837cb6789076cff4fccaa3bd221f823e

jeking3: Sorry, I'm using the geany editor and I have it set to use 2 character spacing to match the rest of the codebase (C++ mostly).  I'll undo the whitespace changes here.

test/tests.json
```diff
@@ -606,5 +606,23 @@
...
+    "transports": [
+      "buffered"
+    ],
+    "sockets": [
+      "ip-ssl"
+    ],
+    "protocols": [
+      "binary"
+    ],
+    "workdir": "secure"
```
jeking3: I'm not certain how that is used and whether it runs as part of Travis CI.  Is there any more documentation for the test/features area?

lib/py/src/transport/sslcompat.py
```diff
@@ -66,6 +66,14 @@ def legacy_validate_callback(cert, hostname):
```
jeking3: In the Ubuntu 14.04 LTS docker image, Python is version 2.7.6.  This means it doesn't have SSLContext.  This code detects that condition and sets a single version that it can use.  I am actually going to change the default to TLSv1.2 for security purposes here.  Any language that cannot properly modify the openssl options to negotiate TLSv1.0 "or higher" should be forced to TLSv1.2.  From what I can tell, csharp, d, and py2 (in the build image) lack the necessary code/knowledge to allow for a SSLv23 handshake and then autonegotiate TLSv1.0 "or higher".

test/secure/test_secure.bash
jeking3: What I'm thinking here is that nothing should allow for SSLv3 as you suggested, however it would be acceptable for csharp and d languages to only allow TLSv1.2, since they lack the ability to specify TLSv1.0 "or later".  So I can relax this test to say, "Accept either TLSv1.0 or later, or TLSv1.2 alone, but never allow SSLv3".  Really I could just simplify the test down to checking that SSLv3 is not allowed and TLSv1.2 works at this point.

test/tests.json
nsuke: Simply moving the same JSON entry (with relative file path adjustment) might get this included to `make crossfeature`.
In that case, known_failure file is also in `features/known_...`.

Otherwise we can leave it as is and plan to relocate later.
I guess I'll need to put some information to `test/README.md` or create `test/features/README.md`

lib/py/src/transport/sslcompat.py
nsuke: As far as I can see, this does not change any of our docker image behaviors for python:
SSLv23 with NO_SSL* is being used for python 3 and debian python 2.

About defaulting to TLSv1.2, Python < 2.7.9 and C# with .NET 3.5 compat do not seem to have TLS 1.1 and 1.2 at all.
The highest versions listed in doc is TLSv1.0.

For C#, once dropping .NET 3.5, it would be able to support TLS 1.0, 1.1 and 1.2 at once because SslProtocols is an explicit "flags" enum that can be combined with pipe like `Tls12 | Tls11 | Tls`.
Though I agree that it's not possible to do forward compatible "SSLv23 minus SSL 2 and 3".

test/secure/test_secure.bash
nsuke: Sounds reasonable to me.

test/tests.json
jeking3: I do not see where crossfeature is executed during Travis CI.  Which build job runs it?

lib/py/src/transport/sslcompat.py
jeking3: I backed out my changes to lib/py and will re-test.

lib/nodejs/lib/thrift/connection.js
```diff
@@ -247,6 +247,11 @@ exports.createConnection = function(host, port, options) {
 };
 
 exports.createSSLConnection = function(host, port, options) {
+  if (!('secureProtocol' in options) && !('secureOptions' in options)) {
+    options.secureProtocol = "SSLv23_method";
+    options.secureOptions = 0x03000000;
```
jeking3: I made this change.

test/tests.json
jeking3: I found it.  I will move it.

test/secure/test_secure.bash
jeking3: I moved these into features/ as you requested, split them apart, and made the TLS check for now pass if any of the TLSv1.1, TLSv1.1, or TLSv1.2 is possible.  For "py" only TLSv1.0 is possible, as with csharp.

jeking3:
@nsuke this is ready for another review as all the code comments were addressed and we have a clean CI build.

PR #1200, opened 2017-02-21
THRIFT-3706: glib client/java server/crosstest multiplexed improvements

jeking3:
- Added crosstest support to prove the java Multiplexed server is backwards compatible with non-multiplexed clients when registering a default processor.  This proves that standard binary/compact/json clients can connect to a Multiplexed binary/compact/json server using new code added in THRIFT-2504.
- Added SecondService as a second named service to the Java Multiplexed TestServer
- Added multiplexed test support to c_glib test_client to test a second named service
- Added testBinary(empty) and [-128..127] test cases to c_glib test_client
- Changed the multiplexed protocol separator from a property to a constant
- Changed ThriftProtocol to manage the lifetime of the Transport object it consumes
- Changed ThriftProtocolDecorator to manage the lifetime of the Protocol object it consumes
- Changed ThriftService to manage the transport, protocol, and handler objects it consumes
- Fixed the dispose and finalize chain up in many places
- Fixed ThriftServer Socket not working properly if the port number is > 32767
- Fixed the lifetime of some objects in test_client

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/protocol/thrift_multiplexed_protocol.c
+++ b/lib/c_glib/src/thrift/c_glib/protocol/thrift_multiplexed_protocol.c
@@ -31,7 +31,6 @@
 enum
 {
   PROP_THRIFT_MULTIPLEXED_PROTOCOL_SERVICE_NAME = 1,
-  PROP_THRIFT_MULTIPLEXED_PROTOCOL_SEPARATOR,
   PROP_THRIFT_MULTIPLEXED_PROTOCOL_END
 };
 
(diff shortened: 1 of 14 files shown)


jeking3:
Build failure attributed to the fix for THRIFT-4076.  I reopened it, I will submit a PR for that and make sure it works, and then rebase this one to fix the Appveyor issue.

lib/c_glib/src/thrift/c_glib/protocol/thrift_multiplexed_protocol.c
```diff
@@ -42,146 +41,119 @@ static GParamSpec *thrift_multiplexed_protocol_obj_properties[PROP_THRIFT_MULTIP
...
-
-	switch (property_id)
-	{
-	case PROP_THRIFT_MULTIPLEXED_PROTOCOL_SERVICE_NAME:
-		if(self->service_name!=NULL)
-			g_free (self->service_name);
-		self->service_name= g_value_dup_string (value);
-		break;
-
-	case PROP_THRIFT_MULTIPLEXED_PROTOCOL_SEPARATOR:
```
gadLinux: Why you did remove the ability to select the separator. It seems that Java and maybe others support it. Think that some protocols may required to change it to something special to make it work. I would leave it.

lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c
```diff
@@ -61,15 +61,15 @@ thrift_protocol_set_property (GObject *object, guint property_id,
   switch (property_id)
   {
     case PROP_THRIFT_PROTOCOL_TRANSPORT:
-      protocol->transport = g_value_get_object (value);
```
gadLinux: Why you duplicate it? The underlaying transport should live as long as the multiplexed one. And must be destroyed after protocol is destroyed. Duplicating the transport may lead to object references hold and maybe memory freeing problems. 
I think this property must hold a reference to it and not a copy. 
The copy can lead to memory freeing problems.

lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c
```diff
@@ -549,12 +549,27 @@ thrift_protocol_init (ThriftProtocol *protocol)
...
+{
+  ThriftProtocol *self = THRIFT_PROTOCOL (gobject);
+
+  g_clear_object(&self->transport);
+
+  /* Always chain up to the parent class; there is no need to check if
+   * the parent class implements the dispose() virtual function: it is
+   * always guaranteed to do so
+   */
+  G_OBJECT_CLASS (thrift_protocol_parent_class)->dispose(gobject);
```
gadLinux: Yes I forgot that! Sorry.

lib/c_glib/src/thrift/c_glib/server/thrift_server.c
```diff
@@ -76,22 +76,22 @@ thrift_server_set_property (GObject *object, guint property_id,
   switch (property_id)
   {
     case PROP_THRIFT_SERVER_PROCESSOR:
-      server->processor = g_value_get_object (value);
```
gadLinux: Same here. Are you sure you must duplicate them?

lib/c_glib/src/thrift/c_glib/transport/thrift_socket.h
```diff
@@ -50,11 +50,8 @@ struct _ThriftSocket
 
   /* private */
   gchar *hostname;
-  gshort port;
+  guint port;
   int sd;
-  guint8 *buf;
```
gadLinux: This is the default implementation of the socket. Are you sure about it? If this is changed I suppose this should be done in it's own ticket.

lib/java/test/org/apache/thrift/test/TestServer.java
```diff
@@ -190,7 +193,9 @@ public static void main(String [] args) {
         if (protocol_type.equals("binary")) {
         } else if (protocol_type.equals("compact")) {
         } else if (protocol_type.equals("json")) {
-        } else if (protocol_type.equals("multiplexed")) {
+        } else if (protocol_type.equals("multi")) {
```
gadLinux: I don't know much about tests. But you can get crazy here if there's a lot of procotols. 
I suggest what I did. Adding a prefix and sufix. The prefix can be matched against multiplexed (multi) the suffix is the name of the protocol of the underlaying implementation. 
This convention doesn't add anything to the code. And makes you be ready for the multiplexed whether a new protocol is added with no effort. 
Convention over configuration.

lib/c_glib/src/thrift/c_glib/protocol/thrift_multiplexed_protocol.c
gadLinux: Suppose a protocol or transport that add a separator to the packet to, for example, add a header of encryption. And the separator is the same. It may be confused because found twice, once for the protocol and another for the multiplexed. Changing the multiplexed separator will fix the possible issue.

jeking3: Given it is a Thrift Protocol Decorator and the field in question can only be a string containing the name of the multiplexed service and colon and name of the original call, I don't see how it would be possible for any protocol to be confused by this always being a colon.  Protocols never see it.  In the decoding phase the call name is a string, so we will always read a string length and then a set of characters, and we will always pass this along to the Multiplexed Processor.  What you suggest means that a "string read" method for a protocol would be confused by a colon which shouldn't ever happen.

In theiry it might be possible for a Transport to be confused, however if a Transport is even looking at the content of the payload at all that is a poorly written transport.  So as I see it, using a fixed string across all languages seems quite reasonable.

Now that said, I didn't check the MultiplexedProtocol implementations to make sure they throw an exception if a colon is used in the multiplexed service name.  That should probably be a requirement for all implementations.

In this case, allowing the colon separator to be changed to something else, seems like too much complexity for no gain.

lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c
jeking3: "should" live as long.  Why should we impose a requirement that the consuming application, using an object oriented-ish language like glib, be required to ensure the underlying transport outlives the protocol that is consuming it?

The copy should not lead to memory issues because the dispose method in this class unreferences and clears the pointer.  This means the consuming application does a g_object_new, that's one reference.  The protocol consumes it, that's a second reference.  If the consuming application were to unref the transport, it would crash.  But by using reference counts on objects the way they were intended to be used, a consuming application could quite literally unref everything but the server before serve() and then unref the server when it returns, and there would be no leaks or problems.  I don't think it's a good practice to force the consuming application to guarantee the lifetime of pointers that reference each-other when the object library itself provides reference counting and garbage collection.  That's why I changed it here and in other classes like ThriftServer.

lib/c_glib/src/thrift/c_glib/transport/thrift_socket.h
jeking3: "buf" is not used anywhere in thrift_socket.c and it was marked as private.  This means it was there for no reason that's useful.

lib/java/test/org/apache/thrift/test/TestServer.java
jeking3: I did it this way to work within the spec:impl naming convention that currently exists in make cross.  See tests.json, specifically "binary:accel" or "compact:accelc".  I wanted to follow the pattern that already existed in the test suite so we have a single use pattern, not two.  In the end the behavior is mostly the same, except by using "multi:binary" on the java server and "binary:multi" on the c_glib client, we end up testing:  c_glib (binary client) => java (multi server) as well as c_glib (multi client wrapping binary) => java (multi server wrapping binary).  Using the "multiplexed-binary" naming convention would not have leveraged the existing logic in crosstest/collect.py to make this happen.

jeking3:
@gadLinux thank you for such a thorough review.  All good questions, hopefully my answers make sense.

jeking3:
Pushed an empty commit to get a clean CI build, despite the fact that the only Travis CI build failure was environmental (couldn't reach a docker server).

lib/c_glib/src/thrift/c_glib/protocol/thrift_multiplexed_protocol.c
gadLinux: Ok. Then I agree.

lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c
gadLinux: Ok. As long as the reference count is guaranteed to be well managed. And it seems because you added the finalize method it should be okay. But I think this breaks RAII. Since the application can get the resource and free it but the resource will outlive the requester inside this class. But I come from C/C++ and maybe RAII doesn't apply quite well here because the reference counter. I will study about it a little bit.

lib/java/test/org/apache/thrift/test/TestServer.java
gadLinux: ok

gadLinux:
Thank you for answers I learnt a lot

PR #1203, opened 2017-02-23
THRIFT-4102 TBufferedTransport performance issue since 0.10.0

atiking:


jeking3:
Wow this build hasn't started for about 18 hours!

PR #1205, opened 2017-02-28
THRIFT-4081: appveyor retooling - build both MSVC2015 and MINGW on appveyor

jeking3:
Also enabled Qt 5.8 for the MSVC build and cleaned up a bunch of smaller things the MINGW build found.


jeking3:
@jfarrell why didn't Travis CI build this job?  Can you kick it?  It failed previously so I want to make sure it passes.

jeking3:
I removed the artifact capture on every build - it might cost money to do this over time (unsure) and the benefit is negligible.

PR #1206, opened 2017-03-02
THRIFT-4029 Accelerated protocols do not build from thrift-py 0.10.0 on PyPI

atiking:


jeking3:
If the build fails try using this technique to kick another if you think it was environmental:

git commit --amend
  don't change anything, just save.  This makes a new identical commit with a new hash.
git push -u origin <branch name> --force
  overwrites the previous push and kicks another build

Thanks.

PR #1208, opened 2017-03-06
THRIFT-4108 : Fix several problems found on library.

gadLinux:
Adjust headers.
Format code.
Make sure finalization is correct.
Fix convenience methods.
Remove global ssl context.

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
+++ b/lib/c_glib/src/thrift/c_glib/transport/thrift_ssl_socket.c
@@ -27,10 +27,14 @@
 #include <openssl/ssl.h>
 #include <pthread.h>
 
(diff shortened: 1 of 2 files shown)


gadLinux:
I added also some documentation and missing include functions.

gadLinux:
@jeking3 This was closed but not merged? Why?

jeking3:
It was merged.  Pull master and look at the log.  When your commit is merged into master I have to amend the commit with "This closes #...." which changes the commit hash, so git may tell you when you attempt to delete the branch that it hasn't been fully merged, but it has.

Jens-G:
The careful reader will notice that it says "asfgit closed this in b79396f 19 hours ago". The b79396f is a link to the commit you are looking for.

PR #1210, opened 2017-03-10
Fix fd leak when failing to dispatch new connections

xiaosuo:
When failing to dispatch new connections to other IO threads other
than the number 0, we returned these connections for reuse without
closing them, so the corresponding fds were leaked forever.

We should close these connections instead.


jeking3:
I opened THRIFT-4129 for this.

PR #1211, opened 2017-03-10
Fix use closed(freed) connections

xiaosuo:
When failing to add tasks into the thread manager, we close the
corresponding connections, then set the flags of these connections,
which have been already freed.

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -591,21 +591,24 @@ void TNonblockingServer::TConnection::transition() {
       // The application is now waiting on the task to finish
       appState_ = APP_WAIT_TASK;
 
+      // Set this connection idle so that libevent doesn't process more
+      // data on it while we're still waiting for the threadmanager to
+      // finish this task
+      setIdle();
+
       try {
         server_->addTask(task);
       } catch (IllegalStateException& ise) {
         // The ThreadManager is not ready to handle any more tasks (it's probably shutting down).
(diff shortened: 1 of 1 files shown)


xiaosuo:
I had tried to find my password, but nothing was received. It seems that Jira doesn't work properly with gmail. Could you help me?

xiaosuo:
@jeking3 you are correct. In fact, we had found this issue, and we should do such things at all the error path when notifying the io threads. I will address this issue in a separated patch.

At the same time, we should setIdle() before calling addTask(). Because the added task maybe expired and destroyed by other threads due to failing to notify IO threads, and the latter setIdle() may corrupt the memory.

jeking3:
This patch as-is causes a counting error, so if you can amend this one with those changes we can get them all fixed together.

jeking3:
@xiaosuo just a reminder, if you have additional fixes (it sounds like you do)

xiaosuo:
@jeking3 Did you mean that we should fix all issues about the number of active processors in this patch instead of a separated one?

jeking3:
If you have the fixes let's get them all in together.  The current fix is not complete and causes counting errors, which lead to - I don't know...

xiaosuo:
I think this patch doesn't introduce any inconsistency. And the correctness of the number of active processors deserves another separated patch. We'd better not mix fixes into a huge patch.

BTW: Fixing the number of active processors may depend on my other patches. So I prefer to submit it after all my patches are merged.

jeking3:
Please rebase this branch against upstream/master and force push to kick off a new build.  I would like to see the CI build pass (or get much closer than it is now) before merging.

jeking3:
THRIFT-4160

PR #1212, opened 2017-03-11
[THRIFT-4130] Release managed HTTP connection back to the underlying pool

jdpgrailsdev:
There is a connection leak in the `THttpClient` when using the Apache `HttpClient` with the `PoolingClientConnectionManager`.  Without calling `releaseConnection` on the `HttpPost` object, the connections are never returned to the pool.  Under heavy load, this can lead to both failures for subsequent calls to be able to get a connection from the pool and connections being held by the underlying OS, eventually resulting in the inability to grab another client port for outgoing connections.  Per the Apache HttpClient examples/documentation:

> In order to ensure correct deallocation of system resources 
> the user MUST either fully consume the response content  or abort request 
> execution by calling HttpGet#releaseConnection().

This might have not been an issue when using the 3.x version of the HttpClient, but it's definitely an issue in the 4.x line.  See https://hc.apache.org/httpcomponents-client-4.2.x/quickstart.html for more details.

### DIFF
--- a/lib/java/src/org/apache/thrift/transport/THttpClient.java
+++ b/lib/java/src/org/apache/thrift/transport/THttpClient.java
@@ -304,6 +304,9 @@ public class THttpClient extends TTransport {
           throw new TTransportException(ioe);
         }
       }
+      if (post != null) {
+        post.releaseConnection();
+      }
     }
   }
 


lib/java/src/org/apache/thrift/transport/THttpClient.java
```diff
@@ -304,6 +304,9 @@ private void flushUsingHttpClient() throws TTransportException {
           throw new TTransportException(ioe);
         }
       }
+      if (post != null) {
+        post.releaseConnection();
```
jeking3: The code on line 295 appears to want to take care of this condition.  Perhaps other exceptions are occurring, or something else is happening like we're not actually consuming all the data that was sent?

From my read of this method "post" will never be null, so the check here and on line 294 are unnecessary.

I would check into the IOException handler above and see if it needs to be expanded, perhaps.

jdpgrailsdev: @jeking3 From what we have seen in testing (when using the pooled connection manager), the call to `releaseConnection` has to happen even when the request/response is successful.  If you think it makes more sense to move this into the `try` block, that can be done.  Also, I believe the `null` check is necessary if a) this code stays in finally block and b) if the call to create the `HttpPost` object fails for some reason (first line in the `try`).

jeking3:
Would you be able to open an Apache Jira ticket against the THRIFT project for this issue? Please see:

Thanks for helping us improve thrift!

jdpgrailsdev:
@jeking3 Sure thing.  Will update this PR with the link once I have it created.

jdpgrailsdev:
@jeking3 https://issues.apache.org/jira/browse/THRIFT-4130

PR #1214, opened 2017-03-18
THRIFT-2026: Eliminate some undefined behavior in C/C++

jbapple-cloudera:
Clients: glib, C++
Patch: Jim Apple <jbapple-impala@apache.org>

This patch fixes some undefined behavior were found using Clang's
UndefinedBehaviorSanitizer (UBSan). To check for undefined behavior,
run /build/docker/scripts/ubsan.sh. This is run during CI builds, as
well.

The examples of the types of undefined behavior fixed in this commit
are:

1. Enumerations exhibit undefined behavior when they have values
   outside of a range dependent on the values of their enumerators, as
   specified in C++14's chapter 7.2 ("Enumeration declarations"),
   paragraph 8.

2. Left shift of negative values, used in zigzag encoding, is
   undefined behavior. See 5.8 ("Shift operators"), paragraph 2 for
   C++ and 6.5.7 ("Bitwise shift operators"), paragraph 4 for C99 and
   C11.

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/protocol/thrift_compact_protocol.c
+++ b/lib/c_glib/src/thrift/c_glib/protocol/thrift_compact_protocol.c
@@ -120,7 +120,7 @@ thrift_bitwise_cast_gdouble (const guint64 v)
 static guint64
 i64_to_zigzag (const gint64 l)
 {
-  return (l << 1) ^ (l >> 63);
+  return (((guint64)l) << 1) ^ (l >> 63);
 }
 
 /**
@@ -130,7 +130,7 @@ i64_to_zigzag (const gint64 l)
 static guint32
 i32_to_zigzag (const gint32 n)
 {
-  return (n << 1) ^ (n >> 31);
+  return (((guint32)n) << 1) ^ (n >> 31);
 }
 
 /**
--- a/lib/cpp/src/thrift/protocol/TCompactProtocol.tcc
+++ b/lib/cpp/src/thrift/protocol/TCompactProtocol.tcc
@@ -387,7 +387,7 @@ uint32_t TCompactProtocolT<Transport_>::writeVarint64(uint64_t n) {
  */
 template <class Transport_>
 uint64_t TCompactProtocolT<Transport_>::i64ToZigzag(const int64_t l) {
-  return (l << 1) ^ (l >> 63);
+  return (static_cast<uint64_t>(l) << 1) ^ (l >> 63);
 }
 
 /**
@@ -396,7 +396,7 @@ uint64_t TCompactProtocolT<Transport_>::i64ToZigzag(const int64_t l) {
  */
 template <class Transport_>
 uint32_t TCompactProtocolT<Transport_>::i32ToZigzag(const int32_t n) {
-  return (n << 1) ^ (n >> 31);
+  return (static_cast<uint32_t>(n) << 1) ^ (n >> 31);
 }
 
 /**
--- a/.travis.yml
+++ b/.travis.yml
@@ -152,6 +152,12 @@ env:
       BUILD_ENV="-e CC=gcc -e CXX=g++"
       DISTRO=debian
 
+    # C and C++ undefined behavior. This wraps autotools.sh, but each binary crashes if
+    # undefined behavior occurs. Skips the known flaky tests.
+    - TEST_NAME="UBSan"
+      SCRIPT="ubsan.sh"
+      BUILD_ARG="--without-haskell --without-nodejs --without-perl --without-python"
+
 matrix:
   include:
     # QA jobs for code analytics and metrics
--- a/build/docker/scripts/ubsan.sh
+++ b/build/docker/scripts/ubsan.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+set -ex
+
+# Wraps autotools.sh, but each binary crashes if it exhibits undefined behavior. See
+# http://releases.llvm.org/3.8.0/tools/clang/docs/UndefinedBehaviorSanitizer.html
+
+# Install a more recent clang than default:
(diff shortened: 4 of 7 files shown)


compiler/cpp/src/thrift/parse/t_field.h
```diff
@@ -68,7 +68,7 @@ class t_field : public t_doc {
 
   int32_t get_key() const { return key_; }
 
-  enum e_req { T_REQUIRED, T_OPTIONAL, T_OPT_IN_REQ_OUT };
+  enum e_req : unsigned { T_REQUIRED, T_OPTIONAL, T_OPT_IN_REQ_OUT };
```
jeking3: We still support C++03 and earlier compilation, so this won't work.  I know it's a burden, but I'm also on some projects that haven't moved to C++11 yet, amazing as it is.

lib/c_glib/src/thrift/c_glib/protocol/thrift_compact_protocol.c
```diff
@@ -120,7 +120,11 @@ thrift_bitwise_cast_gdouble (const guint64 v)
 static guint64
 i64_to_zigzag (const gint64 l)
 {
-  return (l << 1) ^ (l >> 63);
```
jeking3: Couldn't we put ULL qualifiers on everything here to force them to unsigned long long types?

jbapple-cloudera: Well, you can only put those suffixes like "ULL" on literals, and the only literals here are the right-hand sides of the shifts, and the right-hand sides are not the problem. The problem is that `p << q` has undefined behavior when p is negative.

compiler/cpp/src/thrift/parse/t_field.h
jbapple-cloudera: Got it. Should I add a new patch to this pull request or squash them into one patch on this pull request?

lib/c_glib/src/thrift/c_glib/protocol/thrift_compact_protocol.c
jeking3: Sounds to me like a negative input is not allowed; can we change the input type to unsigned then, or throw an exception when it is negative?  If you look inside commit 7edc8faefd391ce11eca3023a35cc54bcb2eb1af I added code in the cpp library (and compiler, following the "throw a string" ugliness) to throw on various conditions like this.  Let's make it so these undefined cases cause an exception rather than trying to clip them.

jbapple-cloudera: Negative input is expected:

https://github.com/apache/thrift/blob/796667b29f91682871cec310571e66b1259fffb0/lib/cpp/src/thrift/protocol/TCompactProtocol.tcc#L385

Also, changing the input to unsigned would actually break this, because the right shift depends on being an arithmetic shift, not a logical one.

lib/cpp/src/thrift/protocol/TCompactProtocol.tcc
```diff
@@ -387,7 +389,11 @@ uint32_t TCompactProtocolT<Transport_>::writeVarint64(uint64_t n) {
  */
 template <class Transport_>
 uint64_t TCompactProtocolT<Transport_>::i64ToZigzag(const int64_t l) {
-  return (l << 1) ^ (l >> 63);
+  const int64_t m = l >> 63;
+  uint64_t p, q;
+  memcpy(&p, &l, sizeof(p));
+  memcpy(&q, &m, sizeof(q));
+  return p = (p << 1) ^ q;
```
nsuke: `return (static_cast<uint64_t>(l) << 1) ^ (l >> 63);` would be much simpler and efficient.
static casting of negative to unsigned value is well defined AFAICT.

compiler/cpp/src/thrift/parse/t_field.h
jbapple-cloudera: I squashed.

lib/cpp/src/thrift/protocol/TCompactProtocol.tcc
jbapple-cloudera: Good point.

lib/c_glib/src/thrift/c_glib/protocol/thrift_compact_protocol.c
jbapple-cloudera: You can read more about zigzag encoding here: https://developers.google.com/protocol-buffers/docs/encoding#types

jbapple-cloudera:
Can this be on hold for a bit? I would like to add the UBSan checks to .travis.yml first, which may take a couple of days for me to have time for.

compiler/cpp/src/thrift/parse/t_field.h
```diff
@@ -68,7 +68,7 @@ class t_field : public t_doc {
 
   int32_t get_key() const { return key_; }
 
-  enum e_req { T_REQUIRED, T_OPTIONAL, T_OPT_IN_REQ_OUT };
+  enum e_req { T_REQUIRED = 0u, T_OPTIONAL, T_OPT_IN_REQ_OUT = ~0u };
```
jeking3: This changes the value of T_OPT_IN_REQ_OUT from before - is this ever on the wire?  Would an older client end up sending "2" here instead of ~0u ?

jbapple-cloudera:
Agreed. I'l get to that as soon as I have a chance and then I'll ping this PR for review when it has addressed that.

jbapple-cloudera:
OK, this new patch does not change the definitions of any `enum`s, only their usages. It also adds undefined behavior checking to the pre-merge CI tests on Travis to discourage regressions.

PR #1215, opened 2017-03-19
Thrift 4126

kufd:
Hello
It is a pull request with required fields validation for php extension
Here is jira issue https://issues.apache.org/jira/browse/THRIFT-4126


kufd:
Hi @jeking3 
I looked into the failed builds. But errors are not related to my changes.
Maybe i should wait for some time and merge changes from thrift repo into my branch.

jeking3:
Yes, I recommend you:

1. Squash your commits (optional, but we'll do it anyway when we merge into apache master)
2. Rebase against upstream master.
3. Force push.

This will trigger a new build against the tip.
If you are not familiar with squashing, it's okay if you skip that part.

kufd:
Now all build jobs are success

kufd:
Yes
I will add functional tests
But it can take a few weeks, because i do not have enough free time now

jeking3:
I'll open a story in Jira for it.

lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
```diff
@@ -728,6 +728,42 @@ inline bool ttypes_are_compatible(int8_t t1, int8_t t2) {
...
+
+    zend_hash_find(fieldspec, "isRequired", 11, (void**)&val_ptr);
+    bool is_required = Z_BVAL_PP(val_ptr);
+
+    zval* prop = zend_read_property(object_class_entry, object, varname, strlen(varname), false TSRMLS_CC);
+
+    if (is_required && Z_TYPE_P(prop) == IS_NULL) {
+        char errbuf[128];
+        snprintf(errbuf, 128, "Required field %s.%s is unset!", object_class_entry->name, varname);
+        throw_tprotocolexception(errbuf, INVALID_DATA);
```
jeking3: I wonder if folks think it would be useful to add a new TProtocolException cause for this case, MISSING_REQUIRED, across all languages.  While technically it is invalid data, I usually think of that as something got garbled.

jeking3:
When this does merge into master, it's probably worth updating the README.md indicating this is a breaking change (for the better).  I would like to take a look at how some of the other runtime libraries for Thrift handle required fields as well to see what the consensus behavior is.

kufd:
There is a falg "validate" in thrift generator for php (validate:        Generate PHP validator methods\n")
 And that flag is used in my changes
I will try  to add tests for both cases: when the flag is set and when not

PR #1216, opened 2017-03-20
TNonblockingServer: Fix using uninitialized event_

xiaosuo:
When there are more than one IO threads, and we have failed to notify
one IO threads, then we have to close the connection. But the event_
in that connection isn't initialized. We should use setIdle() instead.


xiaosuo:
However when closing the connection, we don't check if eventFlags_ is zero.

For a new connection, event_ isn't initialized by the constructor() or init(), and if we fail to pass it to the other IO threads, we will close the connection and use event_ without checking eventFlags_.

jeking3:

jeking3:
THRIFT-4161

PR #1217, opened 2017-03-20
TSocket: Don't close connections when failing to read/write

xiaosuo:
Sockets maybe registered to a event base of libevent, so the
following epoll_ctl(2) may fail due to EBADF, or may
delete other sockets. Chaos occurs.


xiaosuo:
I checked other call sites, and found no issue. Because we will close the connection in the destructor,  we won't leak fds at least.

jeking3:

jeking3:
THRIFT-4162

PR #1218, opened 2017-03-20
export thrift::mem::TBufferTransport

flier:
We need export `thrift::mem::TBufferTransport` to serialize record to a memory based buffer, then send the buffer through other transport like Kafka

### DIFF
--- a/lib/rs/src/transport/mod.rs
+++ b/lib/rs/src/transport/mod.rs
@@ -30,9 +30,9 @@ mod framed;
 mod passthru;
 mod socket;
 
-#[cfg(test)]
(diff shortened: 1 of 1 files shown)


sunchao:
I created a ticket for this: THRIFT-4113. However, a easy workaround is to just create a wrapper around the memory buffer, and implement Read + Write yourself.

flier:
Thanks, we use open source project because we are not intent to reinvent wheel :)

jeking3:

sunchao:
This change looks good to me, even though I'm also relatively new to Rust.

PR #1220, opened 2017-03-26
THRIFT-4069: perl library cleanup - namespaces, versioning, exception fixes

jeking3:


build/docker/centos/Dockerfile
```diff
@@ -74,10 +74,8 @@ RUN yum install -y \
       perl-Bit-Vector \
       perl-Class-Accessor \
       perl-ExtUtils-MakeMaker \
-      perl-Test-Simple \
       perl-IO-Socket-SSL \
-      perl-Net-SSLeay \
-      perl-Crypt-SSLeay
```
jeking3: I have reverted this change locally, since both SSLeay modules are needed for SSL/TLS sockets and for Http(s)Client transport.  When I merge upstream I will drop the changes to this file.

build/docker/debian/Dockerfile
```diff
@@ -157,7 +157,7 @@ ENV PATH /usr/local/go/bin:$PATH
 
 # Haxe
 RUN mkdir -p /usr/lib/haxe && \
-    curl http://haxe.org/website-content/downloads/3.2.0/downloads/haxe-3.2.0-linux64.tar.gz | \
+    wget -O - https://github.com/HaxeFoundation/haxe/releases/download/3.4.2/haxe-3.4.2-linux64.tar.gz | \
```
jeking3: haxe 3.2.0 is no longer available for download, and I needed to build a local image to identify build issues.

jeking3: Will wait for THRIFT-4141 and merge out the Dockerfile change from this pull request.

PR #1223, opened 2017-03-26
THRIFT-4137 vptr @ compiler

Jens-G:


jbapple-cloudera:
@jeking3 It looks to me like this is failing in TravisCI at the UBSan test.

Jens-G:
Yeah, but at least it looks as if we get the compiler ready. 

If there are no objections, I am going commit that part today, With UBSAN still left disabled of course. I only enabled it in the PR to have travis run these tests. 

Since you already made some proposals, could you take care about the boost / libraries part? @jeking3 and me will be happy reviewers, that's for sure.

jbapple-cloudera:
I can take a look at the Boost problem next weekend.

PR #1228, opened 2017-03-30
THRIFT-3978: tighten up pthread mutex implementation, removing asserts and replacing them with exceptions

jeking3:


jeking3:
For anybody interested in reviewing this pull request, I pretty much followed the logic that boost::mutex is using with one exception.  In cases where boost ignores an error condition to avoid throwing in a destructor, I have instead preferred to abort the process.  My thought process on this is that should mutex destruction return EBUSY or EINVAL, it is a design flaw in the implementation using the Mutex class.  EBUSY means we tried to delete the mutex while it was locked.  EINVAL means it may have been destroyed twice.  In either case, getting a core is way more useful than silently letting the issue pass by in a release build.

PR #1230, opened 2017-03-31
Make a patch to fix #4152 at least for structs

gadLinux:
Change type Type_ for Struct_ so structs are correctly linked.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_html_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_html_generator.cc
@@ -698,7 +698,7 @@ int t_html_generator::print_type(t_type* ttype) {
     string type_name = ttype->get_name();
     f_out_ << "<a href=\"" << make_file_link(prog_name + ".html") << "#";
     if (ttype->is_typedef()) {
-      f_out_ << "Typedef_";
+      f_out_ << "Struct_";
     } else if (ttype->is_struct() || ttype->is_xception()) {
       f_out_ << "Struct_";
     } else if (ttype->is_enum()) {


Jens-G:
THRIFT-4152

gnillev:
Sorry to rip open an old issue, but this actually break things for me, since my typedefs are actually generated with the "Typedef_" id! 
(Don't know if anything have changed since this was resolved)

Example:
In a project of mine, we are sometimes using typedefs to simplify imports, so say we have example.thrift containing:
```
import "../path/to/filename_may_change.thrift"

typedef filename_may_change.SomeStruct SomeStruct

struct SomeOtherStruct {
    1: SomeStruct some_struct;
}
```
This allows me to simply use SomeStruct instead of filename_may_change.SomeStruct in the definitions, and when the filename changes (for instance if version is included in the filenames), I only have to update in the import and typedef, instead of multiple places in the thrift file.

However, currently this causes some issues in the generated html, since the actual ids for the typedefs in the html is "example.html#Typedef_SomeStruct" (which is also used in the links in the generated TOC rows). 
However further down in the usages of SomeStruct, the generated html tries to link to "example.html#Struct_SomeStruct", which obviously doesn't exist.

It would be fine for me if it was aligned, so the TOC links and the actual ID's are the same as in the links in Data Structures, but currently at least this is not working for me. 
I can't actually imagine that the TOC links are working properly in your use case, since they are generated with #Typedef_ as well?

(Alternatively, they could link directly to the "filename_may_change.html#[...]", but that is probably more tricky.)

gadLinux:
Maybe you have to open another issue... I can check it since I was not even aware that there was even a typedef for definitions.

PR #1232, opened 2017-04-01
THRIFT-4138: Remove undefined behavior imported from Boost

jbapple-cloudera:
Client: C++

There is undefined behavior in boost::lexical_cast that was fixed in
https://github.com/boostorg/lexical_cast/issues/21, but that fix is
only available in recent Boost releases. This patch removes all uses
of lexical_cast instead.

That removes the last undefined behavior, so this patch also makes
ubsan.sh unconditionally fail on undefined behavior.

### DIFF
--- a/build/docker/scripts/ubsan.sh
+++ b/build/docker/scripts/ubsan.sh
@@ -15,7 +15,7 @@ export CXX=clang++-3.8
 # undefined casting, aka "vptr".
 #
 # TODO: fix undefined vptr behavior and turn this option back on.
-export CFLAGS="-fsanitize=undefined -fno-sanitize-recover=undefined -fno-sanitize=vptr"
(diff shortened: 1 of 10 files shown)


jeking3:
I'm curious - did every case where lexical_cast was used report this error running in ubsan or was it perhaps just one of them?  The original boost defect says, "Undefined behaviour in lexical cast of boost::filesystem::path to std::string".  We're not using boost::filesystem::path in any of these locations - it's usually just a simple numeric conversion.

jbapple-cloudera:
UBSan refuses to report an error more than once, and unfortunately it does not have a way to override this behavior. So, once you've seen one UBSan error at a particular line in a particular function, it won't report it again until the binary exits.

I agree that we are not using boost::filesystem::path. This was a different error from https://github.com/boostorg/lexical_cast/blob/boost-1.54.0/include/boost/lexical_cast.hpp#L1620.

PR #1011, opened 2016-05-17
Fix bug for php close socket

caty:
php must close the socket, anyhow.
otherwise you will get previous reponse when the previous request timeout and the socket is reused.

PR #1233, opened 2017-04-01
THRIFT-4154: properly close socket whether it is persistent or not on close()

jeking3:
Replaces PR #1011.  Resubmitting to get a CI build that passes.


jeking3:
I logged the build failure as THRIFT-4155 as it appears to be an intermittent test failure.  I've seen it twice this week.

PR #1237, opened 2017-04-05
THRIFT-4166: fix recent VC2010 build regression and prevent it in the future with appveyor

jeking3:
I added a MSVC 2010, x86, debug, boost-1.54, qt-5.6 appveyor job to prevent this from happening again.


jeking3:
Appveyor died due to an environmental error getting "ant" with chocolatey.  I have to do an empty commit and kick another one, but I'm going to wait until THRIFT-4071 so it isn't as expensive.

PR #1238, opened 2017-04-05
THRIFT-4071: collapse and remove unnecessary build jobs in Travis CI

jeking3:
Removed build jobs 2 and 4 as they were duplicates of 1 and 3 but on debian instead of ubuntu, this adds too little value to justify the expense of two 40 minute builds.

Added compiler plug-in option to job 6 to make sure it is tested in CI.

Combined jobs 7 and 8 because there is enough time in the build to support it.

Removed build job 9 but kept 10 because it uses debian in light of removing 2 and 4.

Removed build job 12 and 15 as they are redundant to 13 and 14 and add too little value to justify the expense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,22 +47,10 @@ env:
       BUILD_ARG="-'(binary|header|multiplexed)'"
       BUILD_ENV="-e CC=clang -e CXX=clang++ -e THRIFT_CROSSTEST_CONCURRENCY=4"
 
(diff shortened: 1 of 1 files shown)


jeking3:
Although all builds passed, the plug-in builds were not enabled properly for autoconf and a plug-in build with cmake was missing.  Pushing a fix.

jeking3:
The build failure is due to a docker image rebuild because of a submit that was not against master from yesterday/today.

PR #1239, opened 2017-04-05
Fix wrong numActiveProcessors_ when failing to notify any IO thread

xiaosuo:
We should decrease numActiveProcessors_ when failing to notify any
IO thread.


jeking3:
Please open an Apache Jira ticket for each pull request.

jeking3:
We'll re-use THRIFT-4160 for this.

PR #1240, opened 2017-04-06
return full error message when exception happens

DanielYWoo:
If you throw an exception with error message, the error message will be IGNORED and replaced with a generic message like "Internal processing error getUserInfo". The pull request will append the ACTUAL error message to that message. e,g. "Internal processing error getUserInfo, error message: Cannot connect to database"

### DIFF
--- a/lib/java/src/org/apache/thrift/ProcessFunction.java
+++ b/lib/java/src/org/apache/thrift/ProcessFunction.java
@@ -36,12 +36,12 @@ public abstract class ProcessFunction<I, T extends TBase> {
       result = getResult(iface, args);
     } catch(TException tex) {
       LOGGER.error("Internal error processing " + getMethodName(), tex);
-      handleException(seqid, oprot);
+      handleException(seqid, oprot, tex);
       return;
     } catch(RuntimeException rex) {
       LOGGER.error("Internal error processing " + getMethodName(), rex);
       if (handleRuntimeExceptions()) {
-        handleException(seqid, oprot);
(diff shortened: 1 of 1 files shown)


jeking3:
This is related to THRIFT-3607.

DanielYWoo:
Thanks for explanation. In our case, we have a web server layer on top of many layers of backends which communicate in thrift, e,g. 

> web server -> backend A -> backend B -> backend C.

It will be very difficult for trouble shooting on the web server or backend A. Since we will never ever show exception message from the web server to the browser or client app, so there is no way to trigger an exception and see the details of the backend servers. I believe the developers of the web servers or API gateways have the responsibility to keep the details no leaking to the outside of the world. Enforcing security at the thrift layer is fine but I just hope there is an alternative or option to enable this. 

Would you accept a global configurable variable for this?

jeking3:
In a project I work on that uses thrift, the company maintains a set of patches they pull (mostly from master, where defects have been fixed on top of 0.10.0) that they need to ship something that works for them.  If you need to change the core exception handling behavior for your project, you can always do so in your own release engineering workflow regardless of whether a patch gets accepted and merged upstream.  That said, you are always welcome to submit a patch, certainly.  There may be other folks who use thrift who would want the same option, perhaps for standing up a new service for internal QA and getting more rapid identification of issues during testing.

DanielYWoo:
Yes, we probably keep a patch list for this. Thanks Jeking3.

PR #669, opened 2015-10-30
use lua_objlen instead of lua_rawlen when lua version < 5.2

wlgys8:

PR #1241, opened 2017-04-06
THRIFT-4170: re-submitting PR #669 to fix lua backwards compatibility

jeking3:


jeking3:
The only build failure is covered in THRIFT-4071, so I am going to merge this.

PR #1245, opened 2017-04-06
THRIFT-4071: rebalance build job 5 to eliminate CI failures

jeking3:


jeking3:
Got this two builds in a row:

OS Error: Connection timed out, errno = 110, address = storage.googleapis.com, port = 49974

PR #1250, opened 2017-04-10
fix lnk 200 error on windows when compile as lib

xunonxyz:
fix lnk 200 error on windows when compile as lib, can not link the source code such as TServerFramework.cpp


jeking3:
I'm phasing out the static visual studio project (that's checked in) for use of cmake generated project files.  That said, this seems to resolve THRIFT-3832 and for the short while the static project is still checked in, this should resolve some issues for folks.

PR #1251, opened 2017-04-14
Added support for C++ Nonblocking SSL Server

dthaluru:


jeking3:
Is there an Apache Jira ticket for this enhancement?

dthaluru:
@jeking3 I have addressed all issues and updated the PR.  I am not aware of any Apache Jira ticket for this.

dthaluru:
@jeking3 I have added Transport layer for Nonblocking server  as suggested. And I have also added a test for TNonblockingSSLServer.

jeking3:
Thanks!  I am going to need some time to review it further.  The server layer should know nothing about the transport which it looks like you addressed, but similarly the transport layer should know nothing about the server.  There's a new TNonblockingSSLServerSocket class here - this seems to be combining the responsibility of the transport with that of knowing the server type is.

Here are a couple things to consider from a quick review:

1.  I wonder if you need the Nonblocking socket classes; instead if you have TNonblockingServer call TSSLSocket::setLibeventSafe(false) after asking for the socket from the factory - you might not need a Nonblocking factory at all (a nonblocking socket factory combines server and transport concerns).

2. Is there any reason isLibeventSafe() branches only apply to TSSLSocket?  If the same reasoning applies to TSocket as well, then perhaps that logic moves into TSocket so TNonblockingServer uses TSocket the same way as TSSLSocket?

I'm curious what the specific "not libevent safe" mechanisms are.  I'll pick that up from further code review.

dthaluru:
One of the main reason I had separate transport for Nonblocking is, blocking has a feature called interruptible which does not really apply for nonblocking. And additional api's are added for Nonblockingtransport which are not required in blockingtransport.

And currently TSSLSocket underlying implementation is Nonblocking by default and uses select api.
TNonblockingServer implementation uses libevent which inturn calls select api's. We cannot use libevent with select as  there will be a race condition between select and libevent as both poll for events on registered sockets. Ideally there should be another class called TNonblockingSSLSocket which can be used with NonblockingServers.

jeking3:
Interruptable sockets can be controlled by the server during creation, I believe - one should not need another class to use TSocket in a non-interruptable fashion.  You may be right about the select() issue, as the TNonblockingServer really had its own socket implementation it was using.  I'll take a look at the other changes shortly.

PR #1252, opened 2017-04-14
[Ruby] Logging server-side exception message on  ApplicationException::INTERNAL_ERROR

laysakura:
## Problem
`ApplicationException::INTERNAL_ERROR` is often raised, mainly because of incompatibility between Thrift interface and payload (mistakenly using incorrect Thrift types, breaking changes on Thrift interfaces, ...).

But `ApplicationException::INTERNAL_ERROR` does not provide almost any information on what's wrong.

## Solution
Logs `e.message` and `e.backtrace`.

### DIFF
--- a/lib/rb/lib/thrift/processor.rb
+++ b/lib/rb/lib/thrift/processor.rb
@@ -17,10 +17,18 @@
 # under the License.
 # 
 
+require 'logger'
+
 module Thrift
   module Processor
-    def initialize(handler)
+    def initialize(handler, logger=nil)
       @handler = handler
+      if logger.nil?
+        @logger = Logger.new(STDERR)
+        @logger.level = Logger::WARN
+      else
+        @logger = logger
+      end
     end
 
     def process(iprot, oprot)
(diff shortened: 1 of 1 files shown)


jeking3:
It would be acceptable to have the server put this information into the GlobalLogger (if Ruby has the same logging concept as the C++ runtime) so that it ends up in a server log file.  That would not be a security issue.

laysakura:
@jeking3 Thank you for feedback.

> It would be acceptable to have the server put this information into the GlobalLogger (if Ruby has the same logging concept as the C++ runtime) so that it ends up in a server log file. That would not be a security issue.

I got it.

Ruby code does not have GlobalLogger-like feature, so I use the same code as https://github.com/apache/thrift/blob/master/lib/rb/lib/thrift/server/nonblocking_server.rb#L29-L34 to get a logger.

Please check it 🙏

jeking3:
This looks okay to me, but I don't have any ruby experience; hopefully someone else who does can review it.

allengeorge:
@jeking3 If you're willing to take this...I have some ruby experience, and this looks good to me.

jeking3:
Great - well once I can commit again, I will. :)

Jens-G:
Tracked in THRIFT-4198. See [our contrib guide](http://thrift.apache.org/docs/HowToContribute) for details.

PR #1254, opened 2017-04-18
THRIFT-4177 fix java deep copy

afds:
Java compiler produces deep copy constructor that could make shallow copy instead.


afds:
Are you sure? I don't really understand what should I update there.

jeking3:
I'll take a look...

afds:
The test is called TestDeepCopy and it uses entities defined in JavaDeepCopyTest.thrift file.

PR #1255, opened 2017-04-20
THRIFT-4176: Implement threaded server for Rust

allengeorge:
Client: rs

**Overview**

* Replaces single-threaded synchronous server with synchronous threadpool server
* Removes refcounting (simplifies reasoning about thread-safety - i.e. when developing server)
* Reworked types to allow for single-instance ownership and remove refcounting
* Removed `Box` in the user-visible API

**Rationale**

The original version of the library supported only single-threaded operation. To encode the idea that a `Transport` could be shared by a `InputProtocol` and an `OutputProtocol` I used reference-counting via `Rc` in the user-visible API. When I implemented the threaded server I would have had to change all instances of `Rc` to `Arc` for thread-safety. This would incur a large runtime cost (each access of a `*Protocol`, many hundreds while serializing and deserializing messages) for no good reason: after all, while two threads may read/write from the same socket or channel concurrently, only one thread would ever be reading and another thread the writing.

Using inspiration from [Tokio](https://tokio.rs) I pushed the concept of shared ownership *into* a new low-level construct - `TIoChannel`, and allowed it to be `split()` into a readable and writable type, each of which could be owned by upper layers. This allowed me to remove all reference counting and explicit boxing in the user-visible API and allowed for a straightforward implementation of a threaded server. It also made the library easier to use and simplified test code.

Consider the two code samples below:

*Original:*

```rust
use std::cell::RefCell;
use std::rc::Rc;
use thrift::protocol::{TBinaryInputProtocol, TInputProtocol};
use thrift::transport::{TTcpTransport, TTransport};

// create the I/O channel
let mut transport = TTcpTransport::new();
transport.open("127.0.0.1:9090").unwrap();
let transport = Rc::new(RefCell::new(Box::new(transport) as Box<TTransport>));

// create the protocol to decode bytes into types
let mut protocol = TBinaryInputProtocol::new(transport.clone(), true);

// read types from the wire
let field_identifier = protocol.read_field_begin().unwrap();
let field_contents = protocol.read_string().unwrap();
let field_end = protocol.read_field_end().unwrap();
```
`transport` has to be explicitly ref-counted, boxed and cloned into the `protocol`.

*Updated:*

```rust
use thrift::protocol::{TBinaryInputProtocol, TInputProtocol};
use thrift::transport::TTcpChannel;

// create the I/O channel
let mut transport = TTcpChannel::new();
transport.open("127.0.0.1:9090").unwrap();

// create the protocol to decode bytes into types
let mut protocol = TBinaryInputProtocol::new(transport, true);

// read types from the wire
let field_identifier = protocol.read_field_begin().unwrap();
let field_contents = protocol.read_string().unwrap();
let field_end = protocol.read_field_end().unwrap();
```
`transport` can be used directly, without any additional wrapping.

**Testing**

* C++ code compiles cleanly without warnings
* All cross-tests pass
* All library, tutorial, test and cross-test code pass clippy lint checks

**Additional Notes**

As requested in the [THRIFT-4176](https://issues.apache.org/jira/browse/THRIFT-4176) JIRA discussion I've *replaced* the existing single-threaded, synchronous server with a fixed-size threadpool-based server.


allengeorge:
@jeking3 @Jens-G Any chance this could be merged in?

allengeorge:
Updated the PR description with an example. This PR now contains the final state of all types, etc and should make the runtime code a lot cleaner and easier to use!

jeking3:
My only hesitation on these changes is that there are no cross tests for rust right now, so we don't have a good indicator on their stability in general.  We need to update the docker files for Ubuntu to include rust, and get rust into the crosstest (Travis build jobs 1, 2).  You can even follow the java-c_glib multiplexed test example if you want to make it complete.

I'm not very useful reviewing the code because I don't know the language.  As for the appveyor build error, I will fix that.  Some path strings have to be unix-style with later releases of cmake or FindBoost gets "confused" and reports invalid escape sequences.

allengeorge:
@jeking3 That's a very fair point - I didn't realize the cross-tests weren't running. I try to be disciplined in my testing and ensure nothing breaks, but I'm only human; automating things is the right way to go. I'll look at the c_glib example and update the docker images to run the cross tests.

And, I agree that it would be nice to have a second person knowledgeable in the language. I'm not sure how best to make that happen. Any ideas?

jeking3:
Based on Jira comments it looks like @RandyAbernethy might be using it. :)

PR #1258, opened 2017-04-26
THRIFT-4183 Named pipe client blocks forever on Open() when there is …

Jens-G:
…no server at the other end

Client: C#
Patch: Jens Geyer

### DIFF
--- a/lib/csharp/src/Transport/TNamedPipeClientTransport.cs
+++ b/lib/csharp/src/Transport/TNamedPipeClientTransport.cs
@@ -22,6 +22,7 @@
  */
 
 using System.IO.Pipes;
+using System.Threading;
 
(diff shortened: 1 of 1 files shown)


Jens-G:
That travis thing seems pretty reliable. The pobkly difference is in  the commit message, yet it fails where it succeeded minutes ago. Compare https://travis-ci.org/apache/thrift/builds/226143091 against https://travis-ci.org/apache/thrift/builds/226192720.

@jeking3: Maybe. On the other hand, timeouts in C# are always msec, and the name is self-documenting. More comments would be helpful nevertheless, but that would be a much larger task to do it right. Too much for my little addition :-) Appreciate both review & comments though.

PR #1260, opened 2017-05-01
THRIFT-4186 Add travis build for Rust

allengeorge:
Client: rs

* Install Rust in ubuntu/centos docker images
* Fix type bounds on TMultiplexedProcessor
* Add multiplexing support to cross-test server and client
* Run multiplexed tests in rs cross-test
* Temporarily add dart->rs framed as "known failures"

### DIFF
--- a/test/known_failures_Linux.json
+++ b/test/known_failures_Linux.json
@@ -224,5 +224,7 @@
   "hs-py3_json_framed-ip",
   "java-d_compact_buffered-ip",
   "java-d_compact_buffered-ip-ssl",
-  "java-d_compact_framed-ip"
+  "java-d_compact_framed-ip",
+  "rs-dart_binary_framed-ip",
+  "rs-dart_compact_framed-ip"
 ]
--- a/test/rs/src/bin/test_server.rs
+++ b/test/rs/src/bin/test_server.rs
@@ -16,6 +16,10 @@
 // under the License.
 
 #[macro_use]
+extern crate log;
+extern crate env_logger;
+
+#[macro_use]
 extern crate clap;
 extern crate ordered_float;
 extern crate thrift;
@@ -29,17 +33,21 @@ use std::time::Duration;
 use thrift::protocol::{TBinaryInputProtocolFactory, TBinaryOutputProtocolFactory,
                        TCompactInputProtocolFactory, TCompactOutputProtocolFactory,
                        TInputProtocolFactory, TOutputProtocolFactory};
-use thrift::server::TServer;
+use thrift::server::{TMultiplexedProcessor, TServer};
 use thrift::transport::{TBufferedReadTransportFactory, TBufferedWriteTransportFactory,
                         TFramedReadTransportFactory, TFramedWriteTransportFactory,
                         TReadTransportFactory, TWriteTransportFactory};
 use thrift_test::*;
 
 fn main() {
+    env_logger::init().expect("logger setup failed");
+
+    debug!("initialized logger - running cross-test server");
+
     match run() {
-        Ok(()) => println!("cross-test server succeeded"),
+        Ok(()) => info!("cross-test server succeeded"),
         Err(e) => {
-            println!("cross-test server failed with error {:?}", e);
+            info!("cross-test server failed with error {:?}", e);
             std::process::exit(1);
         }
     }
@@ -70,7 +78,7 @@ fn run() -> thrift::Result<()> {
     let workers = value_t!(matches, "workers", usize).unwrap_or(4);
     let listen_address = format!("127.0.0.1:{}", port);
 
-    println!("binding to {}", listen_address);
+    info!("binding to {}", listen_address);
 
     let (i_transport_factory, o_transport_factory): (Box<TReadTransportFactory>,
                                                      Box<TWriteTransportFactory>) =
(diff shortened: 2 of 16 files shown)


allengeorge:
@jeking3 I've updated the travis build with rust support, but I think I'm running into timeouts because now the builds take too long :/ Any ideas what I should do?

allengeorge:
Also, I'm fixing the CentOS issues. I mistakenly assumed that `sh` would be the same in both ubuntu and Centos. Guess I was wrong...

jeking3:
We are definitely getting close to the limit.  There are plenty of things we can do.  We could change Build job #1 to build everything and run all tests (which it does), and test headers, multiplexed, and json transport.  We could change build job #2 to build everything, NOT run unit tests (redundant to job #1), and test binary and compact protocols.   This would probably even things out a bit and add some breathing room for netcore addition to crosstest.

allengeorge:
No problem - I can change it to do that. Should that be in a different commit/pull-request, or merged in this one?

jeking3:
Keeping it separate is fine.  Thanks for offering to make that change to the travis jobs.

test/known_failures_Linux.json
```diff
@@ -224,5 +224,7 @@
   "hs-py3_json_framed-ip",
   "java-d_compact_buffered-ip",
   "java-d_compact_buffered-ip-ssl",
-  "java-d_compact_framed-ip"
+  "java-d_compact_framed-ip",
+  "rs-dart_binary_framed-ip",
```
jeking3: We should file a defect on the poor dart behavior that you described.

test/rs/src/bin/test_server.rs
```diff
@@ -104,35 +104,43 @@ fn run() -> thrift::Result<()> {
...
+            server.listen(&listen_address)
         }
-        "thread-pool" => {
-            TServer::new(
+        t if t.to_owned().starts_with("multi") => {
+            let second_service_processor = SecondServiceSyncProcessor::new(SecondServiceSyncHandlerImpl {},);
+
+            let mut multiplexed_processor = TMultiplexedProcessor::new();
+            multiplexed_processor.register("ThriftTest", Box::new(test_processor));
+            multiplexed_processor.register("SecondService", Box::new(second_service_processor));
```
jeking3: So nice to see more language support for cross test like this!

allengeorge: Glad to add it :)

test/known_failures_Linux.json
allengeorge: OK. Will do. Didn't realize there was a defect list.

allengeorge:
@jeking3 Happy to make the change to the travis jobs! Let me modify this PR with the defect change and then work on rebalancing the travis jobs.

test/known_failures_Linux.json
allengeorge: Oh. If you're talking about a JIRA, I've already filed one: [THRIFT-4187](https://issues.apache.org/jira/browse/THRIFT-4187)

allengeorge:
Please pull the latest commit. I updated to fix a bug in my test server.

jeking3:
I can't do any merges at the moment due to a change in employment, and I need to get any legal hurdles out of the way, but one of the other committers should be able to - assuming the new build passes. :)

allengeorge:
@jeking3 One hopes that...congratulations are in order? :)

jeking3:
Essentially same job as before, we were acquired and cut over on May 1.  Thanks are always appreciated in any form though! :)

allengeorge:
I've rebalanced the matrix. Trying again.

@jeking3 Having a weird problem with the c_glib client. It's using the "binary" or "compact" protocol only when invoked in multi/multic mode.

allengeorge:
I had to split up binary/compact protocols because the builds were timing out with them together. And, AFAICT, the unit tests aren't running for builds 1 and 2.

Also (I didn't change this) but I'm unsure of the utility of some of the builds in the matrix:

- TEST_NAME="C C++ C# D Erlang Haxe Go (automake)"
- TEST_NAME="C C++ Plugin Haskell Perl - GCC (automake)"
- TEST_NAME="Java Lua PHP Ruby Dart Node.js Python (automake)"

Looks like these are actually running the unit tests? If so, then why is C/C++ repeated?

jeking3:
One is clang and one is gcc; I recently lowered us from 25 to 19 build jobs and got rid of the ones we didn't need or were redundant.

allengeorge:
Ah. Gotcha. Sorry - I should have realized that. I'll pay closer attention next time.

allengeorge:
OK. I really don't understand `tests.json`. I'm confident that my code is doing the right thing - it's just that the server is invoked in multiplexed mode and the client is invoked in non-multiplexed mode. Super bizarre.

allengeorge:
@Jens-G @jeking3 Can I get any guidance on how the clients/server get their arguments? I'm confident it's not a bug in the rust server. I can invoke the c_glib client and rust server manually on the command line and they both communicate properly. Somehow the c_glib client gets `binary` while the rust server gets `multi` (and the corresponding arguments for compact).

I looked into `crossrunner/collect.py` but the code is a little tough to decipher, and it's not clear how it's determining an intersection.

jeking3:
I believe that is intentional for backwards compatibility.  A multiplexing server should be able to accept an older client and use the "default" service for it - at least in the Java multiplexing server.  So they should end up using the same protocol (binary or compact) but a multi-binary server should be able to service a binary client.

allengeorge:
@jeking3 Oh. Totally didn't realize that - thanks for clarifying! I've updated my `TMultiplexedProcessor` implementation to have a backwards-compatibility mode, and am re-running the build right now.

allengeorge:
@jeking3 Thanks for pointing me in the right direction. Everything now passes!

I realize you're not committing, but would it be possible for one of the other maintainers - @Jens-G perhaps? - to get this change in? To summarize:

* Added rust cross-test and unit-test support to travis
* Added rust multiplexed client/server cross-tests; fixed some bugs in my multiplexing impl.
* Split up test matrix to avoid timeouts
* Discovered bug in Dart framed transport implementation; filed [THRIFT-4187](https://issues.apache.org/jira/browse/THRIFT-4187)

So, with the exception of the two dart tests, every other rust cross-test and all unit tests pass. I've also fully linted and formatted the rust code. Thank you all!

jeking3:
Whomever does commit, there's a comment at the top of the docker files for ubuntu and centos that need a line removed - now the only language not represented in the images is netcore.

allengeorge:
@Jens-G Thank you!

PR #1263, opened 2017-05-08
D lib updated for x64 build mode on windows

Heromyth:
Support for the x64 build mode on windows with DMD 2.074.0 and the latest Libevent

### DIFF
--- a/lib/csharp/src/Transport/TBufferedTransport.cs
+++ b/lib/csharp/src/Transport/TBufferedTransport.cs
@@ -154,8 +154,12 @@ namespace Thrift.Transport
             {
                 if (disposing)
                 {
-                    inputBuffer.Dispose();
-                    outputBuffer.Dispose();
+                    if(inputBuffer != null)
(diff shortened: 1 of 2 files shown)


jeking3:
Looks like this broke unix builds.  As extra credit, if you are able to add D support to the cmake build environment then we can get it to build on appveyor with each pull request, ensuring it won't regress in the future.

Heromyth:
I'll have a try.

jeking3:
This will also need an Apache Jira ticket opened.

jeking3:
@Heromyth any progress on this?

Heromyth:
So sad, there is a bug in phobos64.lib. It will make the examples failed.

jeking3:
This should be rebased on the current master as I have updated the docker files.  Perhaps if you could open a Jira ticket for this issue and rebase on master and force push, we can get a more recent build.

Heromyth:
It's updated by #1380

PR #1264, opened 2017-05-09
THRIFT-4193 Set default max read buffer size in Java server

roshan:
This is for the non-blocking server. This makes it so that just starting up something with all defaults won't result in it being possible to construct an input that causes it to OOME by just filling up the read buffer.


roshan:
It is configurable using the `maxReadBufferBytes` argument, but the default is just really high.

PR #1267, opened 2017-05-12
THRIFT-4196 Support recursive types in Rust

allengeorge:
Client: rs

* Modify code generator to support recursive types
* Modify rust test (not cross-test) client to test recursive calls


allengeorge:
@Jens-G Any chance this could go in? Thank you?

PR #1268, opened 2017-05-12
THRIFT-4190 Improve C# TThreadPoolServer defaults (part 2 of 2)

Jens-G:
Client: C#
Patch: Jens Geyer


Jens-G:
The NPM timed out, no other errors. If there are no objections against it, I'll commit that shortly. 
So please speak now (or be quiet forever, as they say) ... :-)

PR #1269, opened 2017-05-14
THRIFT-4187 Allow dart framed transport to read incomplete frame

allengeorge:


lib/dart/lib/src/transport/t_framed_transport.dart
```diff
@@ -51,33 +58,112 @@ class TFramedTransport extends TBufferedTransport {
...
+
+      _bodySize = 0;
+      _body = null;
+      _receivedBodyBytes = 0;
+
+      _setReadBuffer(body);
+
+      var completer = _frameCompleter;
+      _frameCompleter = null;
+      completer.complete(new Object());
```
markerickson-wf: `completer` is typed as a `Completer<Uint8List>`, so the Dart analyzer complains about passing `Object` here.

lib/dart/lib/src/transport/t_framed_transport.dart
```diff
@@ -51,33 +58,112 @@ class TFramedTransport extends TBufferedTransport {
...
-    int length = buffer.length;
+    if (_frameCompleter == null) {
+      Uint8List buffer = consumeWriteBuffer();
+      int length = buffer.length;
+
+      _headerBytes.buffer.asByteData().setUint32(0, length);
+      _transport.write(_headerBytes, 0, headerByteCount);
+      _transport.write(buffer, 0, length);
+
+      _frameCompleter  = new Completer<Object>(); // FIXME: .sync?!
```
markerickson-wf: Did you mean to address this `FIXME`?

markerickson-wf:
There are a couple of related unit tests that are failing in `t_socket_transport_test.dart`.  You can run `pub run test test/` from the `lib/dart` directory to see those.

markerickson-wf:
@allengeorge I think this approach looks pretty good at my first pass.  I left a couple of comments, and we'll probably want unit tests around the state transitions to ensure that all paths are covered  (setting / unsetting completers, resetting read counters, etc).  I'll also ask couple of other devs I know to review (they have done some recent work using framed transports in Dart).  Thanks for submitting the PR!

lib/dart/lib/src/transport/t_framed_transport.dart
```diff
@@ -51,33 +58,112 @@ class TFramedTransport extends TBufferedTransport {
...
+            TTransportErrorType.UNKNOWN, "Read a negative frame size: $size");
+      }
+
+      _bodySize = size;
+      _body = new Uint8List(_bodySize);
+      _receivedBodyBytes = 0;
+
+      return true;
+    } else {
+      _registerForReadableBytes();
```
markerickson-wf: Does this create a synchronous cycle if the header bytes are not available?  `_registerForReadableBytes`->`_readFrame`->`readFrameHeader`->`_registerForReadableBytes`->...   It seems like this would cause an unresponsive browser?

markerickson-wf: At first I was reading it as causing a cycle, but `_registerForReadableBytes` will make an async call to `_transport.flush`, so I think this is okay on second thought.

lib/dart/lib/src/transport/t_framed_transport.dart
allengeorge: I was a little worried about returning the frame bytes, but I guess it doesn't matter. Also, I assume that each call to `iterator` returns a completely separate iterator with a view to the underlying buffer.

I'll change it. BTW, it's a little weird that `flush()` returns bytes...

lib/dart/lib/src/transport/t_framed_transport.dart
allengeorge: Ah. Actually I need advice from you as to whether this has to be a `sync` `Completer` or not :) It's not quite clear to me when one would choose one, and whether that's necessary at this point.

lib/dart/lib/src/transport/t_framed_transport.dart
allengeorge: It shouldn't cause a cycle, but I may have missed something? It's the first time I've done anything with Dart.

allengeorge:
@markerickson-wf My bad - I should definitely have run the unit tests. I'll check them and fix them up. Also, I'll look into how to create a reasonable set of unit tests for the framed transport.

lib/dart/lib/src/transport/t_framed_transport.dart
markerickson-wf: Yeah, I think it should be okay, sorry for raising the false alarm on my deleted comment.

lib/dart/lib/src/transport/t_framed_transport.dart
markerickson-wf: Yeah, I don't think you need to type `Completer` at all, and you can just `complete()`.  In this case the Future is just a signal of completion, I agree we don't need to return the bytes.

lib/dart/lib/src/transport/t_framed_transport.dart
markerickson-wf: https://api.dartlang.org/stable/1.23.0/dart-async/Completer/Completer.sync.html

Despite the warning above, I used `sync` [in t_http_transport](https://github.com/apache/thrift/blob/master/lib/dart/lib/src/transport/t_http_transport.dart#L52).  If I recall correctly, this is a symptom of adapting the sync API to an async system.  If two subsequent messages arrive on the socket, the second could stomp on the first before the listener has a chance to react.  i.e. `flush`->`complete message 1`->`complete message 2`->`read`->`read`.  By using a sync completer, I think the first listener should be able to read before the second message is completed.

jeking3:
This pull request needs to be completed.  *nudge*

jeking3:
@allengeorge if you add framed transport in dart to the cross test, that would test a lot... but it would not test error conditions such as this one.  Just a reminder, this needs to move forward.

allengeorge:
Hi James - I’m really sorry for the delay; slammed at work. I’ll work on it this weekend.

Again, super sorry about this

jeking3:
Don't worry, I'm just doing my usual "push things forward" bit. :)  I got the backlog down to 52 open PRs, but that's still about 45 too many for my taste.

jeking3:
This needs to be rebased and completed.

jeking3:
This needs to be rebased and completed.

allengeorge:
Yup. Will do. I'm a gonna set myself a target to get it done by end of next week. I've been lax on this :/

allengeorge:
@jeking3 Fixed up the failing tests. Confirming that precross works.

allengeorge:
@jeking3 I've added unit tests around this behavior and enabled more cross tests! Do you want me to roll that into this PR, or should I create another bug and PR?

allengeorge:
@jeking3 Any chance this could be merged? TY!

jeking3:
I'd suggest waiting to see how the CI build fares given a change was made.  So today is a distinct possibility.  Also, you can commit your own changes.  There's no rule against it. :)

markerickson-wf:
+1

allengeorge:
No problem! I'm sorry I was lax and left it so long :/

Looks like the tests passed, so it's good to merge. Not 100% of the process, so I'll follow up via email.

jeking3:
Remember to resolve https://issues.apache.org/jira/browse/THRIFT-4187 as fixed in 0.12.0.

allengeorge:
@jeking3 Yup - just did that. Is there a way to do it automatically, or is it a manual process (for now)?

PR #1271, opened 2017-05-18
THRIFT-4203  server stop gracefully

dongxiaozhen:
make golang thrift  server  stop gracefully


jeking3:
Thanks - adding THRIFT-4203 to the content as it wasn't in the title at creation, so the robots didn't link it to the Jira ticket yet.

PR #1273, opened 2017-05-23
THRIFT-4206: Fix decoding of strings in containers with py:dynamic and py:utf8strings

elprans:
_read_by_ttype and _write_by_ttype must be using the *element* spec
and not the container spec when determining the correct read/write
handler.


jeking3:
Could you rebase on master and refresh the PR so we get a clean build, now that CI is stable again?  Thanks.

elprans:
Rebased.

PR #1274, opened 2017-05-23
THRIFT-4207: Make sure Python Accelerated protocol does not allow invalid UTF-8

elprans:
The accelerated version of Binary and Compact protocol fails to validate
the input string for UTF-8 fields.

### DIFF
--- a/lib/py/src/ext/protocol.tcc
+++ b/lib/py/src/ext/protocol.tcc
@@ -428,16 +428,17 @@ bool ProtocolBase<Impl>::encodeValue(PyObject* value, TType type, PyObject* type
         return false;
       }
     } else {
-      Py_INCREF(value);
       if (isUtf8(typeargs)) {
         if (PyBytes_AsStringAndSize(value, &str, &len) < 0) {
           return false;
         }
         // Check that input is a valid UTF-8 string.
-        if (!PyUnicode_DecodeUTF8(str, len, 0)) {
+        nval.reset(PyUnicode_DecodeUTF8(str, len, 0));
+        if (!nval) {
           return false;
         }
       }
+      Py_INCREF(value);
       nval.reset(value);
     }
 
--- a/lib/py/src/ext/protocol.tcc
+++ b/lib/py/src/ext/protocol.tcc
@@ -419,6 +419,8 @@ bool ProtocolBase<Impl>::encodeValue(PyObject* value, TType type, PyObject* type
 
   case T_STRING: {
     ScopedPyObject nval;
+    Py_ssize_t len;
+    char *str;
 
     if (PyUnicode_Check(value)) {
       nval.reset(PyUnicode_AsUTF8String(value));
@@ -427,10 +429,19 @@ bool ProtocolBase<Impl>::encodeValue(PyObject* value, TType type, PyObject* type
       }
     } else {
       Py_INCREF(value);
+      if (isUtf8(typeargs)) {
+        if (PyBytes_AsStringAndSize(value, &str, &len) < 0) {
+          return false;
+        }
+        // Check that input is a valid UTF-8 string.
+        if (!PyUnicode_DecodeUTF8(str, len, 0)) {
+          return false;
+        }
+      }
       nval.reset(value);
     }
 
-    Py_ssize_t len = PyBytes_Size(nval.get());
+    len = PyBytes_Size(nval.get());
     if (!detail::check_ssize_t_32(len)) {
       return false;
     }
--- a/lib/py/src/protocol/TProtocol.py
+++ b/lib/py/src/protocol/TProtocol.py
@@ -118,6 +118,8 @@ class TProtocolBase(object):
         pass
 
     def writeString(self, str_val):
+        if isinstance(str_val, bytes):
+            str_val = str_val.decode('utf8')
         self.writeBinary(str_to_binary(str_val))
 
     def writeBinary(self, str_val):
--- a/test/py/FastbinaryTest.py
+++ b/test/py/FastbinaryTest.py
@@ -74,6 +74,9 @@ ooe2.zomg_unicode = u"\xd3\x80\xe2\x85\xae\xce\x9d\x20"\
                     u"\x20\xce\x91\x74\x74\xce\xb1\xe2\x85\xbd\xce\xba"\
                     u"\xc7\x83\xe2\x80\xbc"
 
+ooe_bad = OneOfEach()
+ooe_bad.zomg_unicode = b'\xbe\xef\xff'
+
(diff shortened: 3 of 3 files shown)


jeking3:
Could you rebase on master and refresh the PR so we get a clean build, now that CI is stable again?  Thanks.

elprans:
Rebased.

jeking3:
This had some errors and needs more work - please move it forward.

elprans:
Fixed, but travis seems to be unstable again.

jeking3:
Travis is experiencing an outage event w.r.t Docker Hub that started this morning.

jeking3:
I'll pull the changes into my sandbox and run unit, cross, ubsan, and sca jobs.

jeking3:
I pulled these changes into a sandbox and started the `ubuntu-xenial` docker image, and then ran `build/docker/scripts/autotools.sh` which is what Travis CI would do for one of the unit test jobs.  It failed:
```
make[4]: Entering directory '/thrift/src/test/py'
----------------
 Executing individual test scripts with various generated code directories
 Directories to be tested: gen-py-default, gen-py-slots, gen-py-oldstyle, gen-py-no_utf8strings, gen-py-dynamic, gen-py-dynamicslots
 Scripts to be tested: FastbinaryTest.py, TestFrozen.py, TSimpleJSONProtocolTest.py, SerializationTest.py, TestEof.py, TestSyntax.py, TestSocket.py
----------------

Testing script: /usr/bin/python /thrift/src/test/py/FastbinaryTest.py
----
Testing TBinaryAccelerated
  File "/thrift/src/test/py/FastbinaryTest.py", line 278, in <module>
    do_test(TBinaryProtocolAccelerated, TBinaryProtocol)
  File "/thrift/src/test/py/FastbinaryTest.py", line 244, in do_test
    Test(fast, slow).do_test()
  File "/thrift/src/test/py/FastbinaryTest.py", line 198, in do_test
    self._check_write(hm)
  File "/thrift/src/test/py/FastbinaryTest.py", line 137, in _check_write
    o.write(prot_slow)
  File "/thrift/src/test/py/gen-py-default/DebugProtoTest/ttypes.py", line 637, in write
    iter52.write(oprot)
  File "/thrift/src/test/py/gen-py-default/DebugProtoTest/ttypes.py", line 371, in write
    oprot.writeString(self.zomg_unicode.encode('utf-8') if sys.version_info[0] == 2 else self.zomg_unicode)
  File "/thrift/src/lib/py/build/lib.linux-x86_64-2.7/thrift/protocol/TProtocol.py", line 123, in writeString
    self.writeBinary(str_to_binary(str_val))
  File "/thrift/src/lib/py/build/lib.linux-x86_64-2.7/thrift/protocol/TBinaryProtocol.py", line 131, in writeBinary
    self.trans.write(str)
  File "/thrift/src/lib/py/build/lib.linux-x86_64-2.7/thrift/transport/TTransport.py", line 233, in write
    self._buffer.write(buf)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xd7' in position 0: ordinal not in range(128)
*** FAILED ***
LIBDIR: /thrift/src/lib/py/build/lib.linux-x86_64-2.7
PY_GEN: gen-py-default
SCRIPT: FastbinaryTest.py
  File "./RunClientServer.py", line 321, in <module>
    sys.exit(main())
  File "./RunClientServer.py", line 300, in main
    runScriptTest(options.libdir, options.gen_base, genpydir, script)
  File "./RunClientServer.py", line 101, in runScriptTest
    raise Exception("Script subprocess failed, retcode=%d, args: %s" % (ret, ' '.join(script_args)))
Exception: Script subprocess failed, retcode=1, args: /usr/bin/python /thrift/src/test/py/FastbinaryTest.py
FAIL: RunClientServer.py
==================
1 of 1 test failed
==================
```

lib/py/src/protocol/TProtocol.py
```diff
@@ -118,6 +118,8 @@ def writeDouble(self, dub):
         pass
 
     def writeString(self, str_val):
+        if isinstance(str_val, bytes):
+            str_val = str_val.decode('utf8')
```
nsuke: This would break py2

lib/py/src/ext/protocol.tcc
```diff
@@ -419,18 +419,30 @@ bool ProtocolBase<Impl>::encodeValue(PyObject* value, TType type, PyObject* type
...
+      if (isUtf8(typeargs)) {
+        if (PyBytes_AsStringAndSize(value, &str, &len) < 0) {
+          return false;
+        }
+        // Check that input is a valid UTF-8 string.
+        nval.reset(PyUnicode_DecodeUTF8(str, len, 0));
+        if (!nval) {
+          return false;
+        }
+      }
```
nsuke: Doesn't this affect every user's performance who are passing relatively large utf8-encoded `byte` ?
The problem might be that we're not rejecting `byte` in the first place. Although "fixing" that wouldn't be backward compatible.
What do you think ?

lib/py/src/protocol/TProtocol.py
elprans: It does, yes.  And its absence breaks py3.6.  I'm working on a fix for this.

lib/py/src/ext/protocol.tcc
elprans: I see no other way to check the validity without breaking backwards compat.  On the other hand, this removes a real hazard of garbage data being written into unicode fields.

jeking3:
Question: when do we drop python 2.7 support?
Folks using python 2.7 can continue to use existing shipping versions of thrift.

> Python 3.0 was released in 2008. The final 2.x version 2.7 release came out in mid-2010, with a statement of extended support for this end-of-life release. The 2.x branch will see no new major releases after that. 3.x is under active development and has already seen over five years of stable releases, including version 3.3 in 2012, 3.4 in 2014, 3.5 in 2015, and 3.6 in 2016. This means that all recent standard library improvements, for example, are only available by default in Python 3.x.

jeking3:
Many projects are struggling with this.  StackStorm and Ansible are good examples.  We need to continue to support 2.7 for a while longer.  Until we can drop 2.7 support and fully change everything over, we should either close this out or change it so it works.  From what I've seen in the description, it sounds like even six wouldn't be able to fix this.

lib/py/src/protocol/TProtocol.py
jeking3: Can `six` help here?  `six.binary_type` for example.  Or check for python version and use `bytearray` for py2 and bytes for py3?

jeking3:
As I have been the only contributor to the conversation on this pull request for the last year, I'm closing this.  Feel free to re-open if you have updates.

PR #1278, opened 2017-05-25
THRIFT-4211: Fix logging in thrift library

gadLinux:
It seems we were aborting on error. 

We are now doing properly management. Error is ok and working perfectly.


jeking3:
Would you be able to rebase against master and push so there is one commit (just like there is now), but more importantly it will kick off a new CI build for us.  Thanks.

jeking3:
Please rebase on master and push to get a new Travis CI build, now that it has been cleaned up.

PR #1279, opened 2017-05-25
THRIFT-4212: Fix flush on ssl socket thrift

gadLinux:
Fixes the bug in flush on ssl thrift communications transport


jeking3:
When THRIFT-4211 is clean in CI and merged, you will need to rebase this since it includes THRIFT-4211, and then we can get it into CI for a build then merge.  Thanks.

jeking3:
Let's clear up the other two in the pull request backlog, then rebase this and then see what's left.

PR #1283, opened 2017-05-31
THRIFT-4218: Set TCP_NODELAY for PHP client socket

KlausTrainer:
Client: PHP

Until now, the PHP client hadn't set the `TCP_NODELAY` socket option
like e.g. C++, C#, Ruby, or Java.

This enables `TCP_NODELAY` for the client socket.

This closes #4218

### DIFF
--- a/lib/php/lib/Thrift/Transport/TSocket.php
+++ b/lib/php/lib/Thrift/Transport/TSocket.php
@@ -241,6 +241,9 @@ class TSocket extends TTransport
       }
       throw new TException($error);
     }
+
+    $socket = socket_import_stream($this->handle_);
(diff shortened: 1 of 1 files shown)


Jens-G:
Anyone from the PHP experts who can confirm this?

mfrawley:
I'd just like to add that we deployed a similar fix to this on our infrastructure at Jimdo which led to this PR. We have a large PHP codebase which communicates with some services over Thrift. 
 We support ~20 million websites and our logging and metrics indicate fewer timeouts after this change. We had tried several other tweaks previously and this seems to have had a bigger effect than raising the read/write timeouts (at least by a conservative amount).

PR #1284, opened 2017-05-31
THRIFT-4222: Add Golang NewTServerSocketFromAddrTimeout

zwass:
Adds the ability to create a new server socket from a non-TCP (eg. Unix socket)
address. Modeled after the existing NewTSocketFromAddrTimeout function.

### DIFF
--- a/lib/go/thrift/server_socket.go
+++ b/lib/go/thrift/server_socket.go
@@ -47,6 +47,11 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
 	return &TServerSocket{addr: addr, clientTimeout: clientTimeout}, nil
 }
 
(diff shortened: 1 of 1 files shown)


zwass:
Note that this obviates the need for a library like https://github.com/Wang/thrift_unix_domain/blob/master/server_unix_domain.go, which is mostly a copy of the existing implementation with a minor change to support unix domain socket addresses.

zwass:
I don't think any of the test failures are related. Please let me know if this needs anything else to be merged.

Jens-G:
If you could file a JIRA ticket, that would be awesome :-)

zwass:
JIRA ticket filed: https://issues.apache.org/jira/browse/THRIFT-4222

Jens-G:
Ok, I'll look at it.

PR #1285, opened 2017-06-02
THRIFT-4215/4216 Make transport factories return errors

dcelasun:
This commit changes the signature of `TTransportFactory.GetTransport` from
```go
GetTransport(trans TTransport) TTransport
```
to
```go
GetTransport(trans TTransport) (TTransport, error)
```
so the factory can pass any underlying error to the caller (previously such errors were ignored).

This is a backwards incompatible change for anyone implementing custom transports, but it shouldn't effect anyone using the ones in this library.

Fixes [THRIFT-4215](https://issues.apache.org/jira/browse/THRIFT-4215).

Fixes [THRIFT-4216](https://issues.apache.org/jira/browse/THRIFT-4216).

P.S: Some files have formatting-only changes due to `gofmt`.

### DIFF
--- a/lib/go/test/tests/protocols_test.go
+++ b/lib/go/test/tests/protocols_test.go
@@ -42,7 +42,10 @@ func RunSocketTestSuite(t *testing.T, protocolFactory thrift.TProtocolFactory,
 
 	// client
 	var transport thrift.TTransport = thrift.NewTSocketFromAddrTimeout(addr, TIMEOUT)
-	transport = transportFactory.GetTransport(transport)
+	transport, err = transportFactory.GetTransport(transport)
+	if err != nil {
+		t.Fatal(err)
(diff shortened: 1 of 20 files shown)


dcelasun:
Closing PR until all tests are fixed, I don't want to waste Apache's Travis capacity.

peterbourgon:
For the record, this caused build failures in Go kit; I don't imagine my usage is particularly strange, but maybe it is. ([Link](https://github.com/go-kit/kit/commit/4d3c82de4829f7c3752cf66d00d6ecdf913a18f3).)

dcelasun:
Yes sorry about that, this was an expected breakage as the interface has changed. I suggest using `0.10` instead of `master`, that way you won't get any surprises like this until `0.11` is released.

PR #1287, opened 2017-06-10
THRIFT-4219 Refactor Go HTTP Client

dcelasun:
As discussed in THRIFT-4219, this commit removes support for one-off GET
requests, which brings the Go HTTP client inline with implementations in
other languages.

Constructor functions with "Post" in their names are all deprecated
using the proper format [0] as all constructors now have the POST
behaviour.

Fixes THRIFT-4219.

[0] https://github.com/golang/go/issues/10909#issuecomment-136492606

### DIFF
--- a/lib/go/thrift/http_client.go
+++ b/lib/go/thrift/http_client.go
@@ -46,21 +46,14 @@ type THttpClient struct {
 type THttpClientTransportFactory struct {
 	options THttpClientOptions
 	url     string
-	isPost  bool
 }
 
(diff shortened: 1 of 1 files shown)


dcelasun:
Travis failures are unrelated.

dcelasun:
@Jens-G Thoughts?

Jens-G:
If nobody else wants to vote against it, I'll commit this soon.

dcelasun:
Not to rush anyone, but will this be merged before 0.11?

PR #1289, opened 2017-06-14
fix TypeError when a py3 str is passed to the function.

chaoyann:
str of py3 is not instance of bytes, a Error will rise because cannot write a object to buffer

### DIFF
--- a/lib/py/src/protocol/TBinaryProtocol.py
+++ b/lib/py/src/protocol/TBinaryProtocol.py
@@ -127,6 +127,8 @@ class TBinaryProtocol(TProtocolBase):
         self.trans.write(buff)
 
     def writeBinary(self, str):
+        if not isinstance(str, bytes):
+            str = str.encode()
         self.writeI32(len(str))
         self.trans.write(str)
 


jeking3:
Could you open a Jira ticker for this?

Also could you rebase on master and refresh the PR so we get a clean build, now that CI is stable again?  Thanks.

jeking3:
Could you open a Jira ticker for this?

Also could you rebase on master and refresh the PR so we get a clean build, now that CI is stable again? Thanks.

jeking3:
Based on what I've read about py3 strings, and the fact that thrift strings are designated to be encoded as UTF8, this change seems correct.  I'm not so sure about what happens on the other side (decoding) however my guess is ti is okay.

@nsuke since you did a bunch of work on the python TBinaryProtocol could you look at this?

nsuke:
It's more of a usability problem.

The problem is that the user is passing a non-`byte` instance to `writeBinary`. It's basically the same error as when you pass str `'42'` (not integer `42`) to writeI32.

We can put this kind of conversion to any `writeXxx` functions including `writeBinary`. A downside would be potential surprises when users didn't actually intend to do that.

jeking3:
@nsuke it sounds like you would recommend that we do not accept this fix, and have people do their own conversion before trying to add a string, to make sure it is a byte string (UTF-8 encoded).  Correct?

nsuke:
Basically yes with some note.

It's typically not a UTF-8 encoded binary. To write utf8 strings we can use `writeString`. The point is that `writeBinary` is for writing binaries.

The real problem here is that py2 `str` *is* `byte` but py3 `str` is not. This forces many python users to put this kind of conversions everywhere when migration from 2 to 3. Putting this in Apache Thrift will save their time, while introducing potential surprises to another set of users.

jeking3:
@nsuke there must be some common wisdom and/or tooling to facilitate the conversion from py2 to py3 in this regard - what have other projects done when facing this issue?

jeking3:
Many projects are struggling with this. StackStorm and Ansible are good examples. We need to continue to support 2.7 for a while longer. Until we can drop 2.7 support and fully change everything over, we should either close this out or change it so it works. From what I've seen in the description, it sounds like even six wouldn't be able to fix this.

jeking3:
As I have been the only contributor to the conversation on this pull request for the last year, I'm closing this.  Feel free to re-open if you have updates.

PR #1290, opened 2017-06-16
Update TByteBuffer.java

gx9702:
fix buffer read bug


thomaslee:
@jeking gah, sorry -- believe this was lifted from another code base, possible we fixed this downstream without pushing it back up after the original PR landed. @gx9702 I can go through the full HtC process if you don't have time since I'm the dimwit that messed this up first time around. Let me know either way.

jeking3:
Normally I would ask for a new Jira ticket to be filed by the submitter, but this is pretty obvious and pretty serious, so I did it:

THRIFT-4316

PR #1292, opened 2017-06-19
Using PHP56 const property for contant variables.

Skarlso:
Hi nice folks of Thrift. 

So, before I start trying to figure out how to unit test this, if there is a way to test it, I would like to have some opinions if this a right approach here.

Cheers,
Gergely.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_php_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_php_generator.cc
@@ -63,6 +63,7 @@ public:
     json_serializable_ = false;
     nsglobal_ = ""; // by default global namespace is empty
     psr4_ = false;
+    php56_ = false;
     for (iter = parsed_options.begin(); iter != parsed_options.end(); ++iter) {
       if (iter->first.compare("inlined") == 0) {
         binary_inline_ = true;
@@ -80,6 +81,8 @@ public:
         nsglobal_ = iter->second;
       } else if (iter->first.compare("psr4") == 0) {
         psr4_ = true;
+      } else if (iter->first.compare("php56") == 0) {
+        php56_ = true;
       } else {
         throw "unknown option php:" + iter->first;
       }
@@ -366,6 +369,11 @@ private:
   bool psr4_;
 
   /**
+   * Whether to use PHP56 features like CONST keyword for constant variables
+   */
+  bool php56_;
+
+  /**
    * Whether to generate validator code
    */
   bool validate_;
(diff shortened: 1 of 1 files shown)


Skarlso:
Ruby failure seems unrelated.

```bash
handler000:New TestHandler instance created
Starting the TSimpleServer on port 33908 with framed transport with encryption with json protocol...
Thrift.Transport.TTransportException: System.IO.IOException: EndWrite failure ---> System.Net.Sockets.SocketException: The socket has been shut down
```

compiler/cpp/src/thrift/generate/t_php_generator.cc
```diff
@@ -545,30 +553,41 @@ void t_php_generator::generate_consts(vector<t_const*> consts) {
...
+      for (c_iter = consts.begin(); c_iter != consts.end(); ++c_iter) {
+        string name = (*c_iter)->get_name();
+
+        indent(f_consts) << "public CONST $" << name << " = ";
+        generate_php_doc(f_consts, *c_iter);
+        f_consts << render_const_value((*c_iter)->get_type(), (*c_iter)->get_value());
+        f_consts << ";" << endl;
+      }
+    } else {
+      // Create static property
```
glennpratt: Is it worth maintaining the static property for backwards compatibility. (5.6+ users could opt into the const without having to update all callers).

Skarlso: I think if you are explicity committing to use php56, you ought to go all the way, not? I mean, once you set php56 you are expecting things, and I wouldn't want to keep some old stuff lying around lest I might spot improvement possibilities.

Jens-G:
1) We have a [nice, short & concise contrib guide](http://thrift.apache.org/docs/HowToContribute) that covers all the formalities

2) We also have a [test suite](https://github.com/apache/thrift/tree/master/test) that may be helpful with testing this. It is also run through the pre-commit checks. If you feel additional specific tests should be added, these are usually placed somewhere under [`/lib/php`](https://github.com/apache/thrift/tree/master/lib/php/test)

Skarlso:
@Jens-G Hey Jens. So I read the guide but it does tell me how to run the PHP tests for example. I did try all the things with make, but it still ran everything else as well.

Also, running `make cross` in test folder, like the readme is saying, says there is no target called cross.

```
❯ make cross
make: *** No rule to make target `cross'.  Stop.
```

Skarlso:
So I run test.py. It created over 50 running python, 40 something node and over 60 TestThrift processes, which killed my machine.

jeking3:
If you are using autoconf (bootstrap and configure) then you want to do a ``make -j4`` at the top to ensure compiler and libraries are built, then go into ``lib/py``, ``lib/py.tornado``, and ``lib/py.twisted`` and run ``make check -j4`` to run all the unit tests.  Then from the top level directory you can do "make precross" to get ready for a cross test, and then you can run ``$ THRIFT_CROSSTEST_CONCURRENCY=4 test/test.py -s --client python,py3,cpp,java --server python,py3,cpp,java`` to run cross tests against only those languages - you can be even more restrictive as once you rebase on master, squash and force push then the current Travis CI and Appveyor builds will run the entire test suite.

PR #1293, opened 2017-06-21
THRIFT-2642 Recursive structs don't work in python

econner:
See https://issues.apache.org/jira/browse/THRIFT-2642.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -150,7 +150,7 @@ public:
   void generate_enum(t_enum* tenum);
   void generate_const(t_const* tconst);
   void generate_struct(t_struct* tstruct);
-  void generate_forward_declaration(t_struct* tstruct) override;
+  void generate_forward_declaration(t_struct* tstruct);
   void generate_xception(t_struct* txception);
   void generate_service(t_service* tservice);
 
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -150,6 +150,7 @@ public:
   void generate_enum(t_enum* tenum);
   void generate_const(t_const* tconst);
   void generate_struct(t_struct* tstruct);
+  void generate_forward_declaration(t_struct* tstruct) override;
   void generate_xception(t_struct* txception);
   void generate_service(t_service* tservice);
 
@@ -160,6 +161,7 @@ public:
    */
 
   void generate_py_struct(t_struct* tstruct, bool is_exception);
+  void generate_py_thrift_spec(std::ofstream& out, t_struct* tstruct, bool is_exception);
   void generate_py_struct_definition(std::ofstream& out,
                                      t_struct* tstruct,
                                      bool is_xception = false);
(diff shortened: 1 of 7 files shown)


econner:
As written this will not work.  The spec for a Struct needs to be a reference type such as a list for recursion to work correctly so there is more work to be done here.

econner:
Ok, I've updated the PR to use a list (reference type) instead of a tuple.  I also added some serialization tests for recursive structures. Hopefully everything works now.

lib/py/src/TRecursive.py
```diff
@@ -0,0 +1,63 @@
+# MODIFIED June 20, 2017, Eric Conner
+#
+#
+# Original source copyright 2014-present Facebook, Inc.
+#
```
Jens-G: Where does line 1 to 5 come from? That's not part of the standard ASF license header.

econner: Yea, I'm not very experienced with the whole issue of licensing.  I mainly put these lines because of the fbthrift license and I used this file more or less verbatim from fbthrift.  (https://github.com/facebook/fbthrift/blob/master/LICENSE#L97) (https://github.com/facebook/fbthrift/blob/master/thrift/lib/py/util/Recursive.py)

>       (b) You must cause any modified files to carry prominent notices
          stating that You changed the files; and

      (c) You must retain, in the Source form of any Derivative Works
          that You distribute, all copyright, patent, trademark, and
          attribution notices from the Source form of the Work,
          excluding those notices that do not pertain to any part of
          the Derivative Works; and

But perhaps it is fine to use verbatim since fbthrift is itself a branch of thrift?  Not really sure what the protocol here is...

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -610,11 +623,22 @@ string t_py_generator::render_const_value(t_type* type, t_const_value* value) {
   return out.str();
 }
 
+/** 
+ * Generates the "forward declarations" for python structs.
+ * These are actually full class definitions so that calls to generate_struct
+ * can add the thrift_spec field.  This is needed so that all thrift_spec 
+ * definitions are grouped at the end of the file to enable co-recursive structs.
+ */
+void t_py_generator::generate_forward_declaration(t_struct* tstruct) {
```
econner: This is an interface method that's called by the thrift compiler.  One attribute of this change is to move most of the code that generates python code into ``generate_forward_declaration`` and then output the ``thrift_spec`` via the ``generate_struct`` method.

lib/py/src/ext/types.cpp
```diff
@@ -20,6 +20,8 @@
 #include "ext/types.h"
 #include "ext/protocol.h"
 
+#include <iostream>
```
econner: Mm I believe this was a holdover from testing & using print statements.  It should be safe to remove.  Thanks for catching.

test/py/SerializationTest.py
```diff
@@ -285,6 +290,67 @@ def testIntegerLimits(self):
         for value in bad_values:
             self.assertRaises(Exception, self._serialize, value)
 
+    def testRecTree(self):
+        """Ensure recursive tree node can be created."""
+        children = []
+        for idx in range(1, 5):
```
econner: This is just a test to ensure you can make a tree with some children.  There is no limit on the number of recursions (I guess besides the maximum allowed by python).

PR #1294, opened 2017-06-23
THRIFT-4233 Make THsHaServer.invoker available (get method only) in inherited classes

dmvolod:


dmvolod:
Is there any chance to review this changes?

lib/java/src/org/apache/thrift/server/THsHaServer.java
```diff
@@ -154,7 +154,10 @@ protected static ExecutorService createInvokerPool(Args options) {
 
     return invoker;
   }
-
+  
+  protected ExecutorService getInvoker() {
```
jeking3: This could probably be static, given the method above is also static?  Thoughts?

dmvolod: Yes, you are right. I will close this PR, as I'm unable to add commit to this PR (remove remote repo) and create a new one.
Thank you.

PR #1296, opened 2017-06-25
THRIFT-4234 Travis build fails cross language tests with "Unsupported security protocol type"

econner:
Fix is to upgrade Mono to latest version.

See https://issues.apache.org/jira/browse/THRIFT-4234

This along with https://github.com/apache/thrift/pull/1295 should get the build green again.


Jens-G:
Nice, but not 100%. Still two are saying ..

> npm install
> /bin/bash: npm: command not found
>

PR #1297, opened 2017-06-26
handleRuntimeExceptions

zhouxianjun:
RT:
thrift command: 0.10.0
thrift java lib: 0.10.0
Method does not override method from its superclass
org.apache.thrift.ProcessFunction is no handleRuntimeExceptions method


Jens-G:
Something is wrong here. Could you please review your pull request?

zhouxianjun:
I was wrong, I compiled the master, but I used 0.10.0, I'm sorry.

PR #1298, opened 2017-06-27
THRIFT-4236: Add context support for go server.

taozle:
This commit add the context support for thrift server side in go.
for compatibility, added a use_context option in go's generator option
to tell compiler whether generated code should support context.
Added TProcessor2, TProcessorFactory2, TServer2, TMultiplexedProcessor2
to support ctx as first param.

See: https://issues.apache.org/jira/browse/THRIFT-4236


Jens-G:
I look at it.

PR #1300, opened 2017-06-27
THRIFT-4237 Fix data races in Go TServerSocket

zwass:
Incorrect lock semantics resulted in a race condition in the TServerSocket
implementation. This commit adds a unit test that detects this, and fixes up
the locking semantics.

https://issues.apache.org/jira/browse/THRIFT-4237

### DIFF
--- a/lib/go/Makefile.am
+++ b/lib/go/Makefile.am
@@ -31,7 +31,7 @@ install:
 	@echo '##############################################################'
 
 check-local:
-	$(GO) test ./thrift
+	$(GO) test -race ./thrift
 
 all-local:
 	$(GO) build ./thrift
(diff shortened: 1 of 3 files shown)


zwass:
@Jens-G it appears I've introduced some issue that causes the tests to hang. Is there a good way to run just the Go tests in verbose mode? I've connected directly to a shell in the docker build container but haven't yet found a way.

econner:
Does running ``make check`` from ``/thrift/src/test/go`` not do what you want?

zwass:
```
$ pwd
/Users/zwass/dev/go/src/git.apache.org/thrift.git/lib/go
$ make check
 cd ../.. && /bin/bash /thrift/src/missing automake-1.14 --foreign lib/go/Makefile
/bin/bash: /thrift/src/missing: No such file or directory
make: *** [Makefile.in] Error 127
```

PR #1301, opened 2017-06-27
THRIFT-4240 Fix use of sync.Once in Go TSimpleServer Stop() method

zwass:
A package global sync.Once was used, which ensured that Stop() could only be
called once *per process*. The correct semantics should be once *per server*.

This commit makes the sync.Once a member of the server struct so that the
correct semantics are applied.

https://issues.apache.org/jira/browse/THRIFT-4240


zwass:
If https://github.com/apache/thrift/pull/1302 is merged, this PR can be closed.

PR #1302, opened 2017-06-30
THRIFT-4243 Fix Go TSimpleServer race on wait in Stop() method

zwass:
Fixes the synchronization issues in the existing Stop() method, by ensuring no
further Add calls are made to the WaitGroup at shutdown. See
https://issues.apache.org/jira/browse/THRIFT-4243 for discussion.

This also fixes https://issues.apache.org/jira/browse/THRIFT-4240.

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -32,6 +32,7 @@ import (
  */
 type TSimpleServer struct {
 	quit chan struct{}
+	once sync.Once
 
 	processorFactory       TProcessorFactory
 	serverTransport        TServerTransport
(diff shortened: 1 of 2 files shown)


econner:
Hi @zwass,

It seems one of your recent changes has broken the build. Can you please take a look?
https://travis-ci.org/apache/thrift/builds/249109561

Thanks.

zwass:
Hi @econner,

I believe the issue is in https://github.com/apache/thrift/pull/1300. I have been trying to get assistance with running the tests locally so I can debug this issue since before it was merged. Any help would be greatly appreciated.

PR #1303, opened 2017-07-05
fix framed_transport memory increase when write transport failed

pirDOL:
if p.transport.Write fails, p.buf will not be truncated, which leads to unexpected increasing of thrift client's memory.


Jens-G:
Tracked in THRIFT-4245

PR #1306, opened 2017-07-11
Fix Nodejs generation to auto deal with args class instantiation

wenshin:
I found there is a logic for auto dealing with method args class instantiation, but not using


jeking3:
@wenshin first of all my apologies for having no comments or interactions since you submitted the PR a month ago.  As you may have seen we had some CI build issues which are now resolved, and your PR has a conflict.  Could you rebase this on the current master and push to refresh this PR and kick a build?

Also, changes need Jira tickets.  See:

wenshin:
@jeking3 Thanks for your nice reply, it's duplicate of #1034, I'll close this one.

PR #1307, opened 2017-07-13
THRIFT-4248: Import cstring in TSSLSocket

sokac:
strncpy, memcmp, memset are used in TSSLSocket so cstring needs to be
imported.


sokac:
Correct, <string> is not the same as <string.h>. However, <cstring> (C++) is very similar to <string.h> (C), with only difference that cstring is part of std and usually preferred for C++

https://stackoverflow.com/questions/8380805/difference-between-string-h-and-cstring

PR #1309, opened 2017-07-17
Use build tags to support context.

taozle:
As the comment in https://issues.apache.org/jira/browse/THRIFT-4236, this PR use build tags to support context in go1.7 above.


compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -3746,4 +3746,4 @@ THRIFT_REGISTER_GENERATOR(go, "Go",
                           "    read_write_private\n"
                           "                     Make read/write methods private, default is public Read/Write\n" \
                           "    use_context\n"
-                          "                     Make service method receive a context as first argument.\n")
+                          "                     Make service method receive a context as first argument, only go1.7+ is supported.\n")
```
dcelasun: Can we support `x/net/context` for go<1.7?

lib/go/thrift/simple_server2.go
```diff
@@ -1,180 +0,0 @@
...
-	"log"
-	"runtime/debug"
-	"sync"
-)
-
-/*
- * This is only a simple sample same as TSimpleServer but add context
- * usage support.
- */
-type TSimpleServer2 struct {
```
dcelasun: Why was this removed?

lib/go/thrift/multiplexed_processor.go
```diff
@@ -1,3 +1,5 @@
+// +build go1.7
```
dcelasun: Maybe this file should be named `multiplexed_protocol_go17.go` or something like that to be consistent with `TMultiplexedProcessor` in `multiplexed_protocol.go`?

compiler/cpp/src/thrift/generate/t_go_generator.cc
taozle: i noticed that thrift has no dependency, so i just followed this.

lib/go/thrift/simple_server2.go
taozle: This is just a sample copied from TSimpleServer with little difference, and it is hard to maintain two files with almost the same code.

lib/go/thrift/multiplexed_processor.go
taozle: Agreed, i will fix it.

compiler/cpp/src/thrift/generate/t_go_generator.cc
dcelasun: You do have an explicit dependency to the new `context` package [here](https://github.com/taozle/thrift/blob/0df5965eb2a35793f19de1b2696e3c4afb6b5132/compiler/cpp/src/thrift/generate/t_go_generator.cc#L887). What I'm suggesting is to use `x/net/context` there if it's Go<1.7, so you don't need `use_context` at all.

lib/go/thrift/simple_server2.go
dcelasun: Ok, but as I said in the other comment, if we use `x/net/context` for Go<1.7, then we can have a single `TSimpleServer` with context support for all Go versions.

compiler/cpp/src/thrift/generate/t_go_generator.cc
taozle: Sorry for my poor expression,  the *no dependency* i mean is that *no thirdparty dependency*. `use_context` can't be removed because adding context support breaks the backward compatibility.

dcelasun: Breaking BC can be ok (happened before, see my commits) but if we remove `use_context` we can make sure all RPCs will have the same signature starting with `(ctx *Context, ...`.

To be more clear: I think adding context support is *very* worthwhile and it's worth breaking BC for it (especially for a new release), but it would be better to support `context` for all Go versions, not just Go>1.7

taozle: OK, i am glad to remove the `use_context`, but since then we have either to commit the `x/net/context` in vendor or keep this dep info using package manage tool (glide, govendor ...), i prefer commit the `x/net/context` into vendor since there is no widely adopted package manage tool by now, any advise?

dcelasun: I don't think we should commit it. Right now, we don't have a `vendor` folder [in the library](https://github.com/apache/thrift/tree/master/lib/go). When people use `go get` (or other dependency tools) they'll get the `x/net/context` dependency automatically (if they are on Go<1.7).

What do you think?

taozle: And its also painful to use build tags in the generated code, i think a control flag is still required to decided whether to use `context` or `x/net/context`.

dcelasun: > i think a control flag is still required to decided whether to use `context` or `x/net/context`.

Hmm, I think you are right. Maybe something like `legacy_context`?

taozle: Sounds good, i will update this PR later today.

dcelasun: Great, thanks for working on this!

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -3745,5 +3720,5 @@ THRIFT_REGISTER_GENERATOR(go, "Go",
                           "                     Disable automatic spelling correction of initialisms (e.g. \"URL\")\n" \
                           "    read_write_private\n"
                           "                     Make read/write methods private, default is public Read/Write\n" \
-                          "    use_context\n"
-                          "                     Make service method receive a context as first argument.\n")
+                          "    legacy_context\n"
+                          "                     Use lagacy x/net/context instead of context in go<1.7.\n")
```
dcelasun: Typo s/lagacy/legacy

taozle:
Seems that the failed tests has nothing with this PR.

dcelasun:
Yep, they are unrelated. Thanks for working on this!

@Jens-G what do you think? With this PR, we'll have `context` support for pre-1.7 Go versions as well.

I'll also be sending a PR to update the Go tutorial with the new handler signature.

Jens-G:
> I'll also be sending a PR to update the Go tutorial with the new handler signature.

Please do. Right now the tutorial does not build.

```
handler.go:30:2: cannot find package "golang.org/x/net/context" in any of:
        C:\Go\src\golang.org\x\net\context (from $GOROOT)
        D:\Workdir\src\golang.org\x\net\context (from $GOPATH)
```

dcelasun:
Hmm, turns out @taozle already [updated](https://github.com/apache/thrift/blob/c0d384a38c2b43ee47cef86b1cd054e3f84dc909/tutorial/go/Makefile.am#L37) the Makefile for the tutorial, I just missed it somehow. The `check` target already includes `thirdparty-dep`, so it should work with `make -k check` when you are running Go<1.7. What were your exact steps that resulted in this error? What's your Go version?

Jens-G:
go version go1.8.3 windows/amd64

dcelasun:
Ah, now I see it. The Makefile for the tutorial has the `legacy_context` flag hardcoded, since it's pinned to the Go version included in the Docker container. If you are running it outside the container (and hence with Go 1.7+) you need to remove `legacy_context` and `thirdparty-dep` from the Makefile.

@taozle Maybe there is a way to check the go version before adding the flag?

taozle:
I will try to fix that.

PR #1310, opened 2017-07-19
Better types for callbacks when using typescript

mikemorton:
This adds more specific types to the async callback for client-side calls in the thrift generated .d.ts files.


mikemorton:
I've created THRIFT-4257 on JIRA for this: https://issues.apache.org/jira/browse/THRIFT-4257

PR #1311, opened 2017-07-20
Fix for constant assignments to optional fields in Go.

davinchia:
As per ticket https://issues.apache.org/jira/browse/THRIFT-4253.

As I dug through the code I realised the problem was not only limited to optional strings, and that the generator was not properly generating optionals for all types.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -438,8 +438,17 @@ string t_go_generator::gen_opt_const_values_(t_type* type, t_base_type::t_base t
     case t_base_type::TYPE_I16:
     case t_base_type::TYPE_I32:
     case t_base_type::TYPE_I64:
-      go_value_str += "int}{";
+      go_value_str += "int32}{";
       go_value_str += std::to_string(value->get_integer());
+      break;
+
+    case t_base_type::TYPE_DOUBLE:
+      go_value_str += "float64}{";
+      if (value->get_type() == t_const_value::CV_INTEGER) {
+        go_value_str +=  std::to_string(value->get_integer());
+      } else {
+        go_value_str +=  std::to_string(value->get_double());
+      }
       break;
 
     default:
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -435,10 +435,18 @@ string t_go_generator::gen_opt_const_values_(t_type* type, t_base_type::t_base t
       break;
 
     case t_base_type::TYPE_I8:
+      go_value_str += "int8}{";
+      go_value_str += std::to_string(value->get_integer());
     case t_base_type::TYPE_I16:
+      go_value_str += "int16}{";
+      go_value_str += std::to_string(value->get_integer());
+      break;
     case t_base_type::TYPE_I32:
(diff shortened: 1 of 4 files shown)


davinchia:
https://stackoverflow.com/questions/30716354/how-do-i-do-a-literal-int64-in-go is a good summary of the different ways to assign constant values in Go.

I decided to use an anonymous struct since it seems to be the most self contained and efficient way to do so.

davinchia:
@Jens-G 
This is the first time I'm writing production C++ code so please feel free to tear this apart.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -424,6 +424,52 @@ bool t_go_generator::is_pointer_field(t_field* tfield, bool in_container_value)
...
+  switch (tbase) {
+    case t_base_type::TYPE_BOOL:
+      go_value_str += "bool}{";
+      go_value_str += (value->get_integer() > 0 ? "true" : "false");
+      break;
+
+    case t_base_type::TYPE_I8:
+      go_value_str += "int8}{";
+      go_value_str += std::to_string(static_cast<long long>(value->get_integer()));
+      break;
```
Jens-G: Three questions:
a) Why are these casts needed?
b) Can we move that whole method next to the other method? 
c) Why do we need two separate methods at all?

davinchia: a) The casts are to avoid the `ambiguous calls` errors I was seeing on the integration tests.
b) c) I separated the methods as I had mini unit tests on my machine. I left it this way since I felt the main method was getting slightly too long. Combining both methods will remove the casting problem and I have no strong opinions so I will do that instead.

Jens-G:
Ping

davinchia:
@Jens-G Github must have hidden my comments. My latest commit combined the functions and removed the toString function as per your suggestions.

Not sure why one of the integration tests are failing. Logs show `/usr/include/php5/Zend/zend_hash.h:266:2:` is the cause, which is not something I modified.

jeking3:
@davinchia we've cleared out the CI build issues, so if you could squash your changes to one commit then rebase on master, then do a force push to refresh this pull request, it'll kick off a new build.  Squashing will make it easier to apply the patch when the build completes successfully.

jeking3:
@davinchia we've cleared out the CI build issues, so if you could squash your changes to one commit then rebase on master, then do a force push to refresh this pull request, it'll kick off a new build. Squashing will make it easier to apply the patch when the build completes successfully.

jeking3:
@davinchia we've cleared out the CI build issues, so if you could squash your changes to one commit then rebase on master, then do a force push to refresh this pull request, it'll kick off a new build. Squashing will make it easier to apply the patch when the build completes successfully.

jeking3:
Looks like the build was pretty much clean.  Any objections to a squash and merge?

davinchia:
No objections. Happy to squash and merge!

davinchia:
I don't have write access so I can't do this.

jeking3:
No problem, now that we're on github as master I can squash and merge this with one click.

davinchia:
Thanks @jeking3! The only sad bit about this is since you merged this in, I don't get cred on the contributors page - dang!

jeking3:
You are the commit author, that shows up in "git log".

PR #1313, opened 2017-07-25
THRIFT-4251 Fix JDK Epoll Bug in Thrift of TThreadedSelectorServer model.

Johnny-Liao:
Such as the title. More information see : https://issues.apache.org/jira/browse/THRIFT-4251

### DIFF
--- a/lib/java/src/org/apache/thrift/server/TThreadedSelectorServer.java
+++ b/lib/java/src/org/apache/thrift/server/TThreadedSelectorServer.java
@@ -19,7 +19,15 @@
 
 package org.apache.thrift.server;
 
+import org.apache.thrift.transport.TNonblockingServerTransport;
+import org.apache.thrift.transport.TNonblockingTransport;
+import org.apache.thrift.transport.TTransportException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
+import java.nio.channels.ClosedChannelException;
+import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.nio.channels.spi.SelectorProvider;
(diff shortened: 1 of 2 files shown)


jeking3:
According to the documentation of the defect that triggered this (http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6670302), it indicates this only occurs in kernel 2.4 series, and if you move to kernel 2.6 this issue goes away.  Given this is an operating system defect, I'm inclined to recommend we do not merge this and resolve it as won't fix, because it isn't a defect in Thrift or in Java, but rather in the operating system it is running on.

Johnny-Liao:
My linux kernel is 3.10.0 when this bug occured. And I'm pretty sure that still trouble many people who dependence Thrift to implements theirs server program. Netty uses this approach to mask the underlying problem and why can't we.

jeking3:
There are conflicting reports to what the root cause of this issue really is.  I found:

http://bugs.java.com/view_bug.do?bug_id=6693490 fixed in Java 7b55
http://bugs.java.com/view_bug.do?bug_id=6897993 possibly caused by above, fixed in Java 7b77
http://bugs.java.com/view_bug.do?bug_id=6670302 bug in linux 2.4 kernel fixed in 2.6

Given you reproduced it under Java 8b131 then it sounds like it would be a different issue to what's already been reported, or a regression in one of them.  Folks seem to think this issue was resolved.  I actually found a number of projects that implemented their own solution, but they were all slightly different.

Have you looked into reopening 6670302 with a sample program that recreates it under Java 8b131?   Or, does the sample app used in that defect reproduce on your system possibly?

Johnny-Liao:
I found a report maybe more relevant with this issue. 
See : [http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6403933](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6403933)
It gives a solution what is creates a new Selector to replace the old one.

jeking3:
However, as with all the other issues, they were already resolved, so the fact you are still seeing this with a recent kernel and java version means there's still a bug that needs to be fixed somewhere.  There shouldn't be a workaround in every Java program that uses a Selector().  I would still encourage you to come up with a simple example that demonstrates it without Thrift involved and report it upstream.

Johnny-Liao:
It's not elegant way to rebuild Selector in every Java program, but it's a effective way to fix this bug.  And so many year passed it not really fixed in JDK or kernel. This is one reason for some Frameworks implemented this solution.  After all, the consequences of this bug are too serious.

xiaohu-zhang:
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6403933

xiaohu-zhang:
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6403933
it sounds like have been resolved yet in jdk7

jeking3:
I'd like additional feedback from folks who use Thrift with Java on this.  It seems like the wrong place to fix the issue, but it sounds like many projects are implementing their own fix for it.  I've never seen it happen.

jeking3:
So I asked on the mailing list and it's been crickets, so I'm going to merge this.  It passed all the tests.

xiaohu-zhang:
MY english is poor. could you please explain what is the it's been crickets mean ? thank you very much!

jeking3:
It means nobody replied to my requests for comments, so I assumed nobody
has an issue or nobody cares; whatever the case, I committed it to the
project. :)

xiaohu-zhang:
Thank you ! honestly speaking , netty and lots of framework implementing this fix. I also doubt
It is useful in jdk7. I agree with you ,But if I was you , I will not change the original code,unless give
the unit test to prove it happend in jdk7

lib/java/src/org/apache/thrift/server/TThreadedSelectorServer.java
```diff
@@ -596,6 +601,77 @@ private void select() {
...
+
+      if (selectedNums == 0) {
+        jvmBug++;
+      } else {
+        jvmBug = 0;
+      }
+
+      long selectedTime = afterSelect - beforeSelect;
+      if (selectedTime >= MONITOR_PERIOD) {
+        jvmBug = 0;
```
jbescos: This should be jvmBug++

PR #1314, opened 2017-07-26
THRIFT-4263: Fix use after free bug for thrown exceptions

norrs:
Exceptions thrown through PHPExceptionWrapper are prematurely freed at the end
of the catch block, even though zend_throw_exception_object expects to take
ownership of the value.

Ensure we free return_value in case of exceptions

Test binary deserialization of insufficient data which verifies we can cast
exception to string to verify against memory corruption when transport casts
exceptions.

This closes #4263

### DIFF
--- a/lib/php/src/ext/thrift_protocol/php_thrift_protocol7.cpp
+++ b/lib/php/src/ext/thrift_protocol/php_thrift_protocol7.cpp
@@ -995,7 +995,10 @@ PHP_FUNCTION(thrift_protocol_write_binary) {
     transport.flush();
 
   } catch (const PHPExceptionWrapper& ex) {
-    zend_throw_exception_object(ex);
+    // ex will be destructed, so copy to a zval that zend_throw_exception_object can take ownership of
+    zval myex;
+    ZVAL_COPY(&myex, ex);
+    zend_throw_exception_object(&myex);
     RETURN_NULL();
   } catch (const std::exception& ex) {
     throw_zend_exception_from_std_exception(ex);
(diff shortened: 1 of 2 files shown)


norrs:
Side story:

Ok, I was hoping travis would run the cross compile tests with php7, but sadly it doesn't. As I was hoping the build would fail with the test case before submitting, as this is what we've used for testing it under valgrind.


Currently cross compile builds runs with ubuntu 1404 trusty. For thrift as a project you might consider trying to maybe bump this dependency, as maintenance updates for 1404 is more or less going away this year according to https://www.ubuntu.com/info/release-end-of-life . 

You will also need min 1604 for php7 packages. Sadly the test suite really doesn't run with newer versions, at least python tests seems to start failing. Another problem is the system tests with (cross-compile) doesn't use cmake yet, which seems to have an idea about being able to select which libraries you want to include in the build. 

Also note that PHP5 != PHP7 , and you probably want to focus new development towards PHP7. http://php.net/supported-versions.php . 

Ideally CI should maybe build with latest distros where developers hang and live :-)

Anyhow, me and @zhaakhi rolled out this in production for our systems and haven't noticed any segfaults since we deployed it 24 hours ago. But we did revert https://github.com/apache/thrift/pull/1215 in our fork that we have deployed, since we noticed memory leaks from it. Unsure if it was due to me  not ensuring newer thrift compiler under testing, but didn't want to risk it. I've also left a note about it in https://issues.apache.org/jira/browse/THRIFT-4126 .

norrs:
Test failures seems unrelated, how do we retrigger em to be run again?

PR #1316, opened 2017-07-27
THRIFT-4266: Erlang skip fix

dhull:
The Erlang library has a bug that causes it to crash when attempting to skip over a struct that it receives that does not exist in its thrift definition. This will happen when the client and server are using different versions of the thrift definition where one of the versions adds a struct field to an existing struct.

This pull request has two commits. The first commit adds a test that fails with the existing Erlang library. The second commit fixes the bug.

The bug was that the `erlang_protocol:skip` function expects an atom as its second (Type) argument, but the calls in `skip_struct_loop`, `skip_map_loop`, `skip_set_loop`, and `skip_list_loop` did not convert from the integer type decoded from the buffer to the corresponding atom.


Jens-G:
Anyone amongst the fluently Erlang speaking people who wants to review this?

lib/erl/test/test_omit.erl
```diff
@@ -0,0 +1,82 @@
...
+
+  {P1, ok} = thrift_protocol:write(P0, {{struct, {thrift_omit_with_types, element(1, A)}}, A}),
+  {P2, {ok, O0}} = thrift_protocol:read(P1, {struct, {thrift_omit_without_types, element(1, A)}}),
+  ?assertEqual(element(1, A), element(1, O0)),
+  ?assertEqual(element(2, A), element(2, O0)),
+  ?assertEqual(element(4, A), element(3, O0)),
+
+  {P3, ok} = thrift_protocol:write(P2, {{struct, {thrift_omit_with_types, element(1, B)}}, B}),
+  {_P4, {ok, O1}} = thrift_protocol:read(P3, {struct, {thrift_omit_without_types, element(1, A)}}),
+  %% io:format(user, "~p ~p\n", [ A, O1 ]),
```
djnym: Probably don't need this commented out line.

djnym: And there are about 3-4 other commented out lines which can probably be removed.

djnym:
@Jens-G I gave it a quick review and it looks to do what it claims.  I'm not sure the current process for approval, so just added the comments.  If I need to approve some other way let me know.

lib/erl/test/test_omit.erl
dhull: OK, will do.

djnym:
@Jens-G just following up here, this is good to merge, and it would be great if it could make it with @dhull other fixes in the 0.11.0 release.

PR #1319, opened 2017-07-28
THRIFT-4269: Don't append '.' to Erlang namespace if it ends in '_'.

dhull:
THRIFT-3834 added support for namespaces to the Erlang code generator. However, that support uses a `.` between the namespace name and the type name.  This is inconvenient because, although `.` is a valid character in an Erlang atom, atoms that contain `.` must be quoted.  This means that a struct named MyStruct in the namespace NS will generate a record named `'NS.MyStruct'`.  The rules for naming atoms in Erlang are:

> Atoms begin with a lower-case letter, and may contain alphanumeric characters, underscores (_) or at-signs (@). Alternatively atoms can be specified by enclosing them in single quotes ('), necessary when they start with an uppercase character or contain characters other than underscores and at-signs.

This pull request changes the Erlang code generator so that if an Erlang namespace ends in a `_` then no `.` is added between the namespace and the struct name when creating the record. This preserves the current behavior unless the namespace ends is a `_`, but allow users to override the normal behavior by adding an explicit `_` to the end of their namespace declarations.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_erl_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_erl_generator.cc
@@ -1019,18 +1019,13 @@ string t_erl_generator::argument_list(t_struct* tstruct) {
 }
 
 string t_erl_generator::type_name(t_type* ttype) {
-  string prefix = "";
-  string erl_namespace = ttype->get_program()->get_namespace("erl");
-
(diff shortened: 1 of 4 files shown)


djnym:
I agree with the rationale behind the change and the code looks fine.  It seems like the CI failures were related to client pools and not this code.  I would say this should be merged.

PR #1320, opened 2017-07-29
THRIFT-4270: Generate Erlang mapping functions for const maps and lists

dhull:
The Erlang generator now creates a new constants module containing two functions for each const map and list.

The first function takes a single argument, the key for a const map or the index for a const list, and returns the corresponding value for that key or index; it throws an exception if the key does not exist or the index is out of range.

The second function is similar to the first but takes a default second element, and returns this default instead of throwing an exception if the key does not exist or the index is out of range.

For example, if example.thrift contains the following Thrift definitions:

```thrift
const map<string, string> ALPHABET = { "a" : "apple", "b" : "banana" }
const list<string> NUMBER = [ "one", "two", "three" ]
```

then these calls will succeed:

```erlang
"apple" = example_constants:alphabet("a"),
"three" = example_constants:number(3),
```

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_erl_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_erl_generator.cc
@@ -118,6 +118,8 @@ public:
   void generate_type_metadata(std::string function_name, vector<string> names);
   void generate_enum_info(t_enum* tenum);
   void generate_enum_metadata();
+  void generate_const_function(t_const* tconst, ostringstream& exports, ostringstream& functions);
+  void generate_const_functions();
 
   /**
    * Service-level generation functions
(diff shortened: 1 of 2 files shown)


djnym:
@dhull The CI failure in this case does seem to be related to the code change, digging down into the failed case I see this
```
test/test_const.erl:25: can't find include file "gen-erl/constants_demo_types.hrl"
```

dhull:
@djnym, thanks for pointing out my build problem. I had developed the code in this pull request on top of my code for #1319 but thought it would be better to separate it back out for the pull request; unfortunately I made a mistake.

I have fixed the problem and rebased on the current HEAD and the tests are now passing.

djnym:
@Jens-G this looks good, feel free to merge.

PR #1322, opened 2017-08-01
THRIFT-4246 Multiplexed clients sequence id fix

boivie:
Client: nodejs

Previously, all clients would use the latest created multiplexer
for generating sequence numbers which would create a mismatch
between the mapping of sequence number->service in the connection.

This makes the client instances use the multiplexer that is
bound to it.

### DIFF
--- a/lib/nodejs/lib/thrift/multiplexed_protocol.js
+++ b/lib/nodejs/lib/thrift/multiplexed_protocol.js
@@ -52,17 +52,17 @@ Multiplexer.prototype.createClient = function(serviceName, ServiceClient, connec
   if (ServiceClient.Client) {
     ServiceClient = ServiceClient.Client;
   }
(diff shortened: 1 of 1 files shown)


boivie:
Thanks James. I just rebased it, but travis and AppVeyor fails from unrelated reasons. Can they be retriggered?

boivie:
@jeking3 The CI tests fails every time for seemingly unrelated reasons.

I won't try to solve that.

It's up to you now if you want to bring this one in. I have converted my project to gRPC for several reasons, so I'm not committed in this bugfix any longer.

jeking3:
Hmm, well I tried to rebase this on master and push it back into your branch so it would build again but I cannot - no permission.   So, I'm running cross test in the xenial docker image locally and if it passes I will merge it into master.

PR #1323, opened 2017-08-02
THRIFT-3775: 0.10.0 release

deep1kasingh:
Client: release
Patch: jfarrell

Versioning for 0.10.0 release candidate


Jens-G:
No idea what you think you can achieve with that. Please close this PR, as it will never get merged.

jeking3:
@jfarrell this needs to be closed.  It is invalid.

PR #1324, opened 2017-08-03
THRIFT-4275

sunnypig:
add support for zope.interface only (excluding support for twisted) in python compiler


jeking3:
@sunnypig just a reminder one thing was missed and needs to be updated - see my last comment, at the very least updating the command registration so the option appears in compiler help is mandatory - also recommend you rebase on master to get a clean build from CI.

sunnypig:
Sorry for my belated rely.  I have just updated the registration.  However, I have no idea on how to add a proper unit test,  since I can't find any test code for the option of "py:twisted" which perhaps can be taken for a reference.

jeking3:
That's okay - the twisted tests can be found in ``test/py.twisted`` for reference.

PR #1326, opened 2017-08-04
THRIFT-3821

asuhan:
Going over 2 GB is quite possible when using Thrift for serialization. We can detect the condition and throw an exception.

### DIFF
--- a/lib/cpp/src/thrift/transport/TBufferTransports.cpp
+++ b/lib/cpp/src/thrift/transport/TBufferTransports.cpp
@@ -361,9 +361,13 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
   }
 
   // Grow the buffer as necessary.
-  uint32_t new_size = bufferSize_;
+  uint64_t new_size = bufferSize_;
   while (len > avail) {
     new_size = new_size > 0 ? new_size * 2 : 1;
+    if (new_size > std::numeric_limits<uint32_t>::max()) {
+      throw TTransportException(TTransportException::BAD_ARGS,
+                                "Internal buffer size exceeded 2GB");
+    }
     avail = available_write() + (new_size - bufferSize_);
   }
 
--- a/lib/cpp/test/TMemoryBufferTest.cpp
+++ b/lib/cpp/test/TMemoryBufferTest.cpp
@@ -117,4 +117,17 @@ BOOST_AUTO_TEST_CASE(test_exceptions) {
   BOOST_CHECK_NO_THROW(buf2.write((const uint8_t*)"bar", 3));
 }
 
+#ifndef _WIN32
+// We can't allocate 1 GB of memory in 32-bit environments.
+BOOST_AUTO_TEST_CASE(test_over_two_gb) {
+  TMemoryBuffer buf;
+  std::vector<uint8_t> small_buff(1);
+  std::vector<uint8_t> one_gb(1073741824);
+
+  buf.write(&small_buff[0], small_buff.size());
+  buf.write(&one_gb[0], one_gb.size());
+  BOOST_CHECK_THROW(buf.write(&one_gb[0], one_gb.size()), TTransportException);
+}
+#endif
+
 BOOST_AUTO_TEST_SUITE_END()


lib/cpp/src/thrift/transport/TBufferTransports.cpp
```diff
@@ -361,9 +361,12 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
   }
 
   // Grow the buffer as necessary.
-  uint32_t new_size = bufferSize_;
+  uint64_t new_size = bufferSize_;
   while (len > avail) {
     new_size = new_size > 0 ? new_size * 2 : 1;
+    if (new_size > std::numeric_limits<uint32_t>::max()) {
+      throw TTransportException("Internal buffer size exceeded 2GB");
```
jeking3: I would prefer to see a TTransportException with an error type BAD_ARGS here, however it looks like error types are not being used elsewhere in this module.

jeking3:
Please rebase (and squash to a single commit, if you can) and push here, so the build system can build the change and we can verify it before merging it.  Thanks.

asuhan:
Thanks for the review, I'll address both. It's my preference to squash the commits as well, sounds like I've misinterpreted https://thrift.apache.org/docs/HowToContribute (it says "When bugfixing: add test that will isolate bug before applying change that fixes it").

asuhan:
I've disabled the test for Windows since allocating 1GB on 32-bit is likely to fail with `bad_alloc`.

Test failure looks unrelated, also got a clean run before: https://travis-ci.org/apache/thrift/jobs/263298365.

lib/cpp/src/thrift/transport/TBufferTransports.cpp
```diff
@@ -361,9 +361,13 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
   }
 
   // Grow the buffer as necessary.
-  uint32_t new_size = bufferSize_;
+  uint64_t new_size = bufferSize_;
   while (len > avail) {
     new_size = new_size > 0 ? new_size * 2 : 1;
+    if (new_size > std::numeric_limits<uint32_t>::max()) {
```
jeking3: Shouldn't this be int32_t instead of uint32_t, if we want to test against 2GB?  Sorry I didn't catch this before.

lib/cpp/test/TMemoryBufferTest.cpp
```diff
@@ -117,4 +117,17 @@ BOOST_AUTO_TEST_CASE(test_exceptions) {
   BOOST_CHECK_NO_THROW(buf2.write((const uint8_t*)"bar", 3));
 }
 
+#ifndef _WIN32
+// We can't allocate 1 GB of memory in 32-bit environments.
```
jeking3: This makes me wonder if the TBufferTransport should have a size limit that is configurable, with a default of INT32_MAX, and then the test can make a smaller one like 4KB, and write 4KB and then one byte more, instead of using up 2GB of memory.

lib/cpp/src/thrift/transport/TBufferTransports.cpp
asuhan: We check it post-resize and it fails if we go past 4 GB, therefore pre-resize it must be at most 2GB. Checking against `int32_t` would be overly conservative; we mainly care about avoiding the arithmetic overflow. Maybe the error message could be improved? `"Internal buffer size was already past 2GB when we attempted to resize"` would be a more precise description, but I didn't want to make it overly verbose / obscure.

lib/cpp/test/TMemoryBufferTest.cpp
asuhan: We'd have to write in a loop to trigger buffer resizing if I understand correctly. It'd save the vector allocation in the test, but `TMemoryBuffer` would still grow. I agree a 50% reduction would be better than nothing, for sure.

lib/cpp/src/thrift/transport/TBufferTransports.cpp
asuhan: Basically, if we only go through powers of 2 when resizing (which looks to be the case), the buffer can only grow to 2GB since the maximum value of `uint32_t` is one short of 4GB.

asuhan:
Are there any additional changes I should make to this pull request? As I've said in the inline comment, I believe comparing to `std::numeric_limits<uint32_t>::max()` is actually consistent with what the error message says.

jeking3:
I'd prefer if we refactored it to make the limit configurable with a default matching what it is now; that way tests won't need multiple gigabytes of memory to function properly.  I'll approve and merge this for now as it is an improvement.

PR #1328, opened 2017-08-05
THRIFT-2221: detect C++11 and use std namespace for memory operations (smart_ptr)

jeking3:
I had to close my older pull request because the history got tainted, so I am reopening it with the suggestions from Jira.


lib/cpp/README.md
```diff
@@ -55,15 +54,22 @@ you are using libthriftnb you will also need libevent.
 
 ## Dependencies
 
-boost shared pointers
-http://www.boost.org/libs/smart_ptr/smart_ptr.htm
+If your C++ environment implements C++11 or later, boost will automatically use
```
hcorg: `If your C++ environment implements C++11 or later, boost will automatically use` - boost or Thrift?

hcorg: Even if boost itself provides mapping to `std::` pointers, next sentence suggests that boost is not needed. Some little clean up could help confused users ;) (this is also first place in README.md where boost is mentioned, hence the potential confusion)

lib/cpp/src/thrift/stdcxx.h
```diff
@@ -0,0 +1,137 @@
...
+  using ::boost::static_pointer_cast;
+  using ::boost::weak_ptr;
+
+#else
+
+  using ::std::const_pointer_cast;
+  using ::std::dynamic_pointer_cast;
+  using ::std::enable_shared_from_this;
+  using ::std::make_shared;
+  // yes, they renamed in in std:: for some reason...
```
hcorg: unnecessary comment - `scoped_ptr` and `unique_ptr` are quite different (`scoped_ptr` can't be moved) so it's not like standard committee did something wrong ;) (at least with naming it `unique_ptr`)

lib/cpp/src/thrift/stdcxx.h
```diff
@@ -0,0 +1,137 @@
...
+#pragma message(STRING(__GNUC_MINOR__))
+#pragma message(STRING(__GNUC_PATCHLEVEL__))
+#pragma message(STRING(__cpp_alias_templates))
+#endif
+
+#if defined(BOOST_NO_CXX11_SMART_PTR) || defined(FORCE_BOOST_SMART_PTR)
+#include <boost/smart_ptr.hpp>
+#else
+#include <memory>
+  #if __cplusplus < 201103L
```
octopus-prime: What is this test for? BOOST_NO_CXX11_SMART_PTR told us above that c++11 samart pointers are available.

octopus-prime: What is this test/include for? BOOST_NO_CXX11_SMART_PTR told us above that c++11 smart pointers are available.

lib/cpp/README.md
jeking3: Got it, I can correct this.  I'm still dealing with aligning the stars on this one.  Lots of namespace and using cleanup, and we will no longer work with Visual Studio 2010 and 2012 in C++0x/C++11 mode; 2013 will be required for C++11 because the solution requires template aliases.

lib/cpp/src/thrift/stdcxx.h
jeking3: I removed it in the last PR.

lib/cpp/src/thrift/stdcxx.h
jeking3: Cleaned up in the last push.

lib/cpp/README.md
jeking3: Fixed.

jeking3:
image

Still pushing changes here.  Microsoft didn't make things easy by leaving __cplusplus at an old value even in the latest compilers.

jeking3:
I added a TPipe test because I changed one line of code in TPipe to work around a build issue, and I didn't see a TTransportTest test for TPipe.

jeking3:
The lone build failure is a "D" language/test issue that appears sporadically.  All other builds passed.   As such, I am going to merge this in given it was already reviewed, and the follow-on changes were fairly trivial.

PR #1329, opened 2017-08-08
THRIFT-4279: Fix include path in generated Erlang Thrift sources.

dhull:
The change in THRIFT-4270 inadvertantly modified the include directives in the generated Erlang sources to include the gen-erl directory in the path to the include file.  This commit restores the
previous behavior.

Example include directive before the THRIFT-4270 change:
```erlang
-include("constants_demo_types.hrl").
```

Example include directive after the THRIFT-4270 change:
```erlang
-include("test/gen-erl/constants_demo_types.hrl").
```

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_erl_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_erl_generator.cc
@@ -266,8 +266,8 @@ void t_erl_generator::init_generator() {
                 << "-module(" << program_module_name << "_types)." << endl
                 << erl_imports() << endl;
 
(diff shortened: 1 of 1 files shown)


djnym:
@Jens-G this probably needs to be merged assuming it passes the checks as we don't want to generate bad code.

dhull:
It appears to me that the AppVeyor build failure is due it a problem with the build machine:
```
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.8/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find FLEX (missing: FLEX_EXECUTABLE)
```

PR #1334, opened 2017-08-18
THRIFT-4288: Implement log.js

Equim-chan:
Implement log.js which was empty.
Apply log.js to all existing log operations.

Ref: https://issues.apache.org/jira/browse/THRIFT-4288


jeking3:
@Equim-chan now that the CI build environment has been stabilized, could you rebase on master and refresh this PR so we can get a clean build on it?  Thanks.

Equim-chan:
@jeking3 done.

PR #1335, opened 2017-08-23
Add default message for TApplicationException 

damnever:
Since some implementations may not set message field (field 1).


lib/go/thrift/application_exception.go
```diff
@@ -30,6 +30,17 @@ const (
 	PROTOCOL_ERROR                 = 7
 )
 
+var defaultApplicationExceptionMessage = map[int32]string{
+	UNKNOWN_APPLICATION_EXCEPTION:  "unknown application exception",
+	UNKNOWN_METHOD:                 "unknown method",
+	INVALID_MESSAGE_TYPE_EXCEPTION: "invalid message type",
+	WRONG_METHOD_NAME:              "wrong method name",
+	BAD_SEQUENCE_ID:                "bad sequnce ID",
```
jeking3: Typo in the string

jeking3:
Would it perhaps make sense to change the const list into something that can simply be generated using stringer, and therefore it would always be automatic?

https://godoc.org/golang.org/x/tools/cmd/stringer

damnever:
Type alias is not comparable, would it cause incompatibility problems?

jeking3:
In order to add this I need an Apache Jira ticket under the thrift project.  Do you open one?

damnever:
https://issues.apache.org/jira/browse/THRIFT-4315

PR #1336, opened 2017-08-26
configure.ac, Makefile.am: introduce THRIFT variable to support cross…

tpetazzoni:
…-compilation

The thrift build system currently assumes that the thrift compiler is
always available in $(top_builddir)/compiler/cpp/thrift. However, in a
cross-compilation context, this location contains the thrift compiler
built for the target... which obviously will not run on the build
machine.

In order to support such cross-compilation situation, we introduce the
THRIFT variable as a an argument for the configure script (using
AC_ARG_VAR). If not specified, it defaults to the existing value of
using compiler/cpp/thrift from the build directory, but it can be
overridden when calling ./configure.

Note that $(top_builddir) cannot be used within the configure script,
so we simply use `pwd`, which is the same as the top_builddir.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -81,6 +81,9 @@ AS_IF([test "x$D_IMPORT_PREFIX" = x], [D_IMPORT_PREFIX="${includedir}/d2"])
 AC_ARG_VAR([DMD_LIBEVENT_FLAGS], [DMD flags for linking libevent (auto-detected if not set).])
 AC_ARG_VAR([DMD_OPENSSL_FLAGS], [DMD flags for linking OpenSSL (auto-detected if not set).])
 
+AC_ARG_VAR([THRIFT], [Path to the thrift tool (needed for cross-compilation).])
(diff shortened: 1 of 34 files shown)


jeking3:
Wouldn't we want the value in configure.ac to percolate through?  Why override it in every Makefile.am?

tpetazzoni:
Well, my patch precisely removes the override in every Makefile.am, so I'm not sure to understand your question.

jeking3:
Thanks, it was late when I posted the comment and I misread the change!  If there's any way you could rebase on master, squash, and force push that will kick off another CI build to validate the fix.

jeking3:
I like the changes as proposed; this needs an Apache Jira ticket stating the problem and resolution in order to get merged.

Thanks!

jeking3:
I opened THRIFT-4325 for this.

PR #1337, opened 2017-08-28
THRIFT-4292: Implement TimerManager::remove()

Typz:


jeking3:
Try rebasing on the current master to get a cleaner CI build, and force push to kick a new one here.

lib/cpp/src/thrift/concurrency/TimerManager.cpp
```diff
@@ -290,11 +292,23 @@ void TimerManager::add(shared_ptr<Runnable> task, const struct timeval& value) {
...
+  for (task_iterator ix = taskMap_.begin(); ix != taskMap_.end(); ++ix) {
+    if (ix->second->runnable() == task) {
+      toRemove.push_back(ix);
+    }
+  }
+  if (toRemove.empty()) {
+    throw NoSuchTaskException();
+  }
+  for (std::vector<task_iterator>::iterator ix = toRemove.begin(); ix != toRemove.end(); ++ix) {
+    taskMap_.erase(*ix);
```
jeking3: As written, if a single task exists multiple times in the taskMap then this first erase call invalidates any other iterators in toRemove, and it will lead to undefined behavior. This must be fixed to merge; see my previous comment for a safer way to achieve this.

lib/cpp/src/thrift/concurrency/TimerManager.cpp
```diff
@@ -52,6 +52,8 @@ class TimerManager::Task : public Runnable {
     }
   }
 
+  shared_ptr<Runnable> runnable() const { return runnable_; }
```
jeking3: Consider implementing an ``operator ==`` instead which compares to a ``shared_ptr<Runnable>`` thus hiding the internal details of the ``TimerManager::Task`` and simplifying the caller to ``if (task == ix->second)`` below.  What's there works, just a suggestion - not required to merge.

lib/cpp/src/thrift/concurrency/TimerManager.cpp
```diff
@@ -290,11 +292,23 @@ void TimerManager::add(shared_ptr<Runnable> task, const struct timeval& value) {
 }
 
 void TimerManager::remove(shared_ptr<Runnable> task) {
-  (void)task;
   Synchronized s(monitor_);
   if (state_ != TimerManager::STARTED) {
     throw IllegalStateException();
   }
+  std::vector<task_iterator> toRemove;
```
jeking3: I'd recommend using the prefix/postfix method to walk and erase here in order to avoid iterator invalidation:

    bool found = false;
    for (task_iterator it = taskMap_.begin(); it != taskMap_.end(); ) {
      if (ix->second->runnable() == task) {
        found = true;
        taskMap_.erase(it++);
      } else {
        ++it;
      }
    }

    if (!found) {
      throw NoSuchTaskException();
    }

lib/cpp/test/concurrency/TimerManagerTests.h
```diff
@@ -126,6 +126,38 @@ class TimerManagerTests {
...
+    // Remove one task and wait until the other has completed
+    timerManager.remove(taskToRemove);
+    _monitor.wait(timeout * 2);
+
+    assert(!taskToRemove->_done);
+    assert(task->_done);
+
+    return true;
+  }
+
```
jeking3: Recommend you add a test to remove the same task that's been added to the timerManager more than once.

lib/cpp/src/thrift/concurrency/TimerManager.cpp
Typz: Tasks cannot exist multiple times, only newly allocated tasks are inserted:
```cpp
  taskMap_.insert(
        std::pair<int64_t, shared_ptr<Task> >(timeout, shared_ptr<Task>(new Task(task))));
```

Which is why I implemented it in the same fashion as the run() method. But i can change it.

lib/cpp/src/thrift/concurrency/TimerManager.cpp
Typz: done

lib/cpp/test/concurrency/TimerManagerTests.h
Typz: done

lib/cpp/src/thrift/concurrency/TimerManager.cpp
Typz: done

PR #1338, opened 2017-09-01
Make ssl-open timeout effective in golang client

tudyzhb:
Using **tls.DialWithDialer** just like **net.DialTimeout** which in _socket.go_ to make ssl-open timeout effective.


Jens-G:

PR #1339, opened 2017-09-01
THRIFT-4309 Python print() function

cclauss:


Jens-G:

jeking3:
I'm not certain I see what the purpose of this set of changes is (then again, I don't use python much).  Perhaps if you open a Jira ticket and explain the reasoning for these changes it would help.

jeking3:
@cclauss now that the CI build issues should be squared away, could you rebase on master so we get another CI build?  Thanks.

cclauss:
@jeking3 All of the __print() is a function__ changes are ESSENTIAL Python 3 compatibility.  This is not optional.  See: https://docs.python.org/3/whatsnew/3.0.html

The reason that you are not seeing any problems in the build is because Python is a dynamic, interpreted language.  It does not have a separate compile step like C/C++ and other compiled languages have.  This in turn means that syntax errors are not found at "build" time but that they are only found at runtime.  This in turn explains why flake8 is better tool for spotting these problems than pylint.  flake8 uses the Abstract Syntax Tree process to do simulate runtime while pylint does not.

jeking3:
What you suggest is that our build process is insufficient to catch these things, and I would like to see our build process become sufficient and enhanced with whatever tools are needed to make thrift better.  At the very least we should run modernize on all of the lib/py files, although this PR is a start, it does not do enough to mark THRIFT-4309 complete.  I can merge it, but more work will be necessary to complete everything needed.

cclauss:
This makes sense to me.  Please merge this PR when you are ready but we must do further work in another PR before THRIFT-4309 is complete.

There are three alternative approaches to running the approiate tests that I must consider.

PR #1340, opened 2017-09-01
THRIFT-4295: reworked docker images to resolve build issues

jeking3:
I reworked docker images and travis build jobs, omitting debian for now because it's quite similar to ubuntu.  I brought the number of build jobs down to 12 (from 20) by omitting Python 2.6, mingw32, and consolidating where possible.  This will speed up builds and we don't lose much.  Ubuntu Xenial doesn't have mingw32, nor does debian stretch.  If we need this re-enabled it could be done as part of the centos-7.3 image, just let me know.  I didn't see much reason to keep supporting Python 2.6 when 2.7 is common and 3.5 is out to replace it.

Notes:

For the D language I have openssl and libevent disabled in the xenial image for now.  I ran into issues with the deimos for openssl and the version in the image; deimos supports 1.0.1g and xenial has 1.0.2g in it.  Will address in a separate PR.  The centos/7.3 image still has the demios extensions enabled for openssl and libevent.  I opened a Jira ticket for this already.

For php7 in xenial I had to make a TCP_NODELAY set conditional.  The required methods didn't seem to exist in the OS provided php7 packages.  This didn't affect tests, apparently, as they all passed.

I added versions to every language in the configure output summary.

I removed a generated file that was checked in (mock_handler.go).


jeking3:
Travis CI build failure was environmental (unable to download a module from the dart module service).  Otherwise this is good to go.

lib/php/lib/Thrift/Transport/TSocket.php
```diff
@@ -242,8 +242,10 @@ public function open()
       throw new TException($error);
     }
 
-    $socket = socket_import_stream($this->handle_);
-    socket_set_option($socket, SOL_TCP, TCP_NODELAY, 1);
+    if (function_exists('socket_import_stream') && function_exists('socket_set_option')) {
```
stigsb: In high-traffic situations, this change can create a TIME_WAIT pileup. 0.11.0 is unusable for us because of this issue, please either revert or make it configurable defaulting to off.

jeking3: I'd like to know a little more.  If the functions don't exist they cannot be called.  Is the problem that function_exists is returning the wrong result?

stigsb: My analysis was incorrect, it wasn't your patch causing our issue @jeking3 sorry for jumping the gun.

PR #1000, opened 2016-04-29
THRIFT-3820 detect OTP >= 18 to use new time correction

NOMORECOFFEE:
erlang:now/0 is deprecated BIF.

See the "Time and Time Correction in Erlang" chapter of the ERTS User's Guide for more information.

PR #1346, opened 2017-09-07
THRIFT-4273 erlang:now/0: Deprecated BIF.

walter-weinmann:
Simply following the recommendations here:

[3.7  New Time API](http://erlang.org/doc/apps/erts/time_correction.html#id79939): How to Work with the New API

### DIFF
--- a/lib/erl/src/thrift_reconnecting_client.erl
+++ b/lib/erl/src/thrift_reconnecting_client.erl
@@ -36,7 +36,7 @@
           terminate/2,
           code_change/3 ]).
 
--record( state, { client = nil, 
+-record( state, { client = nil,
                   host,
                   port,
                   thrift_svc,
@@ -226,9 +226,9 @@ timer_fun() ->
   end.
 -else.
 timer_fun() ->
-  T1 = erlang:now(),
+  T1 = erlang:timestamp(),
   fun() ->
-    T2 = erlang:now(),
+    T2 = erlang:timestamp(),
     timer:now_diff(T2, T1)
   end.
 -endif.
--- a/test/erl/src/test_client.erl
+++ b/test/erl/src/test_client.erl
@@ -160,12 +160,11 @@ start(Args) ->
         ClientS4
     end,
 
-  %% Use deprecated erlang:now until we start requiring OTP18
   %% Started = erlang:monotonic_time(milli_seconds),
-  {_, StartSec, StartUSec} = erlang:now(),
+  {_, StartSec, StartUSec} = erlang:timestamp(),
   error_logger:info_msg("testOneway"),
   {Client20, {ok, ok}} = thrift_client:call(Client19, testOneway, [1]),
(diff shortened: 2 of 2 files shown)


walter-weinmann:
This closes #1346

walter-weinmann:
Restart checks.

walter-weinmann:
Restarting the checks.

walter-weinmann:
Restarting the checks.

walter-weinmann:
Restarting the checks.

walter-weinmann:
Restarting the checks.

walter-weinmann:
The checks are really tedious - I needed 5 attempts until everything had gone through without any problems.

jeking3:
@walter-weinmann what was failing - Appveyor or Travis?  There are some known build issues lurking about.  I just eliminated one that popped up occasionally with concurrency_test in C++, however I just found that on Appveyor haskell is failing dependency checks sometimes.

jeking3:
What versions of Erlang would this change eliminate as usable with Thrift?  I am finding it hard to locate the erlang release where the timestamp method was introduced.

walter-weinmann:
The problems were with Travis with varying reasons. I think it would be interesting to look into it. There may be a couple of pull request with bad results in Travis based on Travis internal problems. Similar observations  wit Appveyor - is it really useful to run both ?

walter-weinmann:
The `erlang:timestamp/0` method seems to be a very old kernel methods - from the [release notes](http://erlang.org/doc/apps/kernel/notes.html) (current kernel release is 5.3):

> 1.42  Kernel 2.13.2
> 
> Fixed Bugs and Malfunctions
> 
> A bug when doing io:get_line (among other calls) from a file opened with encoding other than latin1, causing false unicode errors to occur, is now corrected.
> 
> Own Id: OTP-7974
> 
> Improvements and New Features
> 
> Added functionality to get higher resolution timestamp from system. The erlang:now function returns a timestamp that's not always consistent with the actual operating system time (due to resilience against large time changes in the operating system). The function os:timestamp/0 is added to get a similar timestamp as the one being returned by erlang:now, but untouched by Erlangs time correcting and smoothing algorithms. The timestamp returned by os:timestamp is always consistent with the operating systems view of time, like the calendar functions for getting wall clock time, but with higher resolution. Example of usage can be found in the os manual page.
> 
> Own Id: OTP-7971

jeking3:
Building Appveyor right now probably not, since I don't think we do erlang on windows in our CI right now.  In order to get there, we need erlang support in the cmake build environment.  The travis builds are the important ones, although as we discovered with the tutorials rotting, a build is only as good as what it tests.  Hopefully we have enough tests to cover these changes.

walter-weinmann:
@jeking3 please have a look at THRIFT-2289 and THRIFT-2889 - I suppose both can be closed.

NOMORECOFFEE:
Hi, @walter-weinmann. This's patch #1000 do the same thing. :wink:

PR #1352, opened 2017-09-10
THRIFT-4326 Allow safer reuse of BufferedTransport instances in Ruby

benweint:
Addresses [THRIFT-4326](https://issues.apache.org/jira/browse/THRIFT-4326).

## The problem

In the case where a single `Thrift::BufferedTransport` instance is re-used across multiple service calls, certain kinds of malformed responses from one service call can 'leak' into subsequent calls and cause them to fail with a `Thrift::ProtocolException`.

The most easily reproducible example is when a service returns a well-formed Thrift response for the first service call, but with N extra bytes of garbage tacked onto the end.

In such a case, the initial service call will be handled just fine (at least when using the compact protocol), however, the next N service calls that go through the same `Thrift::BufferedTransport` instance will fail with a `Thrift::ProtocolException`. This happens because the `BufferedTransport` doesn't re-set the `@rbuf` instance variable until the read buffer is fully exhausted, so each of the N subsequent service calls will attempt to read one byte identifying the protocol from the remaining buffer, and will get some bogus value from the garbage bytes at the end of the response from the initial service call.

This can also happen if reading from `@rbuf` is interrupted part-way through, while `@index` is still pointing to the middle of the read buffer (e.g. due to a Ruby timeout exception).

## Proposed solution

Re-setting the `@rbuf` instance variable to an empty byte buffer upon every call to `#flush` addresses this problem, and is conceptually similar to what happens in `HttpClientTransport#flush` (where `@inbuf` and `@outbuf` are both always reset).

### DIFF
--- a/lib/rb/lib/thrift/transport/buffered_transport.rb
+++ b/lib/rb/lib/thrift/transport/buffered_transport.rb
@@ -101,7 +101,8 @@ module Thrift
         @transport.write(@wbuf)
         @wbuf = Bytes.empty_byte_buffer
       end
-      
+
+      @rbuf = Bytes.empty_byte_buffer unless @rbuf.empty?
       @transport.flush
     end
   end
--- a/lib/rb/spec/base_transport_spec.rb
+++ b/lib/rb/spec/base_transport_spec.rb
@@ -126,6 +126,17 @@ describe 'BaseTransport' do
       btrans = Thrift::BufferedTransport.new(trans)
       btrans.flush
     end
+
+    it "should reset read buffer on flush even if there's data left in it" do
+      trans = mock("Transport")
+      trans.should_receive(:flush)
+      trans.should_receive(:read).and_return("abc", "xyz")
+
+      btrans = Thrift::BufferedTransport.new(trans)
+      btrans.read(1).should == "a"
+      btrans.flush
+      btrans.read(1).should == "x"
+    end
   end
 
   describe Thrift::BufferedTransportFactory do


benweint:
Thanks @jeking3! I'v just rebased and pushed, so hopefully that'll make CI happy.

jeking3:
@benweint excellent, I'll keep an eye on these.

jeking3:
Sadly we're seeing a lot of timeouts in the Travis environment.  I see one build job stopped responding for 10 minutes and another timed out of a test.

jeking3:
@benweint could you do be a favor and do an empty amend (git commit --amend, and then don't change anything) and force push?  This will generate a new build.  It's build job 4 I am most interested in seeing pass as it runs the ruby tests.

benweint:
Sure thing - I've done that and it looks like it did kick off another build.

jeking3:
If you could rebase one more time and force push, I found out why build jobs 3 and 4 became unstable - I moved nodejs testing from job 3 to 4 in my last PR, and I just undid that.  Sorry about all this...

benweint:
Yup, done. No worries - I know how big hairy test suites go :)

jeking3:
Of course now I see job 2 is failing in nodejs as well.  Spectacular!  Well, let's see how this fares.  If we can get job 4 to pass that would be ideal.  If this one doesn't work I'll run it locally myself in docker and validate it before merging.

jeking3:
@benweint would you be able to rebase on master?  The CI build issues should be resolved now.

benweint:
Yup, rebased!

jeking3:
Ugh, clearly we STILL have some build issues.  They are in js though.  I'll pull your commit into my local system and try to reproduce this build issue and/or get a clean set of cross tests so I can commit this.  Thanks for your patience.

jeking3:
@benweint so this error is reproducible in the ubuntu-xenial docker image, and it looks like it is related to ruby and wasn't happening before, so it's likely related to these changes.
```
*** Following 2 failures were unexpected ***:
rb-nodejs               json              buffered-ip              failure(1)
rb-nodejs               binary            buffered-ip              failure(1)
```

benweint:
Ah, interesting! It looks like the nodejs tests are using a Ruby server as part of some integration testing? I'll see if I can repro this failure locally and figure out what's up. Thanks!

benweint:
Ok, I've finally figured out how to reproduce that test failure locally on macOS. For posterity, here's what I did:

```
PATH="/usr/local/opt/bison/bin:$PATH" ./configure --with-openssl=/usr/local/opt/openssl --with-boost=/usr/local --without-lua --without-csharp --without-c_glib --without-erlang --without-php --without-cpp --without-haskell
make -j3 precross
test/test.py --server rb --client nodejs
```

After doing that, the tests that use a `BufferedTransport` on the Ruby server side fail. Since I'm only dealing with Ruby on the client side in my usage, I didn't realize that `BufferedTransport` can be used to wrap a server socket as well (this is what the Ruby `TestServer.rb` is doing when testing a buffered transport).

The issue stems from my misunderstanding of the semantics of `#flush`. In the nodejs -> Ruby cross test, I think the node client writes multiple requests to the socket before the Ruby server responds to any of them. The Ruby server calls `#flush` after each response to flush the write buffer, but my change drops the read buffer as well, discarding the buffered requests before they've been processed.

I'm going to close this and think about whether there's a better way to accomplish what I'm after.

PR #1355, opened 2017-09-12
Minimal C# library version for .NET Standard 1.4, 1.6, .NET 4.5.1

aloneguid:
This adds another project type to C# library which allows it to be build for .NET Standard 1.4, 1.6 and .NET 4.5.1. The project turns off all dependencies (only the new project, existing ones are untouched) to build the core thrift library.

### DIFF
--- a/lib/csharp/src/Thrift.sln
+++ b/lib/csharp/src/Thrift.sln
@@ -1,6 +1,8 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 11.00
-# Visual Studio 2010
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26730.12
+MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Thrift", "Thrift.csproj", "{499EB63C-D74C-47E8-AE48-A2FC94538E9D}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ThriftTest", "..\test\ThriftTest\ThriftTest.csproj", "{48DD757F-CA95-4DD7-BDA4-58DB6F108C2C}"
@@ -10,6 +12,8 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Thrift.45", "Thrift.45.csproj", "{EBCE35DA-CF6A-42BC-A357-A9C09B534299}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ThriftMVCTest", "..\test\ThriftMVCTest\ThriftMVCTest.csproj", "{891B4487-C7BA-427E-BBC8-4C596C229A10}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ThriftNetStandard", "ThriftNetStandard.csproj", "{2DFDA60B-072C-4265-9CE0-956A3B5131DD}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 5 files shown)


Jens-G:
What is THRIFTCORE? Is this intended tom become another netcore library? What about the existing one? Please explain.

gideonkorir:
I'm also interested in a .net core lib can we upgrade the netcore project to netstandard?

aloneguid:
@Jens-G THRIFTCORE is a simple pre-processor directive which when set excludes files which require dependency on external libraries in .NET Standard, essentially producing a minimal library of Thrift.

At the moment you have csharp library for .NET 3.5 and 4.5. Then, there is another library for .NET Core which to be honest is completely useless as it downloads so many dependencies that it's impossible to use in any web project as they quickly become conflicting. Ideally there must be just one single .NET project which cross-compiles to both .NET classic and .NET Standard. There is no need for .NET Core as .NET Standard covers it (is it maybe coming from misunderstanding of library creators in differences between .net core and .net standard?).

All I'm adding is a new project referencing all existing files. The project simply excludes stuff which is not covered by least common denominator of .NET Standard 1.4, 1.6 and .NET 4.5, but allows you to use Thrift on any device. This can be carefully extended further to support all Thrift features. however it's more complicated with .NET Standard as unlike .NET it doesn't have all socketing, networking etc. built in and needs external references.

I would suggest though having 1 main package with core functionality as I just did, and adding more libraries which implement specific features (sockets, web http etc.) as they require specific dependencies.

aloneguid:
@gideonkorir from what I can see .net core lib was created in times when .net core was in early stages, has some classes renamed and doesn't align with existing csharp project. There is no need to have that library, just cross-compile csharp library with modern .net core sdk.

Jens-G:
I'm fine with combining both flavours into one single c# library etc, as long as you don't break the c# library. The netcore lib is quite new and there's a lot of movement so we surely can throw it out and replace it by sth else. I don't see any conflicts here. Would you say that you two guys able to do that as a team and provide a suitable PR that meets that goal?

aloneguid:
I'm more than happy to do that. @Jens-G in .net ecosystem nuget package manager is quite popular, do you have any guidance how we can automatically publish this as you do with Maven?

Jens-G:
Jake does that.

gideonkorir:
@aloneguid can we target [netstandard 1.1](https://github.com/dotnet/standard/blob/master/docs/versions/netstandard1.1.md)? it covers .net 4.5

lib/csharp/src/Thrift.sln
```diff
@@ -1,6 +1,8 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 11.00
-# Visual Studio 2010
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
```
jeking3: Note that sometime in the future we're going to remove all provided static solution files and rely solely on project creation that comes with cmake, at which point you'll pick what you want at cmake generation time.

aloneguid: is there any strong reason for this?

gideonkorir:
@jeking3 netstandard 2.0 only targets .net framework 461 and .net core 2.0 only [see](https://github.com/dotnet/standard/blob/master/docs/versions.md). Targeting 1.1 targets a bigger audience also I think (unless I got you wrong) there is a difference between netstandard (which is a versioned set of APIs) and .net core (with is the cross platform runtime).

aloneguid:
@gideonkorir 1.1 is fine unless we hit some limitations. Generally 1.4 is the "safe" version of .net standard, which can be  used on embedded devices, clusters, desktops etc. However, it's not compatible with .NET 4.5, which by itself reached end of support stage, and is a pain in the neck in terms of security and compatibility. If you need to support .NET 4.5 it's generally better to cross-compile to .net standard and .net 4.5 instead of just targeting .net standard, currently I do this in https://github.com/elastacloud/parquet-dotnet

gideonkorir:
@aloneguid I actually need 1.4 so if  the rest of the guys here agree then we can target 1.4 as the minumum. I suggested 1.1 do that you wouldn't need to cross compile

jeking3:
Is there a Jira ticket for this work?

aloneguid:
Sorry guys I'm giving up. Personally I think getting a real value from contributing to main repo is not worth the effort due to many showstoppers. This may be a reason why there are so many forked versions exist. I don't mean to offend anyone here.

PR #1356, opened 2017-09-14
THRIFT-2289 Erlang impl of Thrift JSON protocol wrongly writes/expect…

walter-weinmann:
…s true/false for bools.


walter-weinmann:
This closes #1356.

PR #1358, opened 2017-09-14
THRIFT-4328: disable centos-7.3 build jobs which appear to be unstabl…

jeking3:
…e/hang in jenkins


jeking3:
Build job 4 does not fail for me locally, it looks like something happening during "npm install".  I asked on the mailing list for some help.  In the meantime we'll continue to monitor this.

PR #1362, opened 2017-09-15
THRIFT-4331 - C++: TSSLSocket fixes on branch 0.10.0

MartinHaimberger:
fixed issue with large messages, where waitForEvent was called
mutliple times waiting for SSL_read() to get bytes and running
in the retry timeout.

fixed issue where poll was not using the right flags.

Fixes issue THRIFT-4331 on branch 0.10.0.


jeking3:
We don't retroactively fix branches; the community might be interested in this but as a pull request it won't be merged.  @jfarrell this can be closed.

PR #1363, opened 2017-09-15
THRIFT-4331 - C++: TSSLSocket fixes

MartinHaimberger:
fixed issue with large messages, where waitForEvent was called
mutliple times waiting for SSL_read() to get bytes and running
in the retry timeout.

fixed issue where poll was not using the right flags.

Fixes issue THRIFT-4331 on master.

### DIFF
--- a/lib/cpp/src/thrift/transport/TSSLSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLSocket.cpp
@@ -295,8 +295,9 @@ bool TSSLSocket::peek() {
           }
         case SSL_ERROR_WANT_READ:
         case SSL_ERROR_WANT_WRITE:
-          waitForEvent(error == SSL_ERROR_WANT_READ);
-              continue;
+          // in the case of SSL_ERROR_SYSCALL we want to wait for an read event again
+          waitForEvent(error != SSL_ERROR_WANT_WRITE);
+          continue;
         default:;// do nothing
       }
       string errors;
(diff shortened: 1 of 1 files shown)


lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -759,7 +775,9 @@ unsigned int TSSLSocket::waitForEvent(bool wantRead) {
   struct THRIFT_POLLFD fds[2];
   memset(fds, 0, sizeof(fds));
   fds[0].fd = fdSocket;
-  fds[0].events = wantRead ? THRIFT_POLLIN : THRIFT_POLLOUT;
+  // use POLLIN also on write operations too, this is needed for operations
+  // which requires read and write on the socket.
```
jeking3: Interesting, what's an example of this?

MartinHaimberger: hi jeking3,

the comment may not be perfect, but SSL operations, like SSL_connect or SSL_shutdown (where this function is also used) need to perform write and read operations on the socket and we dont know where we are so we need to listen on both socket events.

I could also introduce a new function parameter for that.

jeking3:
If you could do me a favor and rebase on the current master, I finally got all the CI build issues resolved so I'd like to see a clean build before I merge.  There are some minor formatting (indentation) issues as well.

jeking3:
@MartinHaimberger If you could do me a favor and rebase on the current master, I finally got all the CI build issues resolved so I'd like to see a clean build before I merge. There are some minor formatting (indentation) issues as well.

MartinHaimberger:
rebased and indentations should be fixed

PR #1366, opened 2017-09-21
THRIFT-4233 Make THsHaServer.invoker available (get method only) in inherited classes

dmvolod:
…


dmvolod:
@jeking3 , please review. The static is not acceptable to getInvoker as invoker is not static context

PR #1367, opened 2017-09-21
Fix a crash on client close

djnym:
When a client closes a connection to a framed server the server was
crashing because the fact that the transport was framed was being
lost.  Looking through the file I noticed that the block from lines
87-95, looked different from the one from 59-66.  The culprit was
that when an error was occuring in the 59-66 block it was being
propagated up without rewrapping.  That would cause a failure
much further up the chain.  This patch fixes it.


jeking3:
Thanks for the fix - changes need a Jira ticket so we can track them through release notes.  See:

Thanks!

dhull:
This fix looks correct to me. The `read` and `read_exact` functions both pull the wrapped transport out of their state, and need to put the updated wrapped transport back into their state on return. The change fixes that in the error code path in the `read` function.

jeking3:
@dhull thank you!  Now we just need a Jira item for the change, and hopefully it sails through CI.

djnym:
I added a jira ticket yesterday https://issues.apache.org/jira/browse/THRIFT-4339

PR #1379, opened 2017-09-28
THRIFT-4350: enable dotnet-2.0.0-sdk in ubuntu xenial

jeking3:
update netcore projects, get them running with make cross
Client: netcore

This is purely experimental.  I wanted to know what was involved.  I moved everything to the latest version which may not be useful.  I got netcore with dotnet-2.0.0-sdk and netcoreapp2.0 targeted applications runnign in make cross (as client only).


jeking3:
Looks like we need "publish" support or something for that last build job.

danielcrenna:
@jeking3 What csharp gen options are passed to produce source compatible with this netcore lib? Any .thrift file I pass through `csharp:async` for example uses old style begin/end but this netcore lib uses async/await patterns, but I can't find any published reference for C# option to generate corresponding output. Also the namespaces appear to differ. My reference point is thrift-0.10.0.

jeking3:
HI Daniel, great question - I don't have a good answer for you.  I focused my work on making the dotnet 2.0 conversion on the library.  Given all of the tests pass it would appear that the output from the compiler is compatible, at least, if not optimized for netcore 2.  Thoughts?

danielcrenna:
This was just an issue where I was obviously not calling `netcore` as the generator, where `thrift.0.10.0.exe` doesn't have that generator installed. When I built it myself on `master` branch it all is well. Building was a bit of a riddle, had to rename `version.h.in` to `version.h` and rename the `flex` and `bison` to `win_flex` and `win_bison` because of how aliases work in `choco`, but things appear to work now. Thanks for responding :)

PR #1384, opened 2017-10-03
THRIFT-4354 TSocket block on read

RobberPhex:
ref <https://issues.apache.org/jira/browse/THRIFT-4354>


jeking3:
Let's get another CI build going on this.

PR #1391, opened 2017-10-13
Fix segment fault at thrift_protocol extension

RobberPhex:
* while serialize set,
    if type at IDL is scalar, directly use key of array as element.
    if type at IDL isn't scalar, directly use value of array as element.
* update license at config.m4
* drop php5 support
* after call call_user_function, detect EG(exception)
* zend_read_static_property without silent, cause a EG(exception).
* when serialize set at extension, pass sub-fieldspec to inner-serialize.(do not pass nullptr)
* when validate_thrift_object, if there is no isValidate, skip validate.
* add thrift_protocol_read_binary_after_message_begin for deserialize without method name.

---

close #1383 #1385


jeking3:
Is this for THRIFT-4353?  Can you squash it?  Rebase it on master to fix the CI issues.

RobberPhex:
this pr fix [`THRIFT-4356`](https://issues.apache.org/jira/browse/THRIFT-4356) and [`THRIFT-4353`](https://issues.apache.org/jira/browse/THRIFT-4353).

---

And, this work is besed `董菲 <feidong@anjuke.com>`'s work. so I just squashed commits by author(not all-in-one commit).

jeking3:
UBsan build failed due to https://issues.apache.org/jira/browse/THRIFT-2913.

RobberPhex:
build job 4 failed due to https://issues.apache.org/jira/browse/THRIFT-2913.

And, appveyor build queued.
but, same commit at my repo CI is success: https://ci.appveyor.com/project/RobberPhex/thrift/build/1.0.0-dev.58

jeking3:
I saw cross test pass before, so I think we're good.

jeking3:
@jfarrell I see licensing changes in this pull request, can you review?

RobberPhex:
Hi, about license.

`Copyright (C) 2009 Facebook` at file `lib/php/src/ext/thrift_protocol/config.m4` was add by commit f82aee5087bd62989482f5c532cbd80f97a39b7f.
But before this commit, the license at [`php_thrift_protocol.cpp`](https://github.com/apache/thrift/blob/f82aee5087bd62989482f5c532cbd80f97a39b7f^/lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp), [`php_thrift_protocol.h`](https://github.com/apache/thrift/blob/f82aee5087bd62989482f5c532cbd80f97a39b7f^/lib/php/src/ext/thrift_protocol/php_thrift_protocol.h) was Apache License.

lib/php/src/ext/thrift_protocol/config.m4
```diff
@@ -1,25 +1,27 @@
-dnl Copyright (C) 2009 Facebook
-dnl Copying and distribution of this file, with or without modification,
-dnl are permitted in any medium without royalty provided the copyright
-dnl notice and this notice are preserved.
+dnl Licensed to the Apache Software Foundation (ASF) under one
```
jeking3: dreiss's copyright statement says, "provided the copyright notice and this notice are preserved.".  I don't think we should be removing it.  You can move it down to be below the Apache Thrift license, or @jfarrell needs to approve this change.

RobberPhex: OK, I just add APL declaration to `config.m4`

jeking3:
Hmm, I'm just curious about this - given support for php5 was dropped, any idea why the ubuntu-trusty "make check" job succeeded (because it has PHP 5.5.9 in it)?  Is the build job not actually running any unit tests or anything that would parse (and fail, if it really is incompatible) the code?

RobberPhex:
at job `Autotools (Ubuntu Trusty)`, it just run `make check`, so php extension cannot be compiled.

at job `Cross Language Tests`, it needn't php extension.

So, I think there already are some CI problems at php part.

jeking3:
I already discussed the possibility of retiring the trusty build environment to reduce project maintenance.   Maybw now is the time to do it.  That said, I would really like to get input from someone else who uses php regularly.  @Jens-G perhaps... to ensure this change is acceptable (removing php5 support).  I haven't seen any objections in the developer email list but it hasn't been expressly mentioned.

jeking3:
I posted a message on the developer mailing list about removing php5 support.  We'll see what comes of it.  php5 is currently still shipping so I have some reservations about removing support for it, but I want to hear if there are objections.  If there are, can this fix be modified to deal with both?

RobberPhex:
@jeking3 

Currently, I just remove PHP5 support at PHP extension(thrift_protocol), PHP library is still support PHP5.
So, I think this PR is good.

---

If PHP5 support for extension is needed, I will add it back.

PR #1392, opened 2017-10-16
THRIFT-4367 Fix missing positional argument

3013216027:
I think the function process in the base TProcessor is missing a "self" argument.


jeking3:
Please see: https://thrift.apache.org/docs/HowToContribute

This does not have an associated [Thrift Jira](https://issues.apache.org/jira/projects/THRIFT) issue and is not properly named.

Recommend you fix these issues then rebase against master to resolve some of the CI issues.

jeking3:
I opened THRIFT-4367 for this.

PR #1395, opened 2017-10-21
THRIFT-4306 dlang: public imports for dependencies

dhasenan:
If you have an IDL file that imports another IDL file, you can define
a service that accepts, yields, or throws types defined in the
imported file. In this case, the dlang output should ensure that the
types defined in the imported IDL module are visible in the service
interface's module.

The generator failed to do that, resulting in a failure to compile
the generated source in these situations. This was previously
accepted by the compiler due to some long-standing bugs with symbol
visibility, but the compiler bugs have been resolved.

Fixes #4306.


jeking3:
@jfarrell please close this, it was merged in commit 44701e82ad351d7472101ad81aa13e775ffb7384

PR #1396, opened 2017-10-21
THRIFT-4305 Doc comments in dlang output

dhasenan:
Thrift IDL supports documentation comments. D supports doc comments.
So let's copy them over!

Future directions: add function parameter docs. ddoc expects them in
a particular format.


jeking3:
Manually inspecting the result in a sandbox.

PR #1398, opened 2017-10-24
THRIFT-4362 check "read length" in readStringBody(int)

ChristianCiach:
This fixes THRIFT-4362.

If possible, please port this fix to previous versions.


jeking3:
We don't re-release previous versions.  You will need to apply the patch to your own copy of a previous version.

ChristianCiach:
So there won't be a version 0.9.4 or 0.10.1?

jeking3:
We only have resources on this project to handle one release stream (and even then it's a challenge), so we're starting the process of making 0.11.0 with all the changes since 0.10.0 right now.

PR #1399, opened 2017-10-25
THRIFT-4370: build generated code before running static code analysis…

jeking3:
… - fix E722 flake8 issues identified in python code

### DIFF
--- a/lib/py/setup.py
+++ b/lib/py/setup.py
@@ -22,7 +22,7 @@
 import sys
 try:
     from setuptools import setup, Extension
-except:
+except Exception:
     from distutils.core import setup, Extension
 
 from distutils.command.build_ext import build_ext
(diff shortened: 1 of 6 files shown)


lib/py/setup.py
```diff
@@ -22,7 +22,7 @@
 import sys
 try:
     from setuptools import setup, Extension
-except:
+except Exception:
```
RobberPhex: I think we should use distutils.core, and setuptools is feedback,
And we should only process ImportError.
```py
try:
     from distutils.core import setup, Extension
except ImportError:
    from setuptools import setup, Extension
```

jeking3: I have to resubmit anyway to kick off another PR build, but I'd like to leave it as I changed it for now, and it can be improved later through a subsequent PR.  I was just trying to eliminate those pesky E722 errors in flake8.  There are a lot more issues in generated code that flake8 pulls out.

PR #1401, opened 2017-10-25
THRIFT-4373: Derefer PHP zval _TSPEC

sokac:


lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
```diff
@@ -537,6 +537,9 @@ void binary_deserialize(int8_t thrift_typeID, PHPInputTransport& transport, zval
       }
 
       zval* spec = zend_read_static_property(Z_OBJCE_P(return_value), "_TSPEC", sizeof("_TSPEC")-1, false);
+      if (spec) {
```
RobberPhex: After `zend_read_static_property` with `slient=false`, we should detect exception.
If no exception, we needn't detect spec is null.

but we need detect that `Z_TYPE_P(prop) == IS_REFERENCE`

lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
```diff
@@ -699,6 +702,9 @@ void binary_serialize_hashtable_key(int8_t keytype, PHPOutputTransport& transpor
 
 static
 void binary_serialize(int8_t thrift_typeID, PHPOutputTransport& transport, zval* value, HashTable* fieldspec) {
+  if (value) {
```
RobberPhex: `Z_TYPE_P(prop) == IS_REFERENCE`

lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
```diff
@@ -709,6 +715,9 @@ void binary_serialize(int8_t thrift_typeID, PHPOutputTransport& transport, zval*
         throw_tprotocolexception("Attempt to send non-object type as a T_STRUCT", INVALID_DATA);
       }
       zval* spec = zend_read_static_property(Z_OBJCE_P(value), "_TSPEC", sizeof("_TSPEC")-1, true);
+      if (spec) {
```
RobberPhex: I think `Z_TYPE_P(prop) == IS_REFERENCE`

RobberPhex:
If we want process REFERENCE everywhere.
I think we should review every line of `zval` used, is it?

lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
sokac: ZVAL_DEREF checks is reference:
https://github.com/php/php-src/blob/ffc734b2a455e2f2748f18a54ab597f7e0c715ba/Zend/zend_types.h#L944

I removed if statement

lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
sokac: removed if statement

lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp
sokac: removed if statement

sokac:
@RobberPhex how does this look now? I don't have lots of context for other zval values so I'd leave that for another review.

james-johnston-thumbtack:
@RobberPhex  Any word on getting this merged?  Using the extension in PHP 7.0 is kinda broken otherwise... this is fixing a common issue that affected many PHP extensions moving to PHP 7.0.

jeking3:
Need to see if pass the CI build - please squash to a single commit, rebase on master, and force push.

sokac:
@jeking3 squashed and rebased. Looks like CI has some troubles, though

PR #1402, opened 2017-10-25
THRIFT-4372 Pipe write operations across a network are limited to 65,…

Jens-G:
…535 bytes per write

Client: C#
Patch: Jens Geyer

### DIFF
--- a/lib/csharp/src/Transport/TNamedPipeClientTransport.cs
+++ b/lib/csharp/src/Transport/TNamedPipeClientTransport.cs
@@ -21,6 +21,7 @@
  * details.
  */
 
+using System;
 using System.IO.Pipes;
 using System.Threading;
 
@@ -88,7 +89,18 @@ namespace Thrift.Transport
                 throw new TTransportException(TTransportException.ExceptionType.NotOpen);
             }
 
-            client.Write(buf, off, len);
+            // if necessary, send the data in chunks
+            // there's a system limit around 0x10000 bytes that we hit otherwise
+            // MSDN: "Pipe write operations across a network are limited to 65,535 bytes per write. For more information regarding pipes, see the Remarks section."
+            var nBytes = Math.Min(len, 15 * 4096);  // 16 would exceed the limit
+            while (nBytes > 0)
+            {
+                client.Write(buf, off, nBytes);
+
+                off += nBytes;
+                len -= nBytes;
+                nBytes = Math.Min(len, nBytes);
+            }
         }
 
         protected override void Dispose(bool disposing)
--- a/lib/csharp/src/Transport/TNamedPipeServerTransport.cs
+++ b/lib/csharp/src/Transport/TNamedPipeServerTransport.cs
@@ -239,40 +239,51 @@ namespace Thrift.Transport
                     throw new TTransportException(TTransportException.ExceptionType.NotOpen);
                 }
 
-                if (asyncMode)
+                // if necessary, send the data in chunks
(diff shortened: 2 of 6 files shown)


lib/csharp/src/Transport/TNamedPipeClientTransport.cs
```diff
@@ -88,7 +89,18 @@ public override void Write(byte[] buf, int off, int len)
                 throw new TTransportException(TTransportException.ExceptionType.NotOpen);
             }
 
-            client.Write(buf, off, len);
+            // if necessary, send the data in chunks
+            // there's a system limit around 0x10000 bytes that we hit otherwise
+            // MSDN: "Pipe write operations across a network are limited to 65,535 bytes per write. For more information regarding pipes, see the Remarks section."
+            var nBytes = Math.Min(len, 15 * 4096);  // 16 would exceed the limit
```
jeking3: Why not actually use `(2^16)-1` which is the limit?

lib/csharp/src/Transport/TNamedPipeClientTransport.cs
```diff
@@ -88,7 +89,18 @@ public override void Write(byte[] buf, int off, int len)
...
+            while (nBytes > 0)
+            {
+                client.Write(buf, off, nBytes);
+
+                off += nBytes;
+                len -= nBytes;
+                nBytes = Math.Min(len, nBytes);
```
jeking3: Shouldn't this be the same calculation as the one outside the loop?  Technically it doesn't have to be, however I found this confusing to read.

lib/csharp/src/Transport/TNamedPipeServerTransport.cs
```diff
@@ -239,40 +239,51 @@ public override void Write(byte[] buf, int off, int len)
                     throw new TTransportException(TTransportException.ExceptionType.NotOpen);
                 }
 
-                if (asyncMode)
+                // if necessary, send the data in chunks
+                // there's a system limit around 0x10000 bytes that we hit otherwise
+                // MSDN: "Pipe write operations across a network are limited to 65,535 bytes per write. For more information regarding pipes, see the Remarks section."
+                var nBytes = Math.Min(len, 15 * 4096);  // 16 would exceed the limit
```
jeking3: Why not actually use (2^16)-1 which is the limit?

lib/csharp/src/Transport/TNamedPipeServerTransport.cs
```diff
@@ -239,40 +239,51 @@ public override void Write(byte[] buf, int off, int len)
...
+                    else
+                    {
+                        stream.Write(buf, off, nBytes);
+                    }
 
-                    if (eOuter != null)
-                        throw eOuter; // rethrow exception
+                    off += nBytes;
+                    len -= nBytes;
+                    nBytes = Math.Min(len, nBytes);
```
jeking3: Shouldn't this be the same calculation as the one outside the loop?  Technically it doesn't have to be, however I found this confusing to read.

Jens-G:
> Why not actually use (2^16)-1 which is the limit?

Several reasons. First, aligned memory access is always faster. If we subtract 1 byte, we get the worst case. Next, at least on Windows a number of system resources use multiples of 4096, so it is probably not a bad idea to follow that model. That's why I picked this particular value.

jeking3:
Makes sense, thanks for the info.  Every little bit helps us all improve what we do next.

PR #1407, opened 2017-10-29
THRIFT-4377: close socket after processing request

oprudkyi:


jeking3:
@Jens-G I see only one of the three commits here committed in master?

oprudkyi:
@jeking3 , Sorry, there was my fault - haxe php integration was broken by first commit and I added fixes

jeking3:
I wonder if this is why sleep 10 is needed between haxe tests and whether that could be improved now?  I'll apply the other ones and see. :)

PR #1409, opened 2017-10-30
THRIFT-4378: add set method for stopTimeoutUnit in TThreadPoolServer java lib

itegel:


itegel:
Thank you! You are right ! It only works on line 218 and that is almost useless. I will try to fix what you suggested soon.

itegel:
I checked the tests, and find nothing to fix. For old version, it is hard coded the time unit to SECONDS and for default args.stopTimeoutUnit is set to TimeUnit.SECONDS. So it has not change the default behavior.

PR #1410, opened 2017-11-03
THRIFT-82: Add Common Lisp support

uint:
This adds Common Lisp support - library, generator, cross-tests, tutorial code, etc. Currently only SBCL is supported.

The minimal requirements are fulfilled. Cross-tests pass against most languages that I've tried.

### DIFF
--- a/lib/cl/framed-transport.lisp
+++ b/lib/cl/framed-transport.lisp
@@ -1,7 +1,5 @@
 (in-package #:org.apache.thrift.implementation)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/lib/cl/load-locally.lisp
+++ b/lib/cl/load-locally.lisp
@@ -1,7 +1,5 @@
 (in-package #:cl-user)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/lib/cl/test/make-test-binary.lisp
+++ b/lib/cl/test/make-test-binary.lisp
@@ -1,5 +1,3 @@
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/test/cl/implementation.lisp
+++ b/test/cl/implementation.lisp
@@ -1,7 +1,5 @@
 (in-package #:thrift.test-implementation)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/test/cl/make-test-client.lisp
+++ b/test/cl/make-test-client.lisp
@@ -1,7 +1,5 @@
 (in-package #:cl-user)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/test/cl/make-test-server.lisp
+++ b/test/cl/make-test-server.lisp
@@ -1,7 +1,5 @@
 (in-package #:cl-user)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/test/cl/tests.lisp
+++ b/test/cl/tests.lisp
@@ -1,7 +1,5 @@
 (in-package #:thrift-cross)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/tutorial/cl/make-tutorial-client.lisp
+++ b/tutorial/cl/make-tutorial-client.lisp
@@ -1,7 +1,5 @@
 (in-package #:cl-user)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/tutorial/cl/make-tutorial-server.lisp
+++ b/tutorial/cl/make-tutorial-server.lisp
@@ -1,7 +1,5 @@
 (in-package #:cl-user)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;;; you may not use this file except in compliance with the License.
 ;;;; You may obtain a copy of the License at
--- a/tutorial/cl/shared-implementation.lisp
+++ b/tutorial/cl/shared-implementation.lisp
@@ -1,7 +1,5 @@
 (in-package #:shared-implementation)
 
-;;;; Copyright 2017 Rigetti Computing <rigetti.com>
-;;;;
 ;;;; Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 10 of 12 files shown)


jeking3:
Some initial thoughts, before I review all of the files individually:

1. Please squash.
2. Please add [THRIFT-82] at the beginning of the commit description and the pull request description.
3. Please review https://thrift.apache.org/docs/HowToContribute.

I resurrected https://issues.apache.org/jira/browse/THRIFT-82 upon seeing this.  I'll start going through the files.

jeking3:
@jfarrell for some reason there is no recorded Travis CI build for this pull request.  When the author squashes and resubmits hopefully this will kick off a proper build?

compiler/cpp/src/thrift/generate/t_cl_generator.cc
```diff
@@ -0,0 +1,544 @@
+// Copyright (c) 2008- Patrick Collison <patrick@collison.ie>
```
jeking3: This needs the standard Apache Software License statement applied.

lib/cl/externals/bundle-info.sexp
```diff
@@ -0,0 +1,14 @@
+(:CREATION-TIME #A((20) BASE-CHAR . "2017-10-31T11:49:23Z") :REQUESTED-SYSTEMS
```
jeking3: Is this a file that was supposed to be checked in?  It looks like it is machine-specific.

lib/cl/externals/bundle.lisp
```diff
@@ -0,0 +1,161 @@
+(cl:in-package #:cl-user)
```
jeking3: Missing license statement.

lib/cl/externals/software/alexandria-20170830-git/.boring
```diff
@@ -0,0 +1,13 @@
+# Boring file regexps:
```
jeking3: If you checked in alexandria, it should not be part of the submission.  Instead, the build process should install it on demand in the same manner that the nodejs implementation calls npm to install packages during the build.

lib/cl/externals/software/bordeaux-threads-v0.8.5/.travis.yml
```diff
@@ -0,0 +1,44 @@
+language: lisp
```
jeking3: If you checked in bordeaux-threads, it should not be part of the submission.  Instead, the build process should install it on demand in the same manner that the nodejs implementation calls npm to install packages during the build.

lib/cl/externals/software/usocket-0.7.0.1/vendor/OpenTransportUDP.lisp
```diff
@@ -0,0 +1,146 @@
+;;;-*-Mode: LISP; Package: CCL -*-
```
jeking3: As with the other libraries, this should not be checked in.  I have skipped over all of the externals/ due to this.

lib/cl/framed-transport.lisp
```diff
@@ -0,0 +1,136 @@
+(in-package #:org.apache.thrift.implementation)
+
+;;;; Copyright 2017 Rigetti Computing <rigetti.com>
```
jeking3: @jfarrell do we need to check with them to make sure this is okay to submit?

lib/cl/binary-protocol.lisp
```diff
@@ -0,0 +1,255 @@
+(in-package #:org.apache.thrift.implementation)
+
+;;;; This file defines the concrete `binary-protocol` layer for the `org.apache.thrift` library.
+;;;;
+;;;; copyright 2010 [james anderson](james.anderson@setf.de)
```
jeking3: @jfarrell do we need to check with them to make sure this is okay to submit?

lib/cl/externals/bundle-info.sexp
dkochmanski: all things put in `externals` are checked in, so whole directory will be removed (this file included).

lib/cl/framed-transport.lisp
dkochmanski: Our work on Thrift is sponsored by Rigetti, so there is no need for that.

dkochmanski:
Thank you for taking your time for the review

uint:
So about removing the bundled dependencies (`lib/cl/externals`).

There is a nifty package manager for CL (Quicklisp) that works a bit like, say, Cargo for Rust. It's commonly used, but the problem is it isn't available in `PATH` or anything - it's installed in the user's dir normally. There's no good way to detect its presence while building Thrift.

What if we included the Quicklisp installer (literally a single `.lisp` file), installed it locally during the build, and then used it to pull in other dependencies? Would that be fine?

jeking3:
I think it would be better to make a shell script that downloads quicklisp and runs it to install the dependencies, and make the local build in lib/cl depend on successful execution of the shell script.  This makes it similar to the nodejs build which runs npm to do something similar, however the only difference is that npm is installed as part of the docker build.
.

dkochmanski:
Hey, removed all externals and added `ensure-externals.sh` script. Also squashed all commits.

lib/cl/ensure-externals.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+if [ ! -d "externals/" ] ; then
```
jeking3: What happens if externals are partially populated due to a crash?  Can one run the same quicklisp command idempotently?  If so, I would remove this if statement and just always run quicklisp to make sure externals are correct.  You could perhaps make the curl command smarter to only update the file if it has changed - I suspect curl has an option for that, but I don't know...

lib/cl/server.lisp
```diff
@@ -0,0 +1,230 @@
+(in-package #:org.apache.thrift.implementation)
```
jeking3: Does one need the ".implementation" in the namespace?

jeking3:
Without a linux build, won't be able to do any merges...

jfarrell:
Travis requests show an error with "Abuse detected". I've emailed Travis support asking them to look into the issue

lib/cl/framed-transport.lisp
Jens-G: > Our work on Thrift is sponsored by Rigetti, so there is no need for that.

That sounds ... strange.

Jens-G:
@jfarrell: What's your opinion re the (c) things above?

lib/cl/framed-transport.lisp
dkochmanski: Rigetti hired us to implement (and possibly merge) CL support to Thrift. All software under this delivery has to have license required by the project we are contributing to (that is Thrift). Not sure what is strange about that.

lib/cl/server.lisp
dkochmanski: it is a common practice in Common Lisp to have a package dedicated for usage by a programmer (which exports the protocol symbols - library API that is) and having no other symbols of its own and an implementation package which may have other internal symbols and interfaces. This allows avoiding for instance symbol name conflicts (for instance list vs list).

lib/cl/ensure-externals.sh
dkochmanski: it is possible. Fixed in the pushed commit (some extra eval forms are explained in the commit message.

lib/cl/framed-transport.lisp
Jens-G: Well, the term "sponsored" is a bit broad to derive that particular conclusion from it. So I asked for further clarification.

jfarrell: @dkochmanski thank you for working on this and contributing it back the the Apache Thrift. The Apache license header on this should not contain a copyright to Rigetti Computing, details available at https://www.apache.org/legal/src-headers.html#headers part 2

dkochmanski: Right, thank you for pointing me to the document. Last commit removes all copyrights added by us. Note though, that files in lib/cl/ are based on work by James Anderson (and have copyrights as well) and I don't think I can remove them. Should I contact him and ask whenever we can remove his copyright stings too?

jfarrell: That would be great. is it possible fetch his work as a third party downloadable dependency or was it used as the basis for work you did on top of it?

jfarrell:
@dkochmanski Travis support responded saying that no one from TurtleWarePL has logged into https://travis-ci.org/ ever and this is the reason this PR is not getting run within travis. 

From Travis support:

> do you think one of their members could try to log into Travis CI at least once at https://travis-ci.org/ and try closing and reopen one of the rejected Pull Requests?

jeking3:
Why don't we close this PR and open a new one free of any copyright issues or inclusion of third party code.  That should resolve the issue?

dkochmanski:
@jfarrell logged into travis a moment ago for the first time.

@jeking3 I'm waiting for a response from @lisp if we can remove his copyright strings (we've removed ours). Should I close the PR and issue a new one before I receive a reply?

jeking3:
Let's follow @jfarrell 's lead.  Whatever he says, he's managing the travis interaction.

dkochmanski:
closing, will reopen with squashed commits.

dkochmanski:
see https://github.com/apache/thrift/pull/1412

lisp:
good evening; this popped up in my mail due to  mr kochmanski's reference.
what are you waiting for from me?

dkochmanski:
Hello, I've send you an email a few hours ago with a question, if we can remove your "copyright" headers from the code found in de.setf.thrift, because that is what is required by Thrift team (I have included details in the email). Plese see: https://github.com/apache/thrift/pull/1410#discussion_r149655532 .

PR #1411, opened 2017-11-07
Fix remote client for HTTP transport

trotterdylan:
This change fixes two issues:

- Assign parsedUrl to the variable in the outer scope instead of creating a new one. Previously the outer parsedUrl was never assigned and was therefore always empty.
- Create a POST client using NewTHttpPostClient instead of NewTHttpClient since the latter has no requestBuffer to write to the request body.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -2130,7 +2130,7 @@ void t_go_generator::generate_service_remote(t_service* tservice) {
   f_remote << indent() << "var urlString string" << endl;
   f_remote << indent() << "var framed bool" << endl;
   f_remote << indent() << "var useHttp bool" << endl;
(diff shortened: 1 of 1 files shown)


jeking3:
He @trotterdylan please review https://thrift.apache.org/docs/HowToContribute as we need a Jira ticket for this change.

trotterdylan:
Done! https://issues.apache.org/jira/browse/THRIFT-4385

dcelasun:
Please revert the `NewTHttpClient` part of this PR, since `THttpPostClient` is deprecated and is just an alias for `THttpClient` since 0dd82358.

trotterdylan:
Done. Updated the commit message accordingly.

jeking3:
Thanks - once the CI builds pass I will re-review.

jeking3:
@dcelasun can you approve if these changes look good to you?

dcelasun:
@jeking3 LGTM

PR #1418, opened 2017-11-27
Thrift 3877 http server buffering bug oneway

chetmurthy:
This PR is in two commits, so that the reviewer can

(a) apply the testcase commit, elicit the failure by running lib/cpp/tests/UnitTests

(b) apply the (full) second commit, rerun testcase, and see the problem is gone.


test/ThriftTest.thrift
```diff
@@ -409,3 +409,8 @@ struct StructB {
   1: optional StructA aa;
   2: required StructA ab;
 }
+
+service OneWayTest {
```
jeking3: Could we move OneWayTest to its own thrift file, which could be located inside lib/cpp/test because it is only used for that unit test?  I can help in making sure the Makefile.am and the CMakeLists.txt are updated appropriately.

chetmurthy:
Will do.  Please do suggest any style changes that you think appropriate -- I am *fully* bought-into the idea that a project should have a uniform style, and even if I were not to agree with that style, I'd go along with it, b/c ... uniformity has value.

To that end, should I add a block-comment to the test, explaining what the classes do, what's the idea is behind it, etc?

jeking3:
I think it's fine as-is, except for adding to ThriftTest in this case since that is a global test thrift file, and you need one local to lib/cpp/test.  The fact you took the time to add a unit test and did it in a very TDD way makes me quite happy. :)

chetmurthy:
I looked at the AppVeyor logs failing jobs, and cannot see why they failed.  It looks like they just ran outta time?  I have no idea what to do to get these jos to re-run -- could someone enlighten me?

chetmurthy:
OK, I see that the Travis CMake builds all failed.  I figured out how to run cmake on my laptop, but .... I have no idea how to add the dependency that it's asking for.

chetmurthy:
OK, I figured out how to fix cmakelists.txt, to get most of the cmake builds working.  But one still fails.  Then I tried to set up and run docker, to debug that.  And can't figure out how from the instructions here: https://github.com/apache/thrift/blob/master/build/docker/README.md

to no avail.

jeking3:
@chetmurthy try "Raw Commands for Building with Docker" at that page.

Once you are in a docker container:

build/docker/scripts/autotools.sh runs a "make check" build
build/docker/scripts/crosstest.sh runs a "make cross" build

chetmurthy:
Right, the "docker run" command fails with:

chet@twitter:~/Hack/thrift-0.10.0/src/forks/thrift$ dockerrun ubuntu-xenial
Unable to find image 'ubuntu-xenial:latest' locally
docker: Error response from daemon: repository ubuntu-xenial not found: does not exist or no pull access.
See 'docker run --help'.
chet@twitter:~/Hack/thrift-0.10.0/src/forks/thrift$ 

the "refresh.sh" cmd worked fine.

jeking3:
Pulling this into a sandbox to see if it fixes any of the `cpp*http` known test failures.

jeking3:
I'm not certain this resolves THRIFT-3877 entirely - your take?

chetmurthy:
This patch only flxes cpp-cpp (http) failures. B/c the deeper issue is that the Thrift cpp stack (both client & server) is nonstandard.

I think I mentioned that I could improve this a bit, by having the client not wait for the reply from the server, but the server would eventually deliver it anyway.  The client would keep track of of how many replies it had ignored, and when it came time to actually want to see a reply, it woud skip past the requisite # of ignored replies.  Also (of course), you'd want to only ignore replies up to some limit -- or you could deadlock.

Not very hard to do, actually.  And this would also fix cpp(client)-http-<anylang>(server) tests.  BUT it would NOT fix <anylang>(client)-http<anylang>(server) tests, b/c (from what I remember of my little stroll thru code) (almost?) all the other languages' HTTP stacks assume an RPC model for how applications will use the client HTTP stack.  So they're written with "send request, immediately turn around and wait for response" as the model.

Hope this helps.

PR #1420, opened 2017-11-29
THRIFT-4395: fix rust build on xenial

jeking3:
I found this while doing due diligence on another issue (C++ HTTP) - I could no longer run cross test because builds fail in lib/rs.


jeking3:
shoot... will resubmit without THRIFT-3877

PR #1421, opened 2017-11-29
THRIFT-4395: fix rust build on xenial

jeking3:
This will cause all CI builds to fail until resolved.


jeking3:
Well thank goodness for CI builds, I forgot to update the tutorial package file!

PR #1423, opened 2017-11-30
THRIFT-3580 THeader for Haskell (Resubmitting for nsuke)

jeking3:
This closes #820


jeking3:
Seems to be broken when you run the csharp client against it, so I will add that to the exclusions and commit it.

PR #1427, opened 2017-11-30
THRIFT-3657 D TFileWriterTransport close should use non-priority send

jeking3:
Client: D

(resubmit for nsuke)

This closes #884
This closes #1427


jeking3:
Failed due to THRIFT-2913, ignoring that failure.

PR #1430, opened 2017-11-30
Fix generated maven packaging

slandelle:
pom packaging is for empty modules. What you're deploying on maven central is a jar artifact. Your current invalid pom.xml causes build tools such as sbt to not be able to download your library!


Jens-G:
Hi @slandelle, 
thank you for the patch!

[If you could also create a ticket for it as well](https://thrift.apache.org/docs/committers/HowToRelease), that would be awesome.

PR #1433, opened 2017-12-02
fix plugin.thrift, plugin{,output}.cc to use t_const_value correctly

chetmurthy:
[This is a minor bug.  Since it changes types in plugin.thrift, it's not possible to write a before-and-after bug. But conversion_test.cc currently tests this behaviour, so the change in conversion_test.cc is, effectively, the before-and-after test.]

plugin.thrift defines t_const_value as a union. But in plugin_output.cc and plugin.cc, the converters clearly either (a) SET NEITHER of identifier_val & enum_val, or (b) SET BOTH. But these are two different fields in the union. So clearly, the type t_const_value isn't being treated as a union.
I think we need to fix Thrift's treatment of unions, but independently, the plugin should use Thrift's type system in a correct manner. This is easy-to-fix, but since the current plugin relies on a bug, the fix will be a breaking change.

### DIFF
--- a/compiler/cpp/test/plugin/conversion_test.cc
+++ b/compiler/cpp/test/plugin/conversion_test.cc
@@ -269,6 +269,12 @@ void test_const_value(t_const_value* sut) {
     T_CONST_VALUE_CASE(CV_INTEGER, integer);
     T_CONST_VALUE_CASE(CV_DOUBLE, double);
     T_CONST_VALUE_CASE(CV_STRING, string);
+
+#if 0
+  case t_const_value::CV_IDENTIFIER:
+    BOOST_CHECK_EQUAL(sut->const_identifer_val.identifier, sut2->const_identifer_val.identifier)
+      break ;
+#endif
     T_CONST_VALUE_CASE(CV_IDENTIFIER, identifier);
 #undef T_CONST_VALUE_CASE
   case t_const_value::CV_MAP:
--- a/compiler/cpp/src/thrift/plugin/plugin.cc
+++ b/compiler/cpp/src/thrift/plugin/plugin.cc
@@ -258,11 +258,11 @@ THRIFT_CONVERSION(t_const_value, ) {
     T_CONST_VALUE_CASE(string);
   else T_CONST_VALUE_CASE(integer);
   else T_CONST_VALUE_CASE(double);
-  else {
-    T_CONST_VALUE_CASE(identifier);
-    if (from.__isset.enum_val)
-      to->set_enum(resolve_type< ::t_enum>(from.enum_val));
+  else if (from.__isset.const_identifier_val) {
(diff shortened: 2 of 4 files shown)


jeking3:
Please file a Jira ticket per the contribution guidelines.

chetmurthy:
Oh oops, I didn't set the commit-message correctly!  I'll delete and re-submit.  sorry sorry.

compiler/cpp/test/plugin/conversion_test.cc
```diff
@@ -269,6 +269,12 @@ void test_const_value(t_const_value* sut) {
     T_CONST_VALUE_CASE(CV_INTEGER, integer);
     T_CONST_VALUE_CASE(CV_DOUBLE, double);
     T_CONST_VALUE_CASE(CV_STRING, string);
+
+#if 0
```
jeking3: Don't check in dead code.

jeking3:
Is this a breaking change?
Do you use the compiler plug-in api?
I wasn't sure anybody did.

chetmurthy:
If nobody uses the compiler plug-in API, then it's not a breaking change *grin*.  But I think it's a great API to have -- it allows people to develop new Thrifcompiler backends without having to understand all of the current compiler.

And of course, to write those compiler backends in other languages.  I'm in the process of writing a new backend in Ocaml.

jeking3:
I was going to recommend we remove the compiler plug-in API OR remove the monolithic build of the compiler.  It complicates the project, in my opinion unnecessarily, unless we ONLY use the plug-in method.  Having to maintain the "option" of using a plug-in is a build issue for me.

This project is complex enough as it is with 20+ languages to support...
We have a long term goal to move everything to CMake.  Lots of work.  PLUGIN adds branches to that work.

chetmurthy:
Um .... not that I get a vote (*grin*) but I'd go for removing the monolithic build.  Specifically, I feel that writing codegen in C++ (esp. for other languages) is quite cumbersome, and gets in the way of improving those codegens.  I'd be willing to do the work to make that happen.  The point of the plugin, after all, is that you -can- do that, and easily.

Or at least, that's the idea, and I'd be happy to put my money where my mouth is

Jens-G:
> , unless we ONLY use the plug-in method

I personally don't think that is a good idea. I just went through that pain with grpc a while ago and I remember it as being utterly complicated to set up. Monolithic is good.

chetmurthy:
I should have added that another reason I want to plugin interface to ocntinue existing, is that without it, I'd have to invent one, in order to implement the idiomatic JSON wire protocol that I just wrote.  The Thrift generated-code "contract" doesn't have enough information to implement things like an idiomatic JSON wire.  Or YAML.  But with extra type-information, it's a straightforward matter.  And the plugin provides that on-the-nose.

PR #1435, opened 2017-12-02
THRIFT-4399 plugin.thrift t_const_value is not used as a union in C++…

chetmurthy:
… code -- fix this

plugin.thrift defines t_const_value as a union.  But in plugin_output.cc
and plugin.cc, the converters clearly either (a) SET NEITHER of
identifier_val & enum_val, or (b) SET BOTH.  But these are two different
fields in the union.  So clearly, the type t_const_value isn't being
treated as a union.

I think we need to fix Thrift's treatment of unions, but independently,
the plugin should use Thrift's type system in a correct manner.  This is
easy-to-fix, but since the current plugin relies on a bug, the fix will
be a breaking change.

### DIFF
--- a/compiler/cpp/src/thrift/plugin/plugin.cc
+++ b/compiler/cpp/src/thrift/plugin/plugin.cc
@@ -258,11 +258,11 @@ THRIFT_CONVERSION(t_const_value, ) {
     T_CONST_VALUE_CASE(string);
   else T_CONST_VALUE_CASE(integer);
   else T_CONST_VALUE_CASE(double);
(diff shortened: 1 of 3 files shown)


chetmurthy:
BTW, this fix is necessary if I'm going to fix unions so they transport exactly one of their fields.  B/c (of course) right now, the plugin code depends on that not being the case.

I have the C++ code (compiler + lib) ready-to-go-and-tested, but (obv) can't submit a PR until this code goes thru, b/c it would break the build.

PR #1437, opened 2017-12-04
THRIFT-4403: Fix incorrect usage of 'this'

bananer:


jeking3:
Great - I am going to see if this resolves any of the cross test known failures.

lib/js/src/thrift.js
```diff
@@ -599,8 +599,8 @@ Thrift.TWebSocketTransport.prototype = {
           //If the user made calls before the connection was fully
           //open, send them now
           this.send_pending.forEach(function(elem) {
-             this.socket.send(elem.buf);
-             this.callbacks.push((function() {
+             self.socket.send(elem.buf);
```
jeking3: On line 597, self is set to be equal to this, so what does this change actually do?

bananer: In short, `this` inside of a `function()` in JavaScript is not transferred reliably from outer scopes. Try setting a breakpoint at this line and you will see that `self` and `this` are different things here.

It's a bit of a race condition to have this block actually execute, but calling a service method right after opening the transport crashed here repeatedly in firefox.

Long version: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

jeking3: Perhaps line 597 should be removed then?

jeking3: I'm running through cross tests to see if this resolves anything we know failed in the past.

jeking3: We don't run "js" through cross test right now. :|

PR #1440, opened 2017-12-09
Enhancement binary_protocol with frametransport

zhiyu-he:
for frametransport
write method write data-length to server
read method read data-length from server

and there are lots of un-usable memory copy & system call 

i optimize this with an simple buffer, manager de write & read
with the benchmark for write improve 2times & for read improve 3times

link-for testing： https://github.com/zhiyu-he/go_performance/blob/hzy/modify/benchmark/thrift_serializer_test.go

### DIFF
--- a/lib/go/thrift/fast_buffer.go
+++ b/lib/go/thrift/fast_buffer.go
@@ -1,0 +1,257 @@
+package thrift
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+type FastFrameBuffer struct {
+	w         *TFramedTransport //
+	b         []byte
+	wIdx      int               // write-offset, reset when Flush
+	frameSize int
+	rIdx      int     // read-offset, reset when read-new-frame
+	buf       [4]byte // only for write&read data-len
+}
+
+func NewFastBuffer(w *TFramedTransport, size int) *FastFrameBuffer {
+	return &FastFrameBuffer{
+		w:    w,
+		b:    make([]byte, size),
+		rIdx: 0,
+		wIdx: 0,
+	}
+}
+
+func (p *FastFrameBuffer) WritByte(b int8) {
+	if p.wIdx+1 > p.Cap() {
+		p.grow(2*p.Cap() + 1)
+	}
+	p.b[p.wIdx] = byte(b)
+	p.wIdx += 1
+}
+
+func (p *FastFrameBuffer) WriteBytes(b []byte) {
+	if p.wIdx+len(b) > p.Cap() {
+		p.grow(2*p.Cap() + len(b))
+	}
+	copy(p.b[p.wIdx:], b)
+	p.wIdx += len(b)
+}
+
+func (p *FastFrameBuffer) WriteI16(i uint16) {
+	if p.wIdx+2 > p.Cap() {
+		p.grow(2*p.Cap() + 2)
+	}
+	copy(p.b[p.wIdx:], []byte{byte(i >> 8), byte(i)})
+	p.wIdx += 2
+}
+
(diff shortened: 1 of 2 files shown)


jeking3:
Transport and Protocols are supposed to be distinct layers in the Thrift architecture; why did you combine the Binary Transport with the Framed Protocol?

jeking3:
@jfarrell could you please look into why Travis CI didn't execute on this pull request?  It wouldn't have mattered, since this PR has no unit test or cross test that engages it at the moment, but still... it should have run a build.

zhiyu-he:
@jeking3 
but the difference between buffer-transport and frame_transport， is only the data-length.
i think thrift transport only use for raw-socket read&write.

within golang, it's inefficient and cost more cpu time.

 in binary_protocol,  e.g for int64,  int64->temp-buf(binary_protol)->buf(transport-buf)->socket
this is not go enough for efficiency
so i want optimize this part。

Jens-G:
Plus chinese comments should be english.

jeking3:
My initial take on this is that the conflation of transport and protocol in general is not a good thing for Thrift.  It limits the flexibility to use one interchangeably with something else, but it's hard to overlook a 2x to 3x improvement in performance.  Is there any way some of these techniques could be applied back to the original separate classes for these types?

zhiyu-he:
@jeking3 with go programing, [the function call stack may more expensive.](https://github.com/zhiyu-he/go_performance/blob/master/build_in/func_call_test.go)

jeking3:
I would need to see comments from other folks who use thrift go, and also I need to see a linux build pass, so please rebase on master and force push to kick a new build.

lib/go/thrift/fast_buffer.go
```diff
@@ -0,0 +1,257 @@
...
+		l, err := p.w.transport.Read(p.b[i:])
+		if err != nil {
+			return 0, err
+		}
+		i += l
+	}
+	return i, nil
+}
+
+func str2bytes(s string) []byte {
```
dcelasun: At the very minimum, this file needs an `!appengine` build tag and a separate, slower implementation without using `unsafe`, so it can be used in all platforms supported by Go.

lib/go/thrift/fast_buffer.go
```diff
@@ -0,0 +1,257 @@
...
+		return 0, err
+	}
+	frameSize := int(uint32(p.buf[3]) | uint32(p.buf[2])<<8 | uint32(p.buf[1])<<16 | uint32(p.buf[0])<<24)
+	if frameSize < 0 || frameSize > DEFAULT_MAX_LENGTH {
+		return 0, fmt.Errorf("Incorrect frame size (%d)", frameSize)
+	}
+	return frameSize, nil
+}
+
+// TODO 这个方法 是否会有问题？ block? or?
```
dcelasun: All comments need to be in English.

lib/go/thrift/fast_buffer.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			return
+		}
+		_, err = p.readAll(p.frameSize)
+		if err != nil {
+			return
+		}
+	}
+	if p.frameSize < 1 {
+		return 0, fmt.Errorf("Not enought frame size %d to read %d bytes", p.frameSize, 1)
```
dcelasun: s/enought/enough everywhere in the file.

dcelasun:
From a cursory look, while the performance improvements are decent and only small code changes are needed, this doesn't really fit into our current architecture.

Since the new types don't depend on any unexported types we define, I think this should be a separate library.

jeking3:
For reasons stated above I am going to close this PR.  Thanks.

PR #1445, opened 2017-12-13
THRIFT-4423 migrate to psr-4

RobberPhex:
ref <https://issues.apache.org/jira/browse/THRIFT-4423>

### DIFF
--- a/lib/php/test/Makefile.am
+++ b/lib/php/test/Makefile.am
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-PHPUNIT=php ./../../../vendor/bin/phpunit
+PHPUNIT=php $(top_srcdir)/vendor/bin/phpunit
 
 stubs: ../../../test/ThriftTest.thrift  TestValidators.thrift
 	mkdir -p ./packages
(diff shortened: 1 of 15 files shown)


build/docker/run.sh
```diff
@@ -25,6 +25,8 @@ DOCKER_TAG=$DOCKER_REPO:$DISTRO
 
 printenv | sort
 
+composer install
```
jeking3: This looks like it should be part of the php section of the Dockerfile.  By putting it in the run section, we run "composer install" on the host and then start a docker environment which doesn't have access to it.

build/docker/scripts/sca.sh
```diff
@@ -56,7 +56,3 @@ echo  TODOs: `grep -r TODO * | wc -l`
 
 # LoC
 sloccount .
-
-# System Info
```
jeking3: Removing these lines looks unrelated to the task at hand, and they don't take much time, and they provide useful info.

lib/php/test/Makefile.am
```diff
@@ -17,6 +17,8 @@
 # under the License.
 #
 
+PHPUNIT=./../../../vendor/bin/phpunit
```
jeking3: Is this relative path going to work on every platform?

jeking3:
I'll watch for a passing build, and then merge tomorrow if it's good.

jeking3:
That pesky concurrency test!

PR #1446, opened 2017-12-18
Insert files related to eclipse project to gitignore

dmvolod:


jeking3:
In the future please file a Jira ticket for changes.  Given this is trivial and not distributed in the final product, I will waive off the requirement.

PR #1448, opened 2017-12-22
Thrift-4441: Support compilation without Boost

Typz:
The goal of this series of patches is to support compiling Thrift library without boost: in C++11, most of the things are already integrated, and it is often not practical to integrate boost in embedded projects.

These patch use C++11 alternatives when possible, and default to Boost otherwise.


lib/cpp/src/thrift/stdcxx.h
```diff
@@ -102,23 +108,71 @@ namespace apache { namespace thrift { namespace stdcxx {
...
+
+}}} // apache::thrift::stdcxx
+
+///////////////////////////////////////////////////////////////////
+//
+// Atomic
+//
+///////////////////////////////////////////////////////////////////
+
+#if __cplusplus < 201103L
```
jeking3: This is always true on all Microsoft compilers (2010 - 2017).  It isn't the best thing to be conditional on and be portable, unfortunately.

lib/cpp/src/thrift/processor/TMultiplexedProcessor.h
```diff
@@ -165,10 +166,11 @@ class TMultiplexedProcessor : public TProcessor {
     }
 
     // Extract the service name
-    boost::tokenizer<boost::char_separator<char> > tok(name, boost::char_separator<char>(":"));
-
     std::vector<std::string> tokens;
-    std::copy(tok.begin(), tok.end(), std::back_inserter(tokens));
+    std::istringstream tokenStream(name);
+    for (std::string token; std::getline(tokenStream, token, ':');) {
```
jeking3: getline may only be available in C++11 or later

As a project should we consider calling out 0.11.0 as the last version that will support C++03, and simplify this work?

Typz: std::getline is available also in pre-C++11 (though C++11 adds some overloads): http://en.cppreference.com/w/cpp/string/basic_string/getline

I would be OK with C++11, but this may not be applicable to all cases... Maybe a safer approach would be first allow use without boost (relying on C++11: i.e. this patch :-) ), then add C++11 requirement if this is acceptable.

Typz:
I have an issue with UBSan build: it keeps failing in codecvt, which may be due either to a bug in StdLib or simply to that lib not being compiled with UBSan... Any idea how to overcome this, and let the tests pass?

Typz:
@jeking3 : any suggestion?

appveyor.yml
```diff
@@ -45,7 +45,7 @@ environment:
    - PROFILE: MSVC2015
      PLATFORM: x64
      CONFIGURATION: Release
-     BOOST_VERSION: 1.64.0
```
jeking3: Why?

jeking3:
@Typz the example for this conversion on cppreference is different than your code.  Try that?

http://en.cppreference.com/w/cpp/locale/wstring_convert/to_bytes

appveyor.yml
Typz: VS2015 image does not contain Boost 1.64 : it only contains versions up to 1.63.
1.64 and 1.65.1 are included only in VS2017 image. See https://www.appveyor.com/docs/build-environment/#boost for full details

This is not an issue introduced by this bug, but my patch highlights the problem: currently, this build expect Boost 1.64, which is not present, thus the C++ lib is not built.

With my patches, the C++ lib can be built (even without Boost), but the tests still require Boost: so building in this case would require an extra flag to disable tests.

Since the goal is (I think...) to actually build with a recent version of Boost, I downgraded the version the most recent installed version in the image.

Typz:
@jeking3 : This exemple refers to conversions from 'wide' strings to UTF8 or UTF16. I used this exemple: http://en.cppreference.com/w/cpp/locale/codecvt_utf8_utf16 which shows conversion from UTF16 to UTF8.

According to the name of the function, that should be what we expect; but maybe i was indeed mislead by the name, and it should be simply 'wide' to UTF8 ?

jeking3:
Is this done (for now) with the exception of that build issue, since you removed the [WIP] tag?

Typz:
Indeed, it is complete now: the lib can be built without Boost, both through autoconf or cmake.
Tests (based on boost test) and tutorial still require boost, though.

jeking3:
Would it be possible for you to squash this to make the merge easier?

Typz:
Squashed everything in one big patch.

jeking3:
Thanks, that's incredibly helpful when it comes to merging it.

build/cmake/DefineOptions.cmake
```diff
@@ -91,11 +91,26 @@ if(WITH_CPP)
     option(WITH_STDTHREADS "Build with C++ std::thread support" OFF)
     CMAKE_DEPENDENT_OPTION(WITH_BOOSTTHREADS "Build with Boost threads support" OFF
         "NOT WITH_STDTHREADS;Boost_FOUND" OFF)
+
+    set(WITH_CPP_SUPPORT OFF)
```
jeking3: I think we would be better off testing for C++11 features and ensuring that if we find what we need, we can enable certain things.  For example if we ask cmake to check for support of cxx_defaulted_functions then we can enable the boost-less noncopyable.  One can typically check for cxx_nullptr in order to determine if there is smart pointer support, or write a custom check for std::shared_ptr detection.  Relying on checking compiler versions is error-prone and not as portable.

This is a good start to optionally eliminating boost from the C++ runtime library.  I can work on the feature checks as an extension to this work.

jeking3:
I tried to use this work and did a little bit of cmake work to straighten out options.  What I found is that there is a boost dependency in the safe_numeric_cast in TTransportException.hpp and there's another boost dependency in TProtocol.h (for endian detection) that needs to be resolved.

What I would suggest is that you try building on Windows without boost, disabling build of tests and tutorials, so that all you build is the compiler and the C++ library.

You can see what I did in my referenced pull request in my fork.

jeking3:
We've pretty much cleared the way to a C++11 only thrift release post-0.12.0... whether you want to be involved or not is up to you @typz but we'd be happier if you were. :)

helmesjo:
When can this be expected? That is, when is 0.13 planned? :)

jeking3:
We just made a release on January 4.  Best guess right now would be around summer, but it could be sooner if we come to a consensus on it.  I'd love to get to 3 releases yearly if we could.

PR #1450, opened 2017-12-27
THRIFT-4437: JavaScript WebSocket - fix duplicate callbacks

bananer:


jeking3:
Running autotools.sh locally on xenial to see if it was one of those annoying npm transient failures.

PR #1451, opened 2017-12-28
THRIFT-4438: delete trusty and requrie php56

RobberPhex:
* remove support php 5.3, 5.4 and 5.5
* remove test at ubuntu trusty(trusty only have php5.5, haven't php5.6)


build/docker/ubuntu-trusty/Dockerfile
```diff
@@ -1,223 +0,0 @@
...
-#
-# Apache Thrift Docker build environment for Ubuntu Trusty
-# Using all stock Ubuntu Trusty packaging except for:
-# - d: does not come with Ubuntu so we're installing 2.070.0
-# - dart: does not come with Ubuntu so we're installing 1.20.1
-# - dotnetcore, disabled because netcore is for 1.0.0-preview and 2.0.0 is out
-# - haxe, disabled because the distro comes with 3.0.0 and it cores while installing
-# - node.js, disabled because it is at 0.10.0 in the distro which is too old (need 4+)
-# - ocaml, disabled because it fails to install properly
-#
```
jeking3: Recommend adding that php is disabled because we require 5.6 and Trusty comes with 5.5.9, and comment out the PHP lines, but not delete the whole thing.

composer.json
```diff
@@ -16,12 +16,12 @@
         "issues": "https://issues.apache.org/jira/browse/THRIFT"
     },
     "require": {
-        "php": ">=5.3.0"
+        "php": "^5.6 || ^7.0"
```
jeking3: Just looking at the changes here for dropping php 5.3 through 5.5, what's the benefit?  No php code actually changed to take advantage of this, so I'm not sure I see a reason to do it?

RobberPhex:
Considering that trusty haven't reached end of life, so I close this pr, and keep trusty in CI.

jeking3:
Thanks, given there were no structural changes to php other than dropping 5.3 from the compatibility list, I think this is best for now.

PR #1454, opened 2018-01-02
THRIFT-4436: port nodejs changes from THRIFT-3748 to js lib,

bananer:
test for serialization of nested list,
run all tests when building js lib


jeking3:
Suggest you rebase on master, as the fix for the build issue with dart was merged, and it would be nice to see a clean build.

bananer:
rebased as https://github.com/apache/thrift/pull/1457

PR #1457, opened 2018-01-05
THRIFT-4436: port nodejs changes from THRIFT-3748 to js lib (rebased)

bananer:
test for serialization of nested list,
run all tests when building js lib


bananer:
@jeking3 Looks like again these test failures are unrelated to my changes.

PR #1461, opened 2018-01-08
THRIFT-4447: Golang: Fix panic on p.c.Call when using deprecated initializers

johnboiles:
Latest thrift:master can cause panics when using deprecated `New*ClientFactory` and `New*ClientProtocol` functions. This happens because both the Client and the BaseClient have an instance of a `thrift.TClient`. The deprecated methods initialize the BaseClient's TClient, but other methods use the Client's `TClient`.

For example, current thrift master generates structs like this

```
type MyServiceClient struct {
	c thrift.TClient
	*MyServiceBaseClient
}

type MyServiceBaseClient struct {
	c thrift.TClient
}
```

And also a method like this:

```
func NewMyServiceClientFactory(t thrift.TTransport, f thrift.TProtocolFactory) *MyServiceClient {
      return &MyServiceClient{MyServiceBaseClient: NewMyServiceBaseClientFactory(t, f)}
}
```

If that method is used, later calls to service methods will panic, since `p.c` is nil (the actual client was stored in `p.BaseMyServiceClient.c`).

```
func (p *MyServiceClient) DoStuff(ctx context.Context, request *DoStuffRequest) (r *DoStuffResponse, err error) {
	var _args139 DoStuffArgs
	_args139.Request = request
	var _result140 DoStuffResult
	if err = p.c.Call(ctx, "do_stuff", &_args139, &_result140); err != nil { // PANIC
	...
```

The fix in this PR merely sets both instances of `TClient` (which is what happens in the non-deprecated `New*Client` function).

This patch currently fails `make -k check` however, since `src/tutorial/tutorial.go` tries to access a different package's version of the BaseClient.

    src/tutorial/tutorial.go:477:33: bc.c undefined (cannot refer to unexported field or method c)

The fix for that test could possibly be to expose the BaseClient's instance of `c` (by making it a capital and thus exported `C`), or adding an accessor method `C()` or `Client()`.

Possibly a better fix would be to either remove these deprecated methods, or figure out which `TClient` is the correct one to set.

I'm not sure which is preferred, so I'm hoping to get some feedback/input here.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -1878,16 +1878,16 @@ void t_go_generator::generate_service_client(t_service* tservice) {
   f_types_ << indent() << "type " << serviceName << "Client struct {" << endl;
   indent_up();
 
-  f_types_ << indent() << "c thrift.TClient" << endl;
   if (!extends_client.empty()) {
     f_types_ << indent() << "*" << extends_client << endl;
+  } else {
+    f_types_ << indent() << "c thrift.TClient" << endl;
   }
 
   indent_down();
   f_types_ << indent() << "}" << endl << endl;
 
   // Legacy constructor function
-  f_types_ << indent() << "// Deprecated: Use New" << serviceName << " instead" << endl;
   f_types_ << indent() << "func New" << serviceName
              << "ClientFactory(t thrift.TTransport, f thrift.TProtocolFactory) *" << serviceName
              << "Client {" << endl;
@@ -1907,7 +1907,6 @@ void t_go_generator::generate_service_client(t_service* tservice) {
   indent_down();
   f_types_ << indent() << "}" << endl << endl;
   // Legacy constructor function with custom input & output protocols
-  f_types_ << indent() << "// Deprecated: Use New" << serviceName << " instead" << endl;
   f_types_
       << indent() << "func New" << serviceName
       << "ClientProtocol(t thrift.TTransport, iprot thrift.TProtocol, oprot thrift.TProtocol) *"
@@ -1936,15 +1935,24 @@ void t_go_generator::generate_service_client(t_service* tservice) {
   f_types_ << indent() << "return &" << serviceName << "Client{" << endl;
 
   indent_up();
-  f_types_ << indent() << "c: c," << endl;
   if (!extends.empty()) {
     f_types_ << indent() << extends_field << ": " << extends_client_new << "(c)," << endl;
+  } else {
+    f_types_ << indent() << "c: c," << endl;
   }
   indent_down();
   f_types_ << indent() << "}" << endl;
 
   indent_down();
   f_types_ << indent() << "}" << endl << endl;
+
+  if (extends.empty()) {
+    f_types_ << indent() << "func (p *" << serviceName << "Client) Client_() thrift.TClient {" << endl;
+    indent_up();
+    f_types_ << indent() << "return p.c" << endl;
+    indent_down();
+    f_types_ << indent() << "}" << endl;
+  }
 
   // Generate client method implementations
   vector<t_function*> functions = tservice->get_functions();
@@ -1975,7 +1983,7 @@ void t_go_generator::generate_service_client(t_service* tservice) {
       std::string resultName = tmp("_result");
       std::string resultType = publicize(method + "_result", true);
       f_types_ << indent() << "var " << resultName << " " << resultType << endl;
-      f_types_ << indent() << "if err = p.c.Call(ctx, \""
+      f_types_ << indent() << "if err = p.Client_().Call(ctx, \""
         << method << "\", &" << argsName << ", &" << resultName << "); err != nil {" << endl;
 
       indent_up();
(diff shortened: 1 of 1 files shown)


Jens-G:
Hi John,

Questions:
* Why are they deprecated? I probably only overlook sth, so bear with me
* Do we have a JIRA Ticket for this? If not, could you create one?

johnboiles:
> Why are they deprecated? I probably only overlook sth, so bear with me

That's a great question. Looks like they were marked deprecated by @dcelasun in #1382 as a part of THRIFT-4285. That discussion predates my involvement in Thrift.

> Do we have a JIRA Ticket for this? If not, could you create one?

Thanks for the nudge, I was just being lazy. Created as THRIFT-4447

dcelasun:
> Why are they deprecated? I probably only overlook sth, so bear with me

My apologies, it's an unfortunate result of the patch changing authors and things getting mixed up in the process. I should have removed the deprecated comments, especially since those constructors *are* compatible.

The bugfix looks good.

johnboiles:
@dcelasun any thoughts on fixing the tests? Looks like `tutorial/go/src/tutorial/tutorial.go` imports its `BaseClient` from another package. Since `c` isn't public in that other package, my modified initializer fails. 

Probably the simplest fix would just be to add a public accessor for `c`, (e.g. `func (p *BaseClient) C() thrift.TClient`) and us that public access in my modified initializer methods. What do you think?

dcelasun:
```go
type MyServiceClient struct {
	c thrift.TClient
	*MyServiceBaseClient
}

type MyServiceBaseClient struct {
	c thrift.TClient
}
```

@johnboiles Are you getting this second "Base" struct with 0.11 compiler? Because I can't reproduce.

> Also should I remove the // Deprecated comments if these are not actually deprecated?

Yes please.

johnboiles:
> Are you getting this second "Base" struct with 0.11 compiler? Because I can't reproduce.

I've been working off the master branch, but I just tried generation with 0.11 (327ebb6c2b6df8bf075da02ef45a2a034e9b79ba) and it looks like it has the same issue. Here is a direct copy-paste from my generated code with 0.11:

```
// Deprecated: Use NewUserService instead
func NewUserServiceClientFactory(t thrift.TTransport, f thrift.TProtocolFactory) *UserServiceClient {
	return &UserServiceClient{UserServiceBaseClient: NewUserServiceBaseClientFactory(t, f)}
}

// Deprecated: Use NewUserService instead
func NewUserServiceClientProtocol(t thrift.TTransport, iprot thrift.TProtocol, oprot thrift.TProtocol) *UserServiceClient {
	return &UserServiceClient{UserServiceBaseClient: NewUserServiceBaseClientProtocol(t, iprot, oprot)}
}

func NewUserServiceClient(c thrift.TClient) *UserServiceClient {
	return &UserServiceClient{
		c: c,
		UserServiceBaseClient: NewUserServiceBaseClient(c),
	}
}
```

dcelasun:
Ah, you have a service extending another service. I missed that. In that case, I think removing `c thrift.TClient` from `UserServiceClient` (and constructor functions) is the cleanest option. That way, we would only have a `TClient` within the `UserServiceBase` and it can be reused by any extending service. Thoughts?

johnboiles:
You're correct that it's a service extending another service; tbh I missed that too (I'm not the author of the `.thrift` files. I'm also generally new to thrift so still sorting out what should be generated).

So basically the logic would be to only add `c: thrift.TClient` to the `*Client` struct if `extends.empty()` in the generator? I'll take a look and see how hard that looks.

dcelasun:
> So basically the logic would be to only add `c: thrift.TClient` to the `*Client` struct if `extends.empty()` in the generator?

Exactly.

johnboiles:
This is a lot cleaner -- I like it. We still have the same issue with tests, though, if the base service is in a separate package -- you can't access the client as `.c` on the base service since it's lowercased.

Maybe we expose the client with a getter method? 

```
func (p *BaseClient) C() thrift.TClient {
    return p.c
}
```

dcelasun:
Any capitalized method on the client will potentially conflict with an RPC, e.g:

```thrift
service Foo {
    string C()
}
```
will generate uncompilable code. I suggest we use something like `Client_` as the method name. It looks ugly, but there is precedent for it (conflicts with keywords etc).

johnboiles:
Good call, so then:

```
func (p *BaseClient) Client_() thrift.TClient {
    return p.c
}
```
?

dcelasun:
Exactly.

johnboiles:
Done ^. Thanks again for all your direction Dcelasun

dcelasun:
Happy to help. Once Travis is all green, please squash your commits so it'll be easier to merge.

johnboiles:
Travis was green so i've squashed it!

FYI for the future: GitHub can now automatically squash commits in PRs at merge time. If that feature is enabled for the repo it shouldn't be necessary for committers to squash their branch prior to merging, since everything can be squashed via the GitHub UI:

image

dcelasun:
> GitHub can now automatically squash commits in PRs at merge time

For Thrift, PRs are not merged, just closed. You'll see a commit with you as the author but someone else as the committer, like [these](https://github.com/apache/thrift/commits/master).

johnboiles:
> For Thrift, PRs are not merged, just closed. You'll see a commit with you as the author but someone else as the committer, like these.

Ah got it, good to know thanks!

jeking3:
The reasoning for this is that Apache maintains its own master repository; so pull requests here have to be pulled into the master by a committer, and then the github mirror updates.  If the commit message contains a "This closes" or "This fixes" line item then it closes the github issue or pull request when that occurs.  I'll merge this tonight - thanks.

PR #1468, opened 2018-01-13
THRIFT-(1418|1507|3983|4120|4178|4259|4294): Multiple issues

Alex-Vol:
Summary of changes
===============
This pull request changes the build system used to Gradle. The integration uses the recommended and popular option of the Gradle Wrapper to avoid locking the build to the backported versions of Gradle in the Debian repositories. Their respective versions are varying significantly from Ubuntu release to release and cause problems with respect to the more advanced features such as the Maven publication.

With this change the Java library should be buildable on all platforms that support Java and have a usable installation of the tools required.

Issues fixed
=========
[THRIFT-1418](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-1418): Compiling Thrift from source: Class 
org.apache.tools.ant.taskdefs.ConditionTask doesn't
support the nested "typefound" element

[THRIFT-1507](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-1507): Maven can't download resource from central when behind
a proxy and won't use local repository

[THRIFT-3983](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-3983): libthrift is deployed on central with pom packaging instead of jar

[THRIFT-4120](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4120): pom files are not generated or provided in the build

[THRIFT-4178](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4178) Java libraries missing from package when using cmake

[THRIFT-4259](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4259): Thrift does not compile due to Ant Maven task errors

[THRIFT-4294](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4294): Java Configure Fails for Ant >= 1.10

Change details
===========
Refactor CMake install hook to allow using "sudo make install/fast"
which avoids the THRIFT-1507 and THRIFT-1418 issues.

New Gradle based build system for Thrift Java Library

* Add Gradle 4.4.1 Wrapper artifacts to enable builds
Using the Gradle Wrapper helps normalize the builds on all platforms
so we use a consistent build tool independent of package managers.

The Gradle build logic was partitioned into multiple specific
scripts to simplify understanding and maintenance of the build.

This is now hooked into CMake and Autoconf processing steps
and can build/test/publish to Maven. The README.md was updated
to illustrate the new build options and add some documentation
on the requirements for Maven publication.

Cleaned up the CMake files to reduce reliance on file globbing
which is known to cause confusion when multiple users contribute.

* Fix two minor issues in Javadoc and unit test

Return values were undeclared in Javadoc
Test was asserting on the wrong test object instance

* Create simple runner scripts for cross-check

Using Gradle as a java execution wrapper is too heavy.
I simplified the test client/server execution by using a
three generated scripts in the build directory direcly
callable by the cross-check test harness.

* Cleanup the remaining Ant build scripts

Pulled the Maven Ant task properties out of the Java build
since they are no longer used there.
Deleted the no longer used build.xml and build.properties
files from the Java build.
Made each Ant build own the Maven Ant task details in their
build.properties file.

* Fix the build issue with Java SSL in the ubuntu-trusty container

The latest Trusty JDK7 builds seem to have encountered this issue
because the OpenJDK removed the SunEC algorithms.

* Update the developer info as requested in review

Use the generic Apache Thrift developer list for contact information

* Add Clover Code coverage for easy access by developers

Clover plugin for Gradle was applied and configured which enables
code coverage reports to be available on demand via a command line
option. The documentation in the README.md was enhanced to give
the details of this change and how to take advantage of it.


lib/java/test/org/apache/thrift/transport/TestTSimpleFileTransport.java
```diff
@@ -27,21 +27,21 @@
...
     trans_write.write(input_buf);
-    trans_write.write(input_buf,2,2);    
+    trans_write.write(input_buf,2,2);
     trans_write.flush();
     trans_write.close();
-    
+
     //Test read side
     TSimpleFileTransport trans = new TSimpleFileTransport(tempFilePathName.toString(),true, false);
-    assert(trans_write.isOpen());
```
Alex-Vol: This one line was asserting against the wrong object. The trans_write object is quite obviously closed and the trans object was just opened.

lib/java/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
Alex-Vol: This script is autogenerated by the Gradle Wrapper integration. It is expected to exist and is updated as Gradle Wrapper is refreshed to a newer version.

lib/java/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
Alex-Vol: This script is autogenerated by the Gradle Wrapper integration. It is expected to exist and is updated as Gradle Wrapper is refreshed to a newer version.

lib/java/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
Alex-Vol: This control file is autogenerated by the Gradle Wrapper integration. It is expected to exist and is updated as Gradle Wrapper is refreshed to a newer version.

lib/java/CMakeLists.txt
```diff
@@ -17,14 +17,13 @@
...
 
-file(GLOB java_sources src/**/*.java)
-
 if(ANDROID)
     set(android_sources
         android/build.gradle
         android/settings.gradle
         android/src/main/AndroidManifest.xml
     )
+    file(GLOB java_sources src/**/*.java)
```
jeking3: It is better to add a cmake target that always runs the gradle build rather than having cmake attempt to enforce any dependency checking because it's likely to be wrong relative to the language build system.

Also,

https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil

lib/java/build.gradle
```diff
@@ -0,0 +1,495 @@
...
+
+        licenses {
+            license {
+                name 'The Apache Software License, Version 2.0'
+                url "${project.license}"
+                distribution 'repo'
+            }
+        }
+
+        developers {
```
jeking3: This list can be condensed down to:
```
developer {
  id 'dev'
  name 'Apache Thrift Developers'
  email 'dev@thrift.apache.org'
}
```

I believe this information is used elsewhere, for example in CPAN packaging.

lib/java/gradle.properties
```diff
@@ -0,0 +1,35 @@
+# This file is shared currently between this Gradle build and the
+# Ant builds for fd303 and JavaScript. Keep the dotted notation for
+# the properties to minimize the changes in the dependencies.
+thrift.version=1.0.0
```
jeking3: There's a document floating around on the thrift web site about releasing Thrift that points folks at what files to update after cutting a release branch; this should be added.  @Jens-G since you went through this recently can you point that out?

lib/java/gradle.properties
```diff
@@ -0,0 +1,35 @@
...
+mvn.repo=http://repo1.maven.org/maven2
+apache.repo=https://repository.apache.org/content/repositories/releases
+
+# Apache Maven publish
+license=http://www.apache.org/licenses/LICENSE-2.0.txt
+maven-repository-url=https://repository.apache.org/service/local/staging/deploy/maven2
+maven-repository-id=apache.releases.https
+
+# Ant Maven Task Jar Information, not used by this project anymore
+mvn.ant.task.version=2.1.3
```
jeking3: If not used, why do we need it?

lib/java/gradle.properties
```diff
@@ -0,0 +1,35 @@
...
+mvn.ant.task.url=${mvn.repo}/org/apache/maven/maven-ant-tasks/${mvn.ant.task.version}
+mvn.ant.task.jar=maven-ant-tasks-${mvn.ant.task.version}.jar
+
+# Dependency versions
+httpclient.version=4.4.1
```
jeking3: Are these hard or soft dependencies?  There are newer versions of these.

lib/java/build.gradle
Alex-Vol: Sure, that would make the pom file much shorter and less likely to have stale information.

lib/java/gradle.properties
Alex-Vol: This properties file is imported in the remaining Ant builds that I had not enough motivation to redo. The Ant builds that use this reference the JAR files from this project as well. IMHO this is a very poorly constructed side-channel that has no guarantees from what I can see to ensure the Java code has been built to any level before attempting to use the classes.

lib/java/gradle.properties
Alex-Vol: One hurdle at a time! I did not want to open that can of worms yet. Perhaps once we have this building and passing cross-tests we can revisit refreshing dependencies. There are some deprecation warnings in the code that I can see and it is possible they should be addressed.

jeking3:
@Alex-Vol are you registered in the Apache Jira?  @jensg do you know what the requirements are in Apache Jira to be able to have an issue assigned to you?

lib/java/gradle.properties
Alex-Vol: I can have the CMake or autoconf wrapper system provide the version so this becomes a developer placeholder default. I am surprised this project does manual publishing of the Java artifacts. I can see how much the cross language support has increased the build system complexity but there is room for improvement.

Are all the language bits released by a manual build and process?

lib/java/gradle.properties
Alex-Vol: I may try to eliminate these hold overs from the Ant build system because they weaken the overall build system dependency model. Each module in this project should be self describing independently of the rest IMHO.

Unless all the Java related modules could be consolidated under a common Gradle driven project instead of many small separate ones. Then the publication details can live in a common place instead of 4-5 Ant build.xml files independently managed.

Alex-Vol:
The failures are coming from the cross-check system. In my initial approach of the conversion I did not realize that the cross-check build starts a build equivalent just to run a test harness and then it kills the running build. Gradle is not happy having its process killed while in the middle of a "build" step and although it is perfectly capable of recovering it adds runtime overhead.

I am debating changing the harness to either a plain Java JAR file that can be executed directly via a light-weight make harness or a direct java invocation. Will see where this goes.

Alex-Vol:
I will register in Apache Jira to have this properly handled.

lib/java/gradle.properties
jeking3: There is a release process.

https://thrift.apache.org/docs/committers/HowToRelease

Alex-Vol: In that case all we need to provide is the version parameters from Autoconf's own version variables into the execution of all the Java generating tasks. They use this file as their default version to bootstrap for development. This can stay forever at 1.0.0 to facilitate building without make executions. I already wired this project to obey the parsed version for the CMake builds. I can do the same for the Autoconf builds. I do need to find out how the version is expressed in Autoconf makefiles. I saw a PARSED_VERSION variable somewhere, perhaps that is it?

The "make dist" process would then generate all artifacts including the ones for Maven publication and add them to the tar ball. The Maven publication process is intended to be driven by Ant (currently broken) or Gradle and it should correctly post the artifacts as signed artifacts to the Apache Maven staging repository. I believe this could possibly be an automatic process generating "nightlies" for Java and other languages that have a distribution repository available. Nexus has a release promotion process that could then be invoked on any particular "nightly" build that is blessed for release. That would finalize the process that makes the artifacts available at the distribution repositories such as Maven Central.

Alex-Vol:
I have successfully gotten the cross-check tests to pass on Travis. 

What is still broken is the Appveyor builds which seem to hang for some strange reason. No root cause yet.

Also, I encountered [this known issue](https://github.com/travis-ci/travis-ci/issues/8503) that may require a ubuntu-trusty docker image update to fix it up. I suppose that this is something I can handle separately, I will see about that update to the image and perhaps make a pull request.

The make-dist build also fails inexplicably. I can actually run locally and it finished just fine. Seems like CMake on Travis fails to detect the preinstalled Gradle Wrapper in lib/java whereas I can get the same working locally. Any ideas would be appreciated, I cannot tell what is different from the logs.

Alex-Vol:
I believe I found the problem with make-dist. I hope that takes care of the builds because it is the final piece of the puzzle. I wonder if that caused the AppVeryor build failures as well.

Alex-Vol:
I may have figured out the issue with AppVeyor. I was unable to troubleshoot with RDP, AppVeyor has not run a build since 55 hours ago. But I figured out that the configuration in CMakeLists.txt would result in a bad directory configuration with an out of tree build and that may have been why the builds stalled.

I squashed the final changes and added much more in the README.md about the new builds system. I also wired up the PACKAGE_VERSION in Automake to allow a single location to manage the version. I think the same can be done easily for the rest of the Java/Ant based projects and eliminate the multiple files that need editing for release of the final bits. This is deserving an issue on its own though, I do not want to conflate this with any more features.

jeking3:
I don't think folks are going to be too happy about having many levels of new empty directories:

crossTest/java/org/apache/thrift/test
main/java/org/apache/thrift

Every .java file moved.  Ouch?

Alex-Vol:
I felt the need to make this organization like all the other java projects in Github. The files were moved for sure, splitting the cross-test files just makes them stand out. If you think it is a big deal I will just move the files back. Git will oblige.

At least now I know what was bugging AppVeyor. find_program will not return a .bat result even if I search for a .bat file if there is a plain named file in the same location. Somehow cmake/make is fine with executable "gradlew" in Windows but CTest simply fails with "bad command". A little help in the FindGradlew.cmake solved it.

jeking3:
I don't necessarily mind additional organization however I think it would be a good idea for other more active Java folks to weigh in.  Also if this solves other Jira issues, listing them all in the description would be good.

Alex-Vol:
OK, yes, I will list all the issues this solves. Who should we notify to review this change besides you? Can you get more eyes on it now that I finally figured out the AppVeyor magic sauce?

Is there a keyword for JIRA searches that will iterate over the Java issues that could help me identify  as many as this happens to solve? You already pointed at a few and I commented on them.

jeking3:
Once you have the description updated I will email the dev mailing list asking for more eyes.

Alex-Vol:
I am marking JIRA items already fixed by this one as duplicates/related to and will update the description.

Alex-Vol:
I have identified the following issues as duplicates or related to this. 

[THRIFT-4178](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4178)
[THRIFT-4120](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4120)
[THRIFT-3983](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-3983)
[THRIFT-1507](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-1507)
[THRIFT-1418](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-1418)
[THRIFT-4294](https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-4294)

Most are already fixed, the ones relating to running `sudo make install` can be addressed in this PR easily so I am going to tackle the task now.

Regarding the source code move, I will move it back and propose a layout organization as a separate option. It will make reviewing this by a larger group much easier without the noise of the move.

Alex-Vol:
Right after I pushed the last commit with the sources restored to the original locations I found the sonar-project.properties file that had a reference to the new source code layout.

Assuming this is a green build now it is ready for final review. @jeking3 if you could let interested parties in the dev discussion list know about the PR I would appreciate it.

lib/java/CMakeLists.txt
Alex-Vol: Done, no more file globbing!

jeking3:
Well, I did let folks know, but nobody appears to be looking.... is there going to be a change to the "making a release" that needs to be documented?

Alex-Vol:
Yes, as it stands right now we will need to change the "making a release" page to eliminate the manual editing of the gradle.properties file since it is now automated to just require the configure.ac version edit only.

Should I go ahead and add the same integration to the other projects that are now using the Java project gradle.properties file as the origin of their version number? These are I believe the as3, and js projects, perhaps one more. Then we can have less manual steps for a new release.

jeking3:
As a separate pull request, yes. :)

jeking3:
Do you want me to mark all the Jira issues as fixed in 0.12.0 once this is merged?

Alex-Vol:
@jeking3 per our offline discussion, I am closing this and pushing to my branch the update with the commit user email corrected so it does not trigger a new build. You can pick up the changes and merge from my fork.

I am ok with closing the defects as fixed in 0.12.0.

PR #1469, opened 2018-01-14
Thrift 4434: squashed

jeking3:
There were 22 commits to squash - I hope I got them all.  Had to do some cleanup commits after.  Will the CI gods bless us with green checkmarks?


jeking3:
Looks like another problem in the older Trusty environment; this time ruby rack-test updated and dropped support for older ruby environments.

PR #1475, opened 2018-01-19
THRIFT-4464: Fix typo in TNonblockingServer.py

tomtung:


jeking3:
It would be helpful if someone with python experience could double-check this.

allengeorge:
On the face of it that looks correct. There isn't even a `message` in scope, so when it hits that line the code would just break. I'll look a little deeper.

tomtung:
Maybe @nsuke (who wrote made the change originally) can help check?

nsuke:
Thanks @tomtung, looks fine to me.

PR #1476, opened 2018-01-21
Remove premature call to workSocket() in TNonblockingServer

bgedik:


bgedik:
The following tests FAILED:
16 - concurrency_test (Timeout)
20 - TNonblockingSSLServerTest (Timeout)

I think the failure in {{TNonblockingSSLServerTest}} could be related. concurrency_test passes locally for me, so I am guessing that there may be a different problem with it.

jeking3:
Yes, concurrency_test has sporadic failures - it is a lot better than it used to be; seems to be worse on Visual Studio 10.  Perhaps the solution here is to handle EAGAIN properly?

bgedik:
@jeking3 My latest fix worked fine, but failed to compile on windows. I'll rework to something that works for all platforms.

bgedik:
@jeking3 This is ready for review now.

lib/cpp/src/thrift/transport/TSSLSocket.h
```diff
@@ -78,6 +78,7 @@ class TSSLSocket : public TSocket {
   bool peek();
   void open();
   void close();
+  bool hasPendingDataToRead();
```
jeking3: Isn't this what peek() is for?

bgedik: Peek is designed to be blocking for blocking sockets, non-blocking for non-blocking sockets. This one is non-blocking for both. I can replace the existing peek() with this implementation if you prefer that.

jeking3: It would be preferable to have one way to ask, "is there any data I can read" that does not block.

jeking3: Note that the cross test suite currently does not test asynchronous.  It is more of a protocol test; it would be interesting to add one more matrix choice of threaded vs. nonblocking for the languages that support it, and test both against both.

bgedik: @jeking3 You said it would be preferable to have one way to ask, "is there any data I can read" that does not block. I completely agree. But the existing ``peek`` method is not designed to be non-blocking. So I have a few choices here:
  * Update the doxygen comment of ``peek`` to reflect what it does and keep ``hasPendingDataToRead``
  * Add an optional argument to peek that says ``nonBlocking=false`` and if it is provided as ``true``, do what ``hasPendingDataToRead`` used to do and remove ``hasPendingDataToRead``.
  * Change ``peek`` to be always non-blocking.

How about #2 above? I think #3 is not easy, as it will require code changes in other places that I am not comfortable with.

bgedik: I implemented option #2. In terms of your matrix choice suggestion: How can we make that happen? Is that something I can help with? I am bothered by the TNonblockingSeverTest not catching the problem reported in this bug. The moment we switched to 0.11, it started throwing unexpected exceptions for us, when using Java client against C++ server. This should have been caught by the existing tests, but it did not happen. I would like to help with this. So if you have a suggestion for me, please let me know.

jeking3:
I see now, I may have led you down the wrong path with my comments.  So peek() blocks waiting for data the be available to read and the end result of peek is that there is data to read OR the socket disconnected.  What you are looking for is a call to see if the socket has data available but does not block.  As such, those are different code paths...

Perhaps your original code before this last push was correct.  I will take another look.

lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -249,6 +249,17 @@ TSSLSocket::~TSSLSocket() {
   close();
 }
 
+bool TSSLSocket::hasPendingDataToRead() {
+  if (!isOpen()) {
+    return false;
+  }
+  initializeHandshake();
+  if (!checkHandshake())
+    throw TSSLException("SSL_peek: Handshake is not completed");
```
jeking3: This should probably say something other than SSL_peek?

lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -249,6 +249,17 @@ TSSLSocket::~TSSLSocket() {
...
+  // data may be available in SSL buffers (note: SSL_pending does not have a failure mode)
+  return TSocket::hasPendingDataToRead() || SSL_pending(ssl_) > 0;
```
jeking3: Should SSL_pending be checked first for efficiency?  First check the SSL buffers, then if those are clear then check the socket.

lib/cpp/src/thrift/transport/TSSLSocket.h
jeking3: I think my comments were wrong; peek blocks until there is something to do when the socket is a blocking socket.  I think that if the socket knew it was non-blocking then a call to peek() would behave like a call to hasPendingDataToRead, i.e. it would be a non-blocking call.  Perhaps that's a better way to approach it, but I would have to look at the code a little more closely.  I think the original intention of my comment was correct, there should be only one way to peek.  Calling peek() on a non-blocking socket should not block; calling peek() on a blocking socket should block.

bgedik:
Ok, reverted back.

lib/cpp/src/thrift/transport/TSSLSocket.cpp
bgedik: Fixed.

lib/cpp/src/thrift/transport/TSSLSocket.cpp
bgedik: Right. Nice catch.

lib/cpp/src/thrift/transport/TSSLSocket.h
bgedik: Sam applies to ``read``. This goes back to my original point. TSocket should behave differently for blocking vs non-blocking socket. Look at how ``TSocket`` and ``TSSLSocket`` handle ``read``s on a non-blocking socket. They throw exceptions. Luckily for me, when not using SSL, ``TNonBlockingServer`` does not ever call ``read`` on a socket that is not ready to consume data. However, that is not the case for ``SSL``.  Read calls in ``TNonBlockingServer`` are retried by catching the exception and searching for ``"retry"`` in the exception message.

The correct way to fix all this mess is to make ``TSocket`` aware of whether the socket is blocking and non-blocking and behave accordingly. Converting EAGAIN to an exception in the case of non-blocking sockets is a major performance hit. 

My suggesting is to handle the issue of making ``TSocket`` and ``TSSLSocket`` aware of the blocking/non-blocking nature of the socket in a separate Jira issue.

bgedik:
@jeking3 anything I can do to get this into a mergable state?

jeking3:
Squash to one commit, please.

bgedik:
@jeking3 Here: https://github.com/apache/thrift/pull/1497

PR #1479, opened 2018-01-25
THRIFT-4474: generate PHP code use PSR-4 style default

RobberPhex:
ref: https://issues.apache.org/jira/browse/THRIFT-4474

### DIFF
--- a/lib/php/README.md
+++ b/lib/php/README.md
@@ -1,7 +1,6 @@
 Thrift PHP Software Library
 
-License
-=======
+# License
 
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements. See the NOTICE file
@@ -20,8 +19,7 @@ KIND, either express or implied. See the License for the
 specific language governing permissions and limitations
 under the License.
 
-Using Thrift with PHP
-=====================
+# Using Thrift with PHP
 
 Thrift requires PHP 5. Thrift makes as few assumptions about your PHP
 environment as possible while trying to make some more advanced PHP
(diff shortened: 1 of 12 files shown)


jeking3:
Are these changes all backwards compatible?

RobberPhex:
There are some break changes:

1. for generated code, old struct is `Types.php` for all args, results, `<Service>.php` for `ServiceIf`, `ServiceClient`, etc.
  new struct is `<classname>.php` for `<class>`. Maybe cause some code analyzer doesn't work.
2. for ThriftClassloader. before, user use `$loader->registerDefinition('namespace', '<path>')`. 
  At new struct, **it doesn't work**. user can use `$composerLoader->addPsr4('namespace', '<path>')` or `$thriftLoader->registerNamespace('namespace', '<path>')`.
3. for composer loader, generated code loaded via classmap. With new struct, composer's classmap can also load psr-4 code. (of cause psr-4 is recommended).

RobberPhex:
Personally, I think we can mark classmap/old-style PHP compiler deprecated. And can be removed in future.

And, ThriftClassLoader can also be removed in future, to migrate composer autoloader.

jeking3:
If we are making breaking changes, they need to be documented very clearly in the php/README.md file as part of any commit.  See my breaking changes for 0.10.0 and 0.11.0 in the perl/README.md for an example of what I mean.  I'm not certain anybody on the mailing list is listening at this point, but if anyone has an interest in php breaking changes, you need to chime in.

jeking3:
Document all breaking changes in the language README.

jeking3:
Any other php folks out there want to comment or review on the breaking change here?  It looks like the strategy was to change the generator and provide a flag to allow folks to generate older style code for backwards compatibility.

lib/php/README.md
```diff
@@ -20,34 +19,46 @@ KIND, either express or implied. See the License for the
...
-  APC cache is used by the TSocketPool class. If you do not have APC installed,
-  Thrift will fill in null stub function definitions.
+    APC cache is used by the TSocketPool class. If you do not have APC installed,
+    Thrift will fill in null stub function definitions.
+
+# Breaking Changes
+
+## 0.12.0
+
+1. For thrift compiler, `psr4` flag is opened default, generated code match psr4 code struct.
```
dcelasun: Better phrasing:

`PSR-4` (make this a link to https://www.php-fig.org/psr/psr-4/) loader is now the default. If you want to use class maps instead, use `-gen php:classmap`.

lib/php/README.md
```diff
@@ -20,34 +19,46 @@ KIND, either express or implied. See the License for the
...
+
+    If you want old-style directory struct, use `classmap` option. (i.e. `-gen php:classmap`)
+
+2. For ThriftClassloader, if use psr4 code struct, must use `$thriftClassLoader->registerNamespace('namespace', '<path>')` to register autoload.
```
dcelasun: Better phrasing:

If using PSR4, use `$thriftClassLoader->registerNamespace('namespace', '<path>')` instead of `$thriftClassLoader->registerDefinition('namespace', '<path>')`.

dcelasun:
Added some doc comments. The actual code looks fine to me.

lib/php/README.md
RobberPhex: Thank you! It's better phrasing.

---

Fixed by amend comment.

dcelasun:
@RobberPhex can you squash your commits?

RobberPhex:
@dcelasun squashed

PR #1483, opened 2018-01-29
THRIFT-4308: re-enable dlang deimos build support for libevent and openssl

jeking3:
... working around a dmd link order issue.

https://issues.apache.org/jira/browse/THRIFT-4308


jeking3:
Of course the lib/d certificates had to expire on January 29 too...

jeking3:
The concurrency test failed, which is a known intermittent issue.

PR #1484, opened 2018-01-30
TBufferedTransport must have underlying transport

RobberPhex:
ref: https://issues.apache.org/jira/browse/THRIFT-4477

### DIFF
--- a/lib/php/lib/Factory/TStringFuncFactory.php
+++ b/lib/php/lib/Factory/TStringFuncFactory.php
@@ -21,8 +21,9 @@
 
 namespace Thrift\Factory;
 
-use Thrift\StringFunc\Mbstring;
 use Thrift\StringFunc\Core;
+use Thrift\StringFunc\Mbstring;
+use Thrift\StringFunc\TStringFunc;
 
 class TStringFuncFactory
 {
@@ -49,7 +50,7 @@ class TStringFuncFactory
          * Cannot use str* functions for byte counting because multibyte
          * characters will be read a single bytes.
          *
-         * See: http://us.php.net/manual/en/mbstring.overload.php
+         * See: http://php.net/manual/en/mbstring.overload.php
          */
         if (ini_get('mbstring.func_overload') & 2) {
             self::$_instance = new Mbstring();
--- a/lib/php/lib/Transport/TBufferedTransport.php
+++ b/lib/php/lib/Transport/TBufferedTransport.php
@@ -22,6 +22,7 @@
 
 namespace Thrift\Transport;
 
+use Thrift\Exception\TTransportException;
 use Thrift\Factory\TStringFuncFactory;
 
 /**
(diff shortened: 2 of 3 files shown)


lib/php/lib/Factory/TStringFuncFactory.php
```diff
@@ -21,8 +21,9 @@
 
 namespace Thrift\Factory;
 
-use Thrift\StringFunc\Mbstring;
 use Thrift\StringFunc\Core;
+use Thrift\StringFunc\Mbstring;
```
jeking3: These changes seem unnecessary in order to change the line of documentation on line 53 below.

lib/php/lib/Transport/TBufferedTransport.php
```diff
@@ -22,6 +22,7 @@
 
 namespace Thrift\Transport;
 
+use Thrift\Exception\TTransportException;
```
jeking3: I don't see where TTransportException is used in the changes you made.
Is a "use" clause needed for the @inheritdoc directive to work?

lib/php/lib/Transport/TTransport.php
```diff
@@ -22,6 +22,7 @@
 
 namespace Thrift\Transport;
 
+use Thrift\Exception\TTransportException;
```
jeking3: Why is this change needed?

lib/php/lib/Transport/TBufferedTransport.php
```diff
@@ -78,11 +69,26 @@ public function __construct($transport = null, $rBufSize = 512, $wBufSize = 512)
      */
     protected $rBuf_ = '';
 
+    /**
+     * Constructor. Creates a buffered transport around an underlying transport
+     */
+    public function __construct($transport, $rBufSize = 512, $wBufSize = 512)
```
jeking3: It would have been more obvious if the pull request changed only this line, removing the "= null".

RobberPhex: Firstly, remove `=null` at class filed.
And, `$transport=null` should also be removed at function `__construct`.

Sometimes, called `new TBufferedTransport()` will cause $transport_ is null, which cause other funcion doesn't work(`open`, `read`, etc.)

lib/php/lib/Factory/TStringFuncFactory.php
RobberPhex: I sorted use alphabetically, is it unnecessary? Am I expected that change it back?

lib/php/lib/Transport/TBufferedTransport.php
RobberPhex: some function has doc comment `@throws TTransportException`, No one (and IDE) knows which TTransportException, so I think we should add use stmt.

Or, We always use full classname at doc comment, is it?

lib/php/lib/Transport/TTransport.php
RobberPhex: There is doc comment `@throws TTransportException` at Line 45.

No one (and IDE) knows which TTransportException.

lib/php/lib/Factory/TStringFuncFactory.php
jeking3: In general fewer changes are better but in this case no, I don't care.  I do the same cleanup when I'm in files sometimes. :)

lib/php/lib/Transport/TBufferedTransport.php
jeking3: My question was more for learning, and whether or not the "use" is needed for that to work.  I think I understand it now.

RobberPhex:
CI isn't stable.

It's so headache. :(

PR #1486, opened 2018-01-30
THRIFT-4337: Able to set keyStore and trustStore as InputStream in the TSSLTransportFactory.TSSLTransportParameters

dmvolod:
Plus some tests cleanup


dmvolod:
Done rebase.
Thanks, @jeking3 .

jeking3:
Hmm, looks like I broke the build right before this; I'll make sure this carries through and merge it.  Thanks.

PR #1494, opened 2018-02-13
THRIFT-4495: Allow `undefined` for non-required Erlang records fields.

dhull:
As of Erlang 19, the dialyzer static type-analysis tool no longer implicitly adds `undefined` to the allowed types for a field.  This means that dialyzer will now complain about any non-required fields that are not explicitly initialed when creating a new record.


jeking3:
We use these versions in CI today, FYI:

trusty: R16B03
xenial: 18.3 (this is the one we run "make cross" under)
artful: 20.0.4

jeking3:
Hi, please rebase this on the current master and force push to kick a new build - I think the CI build environment is stable again.

jeking3:
I'm running this and 4497 through local tests since it wasn't based on good CI builds.

jeking3:
Ugh, build failure is dlang related, not to this.

PR #1495, opened 2018-02-14
THRIFT-4497: Use map() field type for Erlang type for map struct fields

dhull:
The Thrift Erlang code generator previously generated fields with the `#{}` Erlang type for maps fields.  In the Erlang type specification languages, however, `#{}` specifically means an empty map.  This commit fixes the code to emit `map()` instead, which means the maps keys and values may be of any type.

It would be possible to emit a field type such as `${keytype() => maptype()}`, but this commit does not do that.


jeking3:
Hi, please rebase this on the current master and force push to kick a new build - I think the CI build environment is stable again.

PR #1500, opened 2018-03-07
THRIFT-4509: use nodejs 8.x from nodesource.com in travis builds

bananer:


bananer:
@jeking3 where would the build encounter nodejs 4.2.6 then?

jeking3:
As of yesterday afternoon 4.x is no longer used.

Due to the dependencies, and due to the fact that node.js 4.x LTS ends next month, I moved the "oldest" make check job to use nodejs 6.x (ubuntu-xenial) and the current one uses 8.x (ubuntu-artful) - this is the one that runs make check, make cross, ubsan, cppcheck, etc.

We still need to modernize the code/test for js and nodejs however.

PR #1502, opened 2018-03-07
THRIFT-4509: update grunt to 1.0.2

bananer:
switch from grunt-external-daemon and grunt-shell to grunt-shell-spawn and update grunt to 1.0.2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,6 @@ node_modules
 compile
 test-driver
 erl_crash.dump
-package-lock.json
 project.lock.json
 
 .sonar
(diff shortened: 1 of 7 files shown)


jeking3:
Any special considerations we need to add to the README for js or nodejs if we check in a package-lock.json file?  When distributing through npm should this file also be distributed?

bananer:
The package-lock file keeps dependency versions fixed across all installations. According to [the docs](https://docs.npmjs.com/files/package-lock.json), this does not apply when the package is published on npm, where the file should be excluded.

To update the dependency packages for our builds, one now has to run `npm update` and commit the changed package-lock.json files. This does not impact users of thrift, so I'm not sure if it has to be noted in the Readme or somewhere else.

jeking3:
Okay that sounds good; could you squash and rebase on master to prepare for a merge?

bananer:
Rebased as https://github.com/apache/thrift/pull/1506

jeking3:
In the future you can squash, rebase, and force push to keep the same PR, see:

PR #1504, opened 2018-03-08
Fix wrong @param in comment

victoryifei:
Wrong @ param name can cause warning when `-Wdocumentation` flag is on for objc.


jeking3:
Usually we ask for a Jira ticket for each change but I'm not going to considering this is a small doc change.  In the future please see: https://thrift.apache.org/docs/HowToContribute

PR #1505, opened 2018-03-09
THRIFT-4513: Fixing java thrift compiler to generate constants in stable order.

romanoid:
For description of the issue fixed, see: https://issues.apache.org/jira/browse/THRIFT-4513

### DIFF
--- a/compiler/cpp/src/thrift/audit/t_audit.cpp
+++ b/compiler/cpp/src/thrift/audit/t_audit.cpp
@@ -202,8 +202,8 @@ bool compare_defaults(t_const_value* newStructDefault, t_const_value* oldStructD
          }
       case t_const_value::CV_MAP:
          {
-            const std::map<t_const_value*, t_const_value*> newMap = newStructDefault->get_map();
-            const std::map<t_const_value*, t_const_value*> oldMap = oldStructDefault->get_map();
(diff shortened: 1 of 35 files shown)


dcelasun:
Please add the issue number to your commit message and PR title. See [here](https://thrift.apache.org/docs/HowToContribute) (point 7 under Github).

romanoid:
@jeking3 Yes, thanks, PR had some assumption on Map keys being only primitive for constant maps, which is not true in general. Latest revision should no longer rely on it.

romanoid:
Updated approach:

t_const_value now defines ordering. 
Java generator changed to delegate to native ordering of t_const_value

jeking3:
Please squash and rebase on master, I just merged 4 commits that stabilize the CI builds.

romanoid:
@jeking3 done, thanks, let's see if it passes.

romanoid:
Fix is in, the only failure left is this one, I'm not sure if it is instability of tests or real problem: https://travis-ci.org/apache/thrift/jobs/353050154

jeking3:
Looks like a dlang issue that pops up from time to time.  Just to satisfy myself this change is okay I need to look at generated thrift file comparisons from before and after this change.  I'll do that locally.

romanoid:
Thank you for the review!

PR #1506, opened 2018-03-12
THRIFT-4509: grunt update (rebased)

bananer:
* switch from grunt-external-daemon and grunt-shell to grunt-shell-spawn
* update grunt to 1.0.2
* always use local copy of jquery and qunit
* commit the package-lock files for npm keep versions stable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,6 @@ node_modules
 compile
 test-driver
 erl_crash.dump
-package-lock.json
 project.lock.json
 
 .sonar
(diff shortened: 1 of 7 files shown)


jeking3:
I see, you need to build Java first?

jeking3:
With this merged would you say that THRIFT-4509 is now fixed, or is there more work to do?

bananer:
@jeking3 I think all the problems mentioned in THRIFT-4509 are resolved now. With the package-lock being included now, there should also be less random build failures. The node and js parts of thrift could still use some cleanup and modernization, but I will leave that for a separate ticket.

PR #1507, opened 2018-03-15
THRIFT-4516: Fix "go vet" warnings for Go 1.10

dcelasun:
Opening this here because my free Travis account has its builds killed after 50 minutes.

### DIFF
--- a/lib/go/test/dontexportrwtest/compile_test.go
+++ b/lib/go/test/dontexportrwtest/compile_test.go
@@ -20,7 +20,6 @@
 package dontexportrwtest
 
 import (
-	"fmt"
 	"testing"
 )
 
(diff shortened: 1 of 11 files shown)


lib/go/test/dontexportrwtest/compile_test.go
```diff
@@ -29,10 +28,10 @@ import (
 func TestReadWriteMethodsArePrivate(t *testing.T) {
 	// This will only compile if read/write methods exist
 	s := NewTestStruct()
-	fmt.Sprintf("%v", s.read)
-	fmt.Sprintf("%v", s.write)
+	_ = s.read
```
jeking3: Nice.. I had done everything else but I had no idea how to fix this in golang.  Thanks!

jeking3:
I'll keep an eye on this.  Right now the docker images on docker.io have to be generated manually every time they are updated, otherwise the build jobs take more time than they should.

lib/go/test/dontexportrwtest/compile_test.go
dcelasun: The problem is `s.read` and `s.write` are functions and there just isn't any way to print functions as strings, so I had to find another way to use them without trying to stringify them.

dcelasun:
Failures seem unrelated. Should I merge or do you need to something on docker.io?

jeking3:
Go ahead and merge.  After you merge, I update my fork and kick a build on my account to refresh the images.  It's an optimization that saves ~10 minutes per build job.

PR #1509, opened 2018-03-19
THRIFT-4515: cross server test improvement: graceful test server shutdown

jeking3:
### Problem ###
Test servers are not shut down gracefully - they are SIGKILLed once the client exits.

### Solution ###
Try using SIGINT first, capture the result code and process it properly, SIGKILL if the process seems unresponsive.

### Also ###
- Better subprocess management in cross test requires Python 3.3 or later to run the suite.
- tests.json can declare "stop_signal": <n> in the server stanza to customize this behavior
- Added signal handling to C++, D, and Perl test servers to stop gracefully.  Removed "busy wait" tight loop in C++ test server.
- Cleaned up some C++ library compiler warnings.
- Fixed C++ TSocket error descriptions for domain sockets to use the path instead of host/port.
- Fixed an intermittent issue in concurrency_test.
- Perl server had no "stop()" mechanism which was easy to add.
- Updated SBCL (lisp) to 1.4.5 to try and eliminate these "file not found" sporadic build issues.


jeking3:
The build failure in the artful check job is in lib/d unit tests and is a sporadic issue that needs to be investigated, but will not hold up this work.

PR #1510, opened 2018-03-20
THRIFT-4476: Typecasting problem on floating-point list items, emitting doubles with high precision in Python, JS, Java, C++ and Erlang

ozymaxx:
Squashed version
see the related issue: https://issues.apache.org/jira/browse/THRIFT-4476


ozymaxx:
@jeking3 I squashed all into one commit and opened another branch (I had some problems while changing the history)

PR #1511, opened 2018-03-20
THRIFT-4476: Typecasting problem on double list items, emitting doubles with high precision (for Java, C++, Python, JS and Erlang)

ozymaxx:
All commits have been squashed into one
see the related issue: https://issues.apache.org/jira/browse/THRIFT-4476


jeking3:
Thanks for the squash.  We'll get this merged.

PR #1519, opened 2018-03-24
THRIFT-4214: map<i64,value> key treated as hex value in JavaScript

thexeos:
All strings passed to [node-int64 constructor](https://github.com/broofa/node-int64/blob/c1567475712cb1cfe100c96813c2a2a92e2b42ce/Int64.js#L49) are treated as hex strings. We need to convert all decimal strings to hex representation before passing them to the constructor.

We are making an assumption that no hex string will be passed to `writeI64()` without 0x prefix.


thexeos:
The tests fail at ``TypeError: binary.readI64 is not a function``, even though the function [exists](https://github.com/thexeos/thrift/blob/b5ceb251b200359d93d55386b70e7d6fc4b0fed8/lib/nodejs/lib/thrift/binary_protocol.js#L314).

thexeos:
I'd be happy to fix it, if anybody could explain why this error is happening:

```
/thrift/src/lib/nodejs/test/binary.test.js:86
    assert.equal("0", binary.readI64([0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]).toString());
                             ^
TypeError: binary.readI64 is not a function
    at Test.Should read I64 (/thrift/src/lib/nodejs/test/binary.test.js:86:30)
    at Test.bound [as _cb] (/thrift/src/node_modules/tape/lib/test.js:77:32)
    at Test.run (/thrift/src/node_modules/tape/lib/test.js:96:10)
    at Test.bound [as run] (/thrift/src/node_modules/tape/lib/test.js:77:32)
    at Immediate.next [as _onImmediate] (/thrift/src/node_modules/tape/lib/results.js:71:15)
    at runCallback (timers.js:794:20)
    at tryOnImmediate (timers.js:752:5)
    at processImmediate [as _immediateCallback] (timers.js:729:5)
```

Since `readI64` is [definitely there](https://github.com/apache/thrift/pull/1519/commits/b5ceb251b200359d93d55386b70e7d6fc4b0fed8#diff-7e44d6dd42f5266da83812d53effdb1cR314).

MateuszJeziorski:
@thexeos you looked at wrong file, `readI64` is not defined on `binary` - https://github.com/apache/thrift/blob/master/lib/nodejs/lib/thrift/binary.js

jeking3:
In Jira, the bug is marked as being in JavaScript (lib/js) and not in Node.JS (lib/nodejs), so which is it? :)

thexeos:
@MateuszJeziorski I see. So tests for this should really go into https://github.com/apache/thrift/blob/master/lib/nodejs/test/test-cases.js

@jeking3 I've updated Jira to list Node.js

jeking3:
Any chance this can be completed?

PR #1521, opened 2018-03-27
THRIFT-4531 (#1)

margars:
THRIFT-4531: Fix generated Python read() method for immutable structs with optional members.


margars:
Build failed with Gradle POM parsing error which has nothing to do with the commit.

PR #1522, opened 2018-03-27
THRIFT-4530: add @Nullable annotations to generated Java code

msridhar:
Use our own `org.apache.thrift.annotations.Nullable` annotation to avoid
introducing a third-party dependency

### DIFF
--- a/lib/java/src/org/apache/thrift/annotations/Nullable.java
+++ b/lib/java/src/org/apache/thrift/annotations/Nullable.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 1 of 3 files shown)


msridhar:
Note that while there are no new tests, the code under `tutorial/java/gen-java` includes the annotations and compiles without issue.  Since this change doesn't affect runtime behavior of the generated code I think this might be sufficient.

msridhar:
Looks like the Travis failure is on Common Lisp codegen (https://api.travis-ci.org/v3/job/359136811/log.txt), which has nothing to do with this PR

jeking3:
Yes there is an issue that appears in builds where lisp cannot find a file it just supposedly wrote.  Lisp is new to the project, this is an ongoing thing.

lib/java/src/org/apache/thrift/annotations/Nullable.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.thrift.annotations;
```
jeking3: Any reason why java language uses "annotation" but you selected "annotations" for thrift?  Seems like we should be using "annotation".

msridhar:
I just pushed up another commit with the renaming.  I assume you can do a squash and merge, but if you'd like me to squash just let me know.

dcelasun:
No need, I'll squash once Travis is all green.

msridhar:
Hrm, now I see a concurrency test failure on appveyor.  Doesn't seem related to my change, though I'm not 100% sure.

dcelasun:
I don't think it's relevant either, seems like a flaky test. I think we are good to merge.

@jeking3 any idea on the failing appveyor test, have you seen it before? I'm inclined to ignore it unless you say otherwise.

jeking3:
That is an intermittent test failure that I thought I resolved.  :)  it is C++ only

PR #1523, opened 2018-03-29
 Fix condition where TSimpleServer could exit AcceptLoop() without releasing lock.

mdubbyap:
…easing lock

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -125,26 +125,38 @@ func (p *TSimpleServer) Listen() error {
 	return p.serverTransport.Listen()
 }
 
+func (p *TSimpleServer) innerAccept() (int32, error) {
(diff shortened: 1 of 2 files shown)


dcelasun:
Thanks for the PR. Seems like this was [introduced](https://github.com/apache/thrift/blame/930428438c0b6c8f60560cbb7dcad79042badacb/lib/go/thrift/simple_server.go#L131) by THRIFT-4243 (PR #1302) and unfortunately made its way into the 0.11 release.

dcelasun:
Travis failures are unrelated flaky tests (my own Travis build is fine).

jeking3:
Thanks for checking, we need to stabilize these CI builds so less effort is spent figuring out if a change is at least somewhat ok.

jeking3:
THRIFT-4537

PR #1524, opened 2018-03-30
THRIFT-4535: XML docs; code cleanup (tabs->spaces; String->string)

cwe1ss:
As discussed in [THRIFT-4535](https://issues.apache.org/jira/browse/THRIFT-4535) we'd like to go back to just one .NET library that supports .NET framework (including Mono) AND .NET Core. 

As I had to learn about the library before doing some actual changes, I updated the files to contain proper XML comments (which can be used for IntelliSense) and I did some code cleanup. I did NOT change any code.

Feel free to close this PR, if you prefer to not have such a cleanup commit.

/cc @Jens-G @jeking3 

PS: I'll submit actual code changes once this PR is merged or rejected.


lib/csharp/src/Protocol/TCompactProtocol.cs
```diff
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
  *
- *   http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
```
Jens-G: Why are we doing that change?

Jens-G:
Disclaimer: I did not look at every line and I did not look into the AppVeyor failure. 

Aside from one minor thing everything that I skimmed through LTGM, and my code still compiles :-) so I'm basically fine with.

lib/csharp/src/Protocol/TCompactProtocol.cs
cwe1ss: This was by accident. I fixed it. thank you!

cwe1ss:
The last AppVeyor failure seems to be unrelated:

>The following tests FAILED:
>      30 - HaskellCabalCheck (Failed)

I fixed the license header in another commit so we'll see if the build succeeds now. 

Do you want me to squash the commits or will you do that via Github if you decide to merge the PR?

Jens-G:
Squash please.

PR #1532, opened 2018-04-03
[WIP] THRIFT-4541: Use new project system in lib/csharp

cwe1ss:
See [THRIFT-4541](https://issues.apache.org/jira/browse/THRIFT-4541). 

This is a WIP/PoC and **MUST NOT BE MERGED** (because supported versions must be defined first etc).

This PR updates all project files to the new project system and targets ".NET Standard 2.0" and ".NET 4.5". (Except for the MSBuild-task which currently can't target .NET 4.5 due to its dependencies)

Some notes about it:
* The solution can now be built using the cross-platform .NET Core SDK (v2.0+). (using the cmd `dotnet build`)
* I have not changed the MAKE files yet as I don't yet know how they work. I'll look into this once I know that we'll actually do this change.
* NuGet packages for the main Thrift library and for the MSBuild library can be created with `dotnet pack -c Release`. They will be placed in the "lib\csharp\artifacts" folder.
* `AssemblyInfo.cs` files are no longer required as most things are now defined either in `Directory.Build.props` or directly in the `*.csproj` files.
* **Thrift** project:
  * As there's some classes that depend on "System.Web" I already had to add some #if statements. These classes are only available when the package is consumed on the full .NET framework.
  * I also had to change something in `TNamedPipeServerTransport` as .NET standard doesn't have a NamedPipeServerStream constructor that accepts a PipeSecurity. We'll have to look into how to test this.
* **ThriftMSBuildTask** project:
  * I had to reference NuGet packages for MSBuild and these no longer contain the "Csc" class. I therefore had to change the implementation to manually call "csc". **This is not yet tested though!!** I wanted to show you the general approach first before I invest more time.
  * The project now creates a NuGet package that includes props files etc. I got this from [this how-to](https://www.natemcmaster.com/blog/2017/07/05/msbuild-task-in-nuget/).
* The test projects contain `PreBuildEvents` that generate `*.cs` files for `thrift` files. This currently is executed too late which results in the first build to fail. I still have to look into this or we should discuss using a different approach (e.g. using the MSBuild task)

I'm looking for general feedback on this approach. If the approach is ok, we should discuss in THRIFT-4541, which platforms we'd like to support. We can then either merge this into a branch or I can create a new PR with all changes.

/cc @Jens-G @jeking3


cwe1ss:
Thanks for your response. As noted in my last paragraph, I’m looking for your feedback on my changes to know if I can proceed. I thought that creating a PR would be the easiest option for that.

I’ll not do frequent commits on this PR.

jeking3:
We can look at PRs in your fork as well. :)

cwe1ss:
But since no-one is watching my fork I can only /cc you two (as I don't know anyone else here) and no-one else will have a chance to provide feedback. Also, I won't keep my fork forever so the discussion history would be lost. It would also be great if you guys could help at some point so having this in my fork is not ideal either.

As I've said in my original post, I'd prefer a branch in this repo that I could create the PRs against. This would ensure that everything is in one place, that we can work together and that we don't break the production code until we're done. This seems to be what many other projects do in cases like this.

Your thoughts?

jeking3:
As long as the updates are infrequent I see no issue.

jeking3:
I would recommend focusing on netcore for new development and leaving csharp as legacy.

cwe1ss:
I'm closing this as it's really old and there hasn't been much feedback.

Feel free to use the code if you see fit!

PR #1535, opened 2018-04-09
THRIFT-4548: Supporting TBinaryProtocolAccelerated protocol when using TMultiplexedProcessor in Python

balazskemenes:
Updated TMultiplexedProcessor, and new TMultiplexedBinaryProtocolAccelerated to support TBinaryProtocolAccelerated in Python
Client: [lib/py]

### DIFF
--- a/lib/py/src/TMultiplexedProcessor.py
+++ b/lib/py/src/TMultiplexedProcessor.py
@@ -18,7 +18,7 @@
 #
 
 from thrift.Thrift import TProcessor, TMessageType, TException
-from thrift.protocol import TProtocolDecorator, TMultiplexedProtocol
+from thrift.protocol import TMultiplexedProtocol
 
 
 class TMultiplexedProcessor(TProcessor):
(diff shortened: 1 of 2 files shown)


jeking3:
I also just noticed you removed protocol decoration entirely; if we had cross tests for multi in python (looks like we don't) then this change would have busted all the other multi protocol tests.  I'll be adding multi cross tests for python.

jeking3:
I added multi support to the python cross test client and I have multi, multic, multij working  in the python TestClient against various servers. 

It looks like multiplexed doesn't work with accel or accelc however.  Do the accelerated protocol implementations have a different interface?  Forgive me but I am not a python expert.  Here is the error I am seeing:
```
ERROR: testVoid (__main__.MultiplexedAcceleratedBinaryTest)
  File "/thrift/src/test/py/TestClient.py", line 75, in testVoid
    self.client.testVoid()
  File "/thrift/src/test/py/gen-py/ThriftTest/ThriftTest.py", line 324, in testVoid
    self.send_testVoid()
  File "/thrift/src/test/py/gen-py/ThriftTest/ThriftTest.py", line 330, in send_testVoid
    args.write(self._oprot)
  File "/thrift/src/test/py/gen-py/ThriftTest/ThriftTest.py", line 1695, in write
  File "/thrift/src/lib/py/build/lib.linux-x86_64-2.7/thrift/protocol/TProtocolDecorator.py", line 40, in <lambda>
    return lambda *args, **kwargs: self._wrap(member, args, kwargs)
  File "/thrift/src/lib/py/build/lib.linux-x86_64-2.7/thrift/protocol/TProtocolDecorator.py", line 49, in _wrap
    result = func(self.protocol, *args, **kwargs)
TypeError: function takes exactly 2 arguments (3 given)
```

Fixing this would accomplish your goal of making them work together while allowing for decoration with other protocols.  I suspect you ran into this and that led you to the PR submitted?

I will let you know when the test client is merged, then if you can make the test client work with the accelerated protocols and remove the failing tests from the known test failures, we'll have a viable solution with tests.

jeking3:
You can see the changes I made (running them through CI) here:

https://github.com/jeking3/thrift/pull/14

Specifically the tests.json and TestClient.py file changes.

jeking3:
I think this is obsoleted by #1547

PR #1536, opened 2018-04-11
Add prettier

NameFILIP:
This adds [Prettier](https://prettier.io/). See [why](https://prettier.io/docs/en/why-prettier.html).

The next step will be to run `npm run prettier` to consistently format all JS files. It's better to have that in a separate diff to make merging conflicts easier.


jeking3:
Thanks - please open an Apache Jira THRIFT ticket for this one so we can track it in the release notes for 0.12.0; both PRs can be submitted under that ticket.

NameFILIP:
Created a ticket: [THRIFT-4551](https://issues.apache.org/jira/browse/THRIFT-4551) and included it in the commit message.

PR #1537, opened 2018-04-11
Fix semicolons

NameFILIP:
Add missing and remove extra semicolons


NameFILIP:
I will run prettier instead after https://github.com/apache/thrift/pull/1536 is merged. It'll fix semicolons as well as the other stylistic issues (single vs double quotes, spaces vs tabs, etc)

PR #1549, opened 2018-04-19
thrift-4559: TSSLSocket no longer prints incorrect error for SYSCALL

Sidneys1:
Closes [THRIFT-4559](https://issues.apache.org/jira/browse/THRIFT-4559)

Fixed a case where an `SSL_ERROR_SYSCALL` is encountered and `errno_copy` and `ERR_peek_error()` are both `0`, seemingly indicating no error.

Changes should be backwards compatible.

### DIFF
--- a/lib/cpp/src/thrift/transport/TSSLSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLSocket.cpp
@@ -403,11 +403,16 @@ uint32_t TSSLSocket::read(uint8_t* buf, uint32_t len) {
       break;
     }
     unsigned int waitEventReturn;
+    bool breakout = false;
     switch (error) {
       case SSL_ERROR_ZERO_RETURN:
         throw TTransportException(TTransportException::END_OF_FILE, "client disconnected");
 
       case SSL_ERROR_SYSCALL:
+        if (errno_copy == 0 && ERR_peek_error() == 0) {
+          breakout = true;
+          break;
+        }
         if ((errno_copy != THRIFT_EINTR)
             && (errno_copy != THRIFT_EAGAIN)) {
               break;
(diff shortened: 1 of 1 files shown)


jeking3:
Can you describe the flow that occurs on disconnect?  From what you documented, I'm guessing that it is:

1. TSSLSocket::read called
2. SSL_Read is called.
3. Possibly some data read, possibly SYSCALL error on next loop
4. SYSCALL error handled, breaks out
5. TSSLSocket::read returns indicating any data that was read
5. Outer loop will call TSSLSocket::read again if it needs more data
6. I would expect SSL_Read to return SSL_ERROR_ZERO_RETURN at this point indicating disconnect?

Is that the new disconnect flow when SYSCALL error occurs with 0 errno?

Sidneys1:
That should be the new flow, yes. This is my first contribution to this library, so someone more familiar with the code should probably take a look and ensure that I was understanding the flow of `::read` correctly.

The new `breakout` variable is required because just breaking when `errno` is `0` will still cause an exception to be thrown. With the new variable, `SSL_Read` returns normally as if no data was read, and the next call to `::read` should result in `SSL_ERROR_ZERO_RETURN`.

jeking3:
If TTransport::read returns zero it is considered to be END_OF_FILE:
https://github.com/apache/thrift/blob/master/lib/cpp/src/thrift/transport/TTransport.h#L36

Therefore if we return length 0 from read here as a result of handling `SSL_ERROR_SYSCALL` then it will behave like a normal disconnect.  So it would be okay to simplify your logic to return 0 when SSL_ERROR_SYSCALL occurs rather than use a separate "breakout" statement - however what's there does the same thing AND it passed all the tests... so maybe leave it as is.

I'll merge this.  I spent a fair amount of time analyzing the change and flow.  I believe it is okay.  We don't do a lot of unexpected disconnection testing in any of our test suites which is a testing gap.  I'm sure many of these error-handling branches never get hit in the test suite.

PR #1550, opened 2018-04-27
THRIFT-4561: Remove python socket timeout

sokac:
Client: py

Timeout is already set in HttpClient. The reason for removal is issue
with gunicorn and gevent as described here
https://github.com/benoitc/gunicorn/pull/1616

Tested locally, works as expected


jeking3:
@jfarrell that is an intermittent problem in the lisp solution and we need to engage the original folks who added it to help resolve it.  It looks like lisp writes a temporary file and then cannot find its own file.

PR #1552, opened 2018-05-07
[Erlang]: processor: allow for user controlled request/response cycle

vanjabucic:
While attempting to create a simple HTTP endpoint in Elixir using Plug I noticed that the Erlang library does not have a way to manually process a payload.  
I have done it before in other languages, and the simplest example of that kind of action is done by Ruby http servers @[https://github.com/apache/thrift/blob/master/lib/rb/lib/thrift/server/mongrel_http_server.rb]
As you can see, you can process the thrift message 'manually' regardless of how it was received or how the response is to be sent back.

Assuming this patch is in place, here is an example of usage in Elixir using Plug for handling HTTP thrift requests:
```elixir
  # -----------------------------------------------
  # Low level raw binary handler for thrift packets, raw_data is basically just the POST body
  # returns raw thrift response
  def handle_post(raw_data) do
    # Construct a protocol
    {:ok, mem_transport} = :thrift_membuffer_transport.new(raw_data)
    {:ok, cb_protocol} = :thrift_compact_protocol.new(mem_transport)

    # Have thrift core library process the service request
    # reply is written back into the mem_transport inside the State
    {:ok, state} =
      :thrift_processor.process(
        {"ServiceHandler", cb_protocol, :query_service_thrift, ServiceHandler}
      )

    # Extract data buffer from the State object
    {:thrift_processor, _handler, protocol, _service} = state
    {:protocol, :thrift_compact_protocol, {:t_compact, transport, _, _, _, _}} = protocol
    {:t_transport, :thrift_membuffer_transport, {:t_membuffer, data}} = transport

    case data do
      [_ | _] = reply ->
        {:ok, :erlang.iolist_to_binary(reply)}
      _ ->
        {:error, "No reply was detected!"}
    end
  end
```

This was tested with Ruby client using standard Thrift::HTTPClientTransport.

### DIFF
--- a/lib/erl/src/thrift_processor.erl
+++ b/lib/erl/src/thrift_processor.erl
@@ -19,12 +19,31 @@
 
 -module(thrift_processor).
 
--export([init/1]).
+-export([init/1, process/1]).
 
 -include("thrift_constants.hrl").
 -include("thrift_protocol.hrl").
 
 -record(thrift_processor, {handler, protocol, service}).
+
+% Instead of a loop, do a single-shot processing of a packet
+process({_Server, Proto, Service, Handler}) ->
+    State0 = #thrift_processor{ protocol = Proto,
+                                service = Service,
+                                handler = Handler},
+    {Proto1, MessageBegin} = thrift_protocol:read(Proto, message_begin),
+    State1 = State0#thrift_processor{protocol = Proto1},
+    case MessageBegin of
+        #protocol_message_begin{name = Function,
+                                type = Type,
+                                seqid = Seqid} when Type =:= ?tMessageType_CALL; Type =:= ?tMessageType_ONEWAY ->
+    			case handle_function(State1, list_to_atom(Function), Seqid) of
+       				{State2, ok} -> {ok, State2};
+       				{_State2, {error, Reason}} ->
+                 			apply(Handler, handle_error, [list_to_atom(Function), Reason]),
+                 			thrift_protocol:close_transport(Proto1),
+                 			ok
+    			end
 
 init({_Server, ProtoGen, Service, Handler}) when is_function(ProtoGen, 0) ->
     {ok, Proto} = ProtoGen(),
--- a/lib/erl/src/thrift_processor.erl
+++ b/lib/erl/src/thrift_processor.erl
@@ -28,7 +28,7 @@
 
 % Instead of a loop, do a single-shot processing of a packet
 % allows user to control how the bytes are sent and received
-% see: [https://github.com/apache/thrift/blob/master/lib/rb/lib/thrift/server/mongrel_http_server.rb]
+% see: [https://github.com/apache/thrift/blob/88591e32e710a0524327153c8b629d5b461e35e0/lib/rb/lib/thrift/server/mongrel_http_server.rb]
 process({_Server, Proto, Service, Handler}) ->
     State0 = #thrift_processor{ protocol = Proto,
                                 service = Service,
--- a/lib/erl/src/thrift_processor.erl
+++ b/lib/erl/src/thrift_processor.erl
@@ -44,6 +44,7 @@ process({_Server, Proto, Service, Handler}) ->
                  			thrift_protocol:close_transport(Proto1),
                  			ok
     			end
+    end.
 
 init({_Server, ProtoGen, Service, Handler}) when is_function(ProtoGen, 0) ->
     {ok, Proto} = ProtoGen(),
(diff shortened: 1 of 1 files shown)


jeking3:
The Xenial build job failed due to a known issue:

https://issues.apache.org/jira/browse/THRIFT-4567

The AppVeyor builds failed for unrelated reasons.
Is the change complete and ready to be merged?

nsuke:
Thanks @vanjabucic.
It looks like `process` is basically a stripped off version of `loop`.
Does it make sense to add/keep multiplex support and/or error handling (timeout etc) to `process` ?
Also maybe share the same code path in `loop` ?

vanjabucic:
@nsuke, my use-case is to process a thrift message that was received complete and not chunked, similarly to how ruby or java implementations are done.
To get that done, as you noticed, I cut down the `loop` function to become a single message processor. `timeout` and multiplexing make little sense in this context given that the IO is done by someone else (I am picturing standard req/resp via HTTP, or a async queue of thrift instructions via RabbitMQ)
I tried to imagine sharing code path with `loop` but I can't see it. What did you have in mind?
Not being some erlang master I tried to keep this patch safe by not modifying any existing code, only adding a function that exists in other libraries.

@jeking3, the patch is ready as is

nsuke:
@vanjabucic as we're modifying generic interface, I don't think it's good idea to make an assumption for the underlying transport (that someone else does the IO without propagating errors, unlike existing transports).
Multiplex is about handling multiple services in one wire, so relevence doesn't seem to be different for `loop` and `process`.
Looking at it, sticking to error handler and returning `ok` would be really strange interface.

lib/erl/src/thrift_processor.erl
```diff
@@ -19,13 +19,35 @@
...
+-export([init/1, process/1]).
 
 -include("thrift_constants.hrl").
 -include("thrift_protocol.hrl").
 
 -record(thrift_processor, {handler, protocol, service}).
 
+% Instead of a loop, do a single-shot processing of a packet
+% allows user to control how the bytes are sent and received
+% see: [https://github.com/apache/thrift/blob/master/lib/rb/lib/thrift/server/mongrel_http_server.rb]
```
dcelasun: Please link to a specific commit instead of master (press `y` while on that page) so comments like this survive any future refactor/reorganization.

vanjabucic:
What is the next step here?

jeking3:
@nsuke the author is looking for some guidance on this one.

k32:
I've stumbled upon the same problem while trying to implement Cowboy websocket transport. 
I may be mistaken, but I think your approach will work only for framed transport, when it clear that the entire message is fully received. Which may not always be the case.
Ultimately I would change the whole processor logic to mimic haskell's Attoparsec library. Upon reaching the end of a buffer it should return a continuation which can be restarted when more data arrives. This is the most flexible approach without resorting to a middleman process. Unfortunately, I don't see how it can be implemented without changing pretty much everything.

vanjabucic:
You are right that my current approach does not necessarily handle all possible cases. 
It does (awkwardly) handle a simple request/response scenario (http, zeromq, rabbitmq, etc) where requests are received in **full** and passed along to a processor whose job is to generate a response in **full**.
Other language implementations have this simple in-memory-only pathway that allows for decoupling of byte transport and thrift processors.
I would argue that one of the core capabilities a library has to demonstrate is processing of raw byte arrays.
Handling partial payloads for streaming applications should be another core capability, and I think both of them could be done using a single approach where raw bytes are fed on one end and raw responses are received back via callbacks.
In classical gen_server fashion the thrift data handling process would never end, always waiting to ingest more bytes. On complete-payload-detected it would synchronously processes it, fire a raw response to a registered listener/process, and continue ingesting more bytes.
Full and chunked requests would then be equally handled, and transports and processors would be decoupled better.
Not a quick fix though.

jeking3:
I don't know what to do with this, I don't really know erlang.

jeking3:
@dcelasun can we come to a conclusion on this PR and whether we're going to proceed to close?  Thanks.

dcelasun:
I'm not great with Erlang either. Since there are outstanding questions about this PR, I'd rather just close.

vanjabucic:
The fact is that Erlang implementation is lacking this basic functionality that other language implementations have.
Whether we close this PR or not can we agree that we have a problem and that there should be a separate issue open that details the missing functionality?

jeking3:
Sure, if there's an alternate implementation that we can plug in and enable cross tests on with more transports and protocols supported that would be great.

PR #1556, opened 2018-05-13
Updated ZeroMQ examples to ZeroMQ version >= 3.0.

stefbo:
Some ZeroMQ symbols have changes in version 3.0 of ZeroMQ (before 2014). Current stable branch is 4.2. Additionally, the Python code used `newstyle` language option, which isn't supported in the Thrift compiler anymore. This style is the default now.


jeking3:
The reason why these rotted is because they are in contrib/ - perhaps we should think about pulling this into the library proper?

jeking3:
I opened https://issues.apache.org/jira/browse/THRIFT-4571 for this change.

PR #1558, opened 2018-05-27
THRIFT-4564: Reset buffered transport on serialization errors

bforbis:
Tested locally using the following thrift IDL on thrift v0.10.0
```
struct Foo {
  1: string name
}

service SocketTestService {
  i16 countFoos(1: list<Foo> foos)
}
```

The following client request will fail on the client side, but leave data on the buffer:
```
client.countFoos([new thriftTypes.Foo(), null, new thriftTypes.Foo()]);
```

Using the new code, the client side error will be caught and the buffered transport will be reset, allowing for the socket to be used for a subsequent RPC call without causing a parse error on the server side

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_js_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_js_generator.cc
@@ -1529,16 +1529,7 @@ void t_js_generator::generate_service_client(t_service* tservice) {
 
     std::string messageType = (*f_iter)->is_oneway() ? "Thrift.MessageType.ONEWAY"
                                                      : "Thrift.MessageType.CALL";
-
-    // Serialize the request header
-    if (gen_node_) {
-      f_service_ << indent() << outputVar << ".writeMessageBegin('" << (*f_iter)->get_name()
-                 << "', " << messageType << ", this.seqid());" << endl;
-    } else {
-      f_service_ << indent() << outputVar << ".writeMessageBegin('" << (*f_iter)->get_name()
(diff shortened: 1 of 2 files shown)


bforbis:
I don't think we have a danger with this. The underlying net socket will not get reset, just the wrapping buffered transport. Calling `reset()` on the buffered_transport instance that is responsible for writing will only throw out its internal outBuffers in this case, which will only have up to 1 message at a time before flushing it to the net socket (flush is called at the end of send_Foo RPC call in the generated service module). Since this try/catch is catching serialization errors to the outBuffers, the only thing that would ever be thrown out is this corrupt half-message in the transports outBuffer, which we won't need anymore since serialization has failed.

Node will only serialize and send one message at a time, since constructing the message into the outbuffer and flushing to the socket is done synchronously. Reading is done by an entirely different instance of buffered_transport, so resetting the transport responsible for writing will have no affect on receiving messages.

bforbis:
I think some of the confusion might be around which internal state variables are actually used by the outbound portion of the buffered_transport.

Writing to a buffered transport only uses `this.outBuffers` and `this.outCount`. The rest of the internal state variables like `readCursor` and `writeCursor` are used for reading inbound data, and are done by a different instance of the buffered transport class. It's kind of confusing, because you would think `writeCursor` would be used for sending data, but it is actually used for copying data from the inbound stream to the `inBuf` buffer object.

jeking3:
Okay, thanks, I'm going to rebase it on the current master for you to kick another build.

PR #1559, opened 2018-06-03
THRIFT-4478 Thrift will not build with dlang 2.078 or later

nsuke:
This fixes build errors and deprecation warnings on dmd v2.080.0.


nsuke:
@jeking3 thanks, looks like the part depending on libevent was left out.
Fixed the build error, but haven't touch the warnings.

jeking3:
I need to update the LANGUAGES.md and docker README to reflect the new uppper limit on dlang and check the lib/d README as well to be sure it doesn't still show a limit.

jeking3:
These changes fail to build in Xenial unless I move the version of dlang from 2.073 to 2.075, so I am setting that as the new minimum version.

PR #1560, opened 2018-06-03
THRIFT-4577 Outdated cipher string in python unit test

nsuke:
`test_ciphers` fails if system SSL doesn't support "DES-CBC3-SHA" which is quite old.
Adding an additional modern cipher fixes it.

Also fixes a bug where unit test execution hanged forever on SSL initialization failure.


jeking3:
I just ran into this while updating to Ubuntu Bionic (THRIFT-4579) so I am going to merge this so I can make progress.  I'll swing around and figure out what this Xenial build failure is that's causing all the builds to fail later.

PR #1561, opened 2018-06-06
tutorial/erl/*.sh: change the order of symlinks

emmenlau:
Currently, tutorial/erl/client.sh is a symlink to tutorial/erl/server.sh. During extraction of a thrift archive on Windows, this creates temporarily a dangling symlink, because client.sh is extracted before server.sh, and points to a (then) non-existing file. This breaks tar on Windows:
```
tar: tutorial/erl/client.sh: Cannot create symlink to ‘server.sh’: No such file or directory
tar: Exiting with failure status due to previous errors
```
A simple workaround is to change the order of symlinks in thrift. The archive typically has the files in alphabetical order, so the target will exist when the symlink is extracted.


jeking3:
THRIFT-4580

PR #1562, opened 2018-06-06
THRIFT-4578 Move `TAsyncProtocolProcessor` into main thrift library

stefbo:
Updated CMake and automake files for C++.


jeking3:
The build failures are unrelated to the change.

PR #1564, opened 2018-06-15
THRIFT-4586: Correctly generates NSString with copy instead of strong - Cocoa

CarbonComputed:
Cocoa generator generates NSString with copy instead of strong


jeking3:
I'll need someone who knows the language to code review. and we can ignore the CI because we don't currently test cocoa in CI.

CarbonComputed:
@Jens-G Any thoughts on this?

Jens-G:
Sory, no. I do not feel qualified enough for that task. Don't we have some cocoa users around? Hard to believe ...

jeking3:
CI failures for this PR don't mean much since we don't test cocoa in the CI build today :(

PR #1566, opened 2018-06-24
THRIFT-4591:cannot read call result via lua lib

zhiyong0804:
THRIFT-4591: [cannot read call result via lua lib]
Client: [lua,  lib/lua]

### DIFF
--- a/lib/lua/TFramedTransport.lua
+++ b/lib/lua/TFramedTransport.lua
@@ -100,7 +100,7 @@ function TFramedTransport:flush()
   local tmp = self.wBuf
   self.wBuf = ''
   local frame_len_buf = libluabpack.bpack("i", string.len(tmp))
-  self.trans:write(frame_len_buf)
+  tmp = frame_len_buf .. tmp
   self.trans:write(tmp)
   self.trans:flush()
 end


nsuke:
What is the underlying transport ?
It looks like an issue of underlying transport implementation where data written by multiple `write` call cannot properly be read by receiver side.

zhiyong0804:
@jeking3 the server is writen with c++ library, i donot know if the C++ library should be supported separated reading function, write "ab" is faster than write "a" and then write "b"; otherwise this correction can fix the problem, after all it may not match the root cause.

zhiyong0804:
And the attachment upload to jira show the correction work fine!!!

jeking3:
I believe the error is in the implementation of TFramedTransport in the C++ library.  Specifically, TFramedTransport readSlow() attempts to read a size off the underlying transport and then attempts to read as much as possible from the underlying transport.  If it doesn't read the entire message indicated by the size, it will return a partial read.  The upper layer will then call readSlow again, which will try to read another size off the wire, destroying message integrity.  It's amazing to see such a bug in code that's been around for a long time.  Funny thing is I have been adding support for message bus transports based on TFramedTransport being correct, and obviously it is not.

So the lua library is just fine writing size and then writing payload.  I think that's still better than copying the size and payload into another buffer (increasing memory requirements to write by 2x) and then writing all at once.

So you correctly identified there was a problem but the issue is on the C++ side, not on the lua side.  If another client did the same thing, the C++ server would fail the same way.

That said, I'd like to take over this issue in Jira and I will fix TFramedTransport, hopefully this week.  My current employer is providing me with the time to work on message bus support for thrift, and I need TFramedTransport to work properly.  In fact, I may add a test that specifically targets this issue to see if it is present on any other library implementation.

I'm closing this PR as this change is not necessary.  Thanks.

jeking3:
python library does the same thing as this change so I am going to merge it.  It previously passed all CI tests.

jeking3:
"THRIFT_EAGAIN (unavailable resources)" means your process ran out of resources; you may want to look at your process limits, or see if you are sending a message too large (but not larger than the maximum frame size).

PR #1567, opened 2018-06-25
THRIFT-4496: python specific list of keywords for python generator

nsrtvwls:
Client: py

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_generator.h
+++ b/compiler/cpp/src/thrift/generate/t_generator.h
@@ -42,7 +42,8 @@
  */
 class t_generator {
 public:
-  t_generator(t_program* program) {
+  t_generator(t_program* program)
+    : keywords_(lang_keywords()){
     tmp_ = 0;
     indent_ = 0;
     program_ = program;
@@ -96,7 +97,33 @@ public:
     return escape_string(constval->get_string());
   }
 
+  /**
+   * Check if all identifiers are valid for the target language
+   */
+  virtual void validate_input() const;
+
 protected:
+  virtual std::set<std::string> lang_keywords() const;
+
+  /**
+   * A list of reserved words that cannot be used as identifiers.
+   */
+  const std::set<std::string> keywords_;
+
+  virtual void validate_id(const std::string& id) const;
+
+  virtual void validate(t_enum const* en) const;
+  virtual void validate(t_enum_value const* en_val) const;
+  virtual void validate(t_typedef const* td) const;
+  virtual void validate(t_const const* c) const;
+  virtual void validate(t_service const* s) const;
+  virtual void validate(t_struct const* c) const;
+  virtual void validate(t_field const* f) const;
+  virtual void validate(t_function const* f) const;
+
+  template <typename T>
+  void validate(const std::vector<T>& list) const;
+
   /**
    * Optional methods that may be implemented by subclasses to take necessary
    * steps at the beginning or end of code generation.
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -51,9 +51,8 @@ public:
   t_py_generator(t_program* program,
                  const std::map<std::string, std::string>& parsed_options,
                  const std::string& option_string)
-    : t_generator(program) {
+    : t_generator (program) {
     std::map<std::string, std::string>::const_iterator iter;
-
 
     gen_newstyle_ = true;
     gen_utf8strings_ = true;
@@ -334,6 +333,15 @@ private:
 
   std::string package_dir_;
   std::string module_;
+
+protected:
+  virtual std::set<std::string> lang_keywords() {
+    std::string keywords[] = { "False", "None", "True", "and", "as", "assert", "break", "class",
+          "continue", "def", "del", "elif", "else", "except", "exec", "finally", "for", "from",
+          "global", "if", "import", "in", "is", "lambda", "nonlocal", "not", "or", "pass", "print",
+          "raise", "return", "try", "while", "with", "yield" };
+    return std::set<std::string>(keywords, keywords + sizeof(keywords)/sizeof(keywords[0]) );
+  }
 };
 
 /**
--- a/compiler/cpp/CMakeLists.txt
+++ b/compiler/cpp/CMakeLists.txt
@@ -62,6 +62,9 @@ set(thrift-compiler_SOURCES
     src/thrift/audit/t_audit.cpp
 )
 
+set(thrift_compiler_LANGS
+)
+
 # This macro adds an option THRIFT_COMPILER_${NAME}
 # that allows enabling or disabling certain languages
 macro(THRIFT_ADD_COMPILER name description initial)
@@ -70,6 +73,7 @@ macro(THRIFT_ADD_COMPILER name description initial)
     option(${enabler} ${description} ${initial})
     if(${enabler})
         list(APPEND thrift-compiler_SOURCES ${src})
+        list(APPEND thrift_compiler_LANGS ${name})
     endif()
 endmacro()
 
--- a/compiler/cpp/src/thrift/generate/t_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_generator.cc
@@ -76,6 +76,77 @@ void t_generator::generate_program() {
   close_generator();
 }
 
+std::set<std::string> t_generator::lang_keywords() const {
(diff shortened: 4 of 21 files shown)


compiler/cpp/src/thrift/generate/t_generator.h
```diff
@@ -96,7 +97,47 @@ class t_generator {
...
   }
 
+  /**
+   * Check if all identifiers are valid for the target language
+   */
+  virtual void validate_input() const;
+
 protected:
+  virtual std::set<std::string> lang_keywords() {
+    std::string keywords[] = { "BEGIN", "END", "__CLASS__", "__DIR__", "__FILE__", "__FUNCTION__",
```
jeking3: Would prefer to see code and data in the implementation file, but not a blocker.

compiler/cpp/src/thrift/generate/t_generator.h
```diff
@@ -96,7 +97,47 @@ class t_generator {
...
+        "__LINE__", "__METHOD__", "__NAMESPACE__", "abstract", "alias", "and", "args", "as",
+        "assert", "begin", "break", "case", "catch", "class", "clone", "continue", "declare",
```
jeking3: Should we include C++11 keywords like auto, constexpr, nothrow?

https://en.cppreference.com/w/cpp/keyword

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -334,6 +333,15 @@ class t_py_generator : public t_generator {
 
   std::string package_dir_;
   std::string module_;
+
+protected:
+  virtual std::set<std::string> lang_keywords() {
+    std::string keywords[] = { "False", "None", "True", "and", "as", "assert", "break", "class",
+          "continue", "def", "del", "elif", "else", "except", "exec", "finally", "for", "from",
+          "global", "if", "import", "in", "is", "lambda", "nonlocal", "not", "or", "pass", "print",
```
jeking3: Are "exec" and "print" python keywords?

jeking3:
Another concern I see is that if we do not check for a complete list of keywords during compilation for all languages, when someone wants to add a new language to their build they might run into a problem later on with a language-specific violation.  Fortunately Thrift uses the identifying number and not the name of the item for fields, but not so for names of structures, etc.  So it is possible someone might run into issues in the future adding output for more languages and need to change their thrift file, and need to change other implementations.

Mitigation for this will be to get a list of keywords into every compiler-language implementation and instruct folks to generate code for each language they expect to use, to check for keyword violations early.

It's still not exactly how Thrift is supposed to operate, which is to offer the lowest common denominator across all languages, which really leads us to want to put the entire list of reserved keywords into the .ll file as they were, but grow the list to include keywords for all languages.

I'd like to hear from some other long-time thrifters on this approach and see what they think... should the list of language keywords be language specific or global?

compiler/cpp/src/thrift/generate/t_py_generator.cc
nsrtvwls: "exec" and "print" are python2 keywords. I merged python2 and python3 keyword lists.

nsrtvwls:
we had a discussion with @Jens-G about a global list of keywords in JIRA: https://issues.apache.org/jira/browse/THRIFT-4496?focusedCommentId=16364858&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16364858

compiler/cpp/src/thrift/generate/t_generator.h
nsrtvwls: I don't think it's safe to add keywords to this keyword list, because it will add restrictions to all the languages (except for py) and break things. It's possible to add a complete list of cpp keywords to cpp generator.

compiler/cpp/src/thrift/generate/t_generator.h
nsrtvwls: Done

nsrtvwls:
> It looks like by removing words from the .ll file like "enddeclare" and not reimplementing it elsewhere, we're losing some keyword validation.

Could you please elaborate? In which cases it is better to perform keyword validation in .ll vs generator?

jeking3:
Let's say we only check keywords based on language, as this PR changes the system to do.  If I use a lua keyword but generate C++, I won't get an error.  I build a product.  Now I want to add lua support, I go to generate lua and it complains.  It's a structure name that I cannot use.  I have to change it, and I have to do a bunch of work.  Is it backwards compatible if I change a structure name (unsure)?  If we check for all keywords across all languages up front it avoids future surprises.

nsrtvwls:
It can definitely be a problem. E.g. service function cannot be renamed transparently.  On the other hand, a global list of keywords is hard to maintain. What if thrift decides to support a new language with another set of keywords? Adding to the global list will break things to people who don't even care about this new language.
I like the idea of merging keyword lists from all active generators. Will it work if I show a warning, not an error if an identifier is a keyword in some language that is supported but is not used for generation?

jeking3:
That's probably why it was in the parser to begin with, because one could emit the line number where the error occurred.

jeking3:
I like the notion that we populate the keywords for all languages following your existing changes, and then have the compiler check all of them regardless of what you are outputting and emitting a warning if it isn't something you are not generating, and an error if you are.

compiler/cpp/src/thrift/generate/t_py_generator.cc
Jens-G: > It will be fun to add more keyword lists over time, for example for perl.

You have a strange understanding of fun …

Jens-G:
At least in theory it seems simple. In practice it may require some more work (but only once).

1. Get rid of that global thing or reduce it down to the absolute minimun. Place a big comment on to and below that anybody who is ever going to add a keywored to that list without a really good reason will be punished (methaporically speaking, of course).
2. Create a mechanism flexible enough to suit the needs of all code generators that deals with language-specific Keywords.
3. Every morning repeat the mantra "**The issue is the code generation, not the IDL. It's the generator, not the parser.**"

Jens-G:
I finally had the time to have a closer look at it.

1. general approach is not bad. I still would strive to reduce that global list to the minimum, but maybe we have to do that in a second step.

2. what I find less helpful is the fact that the names are causing a failure(). In certain cases, reserved names can still be used but require a bit of treamtment. For example, in C# it is done by prefixing such a name with `@` to make *any* identifier a valid one, such as `@return` instead of `return`. The generator can do this alone and therefore there is no need to throw at all. In other languages we often added an underscore at the end of the name, which requireds a bit of tracking but otherwise works quite well. 

4. Last not least, the reserved identifiers may also be context dependent. There are keywords, there are reserved words and there are special built-in method and property names used by the runtime that should not be used either, even if the are not a keyword in the strict sense. Context dependent now means, that expecially the latter ones differ between regular structs/classes and exceptions, to name just one possible case.

nsrtvwls:
1. I can remove python keywords not used by other supported languages from the global list. It seems a bit risky though.

2. I really can't see a problem here. With this change the behavior for all generators (except for py) is exactly the same as before: a keyword from the global list causes a failure. As soon as generator learns to make valid ids from keywords, it overrides the validate_input method with no-op or even implements a context dependent validation.

3. I can see your concern here. Speaking of py generator, I haven't found any logic for built-ins, so maybe it's better to start from keywords first. As I've said before, a generator can implement these complex checks in their validate_input methods.

I couldn't find a way to make it easy for all generators to generate valid code from a source with a keyword. There are cases when the original id is needed to suppport protocol and there are cases when the id should be tweaked to make code compileable. It seems like there is no way around going through the code and changing a half of id occurences. After it's done, switching off failures on keywords seems pretty easy.

compiler/cpp/src/thrift/generate/t_generator.h
```diff
@@ -96,7 +97,33 @@ class t_generator {
...
+  virtual std::set<std::string> lang_keywords() const;
+
+  /**
+   * A list of reserved words that cannot be used as identifiers.
+   */
+  const std::set<std::string> keywords_;
+
+  virtual void validate_id(const std::string& id) const;
+
+  virtual void validate(t_enum const* en) const;
```
jeking3: validate looks like an opportunity to use template specialization, since half of them do the same thing.  Just a thought. :)

jeking3:
I believe the net effect is that as long as you generate thrift for all the possible languages you will be using now and in the future, you will get the keyword validation you desire, without worrying about keywords from languages you will never use.  I'm going to push the build forward and see if it will complete, and if not then see if it's just one of those intermittent issues or something we already solved.  There's only one build job outstanding right now.  Then we can merge and move this forward.  Sorry for the delay.

PR #1568, opened 2018-06-27
THRIFT-3950: Memory leak while calling oneway methods

bforbis:
Looks like this was cleaned up for normal JS oneway methods, but not nodejs. This also fixes another issue where promises returned by oneway methods would never resolve (since the callback was never called), so waiting on them would cause the client application to hang forever.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_js_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_js_generator.cc
@@ -1565,7 +1565,14 @@ void t_js_generator::generate_service_client(t_service* tservice) {
                << ".writeMessageEnd();" << endl;
 
     if (gen_node_) {
-      f_service_ << indent() << "return this.output.flush();" << endl;
+      if((*f_iter)->is_oneway()) {
+        f_service_ << indent() << "this.output.flush();" << endl;
+        f_service_ << indent() << "var callback = this._reqs[this.seqid()] || function() {};" << endl;
+        f_service_ << indent() << "delete this._reqs[this.seqid()];" << endl;
+        f_service_ << indent() << "callback(null);" << endl;
+      } else {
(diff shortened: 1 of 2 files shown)


bforbis:
Updated the tests. It's bizarre to me that this was actually tested behavior, as it seems broken.

Is this okay, presuming the tests pass after this build?

bforbis:
Hmm, tests are still failing. I may need to figure out how to run cross-tests locally.

jeking3:
Instructions for running cross tests in a docker container can be found in build/docker/README.md.   The cross tests were not the ones that failed.  Also, oneway thrift APIs do not process responses.  In fact the code generator for oneway will emit a send method which pushes the message out and then is done.  Sending a response may actually confuse some implementations.  If you send a oneway then a twoway right after it, will the twoway response that is read off the wire actually try to consume the oneway response you added here?  I think so, and things will break.

bforbis:
No, this is just calling the callback that is provided from the application for sending the oneway. A subsequent two way request would have it's own callback and wouldn't get mixed up.

bforbis:
This code only affects the clientside, not the server side. This is not about sending a serverside oneway response, it's about executing the supplied clientside callback after the client requested oneway has been sent to the server. The clientside callback will resolve immediately with `null` once the message has been flushed to the server to notify the application that it successfully sent the oneway.

Otherwise, this allows for two bad behaviors which this PR fixes:
1. Supplied callbacks to oneways on the client will build up in the _reqs registry and never get executed since oneways will never receive responses.
2. All client calls return promises that can be waited on with `await` or put in a promise chain with `.then()`. Except oneways will never resolve, since their callback just gets put into a registry and never executed.

bforbis:
Tests should be passing now

PR #1570, opened 2018-06-28
change docker build environment to use a non-root user

jeking3:
The docker build environment was using a root user, now it uses a normal user account (thrift) with the same password to do the build.  This ensures we can build without being root.


jeking3:
This isn't really feasible right now, not until docker allows for mapping in-container UIDs to external ones.  Until then we'll keep building as container root.

PR #1574, opened 2018-07-29
THRIFT-4605: bootstrap.sh fails if automake=1.16.1

beltran:
Version comparison may fail if automake version X.Y.Z


beltran:
I see, thank you for jumping in @jeking3

beltran:
Is the failure a flake? Looks similar to https://api.travis-ci.org/v3/job/390501002/log.txt, from a past commit in master

jeking3:
Some parts of the build are still slightly unstable.  A project with support for 25 programming languages is bound to have issues here and there, especially as the versions of those languages change.  I'll kick the failed jobs and see if it clears.

beltran:
Thanks for merging @jeking3 . Flakes are normal in my experience even in small projects.

PR #1575, opened 2018-08-02
Fix comparison of integers of different signs

kainjow:
This fixes a warning in Xcode/clang "Comparison of integers of different signs: 'int' and 'UInt32' (aka 'unsigned int')"


jeking3:
Normally we like to tag everything with a Jira ticket number - this one is very trivial and isn't tested in CI (because it is cocoa) so I am just going to merge it.

PR #1577, opened 2018-08-06
THRIFT-4614: Generate missing @Nullable annotations for Java iterator getters

msridhar:
THRIFT-4614: Generate `@Nullable` annotations for Java iterator getters
Client: Java compiler

This is a straightforward, compile-time-safe, backward-compatible change.


dcelasun:
Merged, thanks.

PR #1578, opened 2018-08-10
THRIFT-4617: Prepend service-specific struct names with service name

flavray:
Client: rs

If a thrift definition file contains multiple services, with functions
having the same name, there would be conflicting *Args and *Result
structs generated by the rust generator.

This commit adds prefixing of `*Args` and `*Result` structs with
the service name, to resolve any conflict.

e.g:
```thrift
service backend {
  i32 send();
}
```

`SendArgs` -> `BackendSendArgs`
`SendResult` -> `BackendSendResult`

This change is technically non-backwards compatible (if any code decided to call their own structs `<ServiceName><FunctionName>{Args,Result}`, names would clash. However it seems quite unlikely.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -350,6 +350,7 @@ private:
   void render_sync_handler_failed_default_exception_branch(t_function *tfunc);
   void render_sync_handler_send_exception_response(t_function *tfunc, const string &err_var);
   void render_service_call_structs(t_service* tservice);
+  void render_args_struct(t_function* tfunc);
   void render_result_value_struct(t_function* tfunc);
 
   string handler_successful_return_struct(t_function* tfunc);
@@ -472,6 +473,9 @@ private:
 
   // Return the name of the function that users will have to implement to handle incoming service calls.
   string service_call_handler_function_name(t_function* tfunc);
+
+  // Return the name of the struct used to pack the arguments for the thrift service call.
+  string service_call_args_struct_name(t_function* tfunc);
 
   // Return the name of the struct used to pack the return value
   // and user-defined exceptions for the thrift service call.
@@ -2050,7 +2054,7 @@ void t_rs_generator::render_service_call_structs(t_service* tservice) {
   // contains the exceptions as well
   for(func_iter = functions.begin(); func_iter != functions.end(); ++func_iter) {
     t_function* tfunc = (*func_iter);
-    render_struct(rust_struct_name(tfunc->get_arglist()), tfunc->get_arglist(), t_rs_generator::T_ARGS);
+    render_args_struct(tfunc);
     if (!tfunc->is_oneway()) {
       render_result_value_struct(tfunc);
     }
@@ -2297,7 +2301,7 @@ void t_rs_generator::render_sync_send(t_function* tfunc) {
   f_gen_
     << indent()
     << "let call_args = "
-    << rust_struct_name(tfunc->get_arglist())
+    << service_call_args_struct_name(tfunc)
     << " { "
     << struct_fields
     << " };"
@@ -2409,6 +2413,11 @@ string t_rs_generator::struct_to_invocation(t_struct* tstruct, const string& fie
   }
 
   return args.str();
+}
+
+void t_rs_generator::render_args_struct(t_function* tfunc) {
+    string args_struct_name = service_call_args_struct_name(tfunc);
+    render_struct(args_struct_name, tfunc->get_arglist(), t_rs_generator::T_ARGS);
 }
 
 void t_rs_generator::render_result_value_struct(t_function* tfunc) {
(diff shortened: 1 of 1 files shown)


dcelasun:
Thanks for the PR! LGTM, but someone familiar with Rust might also want to take a look. Travis had some timeouts, I've restarted it.

flavray:
Great, thanks!

I'll just ping @allengeorge who (I think) wrote most of the Rust integration 🙂

flavray:
Or maybe @jfarrell, who also owns the [thrift crate](https://crates.io/crates/thrift)? 🙂

allengeorge:
Ah - sorry; was on an extended break! I’ll take a look at it today

Get Outlook for iOS<https://aka.ms/o0ukef>

flavray:
Ah great, thanks @allengeorge! 🙂

flavray:
Hi @allengeorge , any chance you had enough bandwidth to have a look at this change? 🙂

allengeorge:
Sorry Flavien - not yet :/ I’ll make it my top priority tomorrow.

I’m really sorry; I try to be on top of Rust Thrift changes, but this has been a bad month. Again, I really apologize for the delay and *will* get it done tomorrow.

Thank you so much for making the PR!

Get Outlook for iOS<https://aka.ms/o0ukef>

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -350,6 +350,7 @@ class t_rs_generator : public t_generator {
   void render_sync_handler_failed_default_exception_branch(t_function *tfunc);
   void render_sync_handler_send_exception_response(t_function *tfunc, const string &err_var);
   void render_service_call_structs(t_service* tservice);
+  void render_args_struct(t_function* tfunc);
```
allengeorge: Perhaps rename to `render_service_args_struct`?

allengeorge: Or `render_service_call_args_struct`?

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -3212,8 +3221,12 @@ string t_rs_generator::service_call_handler_function_name(t_function* tfunc) {
   return "handle_" + rust_snake_case(tfunc->get_name());
 }
 
+string t_rs_generator::service_call_args_struct_name(t_function* tfunc) {
```
allengeorge: Interesting that the definitions of these two functions are different. For example, why does `service_call_result_struct_name` include the `Result` suffix but the `service_call_args_struct_name` not?

I...assume (?) it's because the arglist has an `Args` suffix already? If so, could you please add a comment stating this?

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -2668,7 +2677,7 @@ void t_rs_generator::render_sync_process_function(t_function *tfunc, const strin
     << "let "
     << (has_non_void_args(tfunc) ? "args" : "_")
     << " = "
-    << rust_struct_name(tfunc->get_arglist())
```
allengeorge: Good change. Now both the `service_call_{args,result}_struct_name` functions take the same parameter 👍

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -2411,6 +2415,11 @@ string t_rs_generator::struct_to_invocation(t_struct* tstruct, const string& fie
   return args.str();
 }
 
+void t_rs_generator::render_args_struct(t_function* tfunc) {
+    string args_struct_name = service_call_args_struct_name(tfunc);
```
allengeorge: I believe you can simply do:

`string args_struct_name(service_call_args_struct_name(tfunc);`

compiler/cpp/src/thrift/generate/t_rs_generator.cc
flavray: You're correct, the `Args` suffix is already present (coming directly from the thrift compiler structs). Good point, I'll add a comment to make it clear 🙂

compiler/cpp/src/thrift/generate/t_rs_generator.cc
flavray: `render_service_call_args_struct` sounds good to me! I'll also change `render_result_value_struct` into `render_service_call_result_value_struct` to stay consistent 🙂

compiler/cpp/src/thrift/generate/t_rs_generator.cc
flavray: Ah thanks, good point! C++ is not my mother tongue 😄

jeking3:
Cross tests work based on the passing CI builds, however @allengeorge has not changed his rejection to an approval yet.

flavray:
Awesome! Thanks for reviewing 🙂

PR #1580, opened 2018-08-11
THRIFT-4618: Use poll() instead of select() in C++ TNonblockingServer if available

st0ke:
Got poll() code from b5ebcd1


st0ke:
"continuous-integration/travis-ci/pr" failed not because of my changes. Please, help.

jeking3:
Sometimes it needs a little help.  We can restart individual jobs.  I have done that.

jeking3:
I don't think non-blocking servers are tested in the current cross test suite...

st0ke:
Do I need to rebase the commit?

jeking3:
Unlikely, there are some stubborn failures.   I will try to rekick.

st0ke:
> I see changes for cmake - what about autoconf?

Done

st0ke:
> I see changes for cmake - what about autoconf?

Done

duynl58:
Select is slow.
Please consider epoll/kqueue instead.

st0ke:
> Select is slow.
> Please consider epoll/kqueue instead.

1. Current realization use select(). I add poll() if available, because FD_SET is failed if descriptor is bigger than 1024.
2. Does epoll/kqueue do it faster than poll on only two descriptors? If so, does anybody need this small boost?

lib/cpp/src/thrift/server/TNonblockingServer.cpp
```diff
@@ -1291,10 +1295,44 @@ bool TNonblockingIOThread::notify(TNonblockingServer::TConnection* conn) {
...
+    pfd.revents = 0;
+    ret = poll(&pfd, 1, -1);
+    if (ret < 0) {
+      return false;
+    } else if (ret == 0) {
+      continue;
+    }
+
+    if (pfd.revents & POLLHUP || pfd.revents & POLLERR) {
+      ::close(fd);
```
jeking3: Should this and the other instances of close instead be THRIFT_CLOSESOCKET like the other code path?

st0ke: Yes, it should. closesocket() is used on win platform.

jeking3:
Sadly I think our coverage on non-blocking is lax, so I'm sure how well this is tested across platforms; and there's no cross test that runs on Windows.

st0ke:
"Travis build" failed in D language tests. Do I need to do something with that?

PR #1581, opened 2018-08-12
THRIFT-4619: Transport SASL Client fo go

beltran:
I've tested this against hive using gssapi and plain mechanisms with [gohive](https://github.com/beltran/gohive/)

### DIFF
--- a/lib/go/thrift/sasl_transport.go
+++ b/lib/go/thrift/sasl_transport.go
@@ -1,0 +1,245 @@
+package thrift
+
+import (
+	"bytes"
+	"context"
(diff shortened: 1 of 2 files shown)


dcelasun:
Thanks for the PR, but I don't think this belongs here upstream, especially with that third party dependency. As far as I can see, gohive doesn't even have a license specified, which is a showstopper on its own.

I recommend maintaining this outside the tree, like [Python](https://github.com/cloudera/thrift_sasl) and [Perl](https://github.com/Perl-Hadoop/Thrift-SASL) communities do.

PR #1584, opened 2018-08-26
THRIFT-4625: Use let/const variable decorators in ES6 Javascript

bforbis:
This PR adds `let` and `const` decorators to generated javascript code when specifying the new es6 compiler option. These changes will fall back to the `var` decorator if not specifying the es6 compiler.


compiler/cpp/src/thrift/generate/t_js_generator.cc
```diff
@@ -1134,7 +1156,12 @@ void t_js_generator::generate_process_function(t_service* tservice, t_function*
   f_service_ << indent() << "if (this._handler." << tfunction->get_name()
              << ".length === " << fields.size() << ") {" << endl;
   indent_up();
-  indent(f_service_) << "Q.fcall(this._handler." << tfunction->get_name() << ".bind(this._handler)";
```
bforbis: This is also fixing a bug in the es6 compiler which references the `Q` package without importing it. ES6 should use the native `Promise.resolve` instead.

bforbis:
I _think_ this is breaking the normal js/es6 tests (without node), but I'm less familiar with how those are set up. Probably the tests need to be set up with a babel transpiler rather than just using the `es6-promise` polyfill.

compiler/cpp/src/thrift/generate/t_js_generator.cc
```diff
@@ -860,14 +883,15 @@ void t_js_generator::generate_js_struct_reader(ostream& out, t_struct* tstruct)
...
 
   // Loop over reading in fields
-  indent(out) << "while (true)" << endl;
+  indent(out) << "while (true) {" << endl;
 
-  scope_up(out);
+  indent_up();
 
-  indent(out) << "var ret = input.readFieldBegin();" << endl;
-  indent(out) << "var fname = ret.fname;" << endl;
```
jeking3: Is fname no longer important?

bforbis: I've removed instances of temporary variables that are not used in any of the generated code. There is currently a lot of clutter right now.

bforbis:
It seems like I ran into an issue where I pulled on a thread and unravelled the entire rug. This requires a decently large overhaul of the test system for javascript, which is currently relying on a very old version of QUnit which relies on a now defunct version of PhantomJS.

QUnit has since moved to using puppeteer on top of headless chrome, which has full ES6 support. The advantage of this is we don't need any transpilation / polyfills, but getting the build dependencies installed across all the docker images is a bit tricky. 

I think I've gotten it resolved for ubuntu-bionic, but ubuntu-xenial is showing some issues (I think due to it having an older version of nodejs)

bforbis:
Builds are passing locally for me, but the latest travis build had an issue installing the Rust dependencies. Looks like it isn't installing within the expected semver range.

bforbis:
@jeking3 any ideas on the Dart dependency issue. I can try to dig in, but I am not sure it's related to this branch.

jeking3:
It's possible something changed in that area.  
It's one of the challenges of supporting 25+ languages in one project.
I re-kicked the build jobs to see if it was transient.

bforbis:
Looks like version 2.0.0 of Dart [was released on August 6th](https://github.com/dart-lang/sdk/releases), so `apt-get install dart/stable` will now install that version of the SDK.

However, the version wanted in /lib/dart/pubspec.yaml is `>=1.13.0 < 2.0.0`.

Is this safe to update?

bforbis:
Pinning the dart versions to 1.x in the Dockerfile seemed to help get the build to pass. There's one more build failure for Dlang on xenial, which I am unsure about. Looks to be that the tests hang when I run it locally.

Anyway, I think this PR should be ready for re-review.

jeking3:
@bforbis the issue you see there is that Xenial has OpenSSL-1.0 and Bionic has OpenSSL-1.1; the dlang deimos for openssl supports only one version, so the Dockerfile for Bionic has been pulling from my branch, and the Dockerfile for Xenial is supposed to be pulling in a tag, but it's pulling in master instead, and 1.1 changes were merged into master upstream for that.  I will submit a separate PR to fix the Xenial Dockerfile then we can merge this.

PR #1585, opened 2018-08-26
THRIFT-4624: Fix refc binary leak

k32:
Client: erl

### DIFF
--- a/lib/erl/src/thrift_compact_protocol.erl
+++ b/lib/erl/src/thrift_compact_protocol.erl
@@ -42,7 +42,8 @@
                            read_stack=[],
                            read_value=?CBOOL_NONE,
                            write_stack=[],
-                           write_id=?ID_NONE
+                           write_id=?ID_NONE,
+                           clone_binaries=true
                           }).
 -type state() :: #t_compact{}.
 -include("thrift_protocol_behaviour.hrl").
@@ -87,7 +88,8 @@ cbool_to_bool(Value) -> Value =:= ?CBOOL_TRUE.
 
 new(Transport) -> new(Transport, _Options = []).
 
-new(Transport, _Options) ->
+new(Transport, Options) ->
+  CloneBinaries = proplists:get_value(clone_binaries, Options, true),
   State  = #t_compact{transport = Transport},
   thrift_protocol:new(?MODULE, State).
 
@@ -329,11 +331,11 @@ read(This0 = #t_compact{read_value = Bool}, bool) ->
   {This0#t_compact{read_value = ?CBOOL_NONE}, {ok, Bool}};
 
 read(This0, ubyte) ->
-  {This1, {ok, <<Val:8/integer-unsigned-big, _/binary>>}} = read_data(This0, 1),
+  {This1, {ok, <<Val:8/integer-unsigned-big, _/binary>>}} = read_data(This0, 1, false),
   {This1, {ok, Val}};
 
 read(This0, byte) ->
-  {This1, Bytes} = read_data(This0, 1),
+  {This1, Bytes} = read_data(This0, 1, false),
   case Bytes of
     {ok, <<Val:8/integer-signed-big, _/binary>>} -> {This1, {ok, Val}};
     Else -> {This1, Else}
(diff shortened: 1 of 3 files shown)


jeking3:
I wonder if @walter-weinmann or @dhull could take a look at this perhaps?

lib/erl/src/thrift_compact_protocol.erl
```diff
@@ -363,26 +365,32 @@ read(This0, double) ->
...
     {#t_compact{}, {ok, binary()} | {error, _Reason}}.
-read_data(This, 0) -> {This, {ok, <<>>}};
-read_data(This = #t_compact{transport = Trans}, Len) when is_integer(Len) andalso Len > 0 ->
-    {NewTransport, Result} = thrift_transport:read(Trans, Len),
+read_data(This, 0, _) -> {This, {ok, <<>>}};
+read_data(This, Len, Clone) when is_integer(Len) andalso Len > 0 ->
+    #t_compact{transport = Trans} = This,
+    {NewTransport, Result0} = thrift_transport:read(Trans, Len),
+    Result = case Result0 of
+                 {ok, Bin} when is_binary(Bin), Clone, Len > 64 ->
```
k32: The last condition should be removed. Cursory look at the BEAM code revealed that only brand new binaries can be allocated on heap, no copying is done for slices

jeking3: Are we missing a test case then? :)

k32: I'm not sure how one can reliably test this, as refc/heap binary dichotomy is hidden in the Erlang VM internals and there is no API to check which is which. The only (known to me) way to detect refc memory leak is through memory region monitoring, and it's beyond scope of unit/component tests. So regression testing is probably the best we can do.

jeking3: It sounded like you wanted to make one more change based on your original comment?

k32: Done

jeking3:
All the tests pass however I'd like one independent code review before merging if possible.

k32:
No hurry

djnym:
The code seems reasonable, but I'm wondering of the need.  Do you actually see refc binary leakage?  I could maybe see that happen if you have long lived connections, from which you subtract a smallish portion of the binary from a much larger thrift object and pass it around, but I'm not totally certain.  Is there a case where you are actually seeing this leakage occuring?  @k32 Have you checked out https://github.com/openx/ox-thrift which @dhull wrote?  It uses the ranch acceptor pool instead of the standard server and is a complete rewrite of the message encoding/decoding.  It's about 5-10x faster at encoding/decoding, so it's an improvement.  The only way to make it faster would be to rewrite the compiler which @dhull has looked at but not done (the idea would be to generate a custom module for encoding/decoding instead of the generic structures which then get interpreted, this should lead to even more efficiency as you'd not have to go through an intermediate representation).  Anyway, maybe David will take a look at this, but I think it's probably fine to merge.  My only real comment would be that this appears to make the default copying the binaries, which could be a surprise if people aren't expecting it and are moving large thrift structures around.  So it might be worthwhile to document that somewhere, not sure where though.

k32:
@djnym Thanks for the suggestion; on the first glance ox-thrift looks very promising. Regarding the defaults: it's just "safety first" approach. This can be changed.

jeking3:
I'm curious to know the answer to:  "Is there a case where you are actually seeing this leakage occuring?"

k32:
I found these errors via code review, while evaluating this library and doing some rapid prototyping. I'm not using Thrift in production. However, I would recommend merging at least part changing list_to_atom to list_to_existing_atom, because (if it wasn't obvious) it fixes a pretty bad DOS vulnerability, allowing attacker to take down the entire Erlang VM by filling up atom table.

jeking3:
Are we all saying the changes in `lib/erl/src/thrift_processor.erl` are good but the others are unnecessary and may force additional copies of things unnecessarily?  Sorry I don't know erlang and no committer has stepped up to review.  Also `ox-thrift` looks interesting, but does not support as many transports or protocols as what exists here today and I assume the interface changed.  If it's that much better though (5-10x faster?) we should have a discussion about bringing that in, if the author desires it.  As for this one, I'm still not sure what to do because nobody made it clear.

jeking3:
@djnym any thoughts about the comments regarding atom and DOS attacks?

jeking3:
If you want to pare down the PR to just the changes in `lib/erl/src/thrift_processor.erl` and it passes the CI build, I can merge that.  I'm not sure the clone changes are needed per comments.

PR #1588, opened 2018-08-29
Update Swift Library and Tests

apocolipse:
Some minor updates to the Swift library (mirrored here [apocolipse/thrift-swift](https://github.com/apocolipse/thrift-swift) for package management usage)
Updates are backwards/forwards compatible with Swift 3/4
Swift native tests added as well

### DIFF
--- a/lib/swift/README.md
+++ b/lib/swift/README.md
@@ -20,6 +20,8 @@ KIND, either express or implied. See the License for the
 specific language governing permissions and limitations
 under the License.
 
+Brought to you by [FiscalNote, Inc](http://www.fiscalnote.com/)
+
 
 ## Build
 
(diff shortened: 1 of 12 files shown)


jeking3:
If any of these changes are worthy of a release note please open a Jira ticket for it.

apocolipse:
@jeking3 No release note needed for this, its minor library updates to squash warnings in Swift 3.1+

Since Swift Foundation is more mature I may go back in and add/update some things here soon that will merit a release note.  I'm also about to push a change to the generator that is somewhat breaking (removing capitalizing of module names for consistency) that may require a note but otherwise shouldn't be a big issue.

As for CI, thats something I can look into but would need some requirements to what's desired.  Ideally, every separate OS from 14.04 - 18.10 for ubuntu would be tested for Swift 3.0.x, 3.1.x, and 4.0.x, macOS as well multiple xcode versions are neeeded to test all the different Swift Versions as well.  
If CI can be limited to "Just latest Swift version on latest Ubuntu/macOS" it's a much simpler task.  Let me know

PR #1589, opened 2018-08-31
THRIFT-4626: Remove checking of remaining bytes in the Go library.

vykulakov:
Client: lib/go

Obtaining the remaining bytes isn't supported with some combinations of protocols and transports in the Go library. For example, the binary protocol doesn't work properly with the zlib transport which wraps the framed transport. In libraries for other languages checking is used for directly reading data from a buffer of an underlying transport. If data isn't enough we just read data from the underlying transport and never throw an error as in the Go library. Also, the buffer for the zlib transport is encapsulated in the Go zlib library and we can't access it to get the remaining bytes. So removing that checking is the most simple and convenient method to solve the problem.

### DIFF
--- a/lib/go/thrift/binary_protocol.go
+++ b/lib/go/thrift/binary_protocol.go
@@ -448,9 +448,6 @@ func (p *TBinaryProtocol) ReadBinary() ([]byte, error) {
 	if size < 0 {
 		return nil, invalidDataLength
 	}
-	if uint64(size) > p.trans.RemainingBytes() {
-		return nil, invalidDataLength
-	}
 
 	isize := int(size)
 	buf := make([]byte, isize)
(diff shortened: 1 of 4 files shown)


dcelasun:
Thanks for the PR, please [open an issue](http://thrift.apache.org/docs/HowToContribute#if-you-discovered-a-defect) first.

dcelasun:
CC @Jens-G since he [introduced the check](https://github.com/apache/thrift/commit/ca8469ec578b13524e387782e7ee72d4150ab542).

vykulakov:
Added the issue: https://issues.apache.org/jira/browse/THRIFT-4626

vykulakov:
It looks like the build failures aren't related to my changes. Who would have to fix these failures?

jeking3:
I agree with you on that, and all the Travis CI builds passed, and the root cause and solution look correct.

PR #1590, opened 2018-09-04
Swift code generator minor updates

apocolipse:
Some minor updates to the Swift code generator.
- Source files no longer capitalized (i.e. `index.thrift` now generates `index.swift` and not `Index.swift`, when namespaced it generates `index/index.swift` instead of `index/Index.swift`).  There was an inconsistency with how module names were capitalized which could cause issues when namespacing and using Swift modules.   This change _may_ be breaking for some use cases, but is otherwise resolved by changing the case of the module name where ever it is an issue.  (for most use cases, it shouldn't cause any issues, just a different named file)

- For cases when handling Errors, module name is appropriately prefixed to Error type name if the Error type name happens to be `Error` in order to disambiguate from Swift.Error.  This is a non-breaking change and fixes compiler errors when namespacing _and_ using a type named `Error`

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_swift_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_swift_generator.cc
@@ -299,10 +299,10 @@ void t_swift_generator::init_generator() {
   string module = get_real_swift_module(program_);
   string out_dir = get_out_dir();
   string module_path = out_dir;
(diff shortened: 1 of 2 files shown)


jeking3:
Gentle reminder, anything worthy of release note level documentation needs a Jira ticket.

PR #1591, opened 2018-09-11
THRIFT-4620: Ensure enough space for for zlib flush marker

domscoyne:
Client: lib/cpp


jeking3:
I don't believe that the cross test framework will test zlib changes; not sure if the unit tests do a good job at that either.

PR #1597, opened 2018-09-20
THRIFT-4639: Use correct sequence number for multiplexed protocol

phlundblom:
Client: lib/nodejs

### DIFF
--- a/lib/nodejs/lib/thrift/multiplexed_protocol.js
+++ b/lib/nodejs/lib/thrift/multiplexed_protocol.js
@@ -45,7 +45,7 @@ function Wrapper(serviceName, protocol, connection) {
 };
 
 function Multiplexer() {
-  this.seqid = 0;
+  this._seqid = 0;
 };
 
 Multiplexer.prototype.createClient = function(serviceName, ServiceClient, connection) {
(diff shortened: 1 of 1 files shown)


phlundblom:
How does one run the NodeJS tests? Unfortunately I don't find it obvious from the Makefiles and/or the Travis outputs

bforbis:
To run the node tests, follow the instructions in the [docker README](https://github.com/apache/thrift/tree/master/build/docker) for creating your test image.

You can then either run the full test suite with `autotools.sh`, or if you want to just run the nodeJS tests you should be able to run `make check` within the node lib directory. Alternatively, if you understand how grunt files work you can just use that directly.

phlundblom:
I honestly don't have time enough to dig into the the test cases. To me it actually seems like the NodeJS Thrift server implementation in one way or another is broken.

I use NodeJS client against Java multiplexed server and that triggers completely different behaviors. Actually without this patch i get JS errors (`TypeError: this.seqid is not a function`) in NodeJS client rather than just the strange sequence number behaviors that appears when running unit tests against NodeJS server.

jeking3:
Author indicated insufficient time to complete the PR, and the changes don't appear to actually change anything other than a variable name.  In addition the tests did not pass.  This will be closed in favor of THRIFT-4405 verifying correctness.

jeking3:
Reopened, will shepherd this through.

jeking3:
I rebased on master to fix up CI issues, if they are not caused by the patch.

jeking3:
I was able to use local changes in THRIFT-4780 to prove this is not necessary; nodejs is sending sequence numbers just fine.

PR #1599, opened 2018-09-21
THRIFT-4678:add noexcept cpp generator option

cyyever:
Because throw() is deprecated in C++11,to avoid compiler warnings.


jeking3:
@cyyever this needs more work, please advise.

cyyever:
> 
> 
> @cyyever this needs more work, please advise.
Modifying the runtime library may cause a lot of file changes.I am not sure such many commits would be accepted.

cyyever:
> 
> 
> @cyyever this needs more work, please advise.

See THRIFT-4678.

jeking3:
Per THRIFT-4441, we're dropping C++03 support going forward, so you can replace these with `noexcept` if you prefer.

PR #1602, opened 2018-10-03
THRIFT-4645: TCurlClient: include failure reason in exception

james-johnston-thumbtack:
When curl_exec fails, more detailed failure information is available by calling curl_error.  Include this error information in the message in the thrown TTransportException.

Also change the comparison of the return value of curl_exec to explicitly check for boolean false, so as to distinguish from an empty response body (per the PHP documentation on this subject).

Changes have been tested and running in production at Thumbtack for the past several days, and it's working as expected.  TCurlClient doesn't seem to have any existing test suite to update.

### DIFF
--- a/lib/php/lib/Transport/TCurlClient.php
+++ b/lib/php/lib/Transport/TCurlClient.php
@@ -219,15 +219,19 @@ class TCurlClient extends TTransport
 
         curl_setopt(self::$curlHandle, CURLOPT_URL, $fullUrl);
         $this->response_ = curl_exec(self::$curlHandle);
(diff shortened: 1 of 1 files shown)


jeking3:
I'd say as long as this doesn't expose any stack frame information (anything that would be considered a security leak) it should be fine.  Can you post an example of the new output?

james-johnston-thumbtack:
No stack information what-so-ever.  Here's an example message:

```
TCurlClient: Could not connect to http://production-users.internal.aws/thrift, Operation timed out after 3001 milliseconds with 0 bytes received
```

PR #1609, opened 2018-10-07
THRIFT-4648: Use correct namespace in type names

simonsouth:
These changes fix the C (Glib) compiler so it uses the correct namespace when generating the name of a type or an included header file:

- The namespace of the target type or include is used, instead of the namespace of the IDL file being processed; and

- An unreliable method of generating a fully-qualified class name, which fails when a type name begins with a lowercase letter, is replaced by code that constructs the class name explicitly using its namespace prefix.

Also fixed is a minor issue whereby `t_c_glib_generator::init_generator()` would output multiple, extraneous comments in a generated file.

This resolves [THRIFT-4648](https://issues.apache.org/jira/browse/THRIFT-4648).


jeking3:
This looks like a breaking change, requiring folks who use namespaces to edit all their handlers and perhaps their build.  Is that true?  If it is, then a release note is necessary.  Certainly something in the `lib/c_glib/README.md` file.  See the `lib/perl/README.md` for a good example of documenting a breaking change.

@Jens-G @jfarrell In most release notes, breaking changes are presented first.  For example when we generate the standard dump from Jira, we should be listing all breaking changes first.  To do this, Jira needs an indicator that something is a breaking change.  In fact, I'd like to see us organize the list of release changes A) by component, B) by severity, C) by type (right now we only organize by type).

simonsouth:
@jeking3 Thanks for pointing this out. Yes, this is a breaking change. I've replaced the commit with one that updates the CHANGES and lib/c_glib/README.md files to indicate this. Please let me know what you think.

jeking3:
Build failure was unrelated.

PR #1610, opened 2018-10-15
THRIFT-4650: fixes incorrectly marking required fields as being set

mindscratch:
Fixes THRIFT-4650 where required fields were incorrectly being marked as having been set even if they weren't.

Client: go


dcelasun:
Thanks for the PR! Would you mind adding a test that covers this change? Take a look at [this commit](https://github.com/apache/thrift/commit/88591e32e710a0524327153c8b629d5b461e35e0) for an example.

mindscratch:
> Thanks for the PR! Would you mind adding a test that covers this change? Take a look at [this commit](https://github.com/apache/thrift/commit/88591e32e710a0524327153c8b629d5b461e35e0) for an example.

What's the best/easiest method to run those tests?

PR #1611, opened 2018-10-17
Version 0.12.0

jfarrell:


jeking3:
Please start using SemVer compliant names in the tags starting with this release, otherwise some third party sites like dub (for dlang) will not be able to build and list our packages.  In this case it should be `v0.12.0` and not `0.12.0`.  I'm not sure the branch name matters in this case so we can maintain the same style we have been using for releases.  So, if you want to add both a `0.12.0` and a `v0.12.0` tag for this release, that works for me.  In addition, if we run into issues integrating with third party sites that require minor build script-only changes, we will introduce additional SemVer compliant tags such as `v0.12.0+dub.0.1` (see [rule 10](https://semver.org/)).  Just a heads up...

Thrift.podspec
```diff
@@ -1,18 +1,18 @@
 Pod::Spec.new do |s|
   s.name          = "Thrift"
-  s.version       = "1.0.0"
+  s.version       = "0.12.0
```
jeking3: Missing trailing quote.

Thrift.podspec
```diff
@@ -1,18 +1,18 @@
...
   s.summary       = "Apache Thrift is a lightweight, language-independent software stack with an associated code generation mechanism for RPC."
   s.description   = <<-DESC
 The Apache Thrift software framework, for scalable cross-language services development, combines a software stack with a code generation engine to build services that work efficiently and seamlessly between C++, Java, Python, PHP, Ruby, Erlang, Perl, Haskell, C#, Cocoa, JavaScript, Node.js, Smalltalk, OCaml and Delphi and other languages.
                     DESC
   s.homepage      = "http://thrift.apache.org"
-  s.license       = { :type => 'Apache License, Version 2.0', :url => 'https://raw.github.com/apache/thrift/thrift-0.9.0/LICENSE' }
+  s.license       = { :type => 'Apache License, Version 2.0', :url => 'https://raw.github.com/apache/thrift/thrift-0.12.0/LICENSE' }
```
jeking3: The old link does not work.  Try `https://raw.githubusercontent.com/apache/thrift/0.12.0/LICENSE`

jeking3: Also remember to modify master with `https://raw.githubusercontent.com/apache/thrift/1.0.0/LICENSE`

composer.json
```diff
@@ -31,7 +31,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-master": "1.0.0-dev"
+            "dev-master": "1.0.x-dev"
```
jeking3: This does not look like the correct change to make, and there is no 0.12.0 designation to be found in the composer file.

contrib/zeromq/csharp/ThriftZMQ.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
-    <ProductVersion>9.0.21022</ProductVersion>
+    <ProductVersion>0.12.0</ProductVersion>
```
jeking3: This might be the Visual Studio version number...

debian/changelog
```diff
@@ -1,8 +1,8 @@
-thrift (1.0.0-dev) stable; urgency=low
-  * update version
-  * fix libthrift0.install
+thrift (0.12.0) stable; urgency=low
 
- -- Roger Meier <roger@apache.org>  Tue, 08 Jan 2013 22:40:12 +0100
+  * update to 0.12.0
+
+ -- Jake Farrell <jfarrell@apache.org>  Wed, 15 Oct 2012 12:00:00 -0500
```
jeking3: 2012

contrib/thrift.spec
```diff
@@ -28,7 +28,7 @@ Name:           thrift
 License:        Apache License v2.0
 Group:          Development
 Summary:        RPC and serialization framework
-Version:        0.11.0
+Version:        0.12.0
```
jeking3: In master this should be changed to 1.0.0-dev?

doap.rdf
```diff
@@ -58,6 +58,16 @@
     <category rdf:resource="http://projects.apache.org/category/network-client" />
     <category rdf:resource="http://projects.apache.org/category/network-server" />
     <release rdf:parseType="Collection">
+      <Version>
```
jeking3: This change should get back into master.

How are we going to separate out the changes that should get back into master from the changes that should not if they are all in the same commit?

contrib/zeromq/csharp/ThriftZMQ.csproj
```diff
@@ -25,7 +25,7 @@
     <UpdateRequired>false</UpdateRequired>
     <MapFileExtensions>true</MapFileExtensions>
     <ApplicationRevision>0</ApplicationRevision>
-    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
+    <ApplicationVersion>0.12.0</ApplicationVersion>
```
jeking3: I think you want to leave %2a, MAYBE... not sure, but it happened below.  That said, it might be a remnant and 0.12.0.0 would be more appropriate.

lib/csharp/test/Multiplex/Server/MultiplexServer.csproj
```diff
@@ -46,7 +46,7 @@
     <UpdateRequired>false</UpdateRequired>
     <MapFileExtensions>true</MapFileExtensions>
     <ApplicationRevision>0</ApplicationRevision>
-    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
+    <ApplicationVersion>0.12.0.%2a</ApplicationVersion>
```
jeking3: Should this be 0.12.0.0?

lib/d/src/thrift/base.d
```diff
@@ -50,7 +50,7 @@ class TCompoundOperationException : TException {
 /// The Thrift version string, used for informative purposes.
 // Note: This is currently hardcoded, but will likely be filled in by the build
 // system in future versions.
-enum VERSION = "1.0.0 dev";
```
jeking3: Should master's version be changed to `1.0.0-dev` ?

lib/haxe/README.md
```diff
@@ -71,8 +71,8 @@ either from the official ASF repo, or via the github mirror.
 
 - To set up any **stable version**, choose the appropriate branch (e.g. `0.10.0`):
 
-    - `haxelib git thrift https://git.apache.org/thrift.git 0.10.0 lib/haxe`
-    - `haxelib git thrift https://github.com/apache/thrift.git 0.10.0 lib/haxe`
+    - `haxelib git thrift https://git.apache.org/thrift.git 0.12.0 lib/haxe`
```
jeking3: This change should get back into master.

lib/lua/Thrift.lua
```diff
@@ -48,7 +48,7 @@ function ttable_size(t)
   return count
 end
 
-version = 1.0
```
jeking3: master should probably be changed to `1.0.0-dev` ?

lib/perl/lib/Thrift.pm
```diff
@@ -31,6 +31,6 @@ use warnings;
 #
 
 package Thrift;
-use version 0.77; our $VERSION = version->declare("v1.0_0");
+use version 0.77; our $VERSION = version->declare("v0.12.0_0");
```
jeking3: I will take care of actually publishing to CPAN once 0.12.0 is complete.

doap.rdf
jfarrell: Master needs to be templated so we are not hand editing 40 files for a release, this is tedious and error prone and not a good path forward. Next release needs to be automated, I'll bring this topic up on the dev@ list and we can come up with a good way to move it forward

jfarrell:
> Please start using SemVer compliant names in the tags starting with this release, otherwise some third party sites like dub (for dlang) will not be able to build and list our packages. In this case it should be `v0.12.0` and not `0.12.0`. I'm not sure the branch name matters in this case so we can maintain the same style we have been using for releases. So, if you want to add both a `0.12.0` and a `v0.12.0` tag for this release, that works for me. In addition, if we run into issues integrating with third party sites that require minor build script-only changes, we will introduce additional SemVer compliant tags such as `v0.12.0+dub.0.1` (see [rule 10](https://semver.org/)). Just a heads up...

We typically have tried to avoid that type of tag and would go with a v0.12.1 release which is why we have had .x branches previously. not sure why this changed along the way, i think composer or one of the other build/packaging tools needed specific versioned branches

jeking3:
We will need a `v0.12.0` tag in order to integrate with the dub (dlang) repository.  There's a good chance something will go wrong and a build metadata update will be needed.  By making an additional commit on the `0.12.0` branch and then tagging that commit `v0.12.0+dub.0.1` then it will satisfy the dub requirements for integration.  I see no need to go through a 0.12.1 release just to fix a little bit of build metadata for one language.

dcelasun:
Also note that Go's new module system [requires](https://github.com/golang/go/wiki/Modules#modules) the `v` prefix for semantic versioning.

jeking3:
AppVeyor has been fixed for the next build cycle.  Even as a committer I do not have permission to re-kick or modify the AppVeyor build for the project.

jeking3:
@dcelasun I read over the SemVer specification and there's nothing about a prefix `v` required or optional in that spec.  Does that mean the go and dlang dub package managers are actually non-compliant?

dcelasun:
@jeking3 yes. Go's internal [semver parser](https://godoc.org/golang.org/x/tools/internal/semver) deviates from the spec:

> This package follows Semantic Versioning 2.0.0 (see semver.org) with two exceptions. First, it requires the "v" prefix. Second, it recognizes vMAJOR and vMAJOR.MINOR (with no prerelease or build suffixes) as shorthands for vMAJOR.0.0 and vMAJOR.MINOR.0.

jeking3:
Okay, so the dub package manager's description just needs to change on their site.  I will update my issue on their site.

jeking3:
@jfarrell this needs to be moved along.  I provided comments 24 days ago and nothing has happened.

dcelasun:
It would be great if we could rebase this branch from master since there has been a lot of changes in the past month.

dcelasun:
@jfarrell can we get this released before the end of the year? *A lot* of changes happened on `master` since this branch was cut and it'd be great if we could release `v0.12.0` with everything included.

jeking3:
I'm going to start this effort once again, and reset the clock for changes, and I will update Jira with the items that are marked "fixed in 1.0" to be "fixed in 0.12.0".

dcelasun:
Thanks, let me know if I can help.

composer.json
jeking3: I found this is correct; it allows someone to require thrift version 1.0.* and it will pull the master branch from the project, instead of a tagged version.

https://getcomposer.org/doc/articles/aliases.md#branch-alias

lib/csharp/test/Multiplex/Server/MultiplexServer.csproj
jeking3: Leaving this alone.

contrib/zeromq/csharp/ThriftZMQ.csproj
jeking3: Fixed locally.  I reverted.

debian/changelog
jeking3: Fixed locally.

Thrift.podspec
jeking3: Fixed locally.  Using the standard apache license link.

jeking3:
When we tag this release we will be tagging both `0.12.0` and `v0.12.0`.  They will be identical.

jamadden:
@jfarrell The PyPI admins have received a request to expand the maintainer list for the [PyPI thrift project](https://pypi.org/project/thrift/) in order to upload 0.12.0 Could you please respond to https://github.com/pypa/warehouse/issues/5346 ?

PR #1261, opened 2017-05-05
THRIFT-4382: Replace the use of Indirect Object Syntax calls to new()

djzort:
Using Indirect Object Syntax to instantiate objects is not recommended as per https://perldoc.perl.org/perlobj.html#Invoking-Class-Methods

I have update the output to follow the recommended Class->new() syntax

PR #1613, opened 2018-10-17
THRIFT-4382: Replace the use of Indirect Object Syntax calls to new()

djzort:
Using Indirect Object Syntax to instantiate objects is not recommended as per https://perldoc.perl.org/perlobj.html#Invoking-Class-Methods I have update the output to follow the recommended Class->new() syntax

See also https://github.com/apache/thrift/pull/1261

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_perl_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_perl_generator.cc
@@ -354,7 +354,7 @@ string t_perl_generator::render_const_value(t_type* type, t_const_value* value) 
   } else if (type->is_enum()) {
     out << value->get_integer();
   } else if (type->is_struct() || type->is_xception()) {
-    out << "new " << perl_namespace(type->get_program()) << type->get_name() << "({" << endl;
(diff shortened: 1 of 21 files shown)


jeking3:
One final question - what perl versions will this syntax be compatible with?  We currently support 5.10.0 or later.  Does this PR change that?

djzort:
i havent introduced anything that would break perl 5.10

djzort:
From appveyor:

```
Downloading JDK from http://download.oracle.com/otn-pub/java/jdk/8u181-b13/96a7b8442fe848ef90c96a2fad6ed6d1/jdk-8u181-windows-x64.exe
ERROR: Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
The install of jdk8 was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\jdk8\tools\chocolateyInstall.ps1'.
 See log for details.
Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
```

Nothing to do with perl

PR #1618, opened 2018-10-30
THRIFT-4656: Fix infinite loop in PHP TCurlClient

sokac:
Infinite loop happens when response body is empty and HTTP status is
200. This patch fixes it.

* tested in production

### DIFF
--- a/lib/php/lib/Transport/TCurlClient.php
+++ b/lib/php/lib/Transport/TCurlClient.php
@@ -170,6 +170,24 @@ class TCurlClient extends TTransport
     }
 
     /**
+     * Guarantees that the full amount of data is read. Since TCurlClient gets entire payload at
+     * once, parent readAll cannot be used.
+     *
+     * @return string The data, of exact length
+     * @throws TTransportException if cannot read data
+     */
+    public function readAll($len)
+    {
+        $data = $this->read($len);
+
+        if (TStringFuncFactory::create()->strlen($data) !== $len) {
+            throw new TTransportException('TCurlClient could not read '.$len.' bytes');
+        }
+
+        return $data;
+    }
+
+    /**
      * Writes some data into the pending buffer
      *
      * @param string $buf The data to write


sokac:
@jeking3 it was the case before THRIFT-4645 got fixed. If response was empty, an exception was thrown.

Alternative is to override readAll. Parent class TTransport, https://github.com/apache/thrift/blob/master/lib/php/lib/Transport/TTransport.php#L76 , will keep calling read until receives all bytes. Since we know TCurlClient won't get any more bytes, an exception can be thrown in case read returns fewer than $len.

To give a little bit of background how this issue was reproduced - due to bug in upstream under some conditions, upstream could respond with 200 OK and empty response. This caused PHP to spin until supervisor terminated it due reached timeout.

Do you think this PR makes sense or should we override readAll? I'm open to suggestions too

jeking3:
I think we need to understand why there's a 0 length response sitting there.  My guess is that it's due to a oneway call followed by a normal call.

sokac:
@jeking3 200 OK with empty response was due an upstream bug not related to Thrift (our framework returned wrong http status code in case of exception, sits lower on stack than Thrift processor; already patched). From what I can tell, Thrift response should never be an empty response.

However, no library shouldn't be stuck in an infinite loop due invalid server response imo. This change:
https://github.com/apache/thrift/pull/1602/files#diff-74eb8d4b20c423c78252e3e3dc847099R227
introduced a change how empty response is handled by TCurlClient. Before it was merged, it would throw TTransportException on empty payload.

jeking3:
I'm still concerned about how oneway requests are handled with HTTP transport.  It would be a HTTP/1.1 200 OK response without any payload...

allengeorge:
Would it be better for the language runtime to return a `204` indicating a successful request but no content?

jeking3:
Sounds like an improvement, but here's the issue I think exists with oneway:

1. Generated client code sends a oneway request.
2. HTTP transport packages it up and send it to the server.  This could be through a proxy.
3. Server receives it.  Since it is a oneway request, my assumption is the server, no matter what the language, has a HTTP transport that issues a response (like 200 OK without content, or 204 as you stated).  If it did not, it wouldn't work through a proxy (or at least, the proxy might behave badly).
An example of this can be found at https://github.com/apache/thrift/blob/master/lib/py/src/server/THttpServer.py#L86.
4. Client has no generated response reader code because it is oneway, so the client code never attempts to get the transport response (which is fair, oneway probably came around before we had a transport with a mandatory response).  However now there's a transport reply sitting in the buffer.
5. The next request sent by the client that is normal (not oneway) would send the request and then read the HTTP 1.1 200 OK response buffered from from the oneway request, and the actual reply to the two-way request would get queued up; however the response processor for the twoway request, seeing no payload, would throw an exception.

sokac:
@jeking3 sorry for spamming. Have you had time to review? I'd be great if we can get this out in v0.12 as current `master` has issue with infinite loop and 0.11 doesn't have.

PR #1620, opened 2018-11-05
THRIFT-4658: TBinaryInputProtocol fails when strict is false

allengeorge:
Client: rs


allengeorge:
Note:

* Added unit tests
* Ran cross-tests in docker

allengeorge:
[puzzled] There appears to be a failure in a sub-build, but it doesn't have to do with the PR.

PR #1621, opened 2018-11-06
THRIFT-4660: Include @javax.annotation.Generated and Nullable as needed for enums

lazaroclapp:
Follow up to THRIFT-4530 and THRIFT-4614.

In that patch, we neglected to add Nullable and Generated annotations to enums, in addition to classes. This PR fixes that.

Client: Java compiler

This is a straightforward, compile-time-safe, backward-compatible change.


lazaroclapp:
cc: msridhar :)

PR #1624, opened 2018-11-07
THRIFT-4662: Rust const string calls function at compile time

jake-ruyi:
Client: rs

Fix with simple test

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -127,7 +127,7 @@ private:
   void render_const_value_holder(const string& name, t_type* ttype, t_const_value* tvalue);
 
   // Write the actual const value - the right side of a const definition.
-  void render_const_value(t_type* ttype, t_const_value* tvalue);
+  void render_const_value(t_type* ttype, t_const_value* tvalue, bool is_owned = true);
 
   // Write a const struct (returned from `const_value` method).
   void render_const_struct(t_type* ttype, t_const_value* tvalue);
@@ -411,6 +411,9 @@ private:
   // Return a string representing the rust type given a `t_type`.
   string to_rust_type(t_type* ttype, bool ordered_float = true);
 
+  // Return a string representing the `const` rust type given a `t_type`
+  string to_rust_const_type(t_type* ttype, bool ordered_float = true);
+
   // Return a string representing the rift `protocol::TType` given a `t_type`.
   string to_rust_field_type_enum(t_type* ttype);
 
@@ -645,8 +648,8 @@ void t_rs_generator::render_const_value(const string& name, t_type* ttype, t_con
     throw "cannot generate simple rust constant for " + ttype->get_name();
   }
 
-  f_gen_ << "pub const " << rust_upper_case(name) << ": " << to_rust_type(ttype) << " = ";
-  render_const_value(ttype, tvalue);
+  f_gen_ << "pub const " << rust_upper_case(name) << ": " << to_rust_const_type(ttype) << " = ";
+  render_const_value(ttype, tvalue, false);
   f_gen_ << ";" << endl;
   f_gen_ << endl;
 }
@@ -673,15 +676,22 @@ void t_rs_generator::render_const_value_holder(const string& name, t_type* ttype
   f_gen_ << endl;
 }
 
-void t_rs_generator::render_const_value(t_type* ttype, t_const_value* tvalue) {
+void t_rs_generator::render_const_value(t_type* ttype, t_const_value* tvalue, bool is_owned) {
   if (ttype->is_base_type()) {
     t_base_type* tbase_type = (t_base_type*)ttype;
     switch (tbase_type->get_base()) {
     case t_base_type::TYPE_STRING:
       if (tbase_type->is_binary()) {
-        f_gen_ << "\"" << tvalue->get_string() << "\""<<  ".to_owned().into_bytes()";
+        if (is_owned) {
+          f_gen_ << "\"" << tvalue->get_string() << "\""<<  ".to_owned().into_bytes()";
+        } else {
+          f_gen_ << "b\"" << tvalue->get_string() << "\"";
+        }
       } else {
-        f_gen_ << "\"" << tvalue->get_string() << "\""<<  ".to_owned()";
+        f_gen_ << "\"" << tvalue->get_string() << "\"";
+        if (is_owned) {
+          f_gen_ << ".to_owned()";
+        }
       }
       break;
     case t_base_type::TYPE_BOOL:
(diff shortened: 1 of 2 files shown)


compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -673,15 +676,22 @@ void t_rs_generator::render_const_value_holder(const string& name, t_type* ttype
...
 
-void t_rs_generator::render_const_value(t_type* ttype, t_const_value* tvalue) {
+void t_rs_generator::render_const_value(t_type* ttype, t_const_value* tvalue, bool isowned) {
   if (ttype->is_base_type()) {
     t_base_type* tbase_type = (t_base_type*)ttype;
     switch (tbase_type->get_base()) {
     case t_base_type::TYPE_STRING:
       if (tbase_type->is_binary()) {
-        f_gen_ << "\"" << tvalue->get_string() << "\""<<  ".to_owned().into_bytes()";
+        if (isowned) {
```
allengeorge: Will there ever be a point where we want the const string/byte-array to be owned?

allengeorge:
@jake-ruyi Thank you for the bug report - and the PR! I've asked a question about the implementation.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
jake-ruyi: Honestly, I'm not really sure about the byte-arrays.  We aren't using them anywhere, so I'm not clear on the use-cases.
I assume it will end up like the regular strings where you might want `map<u8,binary>` or something like that.

jake-ruyi: My goals were:
1. Make `const binary` compile
2. Preserve existing behavior

Presently, neither `const string` nor `const binary` compiles. I'm more interested in getting our stuff compiling than addressing the semantics of `binary`. Should I revert the binary portion?

allengeorge: No - the PR isn't being held up because of const binary. I just didn't have time to look at it yet. I usually pull the branch locally and test in a container to ensure everything works. For one thing, I'm unsure that

```
const FOO: &str = "ladeda";
```

works in rust 1.28.

EDIT. Oh, nm - I see it was [enabled in 1.17](https://rust-lang-nursery.github.io/edition-guide/rust-2018/ownership-and-lifetimes/simpler-lifetimes-in-static-and-const.html)

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -127,7 +127,7 @@ class t_rs_generator : public t_generator {
   void render_const_value_holder(const string& name, t_type* ttype, t_const_value* tvalue);
 
   // Write the actual const value - the right side of a const definition.
-  void render_const_value(t_type* ttype, t_const_value* tvalue);
+  void render_const_value(t_type* ttype, t_const_value* tvalue, bool isowned = true);
```
allengeorge: Could you change parameter name to `owned` or `is_owned`?

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -414,6 +414,9 @@ class t_rs_generator : public t_generator {
   // Return a string representing the rift `protocol::TType` given a `t_type`.
   string to_rust_field_type_enum(t_type* ttype);
 
+  // Return a string representing the `const` rust type given a `t_type`
+  string to_rust_const_type(t_type* ttype, bool ordered_float = true);
```
allengeorge: Good call. Makes sense to split this out from the `to_rust_type`. Could you please put this function declaration right before `to_rust_type`? And do the same for implementation please? It just ensures that code that has similar functionality is in the same part of the file.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -645,8 +648,8 @@ void t_rs_generator::render_const_value(const string& name, t_type* ttype, t_con
     throw "cannot generate simple rust constant for " + ttype->get_name();
   }
 
-  f_gen_ << "pub const " << rust_upper_case(name) << ": " << to_rust_type(ttype) << " = ";
-  render_const_value(ttype, tvalue);
+  f_gen_ << "pub const " << rust_upper_case(name) << ": " << to_rust_const_type(ttype) << " = ";
+  render_const_value(ttype, tvalue, false);
```
allengeorge: Interesting. This works because the call isn't recursive. Only the top-level type will be owned/static. The types in containers will all be owned because of the internal call to `to_rust_type`.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -673,15 +676,22 @@ void t_rs_generator::render_const_value_holder(const string& name, t_type* ttype
   f_gen_ << endl;
 }
 
-void t_rs_generator::render_const_value(t_type* ttype, t_const_value* tvalue) {
+void t_rs_generator::render_const_value(t_type* ttype, t_const_value* tvalue, bool isowned) {
```
allengeorge: Please rename parameter to `owned` or `is_owned`.

lib/rs/test/thrifts/Base_One.thrift
```diff
@@ -37,6 +37,9 @@ const list<double> CommonTemperatures = [300.0, 450.0]
 
 const double MealsPerDay = 2.5;
 
+const string DefaultRecipeName = "Something"
```
allengeorge: This is incredibly minor, but, could you please choose some funny recipe names here? :)

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -3076,6 +3086,21 @@ string t_rs_generator::to_rust_field_type_enum(t_type* ttype) {
   throw "cannot find TType for " + ttype->get_name();
 }
 
+string t_rs_generator::to_rust_const_type(t_type* ttype, bool ordered_float) {
```
allengeorge: Please reorder implementation just before `to_rust_type`

allengeorge:
@jake-ruyi Thank you for the PR. I've some minor comments, but overall this looks good. If you can address them, I'll approve and this can be merged in.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
jake-ruyi: That explains it.  I could have sworn it needed to be `&'static`, but I found a bunch of examples that were `&str` and assumed I was thinking of something else...
Do you think it's worth outputting `&'static str` instead so it works with Rust <1.17?

allengeorge:
No - that’s ok. I’ve set the minimum Rust version to 1.28 for a while now, so I think we should be fine.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
jeikabu: Right

allengeorge:
@jeking3 These changes look good. Could they be merged please? All the cross-tests pass. Two build failures: SBCL (same issue as before) and D.

PR #1625, opened 2018-11-10
THRIFT-4529: Rust enum variants are now camel-cased

allengeorge:
Client: rs

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,8 @@
 Apache Thrift Changelog
 
 Breaking Changes since 0.11.0 [for 0.12.0]:
--------------------------------------------------------------------------------- 
+--------------------------------------------------------------------------------
+    * [THRIFT-4529] - Rust enum variants are now camel-cased instead of uppercased to conform to Rust naming conventions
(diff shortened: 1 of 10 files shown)


allengeorge:
Build failure caused by a PEBCAK on my part (forgot to change enum variant names in cross-tests). Fixing now.

jeking3:
I assume the default setting is to use the previous way, and this is a new way, so it is backwards compatible?

allengeorge:
I did not include a flag for backwards-compatibility. I'd prefer not to, because this change makes the autogenerated code conform to Rust naming conventions.

allengeorge:
Yes - you're right: it is a breaking change. I'll make the README entries, and make sure I do so for any (hopefully rare!) breaking changes I make in the future. Thank you for catching this.

allengeorge:
I've updated the `CHANGES` file, plus added a `COMPATIBILITY` section to the Rust `README` along with transition guidelines. I will do this in the future for any breaking changes.

allengeorge:
@jeking3 I've made the requested changes, and, the only thing that fails in Travis is the SBCL build under autotools :/ Could this be merged please?

PR #1626, opened 2018-11-11
THRIFT-4545: fix haskell build on windows, fix appveyor stale packages

jeking3:


lib/hs/thrift.cabal
```diff
@@ -18,8 +18,8 @@
 --
 
 Name:           thrift
-Version:        1.0.0-dev
-Cabal-Version:  >= 1.24
+Version:        0.99.0
```
jeking3: @jfarrell @Jens-G note: cabal will no longer allow "1.0.0-dev" as a version... this happens with newer versions of cabal.

PR #1629, opened 2018-11-17
THRIFT-4666: Attempt to work around dlang client pool test failure (affects CI)

allengeorge:


allengeorge:
If someone who's dlang-knowledgeable could take a look at this I'd really appreciate it. I think this fixes the sporadic dlang build failures.

PR #1631, opened 2018-11-21
THRIFT-4668: make socket backlog configurable for python

lshgdut:
The backlog parameter of socket.listen was hard coded, so I can't customize it in my code.

Adding `setBacklog`  to `TServerSocket` can resolve my problem.  :)

### DIFF
--- a/lib/py/src/transport/TSocket.py
+++ b/lib/py/src/transport/TSocket.py
@@ -159,6 +159,15 @@ class TServerSocket(TSocketBase, TServerTransportBase):
         self._unix_socket = unix_socket
         self._socket_family = socket_family
         self.handle = None
+        self._backlog = 128
+
+    def setBacklog(self, backlog=None):
+        if not self.handle:
+            self._backlog = backlog
+        else:
+            # We cann't update backlog when it is already listening, since the
+            # handle has been created.
+            logger.warn('You have to set backlog before listen.')
 
     def listen(self):
         res0 = self._resolveAddr()
@@ -183,7 +192,7 @@ class TServerSocket(TSocketBase, TServerTransportBase):
         if hasattr(self.handle, 'settimeout'):
             self.handle.settimeout(None)
         self.handle.bind(res[4])
-        self.handle.listen(128)
+        self.handle.listen(self._backlog)
 
     def accept(self):
         client, addr = self.handle.accept()


lib/py/src/transport/TSocket.py
```diff
@@ -159,6 +159,10 @@ def __init__(self, host=None, port=9090, unix_socket=None, socket_family=socket.
         self._unix_socket = unix_socket
         self._socket_family = socket_family
         self.handle = None
+        self._backlog = 128
```
jeking3: Should backlog be a named parameter with a default of None?

lib/py/src/transport/TSocket.py
```diff
@@ -183,7 +187,7 @@ def listen(self):
         if hasattr(self.handle, 'settimeout'):
             self.handle.settimeout(None)
         self.handle.bind(res[4])
-        self.handle.listen(128)
+        self.handle.listen(self._backlog)
```
jeking3: To preserve the existing behavior, how about `self.handle.listen(self._backlog or 128)`, in combination with a default value of None in the class constructor, this eliminates the need for setBacklog().

lib/py/src/transport/TSocket.py
```diff
@@ -159,6 +159,10 @@ def __init__(self, host=None, port=9090, unix_socket=None, socket_family=socket.
         self._unix_socket = unix_socket
         self._socket_family = socket_family
         self.handle = None
+        self._backlog = 128
+
+    def setBacklog(self, backlog):
```
jeking3: What happens when you call this when it is already listening?

lib/py/src/transport/TSocket.py
lshgdut: We shoud not use `self.handle.listen(self._backlog or 128)` to preserve the existing behavior, because the backlog parameter accept a zero value.
Plz have a look at the python source code below:
```c
// From Python-2.7.14/Modules/socketmodule.c
/* s.listen(n) method */

static PyObject *
sock_listen(PySocketSockObject *s, PyObject *arg)
{
    // some code before
    /* To avoid problems on systems that don't allow a negative backlog
     * (which doesn't make sense anyway) we force a minimum value of 0. */
    if (backlog < 0)
        backlog = 0;
    res = listen(s->sock_fd, backlog);
    // some code more
}
```

I have think about `self._backlog if self._backlog is None else 128 `, and the code seems to be redundant with bad smell. So declaring the _backlog attribute with a default value of 128 in the class constructor seems to be a better way, isn't? On the other hand, `__init__` will be called before listen, while the listen method called, `this.handle.listen` will receive a `_backlog` with a default value of 128,  existing behavior preserved!

lib/py/src/transport/TSocket.py
lshgdut: Yeah, nothing works except `_backlog` changed. I should add some warning for it. Thank for your suggestions!

lshgdut:
The value of `socket.SOMAXCONN`  depends on different operate systems:
- on Window 7, it is `2147483647`
- on Centos or Debian, it is `128`, even I have set a different value to `/proc/sys/net/core/somaxconn`

According to `Python-2.7.14/Modules/socketmodule.c`:
```c
 4967      /* Maximum number of connections for "listen" */
 4968: #ifdef  SOMAXCONN
 4969:     PyModule_AddIntConstant(m, "SOMAXCONN", SOMAXCONN);
 4970  #else
 4971:     PyModule_AddIntConstant(m, "SOMAXCONN", 5); /* Common value */
 4972  #endif
```

The value of `SOMAXCONN` depend on system macro `SOMAXCONN`:
- on Window 7, I can not explain why it is 2147483647
- on linux, `SOMAXCONN` is defined in '/usr/include/bits/socket.h +144', hard code to 128

Linux set  `SOMAXCONN` macro to` core.sysctl_somaxconn` first, then check the sysctl value of 'net.core.somaxconn' existing or not and override it if possible.

When socket listening, if `backlog` is greater than SOMAXCONN, the kernel will take the value of `min(backlog, SOMAXCONN)`. Here is linux kernel code below:
```c
// From linux-3.10.0-693.25.4.el7/net/socket.c +1534
/*
 *      Perform a listen. Basically, we allow the protocol to do anything
 *      necessary for a listen, and if that works, we mark the socket as
 *      ready for listening.
 */

SYSCALL_DEFINE2(listen, int, fd, int, backlog)
{
        struct socket *sock;
        int err, fput_needed;
        int somaxconn;

        sock = sockfd_lookup_light(fd, &err, &fput_needed);
        if (sock) {
                somaxconn = sock_net(sock->sk)->core.sysctl_somaxconn;
                if ((unsigned int)backlog > somaxconn)
                        backlog = somaxconn;

                err = security_socket_listen(sock, backlog);
                if (!err)
                        err = sock->ops->listen(sock, backlog);

                fput_light(sock->file, fput_needed);
        }
        return err;
}
```

So doesn't it need to check the value of `_backlog` or not? May be a warning better?

jeking3:
Build failure is the SBCL issue we know about.

PR #1641, opened 2018-12-06
Fix build with c++17

proller:
random_shuffle is deprecated and removed in c++17
https://en.cppreference.com/w/cpp/algorithm/random_shuffle

### DIFF
--- a/lib/cpp/src/thrift/transport/TSocketPool.cpp
+++ b/lib/cpp/src/thrift/transport/TSocketPool.cpp
@@ -21,6 +21,7 @@
 
 #include <algorithm>
 #include <iostream>
+#include <random>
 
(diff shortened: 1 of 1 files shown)


lib/cpp/src/thrift/transport/TSocketPool.cpp
```diff
@@ -188,7 +189,13 @@ void TSocketPool::open() {
   }
 
   if (randomize_ && numServers > 1) {
-    random_shuffle(servers_.begin(), servers_.end());
+#if __cplusplus >= 201500L // c++17
```
jeking3: 201703L is the correct constant for C++17; this one may include some pre-C++17 standard compilers (6.1 for example) and we don't test with a wide variety of them.

PR #1642, opened 2018-12-07
THRIFT-4679: Remove unused variable declaration

griffinmichl:
Changes are backwards compatible (as this line never did anything)

https://issues.apache.org/jira/browse/THRIFT-4679

### DIFF
--- a/lib/nodejs/lib/thrift/json_protocol.js
+++ b/lib/nodejs/lib/thrift/json_protocol.js
@@ -18,7 +18,6 @@
  */
 
 var Int64 = require('node-int64');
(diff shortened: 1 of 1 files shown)


griffinmichl:
Any idea how I can get this merged into the thrift npm package?

jeking3:
When we complete the 0.12.0 release then we start pushing out to package managers.  It's kind of ad-hoc for the project however.  It looks like @jfarrell maintains the npmjs updates.

griffinmichl:
Sounds good. I think my company also has some fixes to the browser serializer they want to get in (we use our own right now). I'll try to get that moving along so we can get it into that release as well.

PR #1643, opened 2018-12-10
Fix build compiler with cmake

louxiu:
This seems an obvious problem, don't know why it works with others. Maybe I do something wrong. 

I follow with the [README](https://github.com/apache/thrift/tree/master/compiler/cpp#build-using-cmake) to build the compiler using cmake, but failed with "Undefined symbols". It is because `compiler_core` sources is not append to `thrift-compiler_SOURCES`.

 * cmake version: 3.12.3
 * OS: mac os Mojave version 10.14

### DIFF
--- a/compiler/cpp/CMakeLists.txt
+++ b/compiler/cpp/CMakeLists.txt
@@ -113,7 +113,7 @@ THRIFT_ADD_COMPILER(xml     "Enable compiler for XML" ON)
 # we also add the current binary directory for generated files
 include_directories(${CMAKE_CURRENT_BINARY_DIR} src)
 
-if(NOT ${WITH_PLUGIN})
+if(NOT DEFINED WITH_PLUGIN OR NOT ${WITH_PLUGIN})
     list(APPEND thrift-compiler_SOURCES ${compiler_core})
 endif()
 


compiler/cpp/CMakeLists.txt
```diff
@@ -113,7 +113,7 @@ THRIFT_ADD_COMPILER(xml     "Enable compiler for XML" ON)
 # we also add the current binary directory for generated files
 include_directories(${CMAKE_CURRENT_BINARY_DIR} src)
 
-if(NOT ${WITH_PLUGIN})
+if(NOT DEFINED WITH_PLUGIN OR NOT ${WITH_PLUGIN})
```
jeking3: This should not be necessary, as https://github.com/apache/thrift/blob/master/build/cmake/DefineOptions.cmake#L97 ensures it is always defined.  Perhaps you just need to specify `-DWITH_PLUGIN=OFF` ?

louxiu: Oh, I know, because I build the compiler under `compiler/cpp/` and  `DefineOptions.cmake` is at `build/cmake`, so `WITH_PLUGIN` is not defined?

jeking3: I see, I always generate the make environment from the top level, then if I only need the compiler I will cd into the build directory's compiler/cpp and make from there.

PR #1644, opened 2018-12-10
Move to version supporting node 4.x

tinder-speriyasamy:
THRIFT-9999: [Move to version supporting node 4.x]
        Client: [lib/nodejs]


jeking3:
Pretty sure we already support node.js 4.x.

PR #1651, opened 2018-12-27
THRIFT-4546: replace all git-wip-us.apache.org

RobberPhex:


dcelasun:
Why not use Github instead of Gitbox? We already do that for the Go library.

RobberPhex:
https://github.com/apache/thrift is just a mirror of Gitbox.
Technically, the work flow doesn't relay on GitHub. That's why we used git-wip-us, and why we use gitbox.

dcelasun:
It's the other way around, pushes to Github are mirrored to Gitbox. I've personally never pushed to Gitbox.

PR #1652, opened 2018-12-28
Update Cargo.toml, release instructions and release script

allengeorge:
Client: rs

### DIFF
--- a/lib/rs/RELEASING.md
+++ b/lib/rs/RELEASING.md
@@ -1,0 +1,57 @@
+# Publishing the thrift crate
+
+Publishing the Rust thrift crate is straightforward, and involves two major steps:
+
+1. Setting up your [crates.io](https://www.crates.io) account _(one-time)_
+
+2. Packaging/publishing the Rust thrift crate itself
+
+## Set up your crates.io account (one-time)
+
+1. Go to [crates.io](https://www.crates.io) and click the `Log In` button at the top right.
+
+   Log in **as the Github user with write permissions to the thrift repo!**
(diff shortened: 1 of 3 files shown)


lib/rs/RELEASING.md
```diff
@@ -0,0 +1,57 @@
...
+
+#### Automated
+
+Run `./release.sh [THRIFT_RELEASE_VERSION]`.
+
+_Requires you to have stored your credentials in `~/.cargo/credentials`._
+
+#### Manual
+
+1. Edit `Cargo.toml` and update the `version = 1.0` key to `version = [THRIFT_RELEASE_VERSION]`
```
jeking3: As we move forward, master will always have the "next" anticipated release checked in, so that when we cut a release branch it can be merged back into master on completion of the release.  The way we do it today (documented here) makes it nearly impossible to do a merge back into master.

I am planning on having a shell script that tweaks all of the files to update the release number (build/veralign.sh).

So these instructions are valid for now, just wanted to make you aware.

allengeorge: Makes sense - I like that plan. I can update that line. FWIW, the script in this PR (`release.sh`) simply replaces `version = .*$` in `Cargo.toml` with the latest release version, so it doesn't assume that the version is always set to 1.0.

The only thing that will be a pain with releasing is that you have to commit the changes to `Cargo.toml` before packaging and uploading to [crates.io](https://www.crates.io).

I'd be happy to help with trying to automate the releases for various languages.

PR #1658, opened 2018-12-31
THRIFT-4702: Improve AutoExpandingBuffer

BELUGABEHR:


BELUGABEHR:
This change is not backward compatible as the scope of the AutoExpandingBuffer class has been minimized from public to package-level.

PR #1664, opened 2019-01-01
THRIFT-4712: Improve Performance of ShortStack

BELUGABEHR:


jeking3:
Appreciate your changes.  If you could also look at the backlog in Apache Jira and see if there are any existing issues you can tackle, that would also be appreciated.

BELUGABEHR:
Sure thing.  Thanks.

lib/java/src/org/apache/thrift/ShortStack.java
```diff
@@ -26,54 +28,51 @@
...
   }
 
   public short peek() {
-    return vector[top];
+    return vector[top - 1];
   }
 
   public void clear() {
-    top = -1;
+    top = 0;
```
jeking3: Does this actually clear?  I don't think it does.  It resets top to 0 and the next peek() will read whatever was there before.  Shouldn't clear actually clear?  If one peek()s after clear and there's nothing there, what should it do?  Perhaps throw?

BELUGABEHR: Good call.  Well, the peek() method is not used anywhere other than the test framework, so it's a bit of a moot point; it should probably be removed altogether. 

I think this is another example where the class should not be made public since it serves a very specific purpose.  If it were scoped back into the package, it would have to be moved into the same package as the TCompactProtocol since that is what relies on this class.  Thoughts?

jeking3: ShortStack was written in 2010.  Perhaps it would be a good idea to analyze the performance of ShortStack managing 15 shorts vs. java.util.Stack in a more recent version of Java like 1.8.  My guess is by now they've ironed out performance issues in containers like this... ideally we would try to remove ShortStack.java and its test all together.

BELUGABEHR: JDK is still slow because of the auto-boxing of the primitive short value.  In most cases, it wouldn't make a difference, but for a serialization library that needs to be as fast as possible, it's measurable.

BELUGABEHR:
The clear() method does the job.  It sets the pointed back to the first open slot so that it can be overwritten.  This is much faster than clearing the entire buffer or allocating a new one, and for these purposes, works well, now that I also removed the peek() method.

lib/java/src/org/apache/thrift/ShortStack.java
jeking3: Do you have evidence of this or is it just theory?

BELUGABEHR: Ya, I benchmarked it.  I did two different scenarios.  I didn't even bother including the Java JDK Stack implementation in the more aggressive test because it wasn't even close in the more trivial test.

https://issues.apache.org/jira/browse/THRIFT-4712?focusedCommentId=16731695&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16731695

PR #1667, opened 2019-01-02
THRIFT-4301: fix inability to disable py3 with autoconf build

jeking3:


jeking3:
Interesting failure, not finding py3 -- I'm looking into it.

PR #1669, opened 2019-01-03
THRIFT-4368: Fix guaranteed NPE in TBaseAsyncProcessor.java

belugabehr:


lib/java/src/org/apache/thrift/TBaseAsyncProcessor.java
```diff
@@ -43,26 +43,19 @@ public TBaseAsyncProcessor(I iface, Map<String, AsyncProcessFunction<I, ? extend
...
 
         final TProtocol in = fb.getInputProtocol();
         final TProtocol out = fb.getOutputProtocol();
 
-        //Find processing function
+        // Find processing function
         final TMessage msg = in.readMessageBegin();
         AsyncProcessFunction fn = processMap.get(msg.name);
         if (fn == null) {
-            TProtocolUtil.skip(in, TType.STRUCT);
```
jeking3: This should not be removed; if the method cannot be located we need to eat the inbound struct.

lib/java/src/org/apache/thrift/TBaseAsyncProcessor.java
```diff
@@ -43,26 +43,19 @@ public TBaseAsyncProcessor(I iface, Map<String, AsyncProcessFunction<I, ? extend
...
-            in.readMessageEnd();
```
jeking3: Also should not be removed.

lib/java/src/org/apache/thrift/TBaseAsyncProcessor.java
```diff
@@ -43,26 +43,19 @@ public TBaseAsyncProcessor(I iface, Map<String, AsyncProcessFunction<I, ? extend
...
-            if (!fn.isOneway()) {
```
jeking3: Try this instead (or something close):
```
            if (msg.type() == TMessageType.CALL) { // this means it is a two-way request, so we can send a reply
```

keep the original logic.

lib/java/src/org/apache/thrift/TBaseAsyncProcessor.java
```diff
@@ -52,17 +52,19 @@ public boolean process(final AsyncFrameBuffer fb) throws TException {
...
+            // this means it is a two-way request, so we can send a reply
+            if (msg.type == TMessageType.CALL) {
               out.writeMessageBegin(new TMessage(msg.name, TMessageType.EXCEPTION, msg.seqid));
               x.write(out);
               out.writeMessageEnd();
               out.getTransport().flush();
             }
             fb.responseReady();
-            return true;
+            throw x;
```
jeking3: I don't think this is correct.  The exception is intended for the client that made the request, and that was handled.  We could certainly add an else case here and put something into the log as a debug level entry to handle oneway requests that fail... but throwing here will probably stop the server, and it should not.

BELUGABEHR: Yup.  I'm with you now.

I was looking:
https://github.com/apache/thrift/blob/75bac10dbd8d953bb2f556b7dcc3bdf7518e67a6/lib/java/src/org/apache/thrift/server/AbstractNonblockingServer.java#L606-L607

I was thinking that it would be caught here and simply logged, but I overlooked the 'return' statement.  Indeed, a Throwable will cause an exit.

PR #1671, opened 2019-01-03
THRIFT-4714: optimize java TFramedTransports to call write only once per flush

jeking3:
Code calls two writes to the underlying transport per flush: one for the frame size and one for the payload.  The result is that folks use TBufferedTransport on top of TFramedTransport to avoid getting two ethernet frames for small packets.  This means everything is double-buffered!

Based on THRIFT-1121 and THRIFT-959 this may have been an issue for a long time - the fix was easy however.  We make room for the frame size before writing to the buffer, then we only have to pull out the buffer, drop the frame size into the 4 bytes that we pre-allocated, and then write to the underlying transport once.

### DIFF
--- a/lib/java/src/org/apache/thrift/transport/AutoExpandingBufferWriteTransport.java
+++ b/lib/java/src/org/apache/thrift/transport/AutoExpandingBufferWriteTransport.java
@@ -25,10 +25,29 @@ public final class AutoExpandingBufferWriteTransport extends TTransport {
 
   private final AutoExpandingBuffer buf;
   private int pos;
+  private int res;
 
-  public AutoExpandingBufferWriteTransport(int initialCapacity) {
+  /**
+   * Constructor.
+   * @param initialCapacity the initial capacity of the buffer
+   * @param frontReserve space, if any, to reserve at the beginning such
+   *                     that the first write is after this reserve.
+   *                     This allows framed transport to reserve space
+   *                     for the frame buffer length.
+   * @throws IllegalArgumentException if initialCapacity is less than one
+   * @throws IllegalArgumentException if frontReserve is less than zero
+   * @throws IllegalArgumentException if frontReserve is greater than initialCapacity
+   */
+  public AutoExpandingBufferWriteTransport(int initialCapacity, int frontReserve) {
+    if (initialCapacity < 1) {
+      throw new IllegalArgumentException("initialCapacity");
+    }
+    if (frontReserve < 0 || initialCapacity < frontReserve) {
+      throw new IllegalArgumentException("frontReserve");
+    }
     this.buf = new AutoExpandingBuffer(initialCapacity);
-    this.pos = 0;
+    this.pos = frontReserve;
+    this.res = frontReserve;
   }
 
   @Override
@@ -56,11 +75,14 @@ public final class AutoExpandingBufferWriteTransport extends TTransport {
     return buf;
   }
 
-  public int getPos() {
+  /**
+   * @return length of the buffer, including any front reserve
+   */
+  public int getLength() {
     return pos;
   }
 
   public void reset() {
-    pos = 0;
+    pos = res;
   }
 }
--- a/lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
+++ b/lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
@@ -106,8 +106,8 @@ public class TFastFramedTransport extends TTransport {
     this.underlying = underlying;
     this.maxLength = maxLength;
     this.initialBufferCapacity = initialBufferCapacity;
-    writeBuffer = new AutoExpandingBufferWriteTransport(initialBufferCapacity);
     readBuffer = new AutoExpandingBufferReadTransport(initialBufferCapacity);
+    writeBuffer = new AutoExpandingBufferWriteTransport(initialBufferCapacity, 4);
   }
 
   @Override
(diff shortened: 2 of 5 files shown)


jeking3:
Going to leave this here for a day or two for visibility and review.

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
```diff
@@ -158,6 +159,10 @@ private void readFrame() throws TTransportException {
 
   @Override
   public void write(byte[] buf, int off, int len) throws TTransportException {
+    if (writeBuffer.getPos() == 0) {
+	// did not put in ctor because it caused interface changes
+        writeBuffer.write(sizeFiller, 0, 4);
```
allengeorge: (nit) Indentation is incorrect.

allengeorge: This is unfortunate, because it causes an extra conditional check on every `write()` call.

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
```diff
@@ -106,8 +107,8 @@ public TFastFramedTransport(TTransport underlying, int initialBufferCapacity, in
     this.underlying = underlying;
     this.maxLength = maxLength;
     this.initialBufferCapacity = initialBufferCapacity;
-    writeBuffer = new AutoExpandingBufferWriteTransport(initialBufferCapacity);
     readBuffer = new AutoExpandingBufferReadTransport(initialBufferCapacity);
+    writeBuffer = new AutoExpandingBufferWriteTransport(initialBufferCapacity);
```
allengeorge: I suppose it's impossible (or not recommended) to `write(sizeFiller, 0, 4)` here?

Rationale: we do the special case once, and avoid doing a conditional on every `write()` call. Since this call doesn't hit an IO resource it *shouldn't* throw exceptions either.

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
```diff
@@ -172,11 +177,12 @@ public void clear() {
...
-    int length = writeBuffer.getPos();
-    TFramedTransport.encodeFrameSize(length, i32buf);
-    underlying.write(i32buf, 0, 4);
-    underlying.write(writeBuffer.getBuf().array(), 0, length);
+    int length = writeBuffer.getPos() - 4;          // account for frame size reserved space
+    byte[] data = writeBuffer.getBuf().array();
+    TFramedTransport.encodeFrameSize(length, data); // write frame size directly into the buffer
+    underlying.write(data, 0, length + 4);          // write the frame size and payload in one call
     writeBuffer.reset();
+    writeBuffer.write(sizeFiller, 0, 4);            // reserve space for the next frame size
```
allengeorge: I understand the optimization, but worry that now there are two places adding `sizeFiller` to the buffer.

lib/java/src/org/apache/thrift/transport/TFramedTransport.java
```diff
@@ -65,17 +65,26 @@ public TTransport getTransport(TTransport base) {
...
+   */
+  private static final byte[] sizeFiller_ = new byte[] { 0x00, 0x00, 0x00, 0x00 };
+
   /**
    * Constructor wraps around another transport
    */
   public TFramedTransport(TTransport transport, int maxLength) {
     transport_ = transport;
     maxLength_ = maxLength;
+    writeBuffer_.write(sizeFiller_, 0, 4);
```
allengeorge: Q: why were you able to do the initial write in the constructor here, but not in `TFastFramedTransport`?

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
jeking3: Agree, but the alternative was to change all the constructors so they could throw TTransportException.  That's still a viable option but it is a breaking change for some folks who now have to handle the potential (but in all likelihood will never happen) exception.  What do you think it better?  A single size lookup is pretty fast compared to the array copy of writing so I thought this would be acceptable.

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
jeking3: If the only class consuming AutoExpandingBufferWriteTransport is TFastFramedTransport, we could embed the reservation for the size into that.  Thoughts?

lib/java/src/org/apache/thrift/transport/TFastFramedTransport.java
jeking3: Agreed; reset() could do this if it were in the writeBuffer instead.

lib/java/src/org/apache/thrift/transport/TFramedTransport.java
jeking3: The buffer classes are different and write buffer here does not throw TTransportException.

lib/java/src/org/apache/thrift/transport/AutoExpandingBufferWriteTransport.java
```diff
@@ -25,10 +25,20 @@
...
-  public AutoExpandingBufferWriteTransport(int initialCapacity) {
+  /**
+   * Constructor.
+   * @param initialCapacity the initial capacity of the buffer
+   * @param frontReserve space, if any, to reserve at the beginning such
+   *                     that the first write is after this reserve.
+   *                     This allows framed transport to reserve space
+   *                     for the frame buffer length.
+   */
+  public AutoExpandingBufferWriteTransport(int initialCapacity, int frontReserve) {
```
allengeorge: Probably assert that `initialCapacity >= frontReserve`?

Other than that, I like the fact that there's no conditional check on each `write()`

jeking3:
@allengeorge IAE added with tests.

PR #1673, opened 2019-01-04
 Fix: do not reconnect if connection.destroyed, prevent connection leak

zouyixiong:
I can't destroy the connection while the server is still alive, so do this modify.

When will this happen ?

I use consul to register and discover the servers, and save the clients in an array.
But I found sometimes the server is alive, but it's deregistered from consul, so I splice the  deregistered instance, and hope that the connection destroy auto.
If the server is really down, the connection destroy automatically, but the problem is, the server maybe still alive, so the connection never destroy. So the connection is leak, as it was splice out from the clients array, never accessible.
When server register again, new connection will be create, and put into clients array.

So I want to destroy the connection manually before splice it out. But find out I can't, as it will reconnect.

### DIFF
--- a/lib/nodejs/lib/thrift/connection.js
+++ b/lib/nodejs/lib/thrift/connection.js
@@ -221,6 +221,10 @@ Connection.prototype.connection_gone = function () {
   });
 
   this.retry_timer = setTimeout(function () {
+    if (self.connection.destroyed) {
+        return;
+    }
+
     log.debug("Retrying connection...");
 
     self.retry_totaltime += self.retry_delay;
--- a/lib/nodejs/lib/thrift/connection.js
+++ b/lib/nodejs/lib/thrift/connection.js
@@ -222,7 +222,7 @@ Connection.prototype.connection_gone = function () {
 
   this.retry_timer = setTimeout(function () {
     if (self.connection.destroyed) {
-        return;
+      return;
     }
 
     log.debug("Retrying connection...");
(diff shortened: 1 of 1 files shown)


lib/nodejs/lib/thrift/connection.js
```diff
@@ -221,6 +221,10 @@ Connection.prototype.connection_gone = function () {
   });
 
   this.retry_timer = setTimeout(function () {
+    if (self.connection.destroyed) {
+      return;
```
jeking3:
```suggestion
      self.retry_timer = null;
      return;
```

Based on the way the code is written below...

zouyixiong: Thanks for the suggest. Changed.

zouyixiong:
By the way, what time will be the next version release ?

jeking3:
0.12.0 was just released yesterday, so probably 6-or-so months.  If an issue is serious we can make a package update to a particular package manager like npm, however note that right now we're waiting for @jfarrell to add other people as maintainers of the official thrift package, so we haven't pushed 0.12.0 up to npm yet.

zouyixiong:
Is it possible to publish 0.12.0 to npm in the these days ? ( with this pull request )
I want to fix connection leak problem on production env.

jeking3:
We're blocked on @jfarrell enabling more people to do it.  See the libraries page on thrift.apache.org.

PR #1674, opened 2019-01-04
THRIFT-4716: add veralign.sh tool to set versions for releases

jeking3:
This tool will allow the release manager to more easily create a release by automating the update of version numbers in all of the project and packaging files across all of the languages.  Typical usage will be to create a branch to start a release, complete the release, and then submit a pull request to move master to the next version:

    thrift$ ./veralign.sh 1.0.0 1.1.0

This would update most of the files with the correct version information which could be submitted through a pull request.  If a patch release is needed,

    thrift$ ./veralign.sh 0.12.0 0.12.1

`veralign.sh` will modify files but will not check in anything - that's up to the release engineer.  It updates all the files identified in the 0.11.0 to 0.12.0 release cycle.  It is idempotent, which means if you run it like this:

    thrift$ ./veralign.sh 1.0.0 1.0.0

There should be no changes in any files, even though it went through all the motions to update the files.  Also included in the PR is some minor YAML and/or JSON cleanup that the `jq` and `yq` tools decided to make during testing.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,6 +11,5 @@
     "thrift"
   ],
   "license": "Apache v2",
-  "ignore": [
-  ]
+  "ignore": []
 }
--- a/configure.ac
+++ b/configure.ac
@@ -158,7 +158,7 @@ if test "$with_cpp" = "yes";  then
     have_cpp="yes"
   fi
 
-  AX_LIB_EVENT([1.0])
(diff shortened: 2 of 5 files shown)


jeking3:
Looks like I have to tell yg to keep comments... it stripped out license headers from some yaml files.

PR #1675, opened 2019-01-04
Skip native extension build on jruby platforms

zindorsky:
Installing the ruby gem on systems without make fails in the build_ext task.


jeking3:
Thanks - could you open a Jira ticket for this?  This is something that should be in the release notes for the next release.

jeking3:
See THRIFT-4721

PR #1677, opened 2019-01-05
THRIFT-4720:migrate code to c++11

cyyever:
As discussed in THRIFT-4678.

This closes #1448


build/cmake/DefinePlatformSpecifc.cmake
```diff
@@ -97,21 +97,14 @@ elseif(WITH_STDTHREADS)
 endif()
 
 # C++ Language Level
-set(CXX_LANGUAGE_LEVEL "C++${CMAKE_CXX_STANDARD}")
```
jeking3: I believe CXX_LANGUAGE_LEVEL is used in the build summary output which is why it was being built as a string.

build/cmake/DefinePlatformSpecifc.cmake
```diff
@@ -97,21 +97,14 @@ elseif(WITH_STDTHREADS)
...
-if (CMAKE_CXX_STANDARD_REQUIRED)
-  string(CONCAT CXX_LANGUAGE_LEVEL "${CXX_LANGUAGE_LEVEL} [compiler must support it]")
-else()
-  string(CONCAT CXX_LANGUAGE_LEVEL "${CXX_LANGUAGE_LEVEL} [fallback to earlier if compiler does not support it]")
+set(CMAKE_CXX_STANDARD 11)
```
jeking3: By setting the standard outright, one cannot configure it on the build line to C++14, C++17, or C++2a.  Can we set it to 11 if it isn't already set?

cyyever: @jeking3  fixed.

lib/cpp/CMakeLists.txt
```diff
@@ -103,14 +103,7 @@ if(OPENSSL_FOUND AND WITH_OPENSSL)
...
 # WITH_*THREADS selects which threading library to use
-if(WITH_BOOSTTHREADS)
-    set( thriftcpp_threads_SOURCES
-        src/thrift/concurrency/BoostThreadFactory.cpp
-        src/thrift/concurrency/BoostMonitor.cpp
-        src/thrift/concurrency/BoostMutex.cpp
-    )
-    list(APPEND SYSLIBS "${Boost_LIBRARIES}")
-elseif(UNIX AND NOT WITH_STDTHREADS)
+if(UNIX AND NOT WITH_STDTHREADS)
```
jeking3: We don't need pthread raw support any more.  Only use std::threads, get rid of all the options?  Perhaps in the next round?

lib/cpp/Makefile.am
```diff
@@ -106,15 +106,9 @@ libthrift_la_SOURCES = src/thrift/TApplicationException.cpp \
                        src/thrift/server/TThreadPoolServer.cpp \
                        src/thrift/server/TThreadedServer.cpp
 
-if WITH_BOOSTTHREADS
```
jeking3: Hmm, where is the STDTHREADS?

lib/cpp/src/thrift/concurrency/PlatformThreadFactory.h
```diff
@@ -22,9 +22,7 @@
 
 // clang-format off
 #include <thrift/thrift-config.h>
-#if USE_BOOST_THREAD
-#  include <thrift/concurrency/BoostThreadFactory.h>
-#elif USE_STD_THREAD
+#if USE_STD_THREAD
```
jeking3: I'd really recommend getting rid of PosixThreadFactory.h and move the contents of StdThreadFactory.h into PlatformThreadFactory.h, and maybe rename it to TThreadFactory using std::thread only.

lib/cpp/CMakeLists.txt
cyyever: yes，next round

cyyever:
> 
> 
> Definitely getting there; would be nice to eliminate the stdcxx namespace and header (or keep it and just keep the C++11 stuff for now), eliminate the posix and platform thread factory gunk, and just have one TThreadFactory which is the std::thread code.

stdcxx has been removed.

lib/cpp/src/thrift/concurrency/PlatformThreadFactory.h
cyyever: > 
> 
> I'd really recommend getting rid of PosixThreadFactory.h and move the contents of StdThreadFactory.h into PlatformThreadFactory.h, and maybe rename it to TThreadFactory using std::thread only.

Posixthread has policy and priority settings which are lack in std::thread,So deleting it may break many existing codes,and these settings may be important for performance.
Of course I can port these settings to std::thread if native_handle() is pthread,but I would like to change in another PR.

jeking3: Tweaking thread priority is usually a slippery slope... and the pthread code has platform incompatibilities.  We should remove it all if we're going to do this. :)

lib/cpp/src/thrift/server/TServer.h
```diff
@@ -107,99 +107,99 @@ class TServer : public concurrency::Runnable {
...
-    setOutputProtocolFactory(stdcxx::shared_ptr<TProtocolFactory>(new TBinaryProtocolFactory()));
+    setInputTransportFactory(std::shared_ptr<TTransportFactory>(new TTransportFactory()));
+    setOutputTransportFactory(std::shared_ptr<TTransportFactory>(new TTransportFactory()));
+    setInputProtocolFactory(std::shared_ptr<TProtocolFactory>(new TBinaryProtocolFactory()));
+    setOutputProtocolFactory(std::shared_ptr<TProtocolFactory>(new TBinaryProtocolFactory()));
   }
 
-  TServer(const stdcxx::shared_ptr<TProcessorFactory>& processorFactory,
-          const stdcxx::shared_ptr<TServerTransport>& serverTransport)
+  TServer(const std::shared_ptr<TProcessorFactory>& processorFactory,
```
jeking3: Looking forward to using C++11 delegating constructors to solve some of this code bloat.

jeking3:
Turned into quite the pull request.  Are you done making changes?  I'd like to see PosixThreadFactory and PlatformThreadFactory gone.  The whole concept was not done well (it didn't even use inheritance!).  Are you planning to do that after?

cyyever:
Yes,this PR should focus on using c++11 features. I will give a separated PR to remove pthread latter.

发自我的 iPhone

在 2019年1月7日，上午12:14，James E. King III <notifications@github.com<mailto:notifications@github.com>> 写道：


Turned into quite the pull request. Are you done making changes? I'd like to see PosixThreadFactory and PlatformThreadFactory gone. The whole concept was not done well (it didn't even use inheritance!). Are you planning to do that after?

cyyever:
I will give a separated PR for removing pthread later

lib/cpp/src/thrift/server/TServer.h
cyyever: Yes ,these code should be simplified

jeking3:
These changes will break backwards compatibility, so all of these changes will need to be documented as breaking in the lib/cpp/README.md, in the top level CHANGES.  The code will be cleaner and standards compliant in the end, which I think is a good thing.  Folks can continue to use 0.12.0 and earlier to remain compatible with the code they have now.

emmenlau:
I just chip in to say a bit "thank you"! I'm looking forward to a more C++-1-based codebase. Two thumbs up!

jeking3:
I'll follow this up with a change to the cpp readme, top level changes, and top level languages files.

jeking3:
Darnit.  I meant to squash this into one commit, but I rebased it instead!  *grumble*

jeking3:
While this removed the runtime dynamic library dependencies on boost, it did not remove the requirement to have boost headers present to build the runtime library.  While boost is required to run the tests, it shouldn't be required to just build the CPP library.  There are about 40 more mentions of boost in the lib/cpp/src headers that need to be converted to std concepts, if they exist; or in the case of boost::noncopyable, do something else.  Note that the pull request for THRIFT-4441 had a solution for this.

helmesjo:
@jeking3 
I'm a bit torn with what is required... So boost is only required when building the thrift library, not when linking it (runtime)? I'm building my [conan-package](https://github.com/helmesjo/conan-thrift). If I use boost only for building, then my simple test_package.cpp fails with the following (among others):

```
TWinsockSingleton.h(34,10): fatal error C1083: Cannot open include file: 'boost/noncopyable.hpp': No such file or directory [...]
```

So do I understand your above comment correct that there is no need to link boost, but some headers are still required when building my code that includes the generated thrift-stubs?

You mention [THRIFT-4441](https://issues.apache.org/jira/browse/THRIFT-4441) but that has already been closed and is in 0.13 (allegedly), but obviously I still see the above issue (and boost is still required to both build and consume thrift 0.13).

**So basically:** 
What is the state regarding boost? Is there anything in the works to get rid of boost completely? Any light shed on this would be greatly appreciated!

PR #1678, opened 2019-01-06
THRIFT-4725: Change Return Type Signature of Process Methods

BELUGABEHR:


lib/java/src/org/apache/thrift/TBaseAsyncProcessor.java
```diff
@@ -89,7 +89,7 @@ public boolean process(final AsyncFrameBuffer fb) throws TException {
               out.getTransport().flush();
             }
             fb.responseReady();
-            return true;
+            return;
         }
         in.readMessageEnd();
```
jeking3: Check out line 97 below - it looks like responseReady is called before we start processing the response.  Does that lead to optimal behavior or will the async server end up blocking trying to read the reply that doesn't exist?

jeking3:
I'm actually a big fan of removing branches like this, especially if they are not used.  Given how deep the pipelines are in CPUs today, a branch misprediction is expensive, and return value checking requires additional unit testing of both branches.  With exceptions to handle issues both of these problems are eliminated.

It's unfortunate this leads to another breaking change however... I think that's more fodder to call the next release 1.0.0 :)

BELUGABEHR:
@jeking3 Checking this into the V1 branch?  Should I document as broken starting with V1 or V0.12?

jeking3:
A change like this would not be pulled into a 0.12.0 refresh for a language.  This would be in the next major release which is currently 1.0 (although that could change).

jeking3:
Reminder: this needs additional work.

CHANGES.md
```diff
@@ -19,6 +19,7 @@
 - [THRIFT-4730](https://issues.apache.org/jira/browse/THRIFT-4730) - C++: BoostThreadFactory, PosixThreadFactory, StdThreadFactory removed
 - [THRIFT-4732](https://issues.apache.org/jira/browse/THRIFT-4732) - C++: CMake build changed to use BUILD_SHARED_LIBS
 - [THRIFT-4735](https://issues.apache.org/jira/browse/THRIFT-4735) - C++: Removed Qt4 support
+- [THRIFT-4725](https://issues.apache.org/jira/browse/THRIFT-4725) - Java change return type signature of 'process' methods
```
jeking3: In order, next time? :)

BELUGABEHR: Thank you for the approval.  I was doing it in chronological order.  Sorry :)

PR #1683, opened 2019-01-08
THRIFT-4730:remove pthread code

cyyever:
Using C++11 std::thread and std::mutex instead of  pthread to enhance portability across different platforms (especially Windows).


cyyever:
@jeking3 I am not sure whether to remove pthread_mutux* from mutex.h

jeking3:
Is there anything that still uses it?  Shouldn't it all be `<mutex>` now?

cyyever:
Only mutex and related tests use pthread  lib

jeking3:
The goal is to remove any platform-specific thread or mutex concepts and use the C++11 standard, so I would say pthread has to go.

jeking3:
Since we're not eliminating the StdMutex class, we still need a unit test for the resulting Mutex class. (could rename StdMutex.cpp / .h to Mutex.cpp / .h, etc...)  I think your PR removed the unit test which is incorrect.  Removing the pthreads compile-time branches, sure. :)

jeking3:
As long as it gets done.

The appveyor build issue looks environmental.
The travis ones are not.

cyyever:
@jeking3 Not sure why one of the CI tests failed,it seems like an error in node js tests?

jeking3:
I'll kick that job and see if it was transient.

jeking3:
For future breaking changes we need to remember to update lib/cpp/README.md and the top level CHANGES.md file.

jeking3:
The concurrency_test is failing regularly on Appveyor now, in the MSVC2017 job.  This is preventing us from getting any good test runs in CI.  I opened https://issues.apache.org/jira/browse/THRIFT-4739 for this.

PR #1684, opened 2019-01-09
THRIFT-4675: Generate Int64 constants for js

mustafa-cosar:
This PR introduces Int64 constants for javascript using `node-int64` package.

The changes may not be backwards compatible. js thrift library needs `node-int64` (browserified) to work correctly.

@jeking3 can you review the PR?

Thanks,
Mustafa Senol Cosar


jeking3:
I'm kind of confused - we have lib/js, lib/nodejs, lib/ts, and lib/nodets.  I don't know which ones are valid and active any more?

jeking3:
Given this wasn't working properly before, is there an integration test we're missing to have clients send and servers agree on the maximum value for a given thrift numeric primitive type?

jeking3:
How does this affect #1519?

mustafa-cosar:
> How does this affect #1519?

I pulled their changes and ran the tests and they passed. That is all I can say.

mustafa-cosar:
> I'm kind of confused - we have lib/js, lib/nodejs, lib/ts, and lib/nodets. I don't know which ones are valid and active any more?

> Given this wasn't working properly before, is there an integration test we're missing to have clients send and servers agree on the maximum value for a given thrift numeric primitive type?

I'm guessing these questions are not meant for me?

jeking3:
What happens if someone continues to use the old way of specifying a number, i.e. "4" instead of "new Int64(4)" in their js files?

mustafa-cosar:
> What happens if someone continues to use the old way of specifying a number, i.e. "4" instead of "new Int64(4)" in their js files?

`assert.equal(client.testI64(4), 4);` in the test works.

mustafa-cosar:
@jeking3 I documented the breaking changes

jeking3:
AppVeyor issue looks to be environmental.

mustafa-cosar:
> Please comment on whether #1519 is still valid. Thanks.

I think so, I looked at the changes on that PR and they seem to be unrelated to generating int64 constants.

However that made me realize that there is a problem with int64 type map keys. I need to open another PR to fix it.

bgedik:
> I'm kind of confused - we have lib/js, lib/nodejs, lib/ts, and lib/nodets. I don't know which ones are valid and active any more?

`lib/js` and `lib/nodejs`  are for using Thrift on the browser and on nodejs, using JavaScript.

`lib/ts` is for using Thrift on the browser via Typescript, and `lib/nodets` is for using Thrift on nodejs via Typescript. 

Some of these directories contain source code and test. IMO, the following changes should be made, but I am not 100 sure:

- `lib/js` contains `src/thrift.js` (JS code for Thrift library for the browser). I don't see this file being included in the Thrift npm package. So this file seems unused. Instead, the Thrift npm package includes the JS code for nodejs, but to make it usable from the browser, in package.json, it specifies a 'browser' field. This field points to a browser.js file, which defines a subset of the nodejs code to be activated in the browser. IMO, the directory `src` under `lib/js` is unused. Most likely it should be removed. But the tests should stay.
 
- `lib/nodejs` contains `lib/thrift` (JS code for Thrift library for nodejs). This looks fine to me.

- `lib/ts` contains `thrift.d.ts` (type mappings for the above). This is another confusion point. As of now, Thrift type mappings seem to come from DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/thrift/index.d.ts). So I am not sure what this file is used for. 

- `lib/nodets` only contains tests, which seems fine.

cdagli:
@jeking3 @bgedik 

It would be nice if we can extend the current `thrift.d.ts` and configure it in the package.json's `types` so we won't need to use the one in the DefinetelyTyped -seems like it is not maintained very well.

bgedik:
The existing `thrift.d.ts` would need to move under `lib/nodejs`.

jeking3:
I have been looking at the pull request for THRIFT-4214.  It looks like node-Int64 does NOT allow for 64-bit mathematics in node.  It is really not much more than a buffer.  Another package, bignum, lets you do the integer math.  Do we really want to force folks to use a 64-bit integer package that can't do 64-bit integer math?  The cross tests simply spit back what they were passed.  It's an unlikely scenario for an actual user of Thrift.

Is anybody interested in reworking this to use bignum instead?

Thanks, Jim

bgedik:
@jeking3 I think a better alternative could be the 'long' package https://www.npmjs.com/package/long. Bigint would allow results to be non-Longs. 

The reason we kept the original node-int64 was that, we wanted to avoid breaking client code. Thrift was already using node-int64.

jeking3:
I was under the assumption that THRIFT-4214 was already a breaking change.  I looked over the long package and it seems fine.  Similar to bignum in functionality, but it doesn't have the native library requirements (openssl).  bignum actually allows for numbers beyond 64-bit, which is not a requirement for us.

Is this work you could/would-like-to do?

bgedik:
@jeking3 Even before this change, Thrift was already using `node-int64` both in its generated code (for RPCs) and in its runtime library. This PR only added the same to generated constants. We kept `Int64` to be consistent with the rest of Thrift.

What you are suggesting is to change the entire `Int64` support with `Long`. This is certainly better. In fact, we use `Long` in our code base. Whether we are willing to contribute this or not is something we'll discuss and report back.

bgedik:
@jeking3 We won't be able to do this right away. However, we can come back to this once we have more resources.

jeking3:
I'll open a new ticket for you.

PR #1686, opened 2019-01-11
THRIFT-4733: Fix test unix socket address already in use

Junnplus:


lib/py/test/test_sslsocket.py
```diff
@@ -213,11 +213,9 @@ def test_unix_domain_socket(self):
             return
         fd, path = tempfile.mkstemp()
         os.close(fd)
-        try:
-            server = self._server_socket(unix_socket=path, keyfile=SERVER_KEY, certfile=SERVER_CERT)
-            self._assert_connection_success(server, path=path, cert_reqs=ssl.CERT_NONE)
-        finally:
-            os.unlink(path)
```
jeking3: We should still be removing it after the test (finally).  It makes sense to remove it before as you have added.

Junnplus: you're right.

jeking3:
It looks like a real issue - empty commit to kick the build won't change the outcome.

jeking3:
Please rebase on master as the issue seen in these builds was resolved.

PR #1687, opened 2019-01-11
THRIFT-4732: windows cmake refinements and add windows docker build support

jeking3:
This adds initial support for a Windows Docker container which successfully builds the portable windows thrift compiler (statically linked to runtime).

I also did some refactoring on cmake and introduced the potential to get windows to work properly with shared libraries, but it doesn't quite work.  The default for WIN32 is -DBUILD_SHARED_LIBS=OFF and the default for other platforms is ON.

I also removed the defunct build/wincpp which turned into the appveyor framework.

Fixed "cl" (LISP) language generator missing from the cmake file for the compiler.


jeking3:
There is already a backlog item for that: See THRIFT-3973, THRIFT-4025
It involves removing the static projects that are checked in, and then updating the documentation.  Some of the documentation is on the ASF CMS and not in GitHub.  The project docs are a bit of a mess unfortunately, but I agree with the direction.

jeking3:
@cyyever If you want to experiment with building shared libraries on Windows, try cmake -DBUILD_SHARED_LIBS=ON - the CMake project is ready but there are some loose ends to clean up.

PR #1688, opened 2019-01-12
THRIFT-4732:refine cmake scripts 

cyyever:
1.add -wall option for all major compilers.
2.add cmake support in test/c_glib
3.add static analysis targets

### DIFF
--- a/build/cmake/DefinePlatformSpecifc.cmake
+++ b/build/cmake/DefinePlatformSpecifc.cmake
@@ -23,6 +23,31 @@
 # For Debug build types, append a "d" to the library names.
 set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Set debug library postfix" FORCE)
 
+# basic options
+foreach(lang IN ITEMS C CXX)
+  if(CMAKE_${lang}_COMPILER_ID STREQUAL "Clang")
+    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} -Wall")
+    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} -ferror-limit=1")
+  elseif(CMAKE_${lang}_COMPILER_ID STREQUAL "GNU")
+    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} -Wall -Wextra")
+    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} -fmax-errors=1")
+  elseif(CMAKE_${lang}_COMPILER_ID STREQUAL "MSVC")
+    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /MP") # parallel build
+    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /W3") # warning level 3
+    include(CheckCXXCompilerFlag)
+    set(CMAKE_REQUIRED_QUIET ON)
+    check_cxx_compiler_flag("/source-charset:utf-8" res_var)
+    if (res_var)
+      set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /source-charset:utf-8")
+    endif()
+    check_cxx_compiler_flag("/execution-charset:utf-8" res_var)
+    if (res_var)
+      set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /execution-charset:utf-8")
+    endif()
+    add_definitions("-DUNICODE -D_UNICODE")
+  endif()
+endforeach()
+
 # Visual Studio specific options
 if(MSVC)
     # Allow for shared library builds
@@ -66,11 +91,6 @@ if(MSVC)
 
     # Disable boost auto linking pragmas - cmake includes the right files
     add_definitions("-DBOOST_ALL_NO_LIB")
-
-    add_definitions("/MP") # parallel build
-    add_definitions("/W3") # warning level 3
-
-    add_definitions("-DUNICODE -D_UNICODE")
 elseif(UNIX)
   find_program( MEMORYCHECK_COMMAND valgrind )
   set( MEMORYCHECK_COMMAND_OPTIONS "--gen-suppressions=all --leak-check=full" )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,9 @@ include(DefineInstallationPaths)
 # Based on the options set some platform specifics
 include(DefinePlatformSpecifc)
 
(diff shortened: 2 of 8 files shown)


build/cmake/BoostMacros.cmake
```diff
@@ -19,6 +19,18 @@
 
 set(BOOST_MINREV 1.53)
 
+macro(ADD_BOOST_OPTIONS)
+  if (MSVC)
+    add_definitions(/DBOOST_ALL_DYN_LINK)
```
jeking3: We may not need this at all - if someone defines `Boost_USE_STATIC_LIBS` as part of the cmake command line then FindBoost might do the right thing for us.  As it is if someone does define `Boost_USE_STATIC_LIBS` this macro is wrong as these things should not be defined.

compiler/cpp/CMakeLists.txt
```diff
@@ -146,7 +146,7 @@ if(${WITH_PLUGIN})
         src/thrift/plugin/plugin_output.cc
         src/thrift/plugin/plugin.cc
     )
-    TARGET_INCLUDE_DIRECTORIES_THRIFT(thriftc PUBLIC ${Boost_INCLUDE_DIRS})
+    include(BoostMacros)
```
jeking3: Just including BoostMacros doesn't do anything, was there supposed to be something else?

build/cmake/BoostMacros.cmake
cyyever: fixed

build/cmake/BoostMacros.cmake
```diff
@@ -38,4 +38,13 @@ macro(REQUIRE_BOOST_LIBRARIES libs)
 	    "or use -DBUILD_TESTING=OFF")
     message(FATAL_ERROR, ${fatal_message})
   endif()
+  if (MSVC)
```
jeking3: These sections are the same.  They are also wrong.
The code that was removed from DefineOptions.cmake was:
```
if (NOT Boost_USE_STATIC_LIBS)
    add_definitions(-DBOOST_ALL_DYN_LINK)
    add_definitions(-DBOOST_TEST_DYN_LINK)
endif()
```

We need to check the link type before assigning any DYN_LINK fields.
I'm also fairly sure these flags do nothing if you are not on Windows, which is why the NO_LIB directive used to be in DefinePlatformOptions...

build/cmake/DefinePlatformSpecifc.cmake
```diff
@@ -63,14 +78,6 @@ if(MSVC)
...
         set(THRIFT_RUNTIME_POSTFIX "md" CACHE STRING "Set runtime library postfix" FORCE)
     endif(WITH_MT)
-
-    # Disable boost auto linking pragmas - cmake includes the right files
-    add_definitions("-DBOOST_ALL_NO_LIB")
-
-    add_definitions("/MP") # parallel build
-    add_definitions("/W3") # warning level 3
-
-    add_definitions("-DUNICODE -D_UNICODE")
```
jeking3: You removed this line but never added it anywhere else.

compiler/cpp/CMakeLists.txt
jeking3: It's better to let the original reporter resolve the issue they identified in GitHub.

build/cmake/DefinePlatformSpecifc.cmake
cyyever: These lines  were moved to 

> 

  elseif(CMAKE_${lang}_COMPILER_ID STREQUAL "MSVC")
    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /MP") # parallel build
    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /W4") # warning level 4
    set(CMAKE_${lang}_FLAGS "${CMAKE_${lang}_FLAGS} /source-charset:utf-8")
  endif()

in the same cmake file,and I changed W3 to W4.
"-DUNICODE -D_UNICODE" was used to make sure unicode source files could be compiled on MSVC,so I use "/source-charset:utf-8" switch.

build/cmake/BoostMacros.cmake
cyyever: I just moved these lines to the same file with other boost scripts.In fact,I can compile successfully on Windows without these lines,so they could be removed.

build/cmake/DefinePlatformSpecifc.cmake
jeking3: This may exclude anything before MSVC 2015 from working... but that's all CI is covering.  I'm okay with that.  2013 and earlier can have some pretty strange C++ behavior...

cyyever: > 
> 
> This may exclude anything before MSVC 2015 from working... but that's all CI is covering. I'm okay with that. 2013 and earlier can have some pretty strange C++ behavior...

It seems that vs2013 does not have "/source-charset",so I have to return to macros.

cyyever:
> 
> 
> Some issues with linking need to be addressed on Windows.

Have you ever tried to build shared library of master branch successfully on windows? I have encountered some unresolved external symbol errors even with CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS enabled.After spending several hours to investigate this problem,I found that we still need to put dllimport  and dllexport in some places.

jeking3:
See https://issues.apache.org/jira/browse/THRIFT-4759

cyyever:
> 
> 
> See https://issues.apache.org/jira/browse/THRIFT-4759

I would rebase this PR after THRIFT-4759 is solved.

jeking3:
THRIFT-4759 was solved, please rebase and continue. :)

lib/cpp/CMakeLists.txt
```diff
@@ -20,6 +20,8 @@
 # Remove the following once lib/cpp no longer depends on boost headers:
 include(BoostMacros)
 REQUIRE_BOOST_HEADERS()
+set(BOOST_COMPONENTS locale)
+REQUIRE_BOOST_LIBRARIES(BOOST_COMPONENTS)
```
jeking3: I thought we removed all runtime library dependencies on boost linked libraries?

cyyever: no,thrift/protocol/TJSONProtocol.cpp uses boost::locale,not to mention the boost/noncopyable.hpp

cyyever: By repushing this PR,now it would not change the boost code.

jeking3:
Does this close THRIFT-4732 or is there more work to do?

cyyever:
> 
> 
> Does this close THRIFT-4732 or is there more work to do?

To complete THRIFT-4732,I need another PR.

PR #1698, opened 2019-01-17
THRIFT-3587 TTLSSocket now supports connect timeout

coeamyd:
Copied code from TSocket.cs to TTLSSocket.cs to support connect timeouts. This fixes issue THRIFT-3587.

### DIFF
--- a/lib/csharp/src/Transport/TTLSSocket.cs
+++ b/lib/csharp/src/Transport/TTLSSocket.cs
@@ -295,7 +295,33 @@ namespace Thrift.Transport
                 InitSocket();
             }
 
-            client.Connect(host, port);
+            if (timeout == 0)            // no timeout -> infinite
+            {
+                client.Connect(host, port);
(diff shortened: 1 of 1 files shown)


coeamyd:
And heeeere's the pull request :-)

jeking3:
I find it interestingly inefficient to have TTLSSocket **not** deriving from TSocket, otherwise a common method like this timeout handling could be tucked away and reused instead of using code duplication to achieve these means.

coeamyd:
You are absolutely right about this. But, since this is a design decision, I would not want to change this of my own account. The implementation was like this, I just added the missing functionality.

jeking3:
Curious build failure - seems related to a recent commit in lib/csharp, but I thought that passed CI.  @Jens-G did we miss something in #1696?

Jens-G:
Could you rebase and trigger the build again? Should work now.

PR #1703, opened 2019-01-22
THRIFT-4675: Fix i64 map key generation for js

mustafa-cosar:
In the past, i64 map keys were generated as numbers. However it was broken with #1684

Since js keys can only be strings, the numbers were implicitly converted to strings. This PR explicitly uses strings for i64 map keys since some i64 values cannot be represented as a string on javascript.


jeking3:
If this closes #1519 please put "This closes #1519" in the description.

PR #1704, opened 2019-01-22
THRIFT-4715 C# union "data" should be strongly-typed

Jens-G:
Client: C#
Patch: Jens Geyer


compiler/cpp/src/thrift/generate/t_csharp_generator.cc
```diff
@@ -1254,10 +1254,10 @@ void t_csharp_generator::generate_csharp_union_definition(std::ostream& out, t_s
   indent_up();
 
   indent(out) << "public abstract void Write(TProtocol protocol);" << endl;
-  indent(out) << "public readonly bool Isset;" << endl;
-  indent(out) << "public abstract object Data { get; }" << endl;
+  indent(out) << "public readonly int Isset;" << endl;
+  indent(out) << "protected abstract object Data { get; }" << endl;
 
-  indent(out) << "protected " << tunion->get_name() << "(bool isset) {" << endl;
+  indent(out) << "protected " << tunion->get_name() << "(int isset) {" << endl;
```
jeking3: Is this change from bool to int an optimization?

compiler/cpp/src/thrift/generate/t_csharp_generator.cc
```diff
@@ -1254,10 +1254,10 @@ void t_csharp_generator::generate_csharp_union_definition(std::ostream& out, t_s
   indent_up();
 
   indent(out) << "public abstract void Write(TProtocol protocol);" << endl;
-  indent(out) << "public readonly bool Isset;" << endl;
-  indent(out) << "public abstract object Data { get; }" << endl;
+  indent(out) << "public readonly int Isset;" << endl;
+  indent(out) << "protected abstract object Data { get; }" << endl;
```
jeking3: This looks like it would be a breaking change for folks who grab the data directly?  If so, it should be documented in the lib/csharp/README.md and in the top level CHANGES.md file.

compiler/cpp/src/thrift/generate/t_csharp_generator.cc
Jens-G: Depends how you define that … it stores the information which member is set **exactly**. The type alone may not do this:

    union message_piece {
      1: Participant Sender
      2: Participant Receiver
      3. string  Contents
    }

The type alone is not sufficient here, because trying to cast into ```Participant``` alone will not tell you, which case you actually have. 

I admit that the example is a bit artificial, but the construct as such is neither illegal nor absolutely unexpected. We use union constructs to wrap messages sent back and forth using that model, because it makes certain things easier to handle for us. 

Long story short: Yes, there is a good reason behind.

compiler/cpp/src/thrift/generate/t_csharp_generator.cc
Jens-G: I thought a while about this as well. I think I change it back to public, so the user has all the options (and can shoot himself in the foot in the way he likes most 🤣 ).

Jens-G: > Should the union option be the default (on the compiler)? 

Now *that* would certainly be a breaking change.

jeking3: Well perhaps in netstd it should be the default (if it's better).

PR #1710, opened 2019-01-24
THRIFT-4723 Consolidate C#/netcore into new netstd language target

Jens-G:
Client: netstd
Patch: Jens Geyer

Both C# and netcore are left untouched by this. Removal of these parts is a separate step, deprecation should be announced before 0.13.0, but actually be done after 0.13.0 (personal opinion, to be discussed).

lib/netstd/README.md contains a migration roadmap for both cases, listing all the things I ran into while porting over certain Thrift Projects and some of my own projects. It dos not claim to be complete, but it should provide a fairly good overview of the process.


Jens-G:
from https://api.travis-ci.org/v3/job/484155641/log.txt
> [WARNING:generation:1] Unable to get a generator for "perl".

Is that expected? I just checked locally, I do not get any errors with perl or php  codegen.

The other build errors seem not related, but I may have overlooked sth.

jeking3:
So that's pretty bizarre on the perl, looking at the master version of the Makefile.am in compiler/cpp, it is definitely included; however in the build output I don't see t_perl anywhere.

compiler/cpp/Makefile.am
```diff
@@ -100,9 +100,8 @@ thrift_SOURCES += src/thrift/generate/t_as3_generator.cc \
                   src/thrift/generate/t_lua_generator.cc \
                   src/thrift/generate/t_netcore_generator.cc \
                   src/thrift/generate/t_netcore_generator.h \
-                  src/thrift/generate/t_ocaml_generator.cc \
-                  src/thrift/generate/t_perl_generator.cc \
```
jeking3: Looks like you removed perl, ocaml, and php...

lib/netstd/README.md
```diff
@@ -0,0 +1,54 @@
+# Apache Thrift netstd
+
+Thrift client library for Microsoft .NET Standard
```
jeking3: Perhaps additional information on why one would use netstd instead of netcore or csharp (they are deprecated)?

compiler/cpp/Makefile.am
Jens-G: Ooopsie ...

jeking3:
@Jens-G there are 4 commits, so be sure to use the Squash and Merge button.  I approved it, but you can merge it and resolve the ticket. :)

Jens-G:
ok, thanks!

PR #1712, opened 2019-01-25
THRIFT-4760: Configure and install pkgconfig files when using cmake

remleduff:
The pkgconfig files are currently only installed when using autotools.

It's not too hard to install them using cmake, and it's nice for 3rd party projects which don't themselves use cmake


jeking3:
How will this interact with #1713?

build/cmake/DefineInstallationPaths.cmake
```diff
@@ -23,4 +23,9 @@ set(BIN_INSTALL_DIR "bin" CACHE PATH "The binary install dir (default: bin)")
 set(LIB_INSTALL_DIR "lib${LIB_SUFFIX}" CACHE PATH "The library install dir (default: lib${LIB_SUFFIX})")
 set(INCLUDE_INSTALL_DIR "include" CACHE PATH "The library install dir (default: include)")
 set(CMAKE_INSTALL_DIR "cmake" CACHE PATH "The subdirectory to install cmake config files (default: cmake)")
+set(PKGCONFIG_INSTALL_DIR "lib/pkgconfig" CACHE PATH "The subdirectory to install pkgconfig config files (default: lib/pkgconfig)")
 set(DOC_INSTALL_DIR "share/doc" CACHE PATH "The subdirectory to install documentation files (default: share/doc)")
+set(prefix "${CMAKE_INSTALL_PREFIX}")
```
jeking3: Could we move these into ADD_PKGCONFIG_THRIFT so they are not global?

build/cmake/ThriftMacros.cmake
```diff
@@ -17,6 +17,13 @@
 # under the License.
 #
 
+macro(ADD_PKGCONFIG_THRIFT name)
```
jeking3: Could we make this a function so that it can use the prefix, exec_prefix, libdir, and includedir variables without bleed-through?

lib/cpp/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ if(WIN32)
 else()
     TARGET_LINK_LIBRARIES_THRIFT(thrift ${SYSLIBS})
 endif()
+ADD_PKGCONFIG_THRIFT(thrift)
```
jeking3: I wonder if we could just have ADD_LIBRARY_THRIFT call this with the first argument?

remleduff:
> How will this interact with #1713?

I can't think of a reason why it would. They both create files in ${CMAKE_BINARY_DIR}, but I don't know what else is similar about them?

build/cmake/DefineInstallationPaths.cmake
remleduff: That seems to work.

Cmake variables are scoped to the current directory when they are defined in by default, rather than anything related to the function/macro they are defined in though.

build/cmake/ThriftMacros.cmake
remleduff: That seems to work. I'm not quite sure how to verify what is different between the two in this case.

lib/cpp/CMakeLists.txt
remleduff: The pkgconfig file-names don't match the target names. (thriftnb vs thift-nb.pc.in, etc)

I initially tried it this way, but figured being explicit was preferable to a magic find/replace.

build/cmake/DefineInstallationPaths.cmake
remleduff: Hmm, I may be confused. Apparently functions do introduce a new scope. I'm honestly not clear how this works, but I think what you suggest is fine.

lib/cpp/CMakeLists.txt
remleduff: Also see thrift-qt5 below, where 1) the name doesn't match  and 2) the pc.in file lives in a different directory from the ADD_LIBRARY_THRIFT call.

build/cmake/DefineInstallationPaths.cmake
jeking3: Items in DefineInstallationPaths.cmake are global.  I'm probably just being pedantic about this - those names are fairly generic.

PR #1713, opened 2019-01-26
THRIFT-4759: Adding cmake-generated header for symbol visibility

emmenlau:
This PR adds explicit symbol visibiliy with a cmake-based auto-generated header `thrift_export.h` for C++. For shared library builds, the header defines a number of macros. Most importantly it defines `THRIFT_EXPORT` on MSVC builds to ` __declspec( dllimport )` or ` __declspec( dllexport )`. CMake automatically adds the required compile-time define by which the definition of `THRIFT_EXPORT` is adjusted to either ` __declspec( dllimport )` or ` __declspec( dllexport )`.

With this header in place, explicit symbol visibility can be specified with the following modification from:
```
// Without any external symbol visibility:
class SomeClass {
...
}
```
change to:
```
// With full external symbol visibility:
#include "thrift_export.h"
class THRIFT_EXPORT SomeClass {
...
}
```
Its also possible to define partial visibility:
```
// With partial external symbol visibility:
#include "thrift_export.h"
class SomeClass {
    THRIFT_EXPORT int SomeMethod();
}
```

This approach can be transparently combined with the current approach to use `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)` in `build/cmake/DefinePlatformSpecifc.cmake`. Furthermore, the header defines the macros on Unix such that it can be used to hide symbols on Unix platforms (see https://gcc.gnu.org/wiki/Visibility for details).


jeking3:
Looks like more work is needed; perhaps you forgot to add the directory where the generated header is placed during the build?

Also you may need to disable the concurrency_test in the appveyor.yml for now, see my change in #1716 to appveyor.yml: https://github.com/apache/thrift/pull/1716/files#diff-180360612c6b8c4ed830919bbb4dd459 - we've had a lot of appveyor failures since committing the C++11 code; I documented it in THRIFT-4739.

emmenlau:
Interesting! The build works for me locally, but then again I do not build all possible thrift languages. I'll check it and get back to you.

emmenlau:
The build looks good now for at least two platforms on AppVeyor. I'm testing again with your change in #1716 to appveyor.yml: https://github.com/apache/thrift/pull/1716/files#diff-180360612c6b8c4ed830919bbb4dd459 and lets see if that helps to get any further. Thanks!

jeking3:
Both MSVC builds on appveyor right now use BUILD_SHARED_LIBS=OFF.  The MSVC2017 one should be changed in appveyor.yml.

emmenlau:
The build did not go any further with the change. But it does run through on at least 2 platforms, one of which is MSVC. I can also confirm the shared library build works for me on MSVC 2015 and MSVC 2017 Windows 7 x64.
Can the PR progress?

jeking3:
Not when it breaks the build for reasons related to the code submitted.  In this case, cygwin was busted by the changes.

https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/21911295/job/ekouonlbymt2q155#L2605

emmenlau:
Oh I see, I assumed its unrelated. I'll check it tomorrow.

jeking3:
If you rebase on master, the build is fairly clean at this point and master has 5 successful builds in a row.

jeking3:
Cygwin's a different beast, using windows-y rules but with a unix toolchain.  Usually you can "just build like it's unix" but not always.  I suggest cleaning up all the travis failures first.... then see if cygwin is still a problem.

emmenlau:
I see that a significant number of builds uses autotools, is that correct? Supporting autotools would require a different mechanism, so possibly #1714 could be simpler for that scenario.

cyyever:
> 
> 
> Ok I'm at a real loss as to why the build fails on Cygwin for me, but works for #1714. I think the diff-set is mostly identical and I did not touch the failing Cygwin sources. Is it possible that I hit a random bug in Cygwin?

I use "if defined(_MSC_VER )" to check msvc compiler instead of _win32 macro

jeking3:
We don't have cygwin support because of autotools.  We have it because folks still use it and we want to prove we can build on it.

emmenlau:
> I prefer this one where the file is generated and used.

I agree. But how could autotools use the header that is cmake-generated? I see these options:
1. Generate the same header also from autotools. This would be possible but I'm not an expert how to do this.
2. Make the shared library build only available with cmake, and provide a dummy header for autotools that defines the `THRIFT_EXPORT` defines as empty dummies. CMake can place the generated header higher in the search path than the dummy header to be sure the correct one is used.
3. Pre-generate the header twice, once for shared and once for static library build, and store both headers in the source repository as verbatim copies. Then pick the matching header at configure time.

I'm not very fond of either of these choices. If anything then number (2) seems the most feasible and least invasive. But it brings quite some confusion about the duplicate header, and its possible that people will ask for shared library builds from autotools too.

What do yo think?

jeking3:
We are using cmake to build on cygwin, so it should work.

emmenlau:
> We are using cmake to build on cygwin, so it should work.

My comment is not specific to cygwin. If the export-header is cmake-generated, then autotools-users will not have it. That will break *all* non-cmake-based builds.

jeking3:
autotools based builds using gcc or clang have default global visibility and therefore should not be affected.  I must be missing something on my end...  Please explain.

emmenlau:
The problem is that in every source file where symbols should explicitly be exported, we need to do two things:

1. Include the export header: `include "thrift/thrift_export.h"`
2. Add the export definition: `class THRIFT_EXPORT SomeClass {`

Both changes will break autotools builds. Autotools would not have the header and would not have the `THRIFT_EXPORT` definition. We would need to provide some drop-in replacement for both.

emmenlau:
This is my suggestion number (2) from above, yes. Its certainly possible, but I'm not an autotools expert so I won't be able to implement that myself.

jeking3:
I'll try to help out with that.

cyyever:
It would be better to just add the header to source than modify autotools to generate it.And just use my PR.

emmenlau:
Dear @jeking, I usually prefer the same as you suggest, to rely on the build tools for auta-generated headers. But after thinking for a while I agree with @cyyever that *if* you need to support autotools, then his PR is better. Its much easier to understand his PR, and both systems would consistently use the same header.

jeking3:
Merged #1714 instead of this one.

PR #1715, opened 2019-01-26
THRIFT-4757: fix npm identified security vulnerability in transient dependency on sync-exec

jeking3:
The direct dependency `grunt-shell-spawn` was updated to no longer need `sync-exec`.  Also fixed version and package meta for `lib/ts`.


jeking3:
AppVeyor issue is the concurrency test failure, unrelated.

PR #1717, opened 2019-01-26
THRIFT-1549: properly unwrap and close TSSLSocket in python

jeking3:


jeking3:
This needs to be done after https://bugs.python.org/issue29413 is fixed..

PR #1718, opened 2019-01-27
THRIFT-4384: fix concurrent sync in cpp async client code

jeking3:
THRIFT-4384: the StressTest was not being run in concurrency mode so I added that to the cmake build and I implemented the fix proposed in THRIFT-4384 to make it work properly, although it had to change somewhat.

THRIFT-4739: timer manager test 4 was assuming a thread was exiting quickly, it needed additional logic since the notify call used to manage the test behavior was coming from inside the thread, not from the timer manager


jeking3:
I suppose instead of just applying the patch and de-boosting it, I should have actually tried it locally!

PR #1719, opened 2019-01-28
THRIFT-4762: Applied some C++11 refactorings to the runtime library and compiler

szenker:
I basically used clang-tidy to apply the following C++11 refactorings (which make very much sense to me)

- make use of `override` keyword whenever a virtual function is overwritten
- make use of `auto` keyword for iterators
- make use of `auto` keyword when every a redundancy can be avoided, e.g. 
`uint32_t len = static_cast<uint32_t>(str.length());
vs.
`auto len = static_cast<uint32_t>(str.length());`
- replaced `NULL` with `nullptr`
- make use of explicitly-defaulted function definition

Additionally, I applied some more const-correctness to some functions.


jeking3:
Any chance we could fix the delegating constructor mess? :)

jeking3:
I'd prefer those things go into the same commit but if you want to add a second PR with "[ci skip]" at the end of the commit message and pull request title, addressing the three things I mentioned, that would be okay too.  You did get a clean build after all.

szenker:
@jeking3 I fixed all three issues as listed above (extended changelog and added Breaking-Change label)

PR #1732, opened 2019-02-07
THRIFT-4776:Modernize c++11 code by clang-tidy

cyyever:


jeking3:
Recommend building locally with the docker container before pushing up to CI...

lib/cpp/src/thrift/transport/TSSLSocket.cpp
```diff
@@ -318,6 +318,8 @@ bool TSSLSocket::peek() {
     } else if (rc == 0) {
       ERR_clear_error();
       break;
+    } else {
+      break;
```
jeking3: interesting...

cyyever: This case is omitted causing the method to always return false .

PR #1738, opened 2019-02-09
THRIFT-4798: fix py/py3 server, java client with http transport

jeking3:
The java TestClient asks the server to runa  oneway request that
sleeps for 3 seconds.  If the java TestClient sees the duration
of the call exceed one second, it fails the test.  This means the
server did not participate in the "fire and forget" dynamics of
ONEWAY requests.  In this case the THttpServer was processing the
RPC before sending the transport response.  The fix was to enhance
the TProcessor so that the THttpServer has an opportunity to inspect
the message header before processing the RPC.

This is partly due to the violation of the THttpServer in the
layered architecture.  It is essentially implementing a combined
server and transport, whereas there should be a distinct server,
protocol, and transport separation.  Many languages seem to have
this problem where HTTP was introduced.

I can't say this is the best solution possible, but it is one solution.
In general, ONEWAY request dynamics over HTTP are broken in
Apache Thrift and I'm not sure what the best source of action is
with respect to that yet.

- [x] Do you need an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?<details><summary>Expand for guidance...</summary>
    - `Yes` if your change requires a release note.
    - `Yes` if your change is a breaking change.
    - `No` if you change is trivial, such as fixing a typo.
</details>
 
- [x] Is this change worthy of a release note? <details><summary>Examples of Release Note-worthy examples...</summary>
    - Breaking Changes
    - New, Deprecated, or Removed Languages
    - Security Fixes
    - Significant Refactoring
    - Changing how the product is built
</details>

- [ ] Breaking changes have additional requirements: <details><summary>Expand for instructions...</summary>
    - Add or reference an existing Apache Jira THRIFT ticket.
    - Add a `Breaking-Change` label to the Jira ticket.
    - Add a note to the `lib/<language>/README.md` file.
    - Add a line to the `CHANGES.md` file.
</details>

- [x] Does this change require a build? <details><summary>Expand for guidance...</summary>
    - `Yes` for any code change
    - `Yes` for any build script change
    - `Yes` for any docker build environment change
    - `Yes` for any change affecting the cross test suite
    - `No` for documentation-only changes
    - `No` for trivial changes, for example fixing a typo.
    <br/>
    If your change does not require a build, you can add [ci skip] to the end of your commit message.<br/>
    This will avoid costly and unnecessary builds in both the pull request and once it is merged.
</details>


Jens-G:
@nsuke: notsutre what you mean by "drop" exactly: only for C++and py? For all? 

Either way, I still would hesitate to agree. If the HTTP transport needs a reply because HTTP requires it - well, fine.  But that should not affect Thrifts `oneway` semantics: no return value, not even an exception. 

This is from the Whitepaper (`oneway` was called `async` back then)

> Note that void is a valid type for a function return, in addition
> to all other defined Thrift types. Additionally, an async modifier
> keyword may be added to a void function, which will generate
> code that does not wait for a response from the server. Note that a
> pure void function will return a response to the client which guarantees
> that the operation has completed on the server side. With
> async method calls the client will only be guaranteed that the request
> succeeded at the transport layer. (In many transport scenarios
> this is inherently unreliable due to the Byzantine Generals’ Problem.
> Therefore, application developers should take care only to use
> the async optimization in cases where dropped method calls are
> acceptable or the transport is known to be reliable.)

jeking3:
Properties of a oneway request, commonly known as a fire-and-forget message:

- The client may throw an exception during the request call due to protocol or transport level issues in communicating with the server.
- If the transport requires it, a successful but otherwise empty response will be immediately sent on receipt of the oneway request.  In this case the transport must not delay the reply until the RPC is processed (this is what the Java cross test client is testing for).
- There is no thrift protocol level reply from the server.  This means any exception generated by the server will be dropped.  (Perhaps it should be logged, but that's a separate discussion).

I haven't looked, but is there a HTTP mode where the client can tell the server no reply is necessary?  From memory there is no such thing, but things change... At my last company the entire networking layer between systems relied on the "async" philosophy, leaving the application layer responsible to check on and potentially retry idempotent behavior.  It can lead to a more scalable design.

jeking3:
This pull request does what Jens suggests, at least for the THttpServer class we provide for Python.  I suppose we should be doing the same thing to the others.

jeking3:
I created THRIFT-4798 to track this for the release notes.

Jens-G:
> Yang
Ying

Jens-G:
+1 Looks good to me (with the disclaimer, thatI did not test the impl in practice)

PR #1741, opened 2019-02-13
THRIFT-4768 Remove "nullable" option from the code base (netstd ONLY)

Jens-G:
Client: netstd
Patch: Jens Geyer

Removes the (already  disabled) "nullable" option from the netstd generator code

- [x] Does this change require a build?

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_netstd_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_netstd_generator.cc
@@ -140,8 +140,6 @@ static bool type_can_be_null(t_type* ttype)
 
 bool t_netstd_generator::is_wcf_enabled() const { return wcf_; }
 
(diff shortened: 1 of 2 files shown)


compiler/cpp/src/thrift/generate/t_netstd_generator.cc
```diff
@@ -3110,6 +3013,5 @@ THRIFT_REGISTER_GENERATOR(
     "C#",
     "    wcf:             Adds bindings for WCF to generated classes.\n"
     "    serial:          Add serialization support to generated classes.\n"
-    //"    nullable:        Use nullable types for properties.\n"
     "    union:           Use new union typing, which includes a static read function for union types.\n"
```
jeking3: Curious, are we going to make some of these options defaults "true" (like union) for netstd as they are improvements?

PR #1746, opened 2019-02-18
THRIFT-4805: Suppress excessive logging of SASL TTransportExceptions in case of END_OF_FILE

mythrocks:
Two fixes here:

1. Additional logic to properly catch and handle TTransportException.
   Currently, T(SASL)TransportException gets caught and handled in
   the wrong catch-block.
2. The fix for THRIFT-3769 mutes _all_ TTransportExceptions in TThreadPoolServer.
   This might mute legitimate failures. The intent of THRIFT-3769 (and
   THRIFT-2268) was to mute the noise caused by TTransportException.END_OF_FILE.
   This commit lets legitimate failures to be bubbled upwards.


lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
```diff
@@ -343,5 +341,10 @@ public void run() {
         }
       }
     }
+
+    private boolean isIgnorableTSaslTransportException(Exception x) {
+      return (x instanceof TSaslTransportException && ERR_NO_SASL_DATA_IN_STREAM.equals(x.getMessage()))
```
jeking3: Relying on parsing an exception's message is less than ideal.

mythrocks: I'm with you, sir. I could just remove that part because, at the moment, `TSaslTransportException` is thrown only in the `END_OF_FILE` case.
In other words, if `(x instanceof TSaslTransportException)`, then `x.getMessage()` is always `ERR_NO_SASL_DATA_IN_STREAM`.
I'll update appropriately.

jeking3: I'd say get rid of TSaslTransportException all together.  I don't see a need for it?

mythrocks: You're right. I'll simplify this.

mythrocks:
Updated the PR now. I have removed the `TSaslTransportException`, and now check for the `END_OF_FILE` condition directly. Is this more palatable?

lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
mythrocks: Removed `TSaslTransportException`. Added direct checks for `END_OF_FILE`.

lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
```diff
@@ -29,13 +29,13 @@
...
 import org.apache.thrift.TProcessor;
 import org.apache.thrift.protocol.TProtocol;
-import org.apache.thrift.transport.TSaslTransportException;
 import org.apache.thrift.transport.TServerTransport;
 import org.apache.thrift.transport.TTransport;
 import org.apache.thrift.transport.TTransportException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.thrift.transport.TTransportException.END_OF_FILE;
```
jeking3: TTransportException is already imported - suggest using TTransportException.END_OF_FILE in your comparisons to keep with the coding style used in the file.

lib/java/src/org/apache/thrift/transport/TSaslTransport.java
```diff
@@ -310,14 +310,13 @@ public void open() throws TTransportException {
...
-       * If there is no-data or no-sasl header in the stream, throw a different
-       * type of exception so we can handle this scenario differently.
-       */
+      // If there is no-data or no-sasl header in the stream,
+      // log the failure, and clean up the underlying transport.
       if (!readSaslHeader && e.getType() == TTransportException.END_OF_FILE) {
         underlyingTransport.close();
-        LOGGER.debug("No data or no sasl data in the stream");
-        throw new TSaslTransportException("No data or no sasl data in the stream during negotiation", e);
+        if (LOGGER.isDebugEnabled()) {
```
jeking3: Is "isDebugEnabled" necessary when LOGGER.debug is used?

mythrocks: You're right. This check is not necessary for simple logging. 

If a heavier object (i.e. harder to convert `toString()`), it might make sense to do a pre-check. In this case, since it's a simple string, I should forgo the `isDebugEnabled()` check.

lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
mythrocks: Right you are. I've changed this to use `TTransportException.END_OF_FILE` instead.

mythrocks: In the course of resolving \#2 above, I noticed the following discrepancy, between the C++ and Java `TTransportException` types:

In C++:
```cpp
// TTransportException.h
  enum TTransportExceptionType {
    UNKNOWN = 0,
    NOT_OPEN = 1,
    TIMED_OUT = 2,
    END_OF_FILE = 3,
    INTERRUPTED = 4,
    BAD_ARGS = 5,
    CORRUPTED_DATA = 6,
    INTERNAL_ERROR = 7
  };
```
From the corresponding Java class:
```java
// TTransportException.java
  public static final int UNKNOWN = 0;
  public static final int NOT_OPEN = 1;
  public static final int ALREADY_OPEN = 2;
  public static final int TIMED_OUT = 3;
  public static final int END_OF_FILE = 4;
  public static final int CORRUPTED_DATA = 5;
```

`TIMED_OUT` and `END_OF_FILE` are off by 1, and `INTERRUPTED` isn't available among the Java types. This might need fixing separately. For the moment, I'll be muting `TIMED_OUT` and `END_OF_FILE`.

CHANGES.md
```diff
@@ -27,6 +27,7 @@
 - [THRIFT-4709](https://issues.apache.org/jira/browse/THRIFT-4709) - java: changes to UTF-8 handling require JDK 1.7 at a minimum
 - [THRIFT-4712](https://issues.apache.org/jira/browse/THRIFT-4712) - java: class org.apache.thrift.ShortStack is no longer public
 - [THRIFT-4725](https://issues.apache.org/jira/browse/THRIFT-4725) - java: change return type signature of 'process' methods
+- [THRIFT-4805](https://issues.apache.org/jira/browse/THRIFT-4805) - java: Suppress excessive logging of SASL TTransportExceptions in case of END_OF_FILE
```
jeking3: Not sure how this is a breaking change?

lib/java/README.md
```diff
@@ -170,9 +170,12 @@ http://gradle.org/
 
 ## 1.0
```
jeking3: It's really 0.13.0 (it was already there, but should be changed)

PR #1747, opened 2019-02-21
THRIFT-4807: Fix php segfault on reference pass

sokac:
Client: php (extension)

When reference is passed to php extension, it's unreferenced (instead of
derefrenced). The next time one tries to use the same variable PHP
segaults. Even if not used again, php segfaults on
user_shutdown_function_call.

Tested by rerunning code in Jira THRIFT-4807 with the patch - passes


jeking3:
Would you be able to add the example as a test case?

sokac:
It's very challenging as it requires classes defined in library, and not extension. Furthermore, as there are no tests for extension, I don't know does current setup allow compiling native extension and loading it up. It doesn't help I have no experience writing phpt tests, but seems straightforward.

Seems like a big project which I might be able to help with (I can file a Jira ticket for it) but I would suggest to be part of different PR.

Thoughts?

PR #1752, opened 2019-03-08
THRIFT-4536: TryFrom stable

dbcfd:
Submitting this as a draft PR to determine issues that should be created and which parts of PR should be merged.

TryFrom is now stable, so use that instead of try_from crate. Additionally re-export OrderedFloat since
it is used by the generated code. Relax dependencies to reduce downstream
conflicts.
 
### DIFF
--- a/lib/rs/Cargo.toml
+++ b/lib/rs/Cargo.toml
@@ -11,9 +11,8 @@ exclude = ["Makefile*", "test/**", "*.iml"]
 keywords = ["thrift"]
 
 [dependencies]
-byteorder = "~1.2.1"
-integer-encoding = "~1.0.4"
-log = "~0.3.8"
-threadpool = "~1.7.1"
-try_from = "~0.2.2"
-
+ordered-float = "0.5"
+byteorder = "1"
+integer-encoding = "1"
+log = "0.4"
+threadpool = "1.7"
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -550,20 +550,17 @@ void t_rs_generator::render_attributes_and_includes() {
   f_gen_ << endl;
 
   // add standard includes
-  f_gen_ << "extern crate ordered_float;" << endl;
   f_gen_ << "extern crate thrift;" << endl;
-  f_gen_ << "extern crate try_from;" << endl;
   f_gen_ << endl;
-  f_gen_ << "use ordered_float::OrderedFloat;" << endl;
+  f_gen_ << "use thrift::OrderedFloat;" << endl;
   f_gen_ << "use std::cell::RefCell;" << endl;
   f_gen_ << "use std::collections::{BTreeMap, BTreeSet};" << endl;
-  f_gen_ << "use std::convert::From;" << endl;
+  f_gen_ << "use std::convert::{From, TryFrom};" << endl;
   f_gen_ << "use std::default::Default;" << endl;
   f_gen_ << "use std::error::Error;" << endl;
   f_gen_ << "use std::fmt;" << endl;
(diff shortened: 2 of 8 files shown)


jeking3:
Draft PRs don't run the full CI suite.  Please change it to open so we can get a PR build, and @allengeorge will need to review.

dbcfd:
I think since review has been requested, I can't change it from draft. Would you like me to close and reopen?

jeking3:
Let's wait for the review.  @allengeorge any chance you could look at this?

allengeorge:
Hi all - yeap, can take a look. This will end up bumping our min version and moving to Edition 2018. I’m ok with that.


lib/rs/Cargo.toml
```diff
@@ -11,9 +11,8 @@ exclude = ["Makefile*", "test/**", "*.iml"]
...
 
 [dependencies]
-byteorder = "~1.2.1"
-integer-encoding = "~1.0.4"
-log = "~0.3.8"
-threadpool = "~1.7.1"
-try_from = "~0.2.2"
-
+ordered-float = "0.5"
+byteorder = "1"
```
dbcfd: Is 1.X sufficient or does this need to be "1.2"

dbcfd:
Should 2018 edition be specified since it uses stabilized TryFrom?

jeking3:
@allengeorge there's an open question on this one.

allengeorge:
Sorry for taking forever to get around to this. Aight, this opens up a big question: should we significantly bump the minimum Rust version? I'm actually OK with bumping it up to 1.33 because:

1. Most of the rust community moves pretty fast
2. This allows us to target Edition 2018
3. There are already a fair amount of breaking language changes in master (AFAIK)
4. I'd like to start adding futures support when it rolls around.

lib/rs/Cargo.toml
allengeorge: No - you should be able to leave it to 1. There's no need to pin it to 1.2. But why was the ordered-float crate added?

allengeorge:
@dbcfd There's no need to specify edition 2018. I could be wrong, but I believe those two are orthogonal?

allengeorge:
Confirmed: they are orthogonal. There's no need to specify Edition 2018, but, we will end up bumping min Rust version to 1.33

lib/rs/Cargo.toml
dbcfd: Generated code (https://github.com/apache/thrift/blob/3e5a1d9497062e2664a038daa668c3af954f72aa/compiler/cpp/src/thrift/generate/t_rs_generator.cc#L555) uses ordered_float, which means anyone that wants to use thrift also has to bring in ordered_float. Rust crates typically handle this by bringing in the dependency, and re-exporting it as necessary, which is the approach I took here. 

Bringing in thrift alone will now work for generated code, no additional crates are required.

Jens-G:
Could it be that this broke something as well? we get a lot of these errors now:

`error[E0658]:  use of unstable library feature 'try_from' (see issue #33417)`

https://travis-ci.org/apache/thrift/jobs/519745808

dbcfd:
It was supposed to be stable in 1.33, but didn't make it till 1.34
(11/04/2019). Will check travis runs and settings though.

dbcfd:
@allengeorge Should I open up a PR to make https://github.com/apache/thrift/blob/master/build/docker/ubuntu-bionic/Dockerfile#L259 just use rustup and stable? Same for xenial.

allengeorge:
Yes please. And, thank you.

I used to have the images download a pinned stable, but at some point it was changed to the distribution-packaged rust.

The minimum Rust version is going to have to change, and he CHANGES.md file is going to have to list this as a breaking change (as well as the Rust README).

PR #1758, opened 2019-03-14
THRIFT-4821 Normalized the constructors in TServerSocketTransport

animusblue:
…allow parity between the TcpListerer and port options.

Client: netstd
Patch: Kyle Smith

I refactored the constructors in the TServerSocketTransport so that you can use the build-in framed and buffered options while passing in a TcpListener object. Using default values, I was able to reduce the total number of constructors to maintain from 6 to 2 without breaking changes.

### DIFF
--- a/lib/netstd/Thrift/Transport/Server/TServerSocketTransport.cs
+++ b/lib/netstd/Thrift/Transport/Server/TServerSocketTransport.cs
@@ -28,47 +28,27 @@ namespace Thrift.Transport.Server
     public class TServerSocketTransport : TServerTransport
     {
         private readonly int _clientTimeout;
-        private readonly int _port;
         private readonly bool _useBufferedSockets;
         private readonly bool _useFramedTransport;
         private TcpListener _server;
 
-        public TServerSocketTransport(TcpListener listener)
-            : this(listener, 0)
-        {
-        }
-
-        public TServerSocketTransport(TcpListener listener, int clientTimeout)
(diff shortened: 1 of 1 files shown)


lib/netstd/Thrift/Transport/Server/TServerSocketTransport.cs
```diff
@@ -28,47 +28,27 @@ namespace Thrift.Transport.Server
...
-        {
-        }
-
-        public TServerSocketTransport(int port, int clientTimeout, bool useBufferedSockets):
-            this(port, clientTimeout, useBufferedSockets, false)
-        {
-        }
-        
-        public TServerSocketTransport(int port, int clientTimeout, bool useBufferedSockets, bool useFramedTransport)
+        public TServerSocketTransport(int port, int clientTimeout = 0, bool useBufferedSockets = false, bool useFramedTransport = false)
```
jeking3: It would have been cleaner to make a TcpListener here and then call the TcpListener constructor so there is one code path to initialization, not two.

animusblue: True, but you wouldn't be able to rethrow an exception as a TTransportException in this case. I'm not sure if that's a requirement though?

jeking3: Okay then, creating an initialize() method that does what the TcpListener does, and having both call it, would consolidate the code paths.  It isn't necessary, just a suggestion for future development.  Having two constructors initialize the same fields can lead to problems down the road when one constructor is modified and the other one is not.

PR #1761, opened 2019-03-15
convert net order to host order for double type

philomely:
- [ ] Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?  (not required for trivial changes)
- [ ] Did you squash your changes to a single commit?
- [ ] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"

### DIFF
--- a/lib/lua/src/luabpack.c
+++ b/lib/lua/src/luabpack.c
@@ -40,6 +40,17 @@ static int64_t T_ntohll(uint64_t data) {
   return ((uint64_t)d1 << 32) + (uint64_t)d2;
 }
 
+// network order to host order (64-bit)
+static double T_htond(double x){
+  int *Double_Overlay;
+  int Holding_Buffer;
+  Double_Overlay = (int *) &x;
+  Holding_Buffer = Double_Overlay [0];
+  Double_Overlay [0] = htonl (Double_Overlay [1]);
+  Double_Overlay [1] = htonl (Holding_Buffer);
+  return x;
+}
+
 /**
  * bpack(type, data)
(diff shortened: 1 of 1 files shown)


jeking3:
This broke all the lua tests.  My take is that the fix is unnecessary, as the cross-language tests were working before this fix (and they test doubles), and now they are broken.

philomely:
@jeking3 before the fix. lua version can't send double to java or c# correctly 'cause the order issue. not sure other tests

jeking3:
Our cross language tests send doubles from lua client to java server and back:

java-lua                binary            framed-ip                success
java-lua                binary            buffered-ip              success

On the client lua side here is the code:
https://github.com/apache/thrift/blob/master/test/lua/test_basic_client.lua#L147

On the server java side here is the code:
https://github.com/apache/thrift/blob/master/lib/java/test/org/apache/thrift/server/ServerTestBase.java#L89

The reverse where Java is the client:
https://github.com/apache/thrift/blob/master/lib/java/test/org/apache/thrift/test/TestClient.java#L282

and lua is the server:
https://github.com/apache/thrift/blob/master/test/lua/test_basic_server.lua#L57

HOWEVER we do not seem to be running any tests where lua is the server.  That's a bug, but I still feel the round trip from lua client through java server disproves what you said.

Is there something deficient in these tests we're missing?  We are proving that doubles successfully round-trip with these tests.  Are we testing enough values and enough boundaries to prove or disprove what you are saying?

jeking3:
Closing based on comments.

PR #1762, opened 2019-03-15
THRIFT-4825: Aligned the TTlsSocketServerTransport with the TSocketServerTransport

animusblue:
Updated the existing constructor to to use default values for buffering and replaced the simplified constructor with one that allows passing in a TcpListener instance.

### DIFF
--- a/lib/netstd/Thrift/Transport/Server/TTlsServerSocketTransport.cs
+++ b/lib/netstd/Thrift/Transport/Server/TTlsServerSocketTransport.cs
@@ -34,21 +34,15 @@ namespace Thrift.Transport.Server
         private readonly RemoteCertificateValidationCallback _clientCertValidator;
         private readonly int _clientTimeout = 0;
         private readonly LocalCertificateSelectionCallback _localCertificateSelectionCallback;
-        private readonly int _port;
         private readonly X509Certificate2 _serverCertificate;
         private readonly SslProtocols _sslProtocols;
         private readonly Buffering _buffering;
         private TcpListener _server;
-
-        public TTlsServerSocketTransport(int port, X509Certificate2 certificate)
-            : this(port, Buffering.None, certificate)
-        {
-        }
-
+               
         public TTlsServerSocketTransport(
-            int port,
-            Buffering buffering,
+            TcpListener listener,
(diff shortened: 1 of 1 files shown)


animusblue:
As pointed out by jeking [here](https://github.com/apache/thrift/pull/1758/files#r265942159), should I have simply created a TcpListener instance in the port only constructor? You wouldn't be able to rethrow an exception as a TTransportException, but is that a requirement?

jeking3:
In these cases I typically refactor the common initialization code to a single method and call that from both constructors.  It avoids future situations where someone modifies one constructor and not the other.  This typically happens when someone adds a variable and then forgets to initialize it in one of the constructors.  Some languages help protect against this, and it's been a while since I did any meaningful C#, so maybe that's not necessary and the compiler catches it?

animusblue:
Unfortunately .NET does not handle this for us. I could, however, pass null to the TcpListener parameter in the chain, and then use the existing code to set the _server value. It would fix the problem.

```csharp
        public TTlsServerSocketTransport(
            int port,
            X509Certificate2 certificate,
            Buffering buffering = Buffering.None,
            RemoteCertificateValidationCallback clientCertValidator = null,
            LocalCertificateSelectionCallback localCertificateSelectionCallback = null,
            SslProtocols sslProtocols = SslProtocols.Tls12)
            : this(null, certificate, buffering, clientCertValidator, localCertificateSelectionCallback, sslProtocols)
        {
            try
            {
                // Create server socket
                _server = new TcpListener(IPAddress.Any, port);
                _server.Server.NoDelay = true;
            }
            catch (Exception)
            {
                _server = null;
                throw new TTransportException($"Could not create ServerSocket on port {port}.");
            }
        }
```

jeking3:
I'd think that you would want to raise an exception if someone passed in a null TcpListener though.  I see what you mean though.  Anyway, it was just a thought.  It doesn't have to be done.  This is still an improvement as-is, I was just thinking about how to consolidate the initialization.

Jens-G:
@jeking3 the Initializer() way would be great, but we will have to sacrifice `readonly` then, since `readonly` vars can conly be set in a CTOR in C#. We could make it an R/O property though.

PR #1764, opened 2019-03-18
THRIFT-4828: Added the framed/buffered options to the TNamedPipeServerTransport

animusblue:
THRIFT-4828: Added the frame/buffered options to the TNamedPipeServerTransport
Client: netstd
Patch: Kyle Smith

### DIFF
--- a/tutorial/netstd/Server/Program.cs
+++ b/tutorial/netstd/Server/Program.cs
@@ -93,6 +93,8 @@ Options:
         tcp - (default) tcp transport will be used (host - ""localhost"", port - 9090)
         tcpbuffered - tcp buffered transport will be used (host - ""localhost"", port - 9090)
         namedpipe - namedpipe transport will be used (pipe address - "".test"")
+        namedpipebuffered - buffered namedpipe transport will be used (pipe address - "".test"")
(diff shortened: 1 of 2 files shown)


tutorial/netstd/Server/Program.cs
```diff
@@ -272,6 +280,8 @@ private enum Transport
             Tcp,
             TcpBuffered,
             NamedPipe,
+            NamedPipeBuffered,
+            NamedPipeFramed,
             Http,
             TcpTls,
             Framed
```
jeking3: (Understand this isn't part of new code in the PR, however:) @Jens-G Why is "Framed" not "TcpFramed"?  Seems out of place.  Also can one have TcpTlsBuffered, TcpTlsFramed?  Thanks.

animusblue: I can't speak to reason for the existing setup but I know you can have TcpTlsBuffered and TcpTlsFramed as I've used both in testing or production.

Jens-G:
TBH, I don't like T´these options at all, since they break the otherwise modular design. We have buffered and framed transports, so why do we need two extra options that do nothing different?

animusblue:
I'll close this PR in favor of creating a new one that reinstates the original modular design vision.

PR #1773, opened 2019-04-02
THRIFT-4842 - Fix memory leak on set property c_glib

gadLinux:
I've just release a string created in the constructor or previous call to set. 
It was not freed so it was leaked by previous commit.

The original issue is here: https://issues.apache.org/jira/browse/THRIFT-4842


gadLinux:
Any idea why tests are failing?

jeking3:
I saw some fixes go in last week for some of the issues like the Haskell issue, so first try rebasing your pull request and let's see what happens.

jeking3:
CI issues have been resolved; please rebase this one to get a clean build.

kainjow:
https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-free

> If mem is NULL it simply returns, so there is no need to check mem against NULL before calling this function.

PR #1777, opened 2019-04-05
THRIFT-3060: clear the offline queue when written

razvanz:
This PR addresses an issue in the NodeJS module where data would be sent multiple times when the socket reconnects multiple times because the `offline_queue` would never be emptied.

This fix clears the `offline_queue` when the socket reconnects and the queued data is written. 

Fixes [THRIFT-3060](https://issues.apache.org/jira/browse/THRIFT-3060).

### DIFF
--- a/lib/nodejs/lib/thrift/connection.js
+++ b/lib/nodejs/lib/thrift/connection.js
@@ -74,10 +74,7 @@ var Connection = exports.Connection = function(stream, options) {
     this.framePos = 0;
     this.frame = null;
     self.initialize_retry_vars();
-
-    self.offline_queue.forEach(function(data) {
-      self.connection.write(data);
-    });
+    self.flush_offline_queue();
 
     self.emit("connect");
   });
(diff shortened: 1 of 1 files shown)


razvanz:
@jeking3 Having looked [into the details](https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_socket_write_data_encoding_callback), all your 3 points seem to be valid issues. Without using the callback interface, when calling write, the message will be put into some internal queues. This means that if a disconnect happens there will be no transparency over what was delivered and what was left pending into internal queues. 

The solution is to use the callback interface for the `Socket.write` method, thus ensuring transparency over what get's delivered and what not. The `offline_queue` managed by this client adds yet another layer of queuing, which doesn't bring many benefits for the complexity it introduces so it should be avoided. This is however another discussion, which will require a bigger change.

This PR is meant to address a smaller issue in regards to the current implementation, which I think could be merged until a bigger redesign would be ready. Let me know what you think :smiley: .

jeking3:
I agree with you in that this is an improvement.  My comments were really just about the whole offline queue in general, and how it really isn't guaranteeing anything.  I'm not sure something like that belongs in the thrift transport implementation at all.  Given clients are asynchronous already, they should implement their own request timeouts and re-submit requests, and design their systems to have idempotent requests.

PR #1780, opened 2019-04-13
THRIFT-4834 CI error at Haskell: Failed to load interface for 'Network'

Jens-G:
…iled to load interface for `Network' is solved

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -413,6 +413,11 @@ AC_SUBST(CABAL)
 AC_SUBST(RUNHASKELL)
 AM_CONDITIONAL(WITH_HASKELL, [test "$have_haskell" = "yes"])
 
+dnl TODO remove after THRIFT-4834 CI error at Haskell: Failed to load interface for `Network' is solved
+have_haskell=no
+RUNHASKELL=true
+CABAL=true
+
 AX_THRIFT_LIB(go, [Go], yes)
 if test "$with_go" = "yes";  then
   AC_PATH_PROG([GO], [go])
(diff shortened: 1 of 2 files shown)


jeking3:
Perhaps we should announce on the user mailing list that we are deprecating it, and see if anybody wakes up.  I can't believe it's been broken this long.

Jens-G:
Posting the user list was actually a great idea.

cchalmers:
It looks like it doesn't work with the latest `network` (released in January). An easy fix would be to put an upper bound on it:

```
--- a/lib/hs/thrift.cabal
+++ b/lib/hs/thrift.cabal
@@ -42,7 +42,7 @@ Library
   Build-Depends:
     base >= 4, base < 5, containers, ghc-prim, attoparsec, binary, bytestring >= 0.10, base64-bytestring, hashable, HTTP, text, hspec-core > 2.4.0, unordered-containers >= 0.2.6, vector >= 0.10.12.2, QuickCheck >= 2.8.2, split
   if flag(network-uri)
-     build-depends: network-uri >= 2.6, network >= 2.6
+     build-depends: network-uri >= 2.6, network >= 2.6 && < 3.0
   else
      build-depends: network < 2.6
   Exposed-Modules:
```

Jens-G:
@cchalmers can you provide a pull request?

Jens-G:
@cchalmers : Thanks! I gave it a try ...

jeking3:
Seems like this also broke the windows build (we have Haskell on Windows in our CI) in another build (the Rust 1.34 PR)):

https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/23864846

Jens-G:
> Seems like this also broke the windows build (we have Haskell on Windows in our CI) in another build
I don't see the Connection between the two?

Jens-G:
At least this seems to solve *one* issue. Any objections against merging this? Jim?

Jens-G:
Thanks, now lets tackle the Rust issue.

PR #1782, opened 2019-04-17
THRIFT-4849: Do not Ignore InterruptedException

BELUGABEHR:
Do not ignore interrupted exceptions in TNonblockingServer.java


Jens-G:
@BELUGABEHR: any news here?

PR #1783, opened 2019-04-17
THRIFT-4851: Remove All Calls To printStackTrace

BELUGABEHR:
Remove All Calls To printStackTrace


Jens-G:
LGTM, not tested

Jens-G:
Just noticed:is there a specific reason why in some cases it is

    LoggerFactory.getLogger(TServerSocket.class.getName());

and in other cases just

    LoggerFactory.getLogger(TSaslTransport.class);

without "getName()"?

PR #1786, opened 2019-04-22
THRIFT-4846. Topologically sort structs before generation

toddlipcon:
This adds a toposort step prior to emitting generating code for structs.
For many languages (eg Java) this isn't necessary as there is no
"ordering" in the generated code. However, with C and C++, it's
important to define structs prior to their inclusion in another struct.

This also makes Thrift fail when generating C/C++ code with cyclic
references as the toposort is impossible.

A new test includes some structs defined in the "wrong" order to ensure
that the sort produces working code.

I manually tested a file with a true cycle in it, and verified the error
output.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_c_glib_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_c_glib_generator.cc
@@ -106,6 +106,10 @@ public:
   void init_generator() override;
   void close_generator() override;
 
+  bool strict_definition_order() const override {
+    return true;
+  }
+
   /* generation functions */
   void generate_typedef(t_typedef* ttypedef) override;
   void generate_enum(t_enum* tenum) override;
--- a/compiler/cpp/src/thrift/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_cpp_generator.cc
@@ -104,12 +104,14 @@ public:
   void init_generator() override;
   void close_generator() override;
 
-  void generate_consts(std::vector<t_const*> consts) override;
+  bool strict_definition_order() const override {
+    return true;
+  }
 
   /**
    * Program-level generation functions
    */
-
+  void generate_consts(std::vector<t_const*> consts) override;
   void generate_typedef(t_typedef* ttypedef) override;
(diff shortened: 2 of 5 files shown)


Jens-G:
What's blocking this?

toddlipcon:
Not sure. I took a look at the failed test runs for a few minutes and couldn't find anything that indicated issues with this patch, but it's a wall of text and I'm not familiar enough with the test suite to be able to diagnose what went wrong.

Jens-G:
I would suggest to start with the conflict :-)

kainjow:
@toddlipcon I took a look at one of the test outputs and saw this:

```
org.apache.thrift.TestUnsafeBinaries > testSubStructValidation FAILED
4046    junit.framework.AssertionFailedError
4047WARNING: An illegal reflective access operation has occurred
4048WARNING: Illegal reflective access by org.mockito.cglib.core.ReflectUtils$2 (file:/root/.gradle/caches/modules-2/files-2.1/org.mockito/mockito-all/1.9.5/79a8984096fc6591c1e3690e07d41be506356fa5/mockito-all-1.9.5.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
4049WARNING: Please consider reporting this to the maintainers of org.mockito.cglib.core.ReflectUtils$2
4050WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
4051WARNING: All illegal access operations will be denied in a future release
4052
4053org.apache.thrift.TestStruct > testSubStructValidation FAILED
4054    junit.framework.AssertionFailedError at TestStruct.java:383
4055
4056org.apache.thrift.TestReuse > testSubStructValidation FAILED
4057    junit.framework.AssertionFailedError
4058
4059191 tests completed, 3 failed
4060
4061> Task :test FAILED
4062
4063FAILURE: Build failed with an exception.
```

[testSubStructValidation](https://github.com/apache/thrift/blob/af7ecd6a2b15efe5c6b742cf4a9ccb31bcc1f362/lib/java/test/org/apache/thrift/TestStruct.java#L365) failed for Java and this tests StructA and StructB which have been changed in this PR.

PR #1787, opened 2019-04-24
Enables changing 'UserAgent'

pjc0247:
`.Net` does not allow modifying several `known-headers` using `customHeaders`. So, there's no way to change `User-Agent` with the current C# implementation unlike other languages can do.


Jens-G:
Ma I ask you to also file a JIRA ticket for this?

PR #1788, opened 2019-04-24
THRIFT-2059: Support for Enums in Python >= 3.4

constd:
Currently in Python, the generated enum classes for enum thrift objects, are not being used in deserialization (or serialization). This means that the string representations of enums are never used. Moving one step further than the patch contained in THRIFT-2059, this creates python IntEnum classes and retrieves the string representation on deserialization while storing the integer representation on serialization.

- [NO] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [YES] Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?  (not required for trivial changes)
- [YES] Did you squash your changes to a single commit?
- [YES] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"


kainjow:
@constd are you interested in reviving this work? Tests would need to be updated.

One problem now is since the enums aren't Python 3 enums, if you add type annotations and use mypy to check types, it won't detect these correctly.

constd:
> @constd are you interested in reviving this work? Tests would need to be updated.
> 
> One problem now is since the enums aren't Python 3 enums, if you add type annotations and use mypy to check types, it won't detect these correctly, so I think this is a good start.

@kainjow Apologies for my delayed response; yes, I can revive this! Do you know what changed in tests?

honorabel:
Would love to see this work revived! As we've moved fully to python 3, we have a ton of interest in leveraging type annotations for type linting and better code safety.  Having the thrift-generated python be compatible w/ typing would be a huge win :)

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -278,6 +280,7 @@ class t_py_generator : public t_generator {
    * True if we should generate new-style classes.
    */
   bool gen_newstyle_;
+  bool gen_enum_;
```
kainjow, line 283: this needs be initialized like the other variables to false.

kainjow:
I'm recreating the PR at #2489 so we can move forward.

PR #1790, opened 2019-04-28
THRIFT-4854 oneway calls do not work over HTTP

Jens-G:
Client: netstd
Patch: Jens Geyer


jeking3:
This is a generic issue across many of the language implementations.

PR #1802, opened 2019-05-17
THRIFT-4868: Golang: Fix compilation for optional sets with default values

johnboiles:
See [THRIFT-4868](https://issues.apache.org/jira/browse/THRIFT-4868) for details. In short, Golang compilation is broken for optional sets with default values. For example:

`1: optional set<string> with_default = [ "test" ]`

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -3237,8 +3237,12 @@ void t_go_generator::generate_serialize_container(ostream& out,
     t_set* tset = (t_set*)ttype;
     out << indent() << "for i := 0; i<len(" << prefix << "); i++ {" << endl;
     out << indent() << "  for j := i+1; j<len(" << prefix << "); j++ {" << endl;
-    out << indent() << "    if reflect.DeepEqual(" << prefix << "[i]," << prefix << "[j]) { " << endl;
-    out << indent() << "      return thrift.PrependError(\"\", fmt.Errorf(\"%T error writing set field: slice is not unique\", " << prefix << "[i]))" << endl;
+    string wrapped_prefix = prefix;
+    if (pointer_field) {
+      wrapped_prefix = "(" + prefix + ")";
+    }
+    out << indent() << "    if reflect.DeepEqual(" << wrapped_prefix << "[i]," << wrapped_prefix << "[j]) { " << endl;
+    out << indent() << "      return thrift.PrependError(\"\", fmt.Errorf(\"%T error writing set field: slice is not unique\", " << wrapped_prefix << "[i]))" << endl;
     out << indent() << "    }" << endl;
     out << indent() << "  }" << endl;
     out << indent() << "}" << endl;
--- a/test/ThriftTest.thrift
+++ b/test/ThriftTest.thrift
@@ -409,3 +409,7 @@ struct StructB {
   1: optional StructA aa;
   2: required StructA ab;
 }
+
+struct OptionalSetDefaultTest {
+  1: optional set<string> with_default = [ "test" ]
+}


dcelasun:
I can merge this as-is, but I'm thinking it might be better to fix the original error (we shouldn't have a pointer to a slice) even if it's a BC break (we just need to [mention it](https://github.com/apache/thrift/blob/c035eca3c672511779440bd0779ce5bc93d5c327/CHANGES.md#breaking-changes)). It's better for maintainability and will be less surprising for anyone who runs into it.

Thoughts?

johnboiles:
I definitely agree we shouldn’t have a pointer to a slice (or a pointer to a map). Do you think it will be as simple as never returning true from is_pointer for those types? I can try but probably not until next week.

If we’re making breaking changes anyways would it make more sense to use a map type as a set? E.g `map[string]struct{}` for a set of strings? That’s a pretty common way to represent sets in Go.

dcelasun:
> Do you think it will be as simple as never returning true from is_pointer for those types?
Yes, I think that should be enough.

> I can try but probably not until next week.
No rush.

> If we’re making breaking changes anyways would it make more sense to use a map type as a set? E.g `map[string]struct{}` for a set of strings? That’s a pretty common way to represent sets in Go.

Actually, it was originally like that but I changed it to slices because map keys in Go must be [comparable](https://golang.org/ref/spec#Comparison_operators) and there are lots of valid Thrift sets where the values are not valid Go map keys. Using a slice and enforcing uniqueness via `reflect.DeepEqual` provides a better user experience.

johnboiles:
Good to know. Thanks for the context!

johnboiles:
I've gotten started on this but it is more complex than I expected. I can see the light at the end of the tunnel though.

It looks like the `is_pointer_field` method gets also used in different parts of the code to indicate whether a value is optional, and also whether a value needs a `*` or `&` in front of it. I'm going to split out a new `need_pointer_redirect` method to indicate whether the value needs a `*` or `&` in front of it.

|                       | map | opt map | slice | opt slice | string | opt string | int/float | opt int/float | struct | opt struct |
|-----------------------|-----|---------|-------|-----------|--------|------------|-----------|---------------|--------|------------|
| is_pointer_field      | Y   | Y       | Y     | Y         | N      | Y          | N         | Y             | N      | Y          |
| need_pointer_redirect | N   | N       | N     | N         | N      | Y          | N         | Y             | N      | Y          |

johnboiles:
This rabbit hole keeps getting deeper. I'm trying to get a handle on all the special cases with a lookup table that includes optional/required, default values, cpp.ref, types, and how all those factors affect the generated code. It's very complex! This sheet is an attempt at highlighting how things should work when I make the change to not pointers to map/slice types (since they're always pointers).

https://docs.google.com/spreadsheets/d/1Q90gi8zylkf1XTDJiY5ViMCjkTz69PYuk0uOjcI1Yc8/edit?usp=sharing

image

I've also found [THRIFT-2451](https://issues.apache.org/jira/browse/THRIFT-2451) and [THRIFT-2429](https://issues.apache.org/jira/browse/THRIFT-2429) to be helpful reading to get more historical context.

I have a few questions / proposals:
- Can we drop `cpp.ref` support? It doesn't seem to [do anything](https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/generate/t_go_generator.cc#L375) except affect whether native (int/string/bool/double) types are pointers or not. In the case of map/slice it currently also forces those to be pointers pointer types when they don't have a default value, but since maps and slices are already pointers, I don't think this has any practical effect.
- @dcelasun what do you think about merging this change first to fix compilation without making an API-incompatible change. Happy to keep working on the bigger change to not use pointers to maps and slices.

dcelasun:
> Can we drop cpp.ref support?

I think so, yes. @jeking3 any thoughts?

> what do you think about merging this change first to fix compilation without making an API-incompatible change?

Sounds reasonable, given how the scope keeps growing. I'll merge it later today.

dcelasun:
Merged. @johnboiles please open a new PR for the rest. Even if it's not ready, I could provide feedback.

johnboiles:
Thanks @dcelasun will do

dcelasun:
Hey @johnboiles, would you be willing to continue this work? Your lookup table is very useful and getting rid of both `cpp.ref` and pointer-to-pointer cases (maps & slices) would be great, especially since 0.13 has just been released, making this the best possible time for backwards incompatible changes.

PR #1808, opened 2019-06-02
THRIFT-4879 general performance improvements for netstd library

Jens-G:
Client: netstd
Patch: Jens Geyer


Jens-G:
@jeking3 The ruby thing is clearly unrelated, but what really sucks is the mingw update error at AppVeyor. Is there anything I can do about that? Thoughts?

PR #1810, opened 2019-06-12
THRIFT-4612: THeader support in go library

fishy:
Client: go

Implement THeaderTransport and THeaderProtocol, with support of:

* clients:
  - headers
  - framedBinary
  - unframedBinary
  - framedCompact
  - unframedCompact
* transforms:
  - none
  - zlib
* info types:
  - key_value
* wrapped protocols:
  - TBinary
  - TCompact

The support list is in general on par of the THeader implementation in
the python library.

The cross-test passes, except ones related to cpp/nodejs http transport,
which were also failing for non-theader protocols.

This change also fixes two bugs:

1. A small issue in test/go/src/bin/testserver/main.go
2. A bug in TFrameTransport go implementation

  


fishy:
@dcelasun tagging you because you expressed interest in the [JIRA ticket](https://issues.apache.org/jira/browse/THRIFT-4612) :)

fishy:
The failures on travis are due to transient network failures (to pull build dependencies). The corresponding builds are passing on https://travis-ci.com/fishy/thrift/builds/115286920

dcelasun:
Thanks for the PR! From a quick look, this seems really good. Nice catch with the framed transport bug as well.

I'll probably have some time next week to do a proper review.

Jens-G:
I cringe a bit at the terminology used above. We have protocols, layered and endpoint transports. We don't have "sub-protocols" or "transforms"

fishy:
@Jens-G "transforms" are defined in header format: https://github.com/apache/thrift/blob/master/doc/specs/HeaderFormat.md

I used the term "subprotocol" to distinguish from "protocol" in THeaderProtocol. I'm open to suggestions.

lib/go/thrift/application_exception.go
```diff
@@ -28,6 +28,9 @@ const (
 	MISSING_RESULT                 = 5
 	INTERNAL_ERROR                 = 6
 	PROTOCOL_ERROR                 = 7
+	INVALID_TRANSFORM              = 8
```
jeking3: Do these match other language implementations?

lib/go/thrift/simple_server.go
```diff
@@ -187,12 +187,24 @@ func (p *TSimpleServer) processRequests(client TTransport) error {
...
-	}
 	inputProtocol := p.inputProtocolFactory.GetProtocol(inputTransport)
-	outputProtocol := p.outputProtocolFactory.GetProtocol(outputTransport)
+	var outputTransport TTransport
+	var outputProtocol TProtocol
+
+	// for THeaderProtocol, we must use the same protocol instance for
+	// input and output so that the response is in the same dialect that
+	// the server detected the request was in.
+	if _, ok := inputProtocol.(*THeaderProtocol); ok {
```
jeking3: Hmm, interesting, usually the server layer has no special cases for the protocol.

lib/go/thrift/application_exception.go
fishy: Yes:

- python: https://github.com/apache/thrift/blob/823474ec89355f72d3f0720ae5dacc2036d41c03/lib/py/src/Thrift.py#L118-L120
- cpp: https://github.com/apache/thrift/blob/823474ec89355f72d3f0720ae5dacc2036d41c03/lib/cpp/src/thrift/TApplicationException.h#L46-L48

lib/go/thrift/simple_server.go
fishy: Without this logic the outputProtocol won't know it should use a non-default protocol detected on inputProtocol.

- python: https://github.com/apache/thrift/blob/823474ec89355f72d3f0720ae5dacc2036d41c03/lib/py/src/server/TServer.py#L189-L191
- cpp: https://github.com/apache/thrift/blob/823474ec89355f72d3f0720ae5dacc2036d41c03/lib/cpp/src/thrift/server/TNonblockingServer.cpp#L395-L403

fishy:
I renamed `Subprotocol*` into `Protocol*` based on the feedbacks (and also in comments, use "wrapped protocol" to refer to them instead of "subprotocol"). I didn't rename `Transform*` because that name is in header format spec doc.

The commit message was also updated accordingly.

PR #1811, opened 2019-06-13
THRIFT-4797: Golang: Fix import collisions

johnboiles:
See [THRIFT-4797](https://issues.apache.org/jira/browse/THRIFT-4797) for details. In short, Golang compilation fails when a Thrift file includes two files that have the same final namespace component (e.g. `namespace go packagea.common` and `namespace go packageb.common`)

Somewhat based on [fbthrift's implementation](https://github.com/facebook/fbthrift/commit/b6754fe7a6a6e00ce288f14cd4e88cfa3d0cc3e4#diff-803353e952116d4e88caab467a62f35b) though their implementation simply always adds an alias with a unique numerical prefix.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -250,6 +250,7 @@ public:
   std::string go_imports_end();
   std::string render_includes(bool consts);
   std::string render_included_programs(string& unused_protection);
+  std::string render_program_import(const t_program* program, string& unused_protection);
   std::string render_system_packages(std::vector<string> &system_packages);
   std::string render_import_protection();
   std::string render_fastbinary_includes();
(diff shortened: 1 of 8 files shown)


compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -809,17 +812,35 @@ string t_go_generator::render_includes(bool consts) {
...
+        // packageacommon as the alias for packagea/common instead of common=). But just
+        // appending an integer is much simpler code
+        value = tmp(value);
+      }
+      package_identifiers_set_.insert(value);
+      it = package_identifiers_.emplace(go_module, std::move(value)).first;
+    }
+    auto const& package_identifier = it->second;
+
+    result += "\t" + package_identifier + " \"" + gen_package_prefix_ +
```
johnboiles: Actually it'd be better if I didn't include the alias if it's not necessary. I'll go change that.

johnboiles:
Just found another case to handle here. I should also detect collisions with the golang packages that get [imported](https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/generate/t_go_generator.cc#L875) like `bytes` `context` `reflect`

dcelasun:
For stdlib collisions like `context` etc, it would be great if the import list was a vector instead of a hardcoded string so we could reuse it everywhere.

johnboiles:
@dcelasun I made it use a vector for system imports can you take a look? I’ll squash it together if/when it looks good to you!

johnboiles:
@dcelasun squashed and ready to merge!

johnboiles:
Actually hold on merging for a bit. I _may_ have found another edge case

johnboiles:
Yep, confirmed. If the package has a service, and that service’s last namespace component conflicts with a previously imported package, it won’t get aliased. Hopefully I can get a chance to fix tomorrow

johnboiles:
@dcelasun one more commit to review to fix the issue with the remote package import collisions! Hopefully this is it

dcelasun:
LGTM. Tried testing a few different scenarios locally, everything works fine. Thanks for the PR!

PR #1813, opened 2019-06-14
CPP generator test, catching exceptions by ref instead.

devnexen:
Recommended to catch exception by references.

### DIFF
--- a/lib/cpp/test/processor/EventLog.cpp
+++ b/lib/cpp/test/processor/EventLog.cpp
@@ -98,7 +98,7 @@ Event EventLog::waitForEvent(int64_t timeout) {
     while (events_.empty()) {
       monitor_.wait(timeout);
     }
-  } catch (TimedOutException ex) {
+  } catch (const TimedOutException &) {
     return Event(ET_LOG_END, 0, 0, "");
   }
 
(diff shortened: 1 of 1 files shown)


lib/cpp/test/processor/EventLog.cpp
```diff
@@ -98,7 +98,7 @@ Event EventLog::waitForEvent(int64_t timeout) {
     while (events_.empty()) {
       monitor_.wait(timeout);
     }
-  } catch (TimedOutException ex) {
+  } catch (TimedOutException &ex) {
```
jeking3:
```suggestion
  } catch (const TimedOutException &) {
```

lib/cpp/test/processor/EventLog.cpp
```diff
@@ -119,7 +119,7 @@ Event EventLog::waitForConnEvent(uint32_t connId, int64_t timeout) {
       while (it == events_.end()) {
         monitor_.wait(timeout);
       }
-    } catch (TimedOutException ex) {
+    } catch (TimedOutException &ex) {
```
jeking3:
```suggestion
    } catch (const TimedOutException &) {
```

PR #1814, opened 2019-06-17
THRIFT-4887 Clear redundant data in transport after each request to a…

aqingsao:
…void OOM

Client: Java
Patch: Xiaoqing Zhang

Add a safeClear() method in TTransport and it calls clear() method, implement clear() method in TSocket/TFramedTransport etc.

TServiceClient calls TTransport.safeClear() after each request
[Y] Did you create an Apache Jira ticket? (not required for trivial changes)
[Y] Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"? (not required for trivial changes)
[Y] Did you squash your changes to a single commit?
[Y] Did you do your best to avoid breaking changes? If one was needed, did you label the Jira ticket with "Breaking-Change"

ps: a detailed explanation for this bug and how to reproduce it could be found at: https://github.com/aqingsao/thrift-oom
ps:  a load test of 100 concurrencies has proven that the fix works.
ps: This patch only fixes Java, whereas this bug exists in nearly all languages.

### DIFF
--- a/lib/java/src/org/apache/thrift/TServiceClient.java
+++ b/lib/java/src/org/apache/thrift/TServiceClient.java
@@ -85,7 +85,11 @@ public abstract class TServiceClient {
       throw new TApplicationException(TApplicationException.BAD_SEQUENCE_ID,
           String.format("%s failed: out of sequence response: expected %d but got %d", methodName, seqid_, msg.seqid));
     }
-    result.read(iprot_);
+    try {
+      result.read(iprot_);
+    } finally {
+      iprot_.getTransport().safeClear();
+    }
     iprot_.readMessageEnd();
   }
 }
--- a/lib/java/src/org/apache/thrift/TDeserializer.java
+++ b/lib/java/src/org/apache/thrift/TDeserializer.java
@@ -80,7 +80,7 @@ public class TDeserializer {
       trans_.reset(bytes, offset, length);
       base.read(protocol_);
     } finally {
-      trans_.clear();
+      trans_.safeClear();
       protocol_.reset();
     }
   }
@@ -121,7 +121,7 @@ public class TDeserializer {
     } catch (Exception e) {
       throw new TException(e);
     } finally {
-      trans_.clear();
+      trans_.safeClear();
       protocol_.reset();
     }
   }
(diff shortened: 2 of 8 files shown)


Jens-G:
Did you read what I wrote in the ticket?

Jens-G:
Ping

aqingsao:
> Ping
Sorry for the late reply. I've replied a comment in the ticket, and here is a digest:
1. It's fine to reinitialize protocol/transport stack from scratch, but might not close sockets as subsequent requests(maybe another thrift method that works fine) could re-use them. Opening a lot sockets consume a lot resources and could easily lead to Socket Exceptions.
2. If this is an issue, it exist in both client side and server side, my fix is in client side. I need to dive more into source code to see how to fix it in server side, probably Server.Execute() as you suggest.
Any more suggestion is welcome.

lib/java/src/org/apache/thrift/TServiceClient.java
```diff
@@ -85,7 +85,11 @@ protected void receiveBase(TBase<?,?> result, String methodName) throws TExcepti
       throw new TApplicationException(TApplicationException.BAD_SEQUENCE_ID,
           String.format("%s failed: out of sequence response: expected %d but got %d", methodName, seqid_, msg.seqid));
     }
-    result.read(iprot_);
+    try {
```
jeking3: This change is incorrect.  If read throws, the connection is closed.  It means the client sent part of a response but didn't get all of it out, and disconnected, or perhaps in your case, the type of a previously existing field was changed between releases, so the deserializer was looking for one field but got another.  In this case, the deserializer is first looking for a list length, then for each entry in the list it is looking for a string length and a string.  By changing field #2 to a string, it has corrupted the stream in an incompatible way.

jeking3:
I'm going to close this; I don't want it to mistakenly get merged.

lib/java/src/org/apache/thrift/TServiceClient.java
aqingsao: "If read throws, the connection is closed"——Have to say that is is not true and misleading, in most cases people use keep-alive mode in thrift and a connection is shared between multi requests, the connection is still active even if read throws an exception.

PR #1817, opened 2019-06-21
Update language levels, fix erlang and rust in CI

jeking3:
- Include rebar3 in docker image so erlang builds in CI
- Include the correct path in docker image so rust builds in CI
- Updated common lisp (sbcl) to 1.5.3 in docker image
- Updated dlang to 2.086.1 in docker image
- Updated dart to 2.4.0 in docker image
- Updated erlang to 22.0 in docker image
- Updated openjdk to 11.0.3 in docker image
- Updated node.js to 10.16 in docker image
- Updated rust to 1.35.0 in docker image

  


allengeorge:
@jeking3 Sorry for the delay. I'm trying to fix the rust tutorial build failures right now, and I'll try to fix the rust build specifically so that we use a pinned version of rustc as opposed to whatever's on stable at the time the docker container is built.

jeking3:
Issue reproduced locally (gotta love docker containers)... looking into it.

jeking3:
Now the erlang feature test for TLS is breaking.  See what happens when a language rots because it's not being built?

jeking3:
So erlang 22 includes erlang ssl 9.3 which disables TLS 1.0 and 1.1 by default, but TLSv1.2 doesn't seem to be working either.  I'll have to back down to erlang 21 for now unless someone with some erlang knowledge can fix the TLS feature test.

jeking3:
The dlang unit test async_test is failing for ssl and hanging causing those builds to fail.  This seems to be related to the newer version of dlang in use.  I had upgraded from 2.083 to 2.086.  I'll try to bisect it.

jeking3:
One last thing to take care of, the Xenial Dockerfile isn't working on install of cabal.

jeking3:
The issue with the Xenial build is (hopefully) a server error on the haskell server side.  As such, I'm going to merge this.

PR #1821, opened 2019-06-24
THRIFT-4896 - prevent cpp and c_glib include directories from leaking into other targets

kainjow:
`configure.ac` uses `AC_CONFIG_HEADERS` for cpp and c_glib to generate config.h files. However automake by default will add the parent directory of the config.h files to the `DEFAULT_INCLUDES` variable, which appears to be used in all other projects. This leaks the include paths to other targets that probably don't want them. Currently, it's not an issue, but I am looking at adding Objective-C (Cocoa) tests, and these classes use the same Thrift library header file names as the C++ library, which come first in the include path. This ends up trying to compile C++ for Objective-C, which fails.

To fix this, we can turn on the `nostdinc` option, which disables the default include directories. This requires adding back the include directories to some targets.

I tested by compiling:

    --with-c_glib \
    --with-cpp \
    --with-libevent \
    --with-lua \

### DIFF
--- a/compiler/cpp/Makefile.am
+++ b/compiler/cpp/Makefile.am
@@ -21,7 +21,7 @@
 # Please see doc/old-thrift-license.txt in the Thrift distribution for
 # details.
 
-AUTOMAKE_OPTIONS = subdir-objects
+AUTOMAKE_OPTIONS = subdir-objects nostdinc
 
 SUBDIRS = src .
 if WITH_TESTS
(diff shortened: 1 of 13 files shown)


jeking3:
You can use the thrift docker container to leverage the same build environment the CI system uses to test with, in an environment where most of the languages are installed.  See https://github.com/apache/thrift/tree/master/build/docker

kainjow:
@jeking3 that seemed to have fixed it. Once of the Travis jobs failed for seemingly unrelated issues.

PR #1823, opened 2019-06-26
THRIFT-4898 Pipe write operations across a network are limited to 65,…

Jens-G:
…535 bytes per write.

Including improved test cases for netstd and Delphi (acted as the test counterpart)


jeking3:
Is it possible to add a Delphi runtime to the Travis CI image so we can run the cross test with these changes?

PR #1660, opened 2018-12-31
THRIFT-4547: add swift to bionic build image and run unit tests

jeking3:
This is a prerequisite to writing and running swift clients and servers in the cross test.

PR #1827, opened 2019-07-03
THRIFT-4902: Swift 4 and 5 compatibility

Coeur:

This pull request:
- keeps compatibility with Swift 3
- adds support for Swift 4.0, 4.2, 5.0 and 5.1
- fixes the podspec:
  - the pod name must match the filename in current versions of CocoaPods
  - `requires_arc  = true` is already the default in CocoaPods
- `mutating func removeAtIndex` is removed because it wasn't implemented to begin with: it was just calling itself in an infinite loop.

### DIFF
--- a/lib/swift/Sources/TMap.swift
+++ b/lib/swift/Sources/TMap.swift
@@ -36,12 +36,6 @@ public struct TMap<Key : TSerializable & Hashable, Value : TSerializable>: Colle
     return storage.updateValue(value, forKey: key)
   }
   
-  /*
-  public mutating func removeAtIndex(_ index: DictionaryIndex<Key, Value>) -> (Key, Value) {
-    return storage.removeAtIndex(index)
-  }
-  */
-  
   public mutating func removeValueForKey(_ key: Key) -> Value? {
     return storage.removeValue(forKey: key)
   }
--- a/lib/swift/Sources/TMap.swift
+++ b/lib/swift/Sources/TMap.swift
@@ -21,7 +21,7 @@ public struct TMap<Key : TSerializable & Hashable, Value : TSerializable>: Colle
   public typealias Storage = Dictionary<Key, Value>
   public typealias Element = Storage.Element
   public typealias Index = Storage.Index
-  public typealias IndexDistance = Storage.IndexDistance
+  public typealias IndexDistance = Int
   public typealias Indices = Storage.Indices
   public typealias SubSequence = Storage.SubSequence
   internal var storage = Storage()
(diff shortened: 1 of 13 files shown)


Coeur:
> The changes look okay but without cross testing how do we know they work?

From my reputation. 😉

More seriously, with CocoaPods, we have a linter, so it can build the lib for all platforms (ios/macos here) and different Swift versions. Which means that that we can validate this pull request compatibility by running:
```
pod lib lint --allow-warnings --swift-version=3
pod lib lint --allow-warnings --swift-version=4.0
pod lib lint --allow-warnings --swift-version=4.2
pod lib lint --allow-warnings --swift-version=5.0
```
I've now added a travis job for that.

jeking3:
But if we only lint we're not checking I/O against other languages, so adding a cross test harness would be a better choice.

Coeur:
I understand, but that's not the scope of this PR: I'm only adding compatibility with current Swift syntax here. Adding cross test support is the purpose of a separate Jira (THRIFT-4547, #1660).

And I guess that it will be easier to add tests now that the code actually builds for all Swift versions.

jeking3:
Travis CI has Swift 4.2 installed, so if you can provide the cross test changes as a separate PR I would appreciate seeing it cross tested.  I'll watch for build completion.

lib/swift/Sources/TMap.swift
```diff
@@ -33,12 +33,14 @@ public struct TMap<Key : TSerializable & Hashable, Value : TSerializable>: Colle
   }
   
   public mutating func updateValue(_ value: Value, forKey key: Key) -> Value? {
-    return updateValue(value, forKey: key)
+    return storage.updateValue(value, forKey: key)
   }
   
+  /*
```
kainjow: why is this commented out?

Coeur: Maybe, yes.

jeking3:
Rebase on master please.  The only known build issue (after what I just merged) in on Xenial, which doesn't affect this since Swift isn't installed on that image.  Thanks.

Coeur:
@jeking3 test passed

arpwal:
@Coeur seems like master is broken for Swift 5. Could you help? :)

janosvitok:
@arpwal Can you be more specific? What kind of error do you see? Does any of #2073,  #2069, #2059, #2051, #2047, #2038, #1997 help with the problem?

Coeur:
@arpwal Sorry, I'm not available to help currently. Please check janosvitok pull requests, they are likely to help.

PR #1829, opened 2019-07-05
Improve CMake OpenSSL use

kainjow:
The CMake file says OpenSSL is only used for C++ but it's actually used for C GLib too, so move it out from under C++ into a combined C++/GLib check. Update the output to show the status for building OpenSSL including the version. Remove additional find_package() call for OpenSSL since this is already done in DefineOptions.cmake. Lastly, fix test_client missing OpenSSL include dirs and libraries.

- [ ] Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?  (not required for trivial changes)


build/cmake/DefineOptions.cmake
```diff
@@ -53,6 +53,11 @@ endif ()
 CMAKE_DEPENDENT_OPTION(BUILD_AS3 "Build as3 library" ON
                        "BUILD_LIBRARIES;WITH_AS3;HAVE_COMPC" OFF)
 
+# OpenSSL (used by C++ and C GLib)
```
jeking3: Should this be inside:

    if (WITH_CPP AND WITH_C_GLIB)

or something like that?  If you are building just java, for example, you don't need it.

build/cmake/DefineOptions.cmake
```diff
@@ -160,6 +162,8 @@ message(STATUS "  Build as3 library:                          ${BUILD_AS3}")
 MESSAGE_DEP(WITH_AS3 "Disabled by WITH_AS3=OFF")
 MESSAGE_DEP(HAVE_COMPC "Adobe Flex compc was not found - did you set env var FLEX_HOME?")
 message(STATUS)
+message(STATUS "  Build with OpenSSL:                         ${WITH_OPENSSL}")
```
jeking3: Including the OpenSSL version would be nice here.

kainjow: done

PR #1831, opened 2019-07-14
THRIFT-4908: remove reader&writer in Golang's TBinaryProtocol

bengol:
Client: go

  It seems the reader and writer in TBinaryProtocol is not necessary, because this class is serialize/deserialize the struct and write/read the data stream via trans, `reader` & `writer` here are duplicated with `trans` in this struct

- [Y ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [Y ] Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?  (not required for trivial changes)
- [Y ] Did you squash your changes to a single commit?  (not required, but preferred)
- [ Y] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [Y ] If your change does not involve any code, add ` [skip ci]` at the end of your pull request to free up build resources.

### DIFF
--- a/lib/go/thrift/binary_protocol.go
+++ b/lib/go/thrift/binary_protocol.go
@@ -32,8 +32,6 @@ import (
 type TBinaryProtocol struct {
 	trans         TRichTransport
 	origTransport TTransport
-	reader        io.Reader
-	writer        io.Writer
 	strictRead    bool
 	strictWrite   bool
 	buffer        [64]byte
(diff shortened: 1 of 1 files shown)


jeking3:
The original design of Thrift allows for different protocol and transport in each direction, which is why you see reader and writer transports and protocols.  In practice I'm not sure it is ever used this way.  The go code was likely modeled after the Java or C++ protocol code which still maintains the inputProtocol, outputProtocol, inputTransport, outputTransport semantics.  It looks like the constructors were simplified when go was added, so I see nothing wrong with this further simplification.  The appveyor issue with concurrency_test seems to be happening enough we should look into it again but it isn't related to these changes - that is a C++ concern.

dcelasun:
Thanks for the PR. This is indeed harmless, so I'm gonna go ahead and merge this.

PR #1835, opened 2019-07-22
THRIFT-4002: Make generated exception classes immutable by default

elprans:
Currently, the generated exception classes are not hashable under Python 3
because of the generated `__eq__` method.  Exception objects are generally
expected to be hashable by the Python standard library. Post-construction
mutation of an exception object seems like a very unlikely case, so enable
hashing for all exceptions by making them immutable by default.  This also
adds a way to opt-out of immutability by setting the `python.immutable`
annotation to `"false"`.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -65,6 +65,7 @@ public:
     coding_ = "";
     gen_dynbaseclass_ = "";
     gen_dynbaseclass_exc_ = "";
+    gen_dynbaseclass_frozen_exc_ = "";
     gen_dynbaseclass_frozen_ = "";
     import_dynbase_ = "";
     package_prefix_ = "";
@@ -94,8 +95,11 @@ public:
         if( gen_dynbaseclass_exc_.empty()) {
           gen_dynbaseclass_exc_ = "TExceptionBase";
         }
+        if( gen_dynbaseclass_frozen_exc_.empty()) {
+          gen_dynbaseclass_frozen_exc_ = "TFrozenExceptionBase";
+        }
         if( import_dynbase_.empty()) {
-          import_dynbase_ = "from thrift.protocol.TBase import TBase, TFrozenBase, TExceptionBase, TTransport\n";
+          import_dynbase_ = "from thrift.protocol.TBase import TBase, TFrozenBase, TExceptionBase, TFrozenExceptionBase, TTransport\n";
         }
       } else if( iter->first.compare("dynbase") == 0) {
         gen_dynbase_ = true;
@@ -104,6 +108,8 @@ public:
         gen_dynbaseclass_frozen_ = (iter->second);
       } else if( iter->first.compare("dynexc") == 0) {
         gen_dynbaseclass_exc_ = (iter->second);
+      } else if( iter->first.compare("dynfrozenexc") == 0) {
+        gen_dynbaseclass_frozen_exc_ = (iter->second);
       } else if( iter->first.compare("dynimport") == 0) {
         gen_dynbase_ = true;
         import_dynbase_ = (iter->second);
@@ -269,7 +275,16 @@ public:
   }
 
   static bool is_immutable(t_type* ttype) {
-    return ttype->annotations_.find("python.immutable") != ttype->annotations_.end();
+    std::map<std::string, std::string>::iterator it = ttype->annotations_.find("python.immutable");
+
+    if (it == ttype->annotations_.end()) {
+      // Exceptions are immutable by default.
+      return ttype->is_xception();
+    } else if (it->second == "false") {
+      return false;
+    } else {
+      return true;
+    }
   }
 
 private:
@@ -288,6 +303,7 @@ private:
   std::string gen_dynbaseclass_;
   std::string gen_dynbaseclass_frozen_;
   std::string gen_dynbaseclass_exc_;
+  std::string gen_dynbaseclass_frozen_exc_;
 
   std::string import_dynbase_;
 
@@ -742,7 +758,11 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
   out << endl << endl << "class " << tstruct->get_name();
   if (is_exception) {
     if (gen_dynamic_) {
-      out << "(" << gen_dynbaseclass_exc_ << ")";
+      if (is_immutable(tstruct)) {
+        out << "(" << gen_dynbaseclass_frozen_exc_ << ")";
+      } else {
+        out << "(" << gen_dynbaseclass_exc_ << ")";
+      }
     } else {
       out << "(TException)";
     }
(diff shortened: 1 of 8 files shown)


compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -269,7 +275,16 @@ class t_py_generator : public t_generator {
   }
 
   static bool is_immutable(t_type* ttype) {
-    return ttype->annotations_.find("python.immutable") != ttype->annotations_.end();
+    std::map<std::string, std::string>::iterator it = ttype->annotations_.find("python.immutable");
+
+    if (it == ttype->annotations_.end()) {
```
jeking3: I'm confused as to how ttype->annotations_ would get populated with the correct data.  Below in DebugProtoTest.thrift containing MutableException, it says (python.immutable = "false").  Wouldn't that populate ttype->annotations with an entry where the key is "python.immutable" and the value is "false"?  The code here assumes if an entry is found then it is immutable and does not check the value.

Also if someone is going to set an attribute like "python.immutable" to false, that would mean the default without an attribute means it is immutable.  That is what the code does here.  So wouldn't it be better to look for an attribute of "python.mutable" = <any value> and if it is there, it is not immutable?

elprans: > The code here assumes if an entry is found then it is immutable and does not check the value.

But it does: `else if (it->second == "false")`

> Also if someone is going to set an attribute like "python.immutable" to false, that would mean the default without an attribute means it is immutable.

Only for exceptions.

> So wouldn't it be better to look for an attribute of "python.mutable"

Those annotations are already "boolean" flags, so I see no benefit of adding a reverse spelling as opposed to setting a boolean to `"false"`

cjbarth:
This is still important to me. Is there anything that I can help with to move this along?

cjbarth:
This is still important to me. Is there anything that I can help with to move this along?

dcelasun:
LGTM, but I don't personally use the Python library, so I'll leave this open for comments from other maintainers for 24 hours before merging.

@elprans can you please rebase from master?

dcelasun:
Merged as 45a9827 and b40f5c2.

dcelasun:
Reverted due to failing tests. https://travis-ci.org/apache/thrift/builds/623369529

cjbarth:
@elprans One of these failures is very easy to fix:

`./lib/py/src/Thrift.py:20:1: F401 'sys' imported but unused`

The other is a little more complex because it depends on the mutability of the exceptions. So, does anyone know if this test is accurate in what it tests? I ask because it appears that the two tests ([here](https://github.com/apache/thrift/blob/6e023df1ded255dda00eb4c041c201e66c8d1fbc/test/py.tornado/test_suite.py#L83) and [here](https://github.com/apache/thrift/blob/6e023df1ded255dda00eb4c041c201e66c8d1fbc/test/py.twisted/test_suite.py#L77)) are the ones modifying the exception, not any underlying Thrift code. @dcelasun do you know anything about this?

```
./test_suite.py
..ERROR:root:Unexpected exception in handler
  File "/thrift/src/test/py.tornado/gen-py.tornado/ThriftTest/ThriftTest.py", line 1692, in process_testException
    yield gen.maybe_future(self._handler.testException(args.arg))
  File "./test_suite.py", line 86, in testException
    x.errorCode = 1001
  File "/thrift/src/test/py.tornado/gen-py.tornado/ThriftTest/ttypes.py", line 839, in __setattr__
E...ERROR:root:Exception in oneway handler
  File "/thrift/src/test/py.tornado/gen-py.tornado/ThriftTest/ThriftTest.py", line 1744, in process_testOneway
    yield gen.maybe_future(self._handler.testOneway(args.secondsToSleep))
  File "./test_suite.py", line 102, in testOneway
    raise Exception('testing exception in oneway method')
Exception: testing exception in oneway method
....
======================================================================
ERROR: test_exception (__main__.ThriftTestCase)
  File "/usr/lib/python2.7/dist-packages/tornado/testing.py", line 136, in __call__
    result = self.orig_method(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/tornado/testing.py", line 529, in post_coroutine
    timeout=timeout)
  File "/usr/lib/python2.7/dist-packages/tornado/ioloop.py", line 458, in run_sync
    return future_cell[0].result()
  File "/usr/lib/python2.7/dist-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib/python2.7/dist-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "./test_suite.py", line 211, in test_exception
    yield self.client.testException('Xception')
  File "/usr/lib/python2.7/dist-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/usr/lib/python2.7/dist-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
TApplicationException: Internal error```

**Comment by dcelasun** [comment]
>  @dcelasun do you know anything about this?

Unfortunately no, I'm not really familiar with the Python library. Judging by `git blame`, those tests are 5-7 years old and if they are no longer relevant, I'd be fine with removing them.

But first I need confirmation from someone familiar with the code.

**Comment by cjbarth** [comment]
@bufferoverflow It appears that you and @cpiro may have originally written these tests. Would either of you happen to know why this test modifies an exception? Or, perhaps shed light on what is being tested here (maybe we don't even need theses tests anymore)?

**Comment by cpiro** [comment]
As you've guessed, I ripped off that test code (and large chunks of the rest of `py.tornado`) from `py.twisted`. These tests check whether exceptions have the proper behavior; they should stay in. In both cases, it should be fine to construct the `Xception`s all at once instead of with mutations, i.e. 
```diff
-            x = Xception()
-            x.errorCode = 1001
-            x.message = s
-            raise x
+            raise Xception(1001, s)
```         
The original `py.twisted` test didn't do this either because `Xception.__init__` didn't support it historically (I don't recall) or for no good reason. Note that I struggled and failed to get `master` to build and the tests to run, so someone else should verify that these fixes are correct.

cjbarth:
@elprans would you be in a position to update this branch with the adjustment that @cpiro suggests here for both `py.twisted` and `py.tornado` in addition to removing that unneeded import? I feel like we're really close on getting this landed :)

bufferoverflow:
> @bufferoverflow It appears that you and @cpiro may have originally written these tests. Would either of you happen to know why this test modifies an exception? Or, perhaps shed light on what is being tested here (maybe we don't even need theses tests anymore)?

@cjbarth I'm very sorry, no clue.

dcelasun:
@cjbarth would you like to open a new PR with the suggested fixes? If you do, please keep @elprans' commits intact and push a new commit on top. This way, we can preserve authorship.

cjbarth:
Thank you @elprans for pushing those changes. That seems to correct the previous problems. Now there appears to be a problem with the CI build itself. It says it can't open a file that clearly exists. @dcelasun Would you happen to know what might cause this sort of error, given that the file exists and wasn't modified by this PR?

```
test 421
        Start 421: StalenessCheckTest
421: Test command: /cygdrive/c/Python27/python.exe "/cygdrive/c/projects/thrift/compiler/cpp/test/compiler/staleness_check.py" "/cygdrive/c/projects/build/CYGWIN/x64/compiler/cpp/bin/thrift.exe"
421: Test timeout computed to be: 300
421: C:\Python27\python.exe: can't open file '/cygdrive/c/projects/thrift/compiler/cpp/test/compiler/staleness_check.py': [Errno 2] No such file or directory
421/445 Test #421: StalenessCheckTest ...............***Failed    0.05 sec
```

dcelasun:
Appveyor builds for thrift are highly unstable and we don't really block merges if they fail. Since Travis is green, I'll merge this shortly.

PR #1840, opened 2019-08-02
THRIFT-4914: Add THeader to context object for server reads

fishy:
Client: go

This is the first part of THRIFT-4914, which handles the server reading
part in the requests (client -> server direction).

In TSimpleServer implementation, when the protocol is THeaderProtocol,
auto add all the headers read into the context object before passing
that to processor, so the processor code can access them from the
context object directly, via the new helper functions added in
header_context.go file.

  

### DIFF
--- a/lib/go/thrift/header_context.go
+++ b/lib/go/thrift/header_context.go
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


lib/go/thrift/header_context.go
```diff
@@ -0,0 +1,81 @@
...
+	headerKeyList int
+)
+
+// Values for headerKeyList.
+const (
+	headerKeyListRead headerKeyList = iota
+)
+
+// SetContextHeader sets a header in the context.
+func SetContextHeader(ctx context.Context, key, value string) context.Context {
```
dcelasun: `SetHeader` sounds better, no need to mention `Context` in the function name. Same goes for all the other functions in this file.

fishy: @dcelasun My main concern is that we have functions sound similar, like [`thrift.(*THeaderProtocol).SetWriteHeader`](https://godoc.org/github.com/apache/thrift/lib/go/thrift#THeaderProtocol.SetWriteHeader). Although there's no conflicts this might look confusing and making it more explicit might be better. Thoughts?

fishy: I did it in a fixup commit. If we decided that we do want the rename I'll squash the branch.

dcelasun: Since the other functions belong to `THeaderProtocol` and these new ones are on the package level, I don't think there'll be much confusion.

If you have any other changes you'd like to do, please go ahead. Otherwise I'm gonna squash & merge.

fishy: I squashed them. Thanks for the review!

PR #1843, opened 2019-08-09
THRIFT-4920: Fix generation of binary string constants

elprans:
Currently, the thrift compiler generates regular string literals for
string constants marked as `binary`.  This makes the resulting type
inconsistent under Python 3.


dcelasun:
Looks like this PR fell through the cracks, sorry about that.

PR #1845, opened 2019-08-15
THRIFT-4914: Send context THeaders for client writes

fishy:
Client: go

This is the second part of THRIFT-4914, which handles the client writing
part in the requests (client -> server direction).

In TStandardClient, when the context has write headers set, and the
protocol is THeaderProtocol, automatically extract all headers from the
context object and set to THeaderProtocol to send over the wire.

Client code can set headers into the context object by using the helper
functions in header_context.go.

Note that we have separated keys for read and write header key list, so
that for code that's both a server and a client (example: a server that
calls other upstream thrift servers), they don't automatically forward
all headers to their upstream servers, and need to explicitly set which
headers to forward.

In order to make auto forwarding easier, also add SetForwardHeaders
function to TSimpleServer, which will help the users to auto forward
selected headers.

  

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -42,6 +42,9 @@ type TSimpleServer struct {
 	outputTransportFactory TTransportFactory
 	inputProtocolFactory   TProtocolFactory
 	outputProtocolFactory  TProtocolFactory
+
+	// Headers to auto forward in THeaderProtocol
+	forwardHeaders []string
 }
 
 func NewTSimpleServer2(processor TProcessor, serverTransport TServerTransport) *TSimpleServer {
@@ -123,6 +126,26 @@ func (p *TSimpleServer) OutputProtocolFactory() TProtocolFactory {
 
 func (p *TSimpleServer) Listen() error {
 	return p.serverTransport.Listen()
+}
+
+// SetForwardHeaders sets the list of header keys that will be auto forwarded
+// while using THeaderProtocol.
+//
+// "forward" means that when the server is also a client to other upstream
+// thrift servers, the context object user gets in the processor functions will
+// have both read and write headers set, with write headers being forwarded.
+// Users can always override the write headers by calling SetWriteHeaderList
+// before calling thrift client functions.
+func (p *TSimpleServer) SetForwardHeaders(headers []string) {
+	size := len(headers)
+	if size == 0 {
+		p.forwardHeaders = nil
+		return
+	}
+
+	keys := make([]string, size)
+	copy(keys, headers)
+	p.forwardHeaders = keys
 }
 
 func (p *TSimpleServer) innerAccept() (int32, error) {
@@ -235,6 +258,7 @@ func (p *TSimpleServer) processRequests(client TTransport) error {
 				return err
 			}
 			ctx = AddReadTHeaderToContext(defaultCtx, headerProtocol.GetReadHeaders())
+			ctx = SetWriteHeaderList(ctx, p.forwardHeaders)
 		}
 
 		ok, err := processor.Process(ctx, inputProtocol, outputProtocol)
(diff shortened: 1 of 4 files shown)


dcelasun:
> we have separated keys for read and write header key list, so
that for code that's both a server and a client (example: a server that
calls other upstream thrift servers), they don't automatically forward
all headers to their upstream servers, and need to explicitly set which
headers to forward.

I'd rather we do this the other way around. A common use case is passing a trace ID across services to track a request across multiple hops and it'd be great if this happened without user intervention.

If there are headers one doesn't want to forward, it's easy to remove them manually.

fishy:
> > we have separated keys for read and write header key list, so
> > that for code that's both a server and a client (example: a server that
> > calls other upstream thrift servers), they don't automatically forward
> > all headers to their upstream servers, and need to explicitly set which
> > headers to forward.
> 
> I'd rather we do this the other way around. A common use case for headers is passing a trace ID across services to track a request across multiple hops and it'd be great if this happened without user intervention.
> 
> If there are headers one doesn't want to forward, it's easy to remove them manually.

I think from a security perspective automatically forward all headers sound like a bad idea. Yes auto forward tracing headers is a common use case, but there might be other headers that's inappropriate to forward.

Other reasons I don't want to auto forward are:

1. From https://golang.org/pkg/context/ you can see that as of go 1.12 there's no way to actually remove a value from a context object (a hack to workaround that is that instead of storing a string value, we store a struct with a string and a deleted flag so that when we need to remove a value we just update it with deleted flag set to true, or just treat empty strings as deleted, or something like that)
2. We actually call the reading helper functions [`GetReadHeaderList`](https://godoc.org/github.com/apache/thrift/lib/go/thrift#GetReadHeaderList)/[`SetReadHeaderList`](https://godoc.org/github.com/apache/thrift/lib/go/thrift#SetReadHeaderList). If we want to do auto forward we would need to rename it into `GetHeaderList`/`SetHeaderList` (because there's only one list now), which is a breaking change (but maybe that's OK since that's not in a release yet)

In the current implementation, it would not be too hard for users to forward all/some headers if they have some middleware setup. They just need to add an extra step like:

```
ctx = thrift.SetWriteHeaderList(ctx, thrift.GetReadHeaderList(ctx))  // forward all
```

or

```
var AutoForwardHeaders = []string{
  "Trace",
  "others",
}

ctx = thrift.SetWriteHeaderList(ctx, AutoForwardHeaders)
```

Alternatively, an idea is that we add a field and a function `TSimpleServer`, to setup the header list that auto forwards, and call `SetWriteHeaderList` where we currently inject the read headers into `ctx` right now. So users only need to setup that once before serving and it will work for all requests.

fishy:
>Alternatively, an idea is that we add a field and a function TSimpleServer, to setup the header list that auto forwards, and call SetWriteHeaderList where we currently inject the read headers into ctx right now. So users only need to setup that once before serving and it will work for all requests.

I implemented that in the fixup commit. Please take another look.

lib/go/thrift/simple_server.go
```diff
@@ -125,6 +128,26 @@ func (p *TSimpleServer) Listen() error {
 	return p.serverTransport.Listen()
 }
 
+// SetForwardHeaders sets the list of header keys that will be auto forwarded
+// while using THeaderProtocol.
+//
+// "forward" means that when the server is also a client to other upstream
+// thrift servers, the context object user get in the processor functions will
```
dcelasun: ... the context object *the user gets* in the processor ...

lib/go/thrift/simple_server.go
```diff
@@ -125,6 +128,26 @@ func (p *TSimpleServer) Listen() error {
...
+// have both read and write headers set, with write headers are from the forward
```
dcelasun: ... *with write headers being forwarded.*

dcelasun:
> there's no way to actually remove a value from a context object

> we would need to rename it into GetHeaderList/SetHeaderList (because there's only one list now), which is a breaking change

Both good points, thanks. I've just added some minor comments, then we are good to merge.

fishy:
I updated the comments based on your suggestion. Please take another look (I also updated the commit message to include the TSimpleServer change).

PR #1847, opened 2019-08-19
THRIFT-4919 THttpTransport/THttpClientTransport have bad timeout code

Jens-G:
THRIFT-4919 THttpTransport.cs (netstd) and THttpClientTransport (netcore) have bad timeout code
Patch: Jens Geyer

### DIFF
--- a/lib/netcore/Thrift/Transports/Client/THttpClientTransport.cs
+++ b/lib/netcore/Thrift/Transports/Client/THttpClientTransport.cs
@@ -63,6 +63,7 @@ namespace Thrift.Transports.Client
         public int ConnectTimeout
         {
             set { _connectTimeout = value; }
+            get { return _connectTimeout; }
         }
 
         public override bool IsOpen => true;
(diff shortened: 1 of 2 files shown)


miegie-a:
The commit only added a getter and fixes the timespan conversion from milliseconds.
There is no code fixing the issue that the class does not allow users from changing the timeout in the class instance since http client is created at class construction.

Setting connection timeout property should allow one to update the internal httpclient field’s timeout value, after class construction.

VisiG:
If an thrift server blocks a request for a special time (e.g. larger than 30 s) there will be a time-out. Seems to be this issue here. Please make this field readable and maybe also configurable.

It took me 2 hours to configure out why there is a time-out on the client side.

VisiG:
Please provide a property for the timeout in the netstandard libray.

It took me 2 hours to figure out why there is a 30s timeout. There is no way to change it without reflection in version 0.13.0

PR #1868, opened 2019-09-06
THRIFT-4951 clean up unsupported senum base type from 0.9.1

Issacpeng:
clean up unsupported senum base type from 0.9.1

  

### DIFF
--- a/test/AnnotationTest.thrift
+++ b/test/AnnotationTest.thrift
@@ -49,14 +49,6 @@ enum weekdays {
   SATURDAY ( weekend = "yes" )
 } (foo.bar="baz")
 
-/* Note that annotations on senum values are not supported. */
-senum seasons {
-  "Spring",
-  "Summer",
-  "Fall",
-  "Winter"
-} ( foo = "bar" )
-
 struct ostr_default {
   1: i32 bar;
 }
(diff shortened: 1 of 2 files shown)


Jens-G:
But sernum is not unsupported? It is only deprecated.

Jens-G:
I might not have been clear about this, but I'm against removing a working test for a still existing, although deprecated, feature. What added value does converting a tested feature into an untested feature provide?

IMHO this one should be reverted and should have been closed from the beginning.

dcelasun:
> But sernum is not unsupported? It is only deprecated.

Sorry, I thought it was unsupported. I'd be happy to revert it, but perhaps it's a good time to retire senum for good? It's been more than 5 years since the deprecation. We should either remove it or un-deprecate it.

Jens-G:
Agree,  but as I wrote in JIRA we should first do our homework and check if fbthrift still supports it, for sake of compatibility. Could be worth filing an fbthrift ticket, if they still have it, so we get rid of it on both ends. And while we're at it, we could also do the same for ```slist```

Issacpeng:
@Jens-G @dcelasun  i double checked senum and slist in fbthrift, find 2 remove commits, hope it can work for us~

1.senum:  https://github.com/facebook/fbthrift/commit/e03dab8439561afec565f1b87a546689a5a397e7
2.slist:
https://github.com/facebook/fbthrift/commit/9681401a2ce63a85e37c4436b55fd1d10a0310f8

PR #1878, opened 2019-09-21
THRIFT-2530: Fix TIOStreamTransport#isOpen

qinghui-xu:
isOpen always returning true (even after close is called) should be considered as anti-pattern behavior.


dcelasun:
This can unexpectedly break callers, please add it to [CHANGES.md](https://github.com/apache/thrift/blob/master/CHANGES.md#breaking-changes).

qinghui-xu:
Done.

dcelasun:
Merged, thanks!

PR #1904, opened 2019-10-23
THRIFT-4984: Ignore EOF error in TSimpleServer ReadFrame call

fishy:
Client: go

Since we are already ignoring some other EOF errors in
TSimpleServer.processRequest[0], this would help us reducing the spammy
and scary logs.

Also give the user more control on how TSimpleServer does the logging.

[0]: https://github.com/apache/thrift/blob/cecee50308fc7e6f77f55b3fd906c1c6c471fa2f/lib/go/thrift/simple_server.go#L265-L266

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -20,6 +20,7 @@
 package thrift
 
 import (
+	"io"
 	"log"
 	"runtime/debug"
 	"sync"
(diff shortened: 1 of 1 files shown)


dcelasun:
Please don't introduce a custom logger, the Go ecosystem already has too many of those. Just use `log`, users can redirect it with `log.SetLogger` if they want to.

fishy:
@dcelasun I removed the logging part, I still think we should make that change but I'll open a new JIRA ticket to discuss that. Please take another look.

PR #1905, opened 2019-10-23
THRIFT-4252: Close sockets when shut down server

qinghui-xu:
In TThreadPoolServer, threads are blocking in io with open sockets,
as long as clients don't close the connection, server threads are
never stopped even after a shutdown is called on server (because
they are blocked waiting for io).
To be able to stop all server threads properly, server should
proactively close sockets once a shutdown is initiated.

  

### DIFF
--- a/lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
+++ b/lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
@@ -171,7 +171,7 @@ public class TThreadPoolServer extends TServer {
   	}
 
   	execute();
-  	if (!waitForShutdown()) {
+    if (!waitForShutdown()) {
   	  LOGGER.error("Shutdown is not done after " + stopTimeoutVal + stopTimeoutUnit);
     }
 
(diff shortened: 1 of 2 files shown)


lib/java/src/org/apache/thrift/server/TThreadPoolServer.java
```diff
@@ -173,13 +171,14 @@ public void serve() {
   	}
 
   	execute();
-  	waitForShutdown();
-    
+  	if (!waitForShutdown()) {
+  	  LOGGER.error("Shutdown is not done after " + stopTimeoutVal + stopTimeoutUnit);
+    }
```
kainjow: whitespace looks off here?

qinghui-xu: In fact, it seems the closing bracket is correctly indented (with white spaces).
The problem is from previsous lines (especially the existent code base), it is using tabulation instead of white space for indentation.
I'll fix my code here and then eventually push another PR for fixing existent code base.

dcelasun:
LGTM. Restarted the flaky Travis test, will merge once it's green.

PR #1906, opened 2019-10-23
THRIFT-4985: Cleanup logging in go library

fishy:
Client: go

As discussed in https://issues.apache.org/jira/browse/THRIFT-4985:

1. Remove logging in TZlibTransport
2. Add a new Logger type to help users to better control logging in this
library
3. Use the new Logger type in TSimpleServer and TDebugProtocol
4. Remove panic recovery from TSimpleServer

### DIFF
--- a/lib/go/thrift/debug_protocol.go
+++ b/lib/go/thrift/debug_protocol.go
@@ -21,23 +21,40 @@ package thrift
 
 import (
 	"context"
-	"log"
+	"fmt"
 )
 
 type TDebugProtocol struct {
 	Delegate  TProtocol
 	LogPrefix string
+	Logger    Logger
 }
 
 type TDebugProtocolFactory struct {
 	Underlying TProtocolFactory
 	LogPrefix  string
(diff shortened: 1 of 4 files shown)


lib/go/thrift/debug_protocol.go
```diff
@@ -21,247 +21,269 @@ package thrift
...
 
 type TDebugProtocolFactory struct {
 	Underlying TProtocolFactory
 	LogPrefix  string
+	Logger     Logger
 }
 
+// NewTDebugProtocolFactory creates a TDebugProtocolFactory.
+//
+// Deprecated. Please use NewTDebugProtocolFactoryWithLogger or the struct
```
dcelasun: Should be `Deprecated: Please...` to be detectable by IDEs and static analysis tools (see [here](https://github.com/golang/go/wiki/Deprecated)).

fishy: Done.

PR #1910, opened 2019-11-02
fix typo in cmake

Jens-G:
One Appveyor build is broken for a while now

### DIFF
--- a/lib/hs/CMakeLists.txt
+++ b/lib/hs/CMakeLists.txt
@@ -37,7 +37,7 @@ set(haskell_sources
 )
 
 if(BUILD_TESTING)
-    list(APPEND haskell_soruces
+    list(APPEND haskell_sources
         test/Spec.hs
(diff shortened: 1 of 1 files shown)


kainjow:
It looks like this might be the error?

> [00:10:32]   Building Haskell library
[00:10:33]   Downloading the latest package list from hackage.haskell.org
[00:10:49]   cabal.exe: --enable-tests was specified, but tests can't be enabled in a
[00:10:49]   remote package
[00:10:49]   
[00:10:49] C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with code 1. [C:\projects\build\MSVC2015\x86\lib\hs\haskell_library.vcxproj]

kainjow:
It might be good to merge this on its own so we can isolate it away from other fixes needed?

PR #1922, opened 2019-11-06
THRIFT-4984: Handle wrapped io.EOF errors

fishy:
Client: go

TCompactProtocol (which is used by THeaderTransport to read headers)
could wrap the underlying error with TProtocolException, which breaks
err == io.EOF test in some cases.


fishy:
@dcelasun Sorry we have to do this yet again :(

PR #1923, opened 2019-11-06
THRIFT-4914: Add TResponseHelper

fishy:
Client: go

This is the third part of THRIFT-4914, which handles the server writing
part in the response (server -> client direction).

Define a new type, TResponseHelper, which only contains THeader related
functions for now, but can be extended for other functions in the
future.

In TSimpleServer, inject a TResponseHelper into the context object
passed into the handler functions. Handler function code could retrieve
the injected TResponseHelper to set headers to be written to the client.

### DIFF
--- a/lib/go/thrift/response_helper.go
+++ b/lib/go/thrift/response_helper.go
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package thrift
+
+import (
+	"context"
+)
+
+// See https://godoc.org/context#WithValue on why do we need the unexported typedefs.
+type responseHelperKey struct{}
+
+// TResponseHelper defines a object with a set of helper functions that can be
+// retrieved from the context object passed into server handler functions.
+//
+// Use GetResponseHelper to retrieve the injected TResponseHelper implementation
+// from the context object.
+//
+// The zero value of TResponseHelper is valid with all helper functions being
+// no-op.
+type TResponseHelper struct {
+	// THeader related functions
+	*THeaderResponseHelper
+}
+
+// THeaderResponseHelper defines THeader related TResponseHelper functions.
+//
+// The zero value of *THeaderResponseHelper is valid with all helper functions
+// being no-op.
+type THeaderResponseHelper struct {
+	proto *THeaderProtocol
+}
+
+// NewTHeaderResponseHelper creates a new THeaderResponseHelper from the
+// underlying TProtocol.
+func NewTHeaderResponseHelper(proto TProtocol) *THeaderResponseHelper {
+	if hp, ok := proto.(*THeaderProtocol); ok {
+		return &THeaderResponseHelper{
+			proto: hp,
+		}
+	}
+	return nil
+}
+
+// SetHeader sets a response header.
+//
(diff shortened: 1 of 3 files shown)


fishy:
@dcelasun

lib/go/thrift/response_helper.go
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+package thrift
+
+import (
+	"context"
+)
+
+// See https://godoc.org/context#WithValue on why do we need the unexported typedefs.
+type responseHelperKeyType int
```
dcelasun: Let's use `struct{}` as the type, which is more common and doesn't allocate memory, unlike `int`.

lib/go/thrift/response_helper.go
```diff
@@ -0,0 +1,111 @@
...
+type nopHeaderResponseHelper int
+
+func (nopHeaderResponseHelper) SetHeader(key, value string) {}
+func (nopHeaderResponseHelper) ClearHeaders()               {}
+
+type headerHelper struct {
+	proto *THeaderProtocol
+}
+
+func newHeaderHelper(proto TProtocol) THeaderResponseHelper {
```
dcelasun: Return the concrete type, not an interface.

lib/go/thrift/response_helper.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type nopResponseHelper struct {
+	nopHeaderResponseHelper
+}
+
+var nopHelper TResponseHelper = nopResponseHelper{}
+
+// THeaderResponseHelper defines THeader related TResponseHelper functions.
+type THeaderResponseHelper interface {
```
dcelasun: Not sure if we need the interface at all. What if we export `headerHelper` and let it handle the no-op case? For example:

```go
type HeaderHelper struct {
	proto *THeaderProtocol
}

func (h *HeaderHelper) SetHeader(key, value string) {
        if h != nil {
	    h.proto.SetWriteHeader(key, value)
        }
}

func (h *HeaderHelper) ClearHeaders() {
        if h != nil {
	    h.proto.ClearWriteHeaders()
        }
} 
```

Which would allow us to simplify the getter:
```go
func GetResponseHelper(ctx context.Context) (helper *HeaderHelper, ok bool) {
	if v := ctx.Value(responseHelperKey); v != nil {
		helper, ok = v.(TResponseHelper)
	}
	return
}
```

lib/go/thrift/response_helper.go
fishy: Good point. Done.

lib/go/thrift/response_helper.go
fishy: We do need an interface if we want to embed it into the `TResponseHelper` interface, because you cannot embed a struct into an interface. I think we do want embedding because that way makes `TResponseHelper` extensible in the future.

But I like the idea of not having a separated `nopHeaderHelper`, and made that change. I'm not sure if we want to export `headerHelper` type (and `newHeaderHelp` function). Looking at the existing library code we didn't seem very consistent on whether to export those helper implementations. I didn't export them now but I don't feel strongly either way, but if we do decide to export them, we might need a better name (`HeaderHelper` sounds too generic, but we do need the name `THeaderResponseHelper` for the interface, not the concrete type).

dcelasun: > We do need an interface if we want to embed it into the `TResponseHelper` interface

Only if we want `TResponseHelper` to be an interface. Why can't it be a struct? I think it's better if callers define interfaces if/when they need them, instead of the library exporting it.

In this case, we only provide optional helper functions so if the user wants a different implementation, they can write one and perhaps define their own interface which is implemented by our `TResponseHelper` as well.

> (`HeaderHelper` sounds too generic, but we do need the name `THeaderResponseHelper` for the interface, not the concrete type)

Not having interfaces would solve this as well.

fishy: OK I converted interfaces into structs. PTAL.

lib/go/thrift/response_helper.go
```diff
@@ -0,0 +1,94 @@
...
+//
+// The zero value of *THeaderResponseHelper is valid with all helper functions
+// being no-op.
+type THeaderResponseHelper struct {
+	proto *THeaderProtocol
+}
+
+// NewTHeaderResponseHelper creates a new THeaderResponseHelper from the
+// underlying TProtocol.
+func NewTHeaderResponseHelper(proto TProtocol) *THeaderResponseHelper {
```
dcelasun: Let's go with `NewTHeaderResponseHelper(proto *THeaderProtocol) *THeaderResponseHelper` so we won't need the type check.

lib/go/thrift/response_helper.go
dcelasun: Looks good!

PR #1926, opened 2019-11-09
THRIFT-4914: Add GetResponseHeadersFromClient helper function

fishy:
Clients: go

This is the fourth and final part of THRIFT-4914, which handles the
client reading part in the response (server -> client direction).

### DIFF
--- a/lib/go/thrift/header_protocol.go
+++ b/lib/go/thrift/header_protocol.go
@@ -303,3 +303,17 @@ func (p *THeaderProtocol) ReadBinary() (value []byte, err error) {
 func (p *THeaderProtocol) Skip(fieldType TType) error {
 	return p.protocol.Skip(fieldType)
 }
+
+// GetResponseHeadersFromClient is a helper function to get the read THeaderMap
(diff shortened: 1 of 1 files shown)


lib/go/thrift/header_protocol.go
```diff
@@ -303,3 +303,17 @@ func (p *THeaderProtocol) ReadBinary() (value []byte, err error) {
 func (p *THeaderProtocol) Skip(fieldType TType) error {
 	return p.protocol.Skip(fieldType)
 }
+
+// GetResponseHeadersFromClient is a helper function to get the read THeaderMap
+// from the last response received from the given client.
+//
+// If the last response was not sent over THeader protocol,
+// an empty map will be returned.
```
dcelasun: Should be `nil will be returned` since nil maps have slightly different semantics. We could return an initialized empty map instead, but that seems like an unnecessary allocation.

fishy: Changed to `a nil map will be returned`.

PR #1927, opened 2019-11-12
THRIFT-5003: Websocket Connection in Browsers with nodejs code

alabama:
After some changes to the ws_connection.js file in the nodejs code, I could create a version of the WSConnection, createWSConnection and createWSClient which is able to run similar code of WebSocket connections in the browser and NodeJS.
Solution is package, `isomorphic-ws` which provides the proper Websocket libs depending on the env (browser or node)

https://issues.apache.org/jira/browse/THRIFT-5003
- [] If your change does not involve any code, add ` [skip ci]` at the end of your pull request to free up build resources.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,14 +33,19 @@
   "browser": "./lib/nodejs/lib/thrift/browser.js",
   "main": "./lib/nodejs/lib/thrift",
   "engines": {
-    "node": ">= 4.1.0"
+    "node": ">= 8.16.2"
   },
   "dependencies": {
+    "browser-or-node": "^1.2.1",
+    "isomorphic-ws": "^4.0.1",
     "node-int64": "^0.4.0",
     "q": "^1.5.0",
-    "ws": "^5.0.0"
+    "ws": "^5.2.2"
   },
   "devDependencies": {
+    "@types/node": "^10.12.6",
+    "@types/node-int64": "^0.4.29",
+    "@types/q": "^1.5.1",
     "buffer-equals": "^1.0.4",
     "commander": "^2.14.1",
     "connect": "^3.6.6",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -33,14 +33,19 @@
   "browser": "./lib/nodejs/lib/thrift/browser.js",
   "main": "./lib/nodejs/lib/thrift",
   "engines": {
-    "node": ">= 4.1.0"
```
dcelasun: This is a massive version jump of ~4 years. Are we sure want this?

At the very least [LANGUAGES.md](https://github.com/apache/thrift/blob/master/LANGUAGES.md) should be updated to reflect this.

dcelasun:
Besides the Node version issue, this looks OK. I don't personally use the js/node clients so I'd appreciate another +1.

package.json
emmenlau: I assume you are right, we may not require this new node version, its just the one we tested with. I'll try to clarify what is required.

alabama: The version jump of ~4 years is huge, I understand but do you think it is a good idea to work with versions which are no longer maintained? 

If this is a necessary requirement I could switch to node version '6.17.1'  which was the last release of LTS version 'Boron' and has some security fixes in it: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V6.md#6.17.1

dcelasun: I'd be happy to raise the minimum version to the latest supported LTS release.

alabama: The latest supported LTS release would be 8.16.2: https://nodejs.org/en/about/releases/ (the end of life will be in 2 weeks 2019-12-31 😅) 

I just mentioned 6.17.1 because the version jump wouldn't be that big, but I still won't recommend to use that. If you think we should be careful with these version changes, we should go for that. I'll update [LANGUAGES.md](https://github.com/apache/thrift/blob/master/LANGUAGES.md) and packages.json

dcelasun: OK then, let's update the min version to 10x which should be supported until mid-2021.

alabama:
Like we discussed before I updated the node version to 10.18.0 after testing it on my machine. What is left to merge this PR to master?

emmenlau:
This PR is quite relevant for us. @dcelasun could you kindly take a look if this can be merged?

dcelasun:
Merged, apologies for the delay.

PR #1930, opened 2019-11-13
THRIFT-5004, THRIFT-5005, THRIFT-5006, THRIFT-5007

Jens-G:
All Delphi changes, Since Delphi is not really connected to CI yet I do all four patches in one PR


Jens-G:
Adding some more test cases, will send new PR.

PR #1933, opened 2019-11-16
THRIFT-5012 Centralize configuration aspects into a commonly used con…

Jens-G:
…figuration object

Client: Delphi
Patch: Jens Geyer


Jens-G:
On hold, since I want to do some more tests before I merge this.

PR #1934, opened 2019-11-17
THRIFT-5013: Use Java Objects RequireNonNull

belugabehr:
Use Java Objects RequireNonNull. Made for this scenario.

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-


compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -1060,11 +1060,15 @@ void t_java_generator::generate_union_getters_and_setters(ostream& out, t_struct
     }
     indent(out) << "public void set" << get_cap_name(field->get_name()) << "("
                 << type_name(field->get_type()) << " value) {" << endl;
+
+    indent(out) << "  setField_ = _Fields." << constant_name(field->get_name()) << ";" << endl;
+
     if (type_can_be_null(field->get_type())) {
-      indent(out) << "  if (value == null) throw new java.lang.NullPointerException();" << endl;
+      indent(out) << "  value_ = java.util.Objects.requireNonNull(value);" << endl;
```
ctubbsii: This is nice, but the method can take an optional argument to have a string message. It might be good to put the name of the field in the message:

```suggestion
      indent(out) << "  value_ = java.util.Objects.requireNonNull(value,\"" << "_Fields." << constant_name(field->get_name()) << "\");" << endl;
```

(or similar)

ctubbsii: Also, this will only work with Java 7 or later. I'm not sure what version of Java libthrift is targeting. Do we know it's definitely for 7 or later? I'm unable to confirm using my normal method (inspecting the pom.xml) because the pom.xml is bolted on afterwards, and libthrift isn't built with Maven (AFAICT).

belugabehr: My expectation is that Thirft is running on JDK 1.8 or higher.

https://issues.apache.org/jira/browse/THRIFT-4694

ctubbsii: Perfect. Thanks.

belugabehr:
@ctubbsii I generated the output Java, compiled it, and it looks correct.

```
 public void setName(java.lang.String value) {
    setField_ = _Fields.NAME;
    value_ = java.util.Objects.requireNonNull(value,"_Fields.NAME");
  }
```

belugabehr:
How do we kick off CI again?

kainjow:
> How do we kick off CI again?

Create an empty commit or merge master.

ctubbsii:
Appveyor CI seems to be failing for an unrelated reason. The test case in `compiler/cpp/test/compiler/staleness_check.py` seems to be failing, possibly because it is trying to run a python3 test using python2.7.

belugabehr:
@ctubbsii Are you willing and able to provide a +1 on this change?

ctubbsii:
@belugabehr I gave my +1 above already, but I'm not a committer on Thrift, so my opinion is just advisory.

belugabehr:
@dcelasun And this one as well? :)

dcelasun:
Thanks for the PR and apologies for the delay in reviews.

PR #1937, opened 2019-11-18
THRIFT-5016: Do Not Check 'other' For Null in Equals

belugabehr:
There is a null check before instanceof in the generated Java equals method. This check is superfluous.  Remove it.

https://stackoverflow.com/questions/2950319/is-null-check-needed-before-calling-instanceof


Jens-G:
+1

Jens-G:
Still open ... any issues?

belugabehr:
Will rebase and run test again.

belugabehr:
@dcelasun Can you also lend a hand here for merge?

Jens-G:
> Will rebase and run test again.
Thanks.

PR #1938, opened 2019-11-18
Fix UTF-8 Decoding Errors

eoinmiller-r7:
If a binary string contains non-utf8 decodable characters, this will cause an exception.

### DIFF
--- a/lib/py/src/compat.py
+++ b/lib/py/src/compat.py
@@ -37,7 +37,7 @@ else:
     from io import BytesIO as BufferIO  # noqa
 
     def binary_to_str(bin_val):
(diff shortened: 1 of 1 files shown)


Jens-G:
Can we have a ticket with a test case?

Jens-G:
> Can we have a ticket with a test case?
> http://thrift.apache.org/docs/HowToContribute

Any news?

Jens-G:
No explanation, no ticket, no merge.

eoinmiller-r7:
The explanation was provided:

If a binary string contains non-utf8 decodable characters, this will cause an exception.

PR #1940, opened 2019-11-20
THRIFT-5019: Duplicate imports from multiple includes from a namespace

dcelasun:
If a thrift file includes two files from the same namespace into a
separate file, the generated Go code has duplicate imports for that
namespace. This fixes that.

Client: go


dcelasun:
Tests are [all green on my fork](https://travis-ci.org/dcelasun/thrift/builds/614535616), so I'll merge this right away.

PR #1944, opened 2019-11-21
THRIFT-5010: BinaryPrimitives.Read/WriteInt32BigEndian should be used to convert to/from network byte order

zembord:


Jens-G:
Thanks for the PR,  am looking at it.

PR #1949, opened 2019-11-25
THRIFT-5032 netstd: Use PascalCase for Properties.

ptsneves:
According to microsoft guidelines properties should be
Pascal cased. See [1].
[1] https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions


ptsneves:
It seems my changes are failing in AppVeyor in unrelated topics.
```/cygdrive/c/projects/thrift/compiler/cpp/test/compiler/staleness_check.py``` I did not touch this file yet it says that it is missing.

PR #1950, opened 2019-11-26
THRIFT-5034 Allow for Reverse Connection servers

ptsneves:
The changes at this point are very shallow as they only include a reference implementation for Python. I will in the course of the discussion add tests and some reference implementations in other languages.

EDIT: The initial commit was wrong and included other things i am working on. Disregard them and only regard the single patch with Jira ID number

### DIFF
--- a/lib/py/src/server/TReverseTunnelServer.py
+++ b/lib/py/src/server/TReverseTunnelServer.py
@@ -1,0 +1,67 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+from six.moves import queue
+from thrift.transport import TTransport
+
+
+class TReverseTunnelServer(TTransport.TTransportBase):
+    """A transport that allows RPC Servers to use Client Transports.
+
+    This transport is useful for reverse connections, where the server/client
+    transport architecture is the reverse of server/client architecture for
+    the RPC service. One example where this transport is useful
+    is when there is RPC service behind a restrited network that does not allow
+    incoming connections.
+
+    The current implementation of this transport is compatible with
+    ThreadedPoolServer and other concurrent transport servers.
+    As transport servers expect that the accept method is a blocking one,
+    we use a queue for job control.
+    """
+
+    def __init__(self, transport, acceptable_clients=10):
+        self._transport = transport
+        self._connections = queue.Queue(acceptable_clients)
+
+    def listen(self):
+        pass
+
+    def accept(self):
+        self._transport.open()
+        self._connections.put(None)
+        return self._transport
+
+    def close(self):
+        with self._connections.mutex:
+            self._connections.queue.clear()
+
+    def flush(self):
+        # We assume that once we flush we are done with a message
+        self._connections.get()
+        self._transport.flush()
+
+
+class TReverseTunnelServerFactory(object):
+    def __init__(self, transport, acceptable_clients=10):
+        self._trans = transport
+        self._acceptable_clients = acceptable_clients
+
+    def getTransport(self, trans):
+        return TReverseTunnelServer(self.trans, self._acceptable_clients)
--- a/lib/py/src/server/__init__.py
+++ b/lib/py/src/server/__init__.py
@@ -17,4 +17,4 @@
 # under the License.
 #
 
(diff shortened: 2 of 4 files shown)


lib/py/src/transport/TReverseTunnelServer.py
```diff
@@ -0,0 +1,40 @@
+import queue
+from .TTransport import TTransportBase, TTransportException, TServerTransportBase
+
+
+
+class TReverseTunnelServerFactory(object):
+    """Factory transport that builds framed transports"""
```
Jens-G: comment needs a fix, I guess

lib/py/src/transport/TReverseTunnelServer.py
```diff
@@ -0,0 +1,40 @@
+import queue
```
Jens-G: ASF header again

Jens-G: .

ptsneves:
From what i see the travis-ci is failing only on some c++ thing and these changes are unrelated to c++.

RandyAbernethy:
The Travis complaints are style based but related to this patch:

`# Python code style
flake8
./test/py/TestReverseTunnel.py:35:1: E302 expected 2 blank lines, found 1
./test/py/TestReverseTunnel.py:37:9: F401 'glob' imported but unused
./test/py/TestReverseTunnel.py:57:13: E265 block comment should start with '# '
./test/py/TestReverseTunnel.py:58:13: E265 block comment should start with '# '
./test/py/TestReverseTunnel.py:92:1: E305 expected 2 blank lines after class or function definition, found 1
./lib/py/src/server/TReverseTunnelServer.py:22:1: E265 block comment should start with '# '
./lib/py/src/server/TReverseTunnelServer.py:24:1: E302 expected 2 blank lines, found 1
./lib/py/src/server/TReverseTunnelServer.py:67:9: E265 block comment should start with '# '
./lib/py/src/server/TReverseTunnelServer.py:71:1: E302 expected 2 blank lines, found 1
The command "build/docker/run.sh" exited with 1.`

Simple to fix so probably worth fixing.

lib/py/src/server/TReverseTunnelServer.py
```diff
@@ -0,0 +1,77 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+from six.moves import queue
+from thrift.transport import TTransport
+#from thrift.transport.TTransport import TTransportBase, TTransportException, TServerTransportBase
```
dcelasun: Don't leave commented imports.

lib/py/src/server/TReverseTunnelServer.py
```diff
@@ -0,0 +1,77 @@
...
+
+class TReverseTunnelServer(TTransport.TTransportBase):
+    """A transport that allows RPC Servers use Client Transports.
+
+    This useful for reverse connections, where
```
dcelasun: Grammar and structure fixes to this whole section:

```
A transport that allows RPC servers to use client transports.

This is useful for "reverse connections" where a primary server
passes RPC calls to secondary servers attached to it. One example
would be an RPC service in a private network that doesn't allow
public traffic.

The current implementation of this Transport is compatible with
ThreadedPoolServer and other concurrent servers as it blocks calls
to accept() based on the number of active connections.

The idea is that for a server, a flush means a finished RPC call. This
isn't mentioned in the specification, but is a good heuristic for signalling
the accept() caller.
```

ptsneves: I made some changes but could you be more specific, specially on the structure?

lib/py/src/server/TReverseTunnelServer.py
ptsneves: done

lib/py/src/server/TReverseTunnelServer.py
dcelasun: I mean my version fixes some typos and grammar issues, and it reads better (to me, at least). If you don't have any objections to it, can you use it as is?

dcelasun:
@ptsneves any interest in contributing this for other languages? I'm still tenatively +1 on this, but I think it's important to have at least two implementations of brand new, out-of-spec features like this.

lib/py/src/server/TReverseTunnelServer.py
ptsneves: Ahh of course, i thought it was a quote not a suggestion to use that.

ptsneves:
@dcelasun Yes, I will try to submit it. I can do it in C++ or C#. Do you want cross language tests also? If so I do not know how to do this, but if you provide me with a hint i can do it.

dcelasun:
C++ or C# sounds fine, whichever you are more comfortable with.  Cross tests would be great! They are defined [here](https://github.com/apache/thrift/blob/0ebf3ce99d74bc77912536dc3ad95dac10ebf12a/test/tests.json). For languages not covered by your test, you'll have to add known failures [here](https://github.com/apache/thrift/blob/0ebf3ce99d74bc77912536dc3ad95dac10ebf12a/test/known_failures_Linux.json). See a recent example [here](https://github.com/apache/thrift/commit/f1c53418a794f3d740b809ece822ce79c0bb62a3).

If you have any questions or need help, feel free to join our [Slack channel](https://s.apache.org/slack-invite) #thrift and ping me there.

PR #1951, opened 2019-11-26
THRIFT-5030 Seqid verification

ptsneves:
One of the proposals for seqid verification based on code generation.


ptsneves:
Needs rework

PR #1953, opened 2019-11-27
THRIFT-5000 Thrift docker image publish on releases appveyor pipeline…

cspwizard:
THRIFT-5000 Thrift docker image publish on releases appveyor pipeline added

Added appveyor pipeline. Some additional configuration needed:
secret environment variable `login` should be set for the pipeline with content:
`echo docker-hub-api-key | docker login --username thrift --password-stdin` where `docker-hub-api-key` should be replaced with appropriate api key value.

`[skip ci]`

### DIFF
--- a/build/docker/publish/Dockerfile
+++ b/build/docker/publish/Dockerfile
@@ -1,0 +1,14 @@
+FROM gcc:8 as build
+
+RUN apt update && apt upgrade -y && apt install flex bison cmake -y
+
+WORKDIR /usr/src/thrift
+
+COPY . /usr/src/thrift
+
+RUN mkdir cmake-build && cd cmake-build && cmake -DWITH_SHARED_LIB=off .. && make
+
+FROM ubuntu 
+COPY --from=build /usr/src/thrift/cmake-build/bin/thrift /bin/thrift
+WORKDIR /thrift
+ENTRYPOINT ["thrift"]
--- a/docker-hub.appveyor.yml
+++ b/docker-hub.appveyor.yml
@@ -11,4 +11,4 @@ environment:
 build_script:
   - sh: docker build compiler/cpp -f build/docker/publish/Dockerfile -t thrift:$APPVEYOR_REPO_BRANCH
   - sh: eval $login
-  - sh: docker push thrift:$APPVEYOR_REPO_BRANCH
+  - sh: docker push apache/thrift:$APPVEYOR_REPO_BRANCH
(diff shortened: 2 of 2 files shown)


docker-hub.appveyor.yml
```diff
@@ -0,0 +1,14 @@
...
+image: Ubuntu
+clone_depth: 1
+environment:
+  matrix:
+    - login:
+        secure: echo docker-hub-api-key | docker login --username thrift --password-stdin
+build_script:
+  - sh: docker build compiler/cpp -f build/docker/publish/Dockerfile -t thrift:$APPVEYOR_REPO_BRANCH
+  - sh: eval $login
+  - sh: docker push thrift:$APPVEYOR_REPO_BRANCH
```
jeking3, line 14: The official repository is at apache/thrift not thrift.

build/docker/publish/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM gcc:8 as build
+
+RUN apt update && apt upgrade -y && apt install flex bison cmake -y
+
+WORKDIR /usr/src/thrift
+
+COPY . /usr/src/thrift
+
+RUN mkdir cmake-build && cd cmake-build && cmake -DWITH_SHARED_LIB=off .. && make
```
jeking3, line 9: Is there a way to build, drop the build directory, and squash the layers so you don't have to ship all the intermediary bits that make the container larger?

build/docker/publish/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM gcc:8 as build
```
jeking3, line 1: This is a fairly large container to start from.

build/docker/publish/Dockerfile
cspwizard: it's already done there output image is based on ubuntu image and contains only thrift binary
`FROM ubuntu 
COPY --from=build /usr/src/thrift/cmake-build/bin/thrift /bin/thrift
WORKDIR /thrift
ENTRYPOINT ["thrift"]`

build/docker/publish/Dockerfile
cspwizard: maybe, but it has almost everything that is needed for build

docker-hub.appveyor.yml
cspwizard: fixed

PR #1955, opened 2019-11-27
THRIFT-5012 Centralize configuration aspects into a commonly used configuration object [ci skip]

Jens-G:
Client: Delphi
Patch: Jens Geyer


Jens-G:
merged

PR #1964, opened 2019-12-18
THRIFT-5078 Handle named pipe clients quickly disconnecting

nickcollier:
When using the TPipeServer transport, if a client connects but disconnects before the acceptImpl method is able to create a TPipe instance the resulting exception causing the entire transport to stop.

You can cause this issue using the SysInternals accesschk utility (https://docs.microsoft.com/en-us/sysinternals/downloads/accesschk) to query the permissions of the named pipe.

This has also been reported in osquery https://github.com/osquery/osquery/issues/4280


Jens-G:
Could you rebase, also to trigger CI again?

PR #1966, opened 2019-12-20
cpp support custom header in http client transport

farmerworking:
cpp support custom header in http client transport


Jens-G:
I'd like an JIRA ticket for this explaining what the actual issue is. Also a test case might be helpful.

Jens-G:
Please rebase and create a ticket. Thank you.

dcelasun:
The C++ library supports THeader, which is more or less transport agnostic and should be the preferred way of setting headers.

This PR on the other hand, effectively bypasses the thrift protocol. If you really think this needs to be in thrift, please open a Jira ticket as advised and discuss.

PR #1987, opened 2020-01-17
THRIFT-5069: Make TDeserializer resource pool friendly

fishy:
Client: go

This change improves performance when using TDeserializer with a
resource pool. See https://issues.apache.org/jira/browse/THRIFT-5069 for
more context.

Also add TSerializerPool and TDeserializerPool, which are thread-safe
versions of TSerializer and TDeserializer. Benchmark result shows that
they are both faster and use less memory than the plain version:

    $ go test -bench Serializer -benchmem
    goos: darwin
    goarch: amd64
    BenchmarkSerializer/baseline-8            577558              1930 ns/op             512 B/op          6 allocs/op
    BenchmarkSerializer/plain-8               452712              2638 ns/op            2976 B/op         16 allocs/op
    BenchmarkSerializer/pool-8                591698              2032 ns/op             512 B/op          6 allocs/op
    PASS

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,10 +7,16 @@
 - [THRIFT-4990](https://issues.apache.org/jira/browse/THRIFT-4990) - Upgrade to .NET Core 3.1 (LTS)
 - [THRIFT-4981](https://issues.apache.org/jira/browse/THRIFT-4981) - Remove deprecated netcore bindings from the code base
 - [THRIFT-5006](https://issues.apache.org/jira/browse/THRIFT-5006) - Implement DEFAULT_MAX_LENGTH at TFramedTransport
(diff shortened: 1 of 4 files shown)


dcelasun:
Thanks for the PR! I think this is an acceptable breaking change, you just need to mention it in [this file](https://github.com/apache/thrift/blob/master/CHANGES.md#0140).

Ignore the Travis failure, it's a known issue we are working on.

fishy:
@dcelasun Done.

dcelasun:
Can you also rebase from master? There is a fix for Travis failures.

fishy:
Rebased latest master.

fishy:
@dcelasun Added `TSerializerPool` & `TDeserializerPool` into this PR and also added tests and benchmark tests. PTAL.

dcelasun:
LGTM.

PR #1994, opened 2020-01-25
THRIFT-5076: Improve CMake OpenSSL usage

kainjow:
The CMake file says OpenSSL is only used for C++ but it's actually used for C GLib too, so move it out from under C++ into a combined C++/GLib check.

- Update the output to show the status for building OpenSSL including the version.
- Remove additional find_package() call for OpenSSL since this is already done in DefineOptions.cmake.
- Fix test_client missing OpenSSL include dirs and libraries.

Originally #1829


kainjow:
@Jens-G @dcelasun any questions/comments about this?

PR #1995, opened 2020-01-27
Fix typo for python 3 support

cfriedt:
`--with-py3` was not working as expected before, so I did a quick investigation and found there was a typo.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -325,7 +325,7 @@ AM_CONDITIONAL(WITH_TWISTED_TEST, [test "$have_trial" = "yes"])
 have_py3="no"
 AX_THRIFT_LIB(py3, [Py3], yes)
 if test "$with_py3" = "yes"; then
-  if $PYTHON --version 2>&1 | grep -q "Python 2"; then
+  if $PYTHON --version 2>&1 | grep -q "Python 3"; then
     AC_PATH_PROGS([PYTHON3], [python3 python3.7 python37 python3.6 python36 python3.5 python35 python3.4 python34])
     if test -n "$PYTHON3"; then
       have_py3="yes"


cfriedt:
I should note that the checks were not successful because of this

```
[00:01:05] Download of win_flex_bison-2.5.18.zip (954.72 KB) completed.
[00:01:08] Hashes match.
[00:01:08] Extracting C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\winflexbison3\2.5.18.20190508\win_flex_bison-2.5.18.zip to C:\ProgramData\chocolatey\lib\winflexbison3\tools...
[00:01:08] C:\ProgramData\chocolatey\lib\winflexbison3\tools
[00:01:09]  ShimGen has successfully created a shim for win_bison.exe
[00:01:09]  ShimGen has successfully created a shim for win_flex.exe
[00:01:09]  The install of winflexbison3 was successful.
[00:01:09]   Software installed to 'C:\ProgramData\chocolatey\lib\winflexbison3\tools'
[00:01:09] 
[00:01:09] Chocolatey installed 1/1 packages. 
[00:01:09]  See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
[00:01:12] Error downloading remote file: One or more errors occurred.
[00:01:12] Inner Exception: Remote server returned 404: Not Found
```
Not because my change broke something.

dcelasun:
Thanks for the PR!

janosvitok:
@cfriedt would you describe in more detail what was the problem with --with-py3? Would you describe also the environment (OS, python distribution, version)?

It seems to me that your change broke CI (i.e. linux with /usr/bin/python = 2.x, /usr/bin/python3=3.x).  If I revert your change, CI is working again, but that breaks --with-py3.

I need to know about your problem so we can fix both problems ( and CI).

Thanks

cfriedt:
@janosvitok  `--with-py3` is there to enable python 3 support.

Prior to this patch, `configure` was failing to detect python3 support and would fail even when explicitly passing `--with-py3`.

Upon inspection, it was pretty obvious that `configure` would only accept a Python 2 version when specifying `--with-py3`, which is not the desired behaviour. To me, it seemed very clearly to be a typo. I only have python3, since python2 is now deprecated, so it was quite obvious that there was a problem.

Have you looked at the diff?

janosvitok:
What I see in docker ubuntu image is: with your patch python3 is not detected.
This is because /usr/bin/python returns Python 2.7.x and the rest is skipped.
In this case configure should check and find /usr/bin/python3 and/or /usr/bin/python3.6

What is your operating system and how did you install python so I can try to reproduce?

It is clear that there are two kinds of setup that are not properly handled:
1. both python 2 and 3 installed (the docker ubuntu case): /usr/bin/python is 2.x, /usr/bin/python3 is 3.x
2. only python 3 is installed (your case): both /usr/bin/python and /usr/bin/python3 are 3.x

Without your patch 1. is handled ok, and 2. not. With your patch vice versa.

janosvitok:
What I see in docker ubuntu image is: with your patch python3 is not detected.
This is because /usr/bin/python returns Python 2.7.x and the rest is skipped.
In this case configure should check and find /usr/bin/python3 and/or /usr/bin/python3.6

What is your operating system and how did you install python so I can try to reproduce?

It is clear that there are two kinds of setup that are not properly handled:
1. both python 2 and 3 installed (the docker ubuntu case): /usr/bin/python is 2.x, /usr/bin/python3 is 3.x
2. only python 3 is installed (your case): both /usr/bin/python and /usr/bin/python3 are 3.x

Without your patch 1. is handled ok, and 2. not. With your patch vice versa.

cfriedt:
I'm just using Ubuntu (bionic) in a docker container. After firing up the container, run
```
apt update
apt install python3
```
Install minimal dependencies of thrift (e.g. build-essential autoconf automake pkg-config libtool, flex, bison, git, etc). Then check out the sources and build.

janosvitok:
@cfriedt: #2055 contains new version of the code. I made two if's: if $python is 2 then search for $python3; and if $python is 3 then $python3 = $python and have_py3="yes"

Now there's code that should always execute under python2... another can of worms :)

PR #1997, opened 2020-01-30
THRIFT-5080: Generate async server implementation for Swift

davidfurey:
Adds additional flags to the compiler to generate an interface which allows swift servers to respond to client requests asynchronously.

This is based on the existing server implementation, and the async client implementation.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_swift_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_swift_generator.cc
@@ -55,6 +55,7 @@ public:
 
     log_unexpected_ = false;
     async_clients_ = false;
+    async_servers_ = false;
     debug_descriptions_ = false;
     no_strict_ = false;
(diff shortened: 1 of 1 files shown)


compiler/cpp/src/thrift/generate/t_swift_generator.cc
```diff
@@ -67,6 +68,11 @@ class t_swift_generator : public t_oop_generator {
         log_unexpected_ = true;
       } else if( iter->first.compare("async_clients") == 0) {
         async_clients_ = true;
+      } else if( iter->first.compare("async_servers") == 0) {
+        if (gen_cocoa_ == true) {
```
Jens-G: Why can't we just write `if (gen_cocoa_ )`?

davidfurey: Good point, thanks.

PR #1999, opened 2020-02-03
[threadFactory] Propagate caught exception instead of rethrowing a copy

gaurav1086:
Propagate exception instead of rethrowing -> In the catch block, use throw; instead of throw e;


allengeorge:
I’m not too knowledgeable about the C++ library, but, that aside: why is this change being made? Is there a JIRA describing the problem and the proposed fix?

gaurav1086:
@allengeorge , thanks for the review. This is a performance fix. There is no jira. Thx.

PR #2000, opened 2020-02-04
THRIFT-5082: Add a Class reference for PHP enum $_TSPEC

spwin:
Issue described in the ticket: https://issues.apache.org/jira/browse/THRIFT-5082

This adds the class reference to PHP `$_TSPEC`

**BEFORE**
```
5 => array(
  'var' => 'group_type',
  'isRequired' => true,
  'type' => TType::I32,
  ),
```

**AFTER**
```
5 => array(
  'var' => 'group_type',
  'isRequired' => true,
  'type' => TType::I32,
  'class' => '\thrift\groups\GroupType',
  ),
```


mmuszynskifln:
Nice!

spwin:
bump

dcelasun:
Thanks for the PR @spwin! Can you please add a test case so this won't regress in the future?

PR #2001, opened 2020-02-05
THRIFT-5083: Reset context on new messages.

ptsneves:
Client: netstd

In TJSONProtocol writeMessageEnd and readMessageBegin potentially can
have context leftovers of the previous calls. Ensure that the leftovers
are cleared so that we are not starting or leaving undefined states.
This is a port of java #3743 to netstd.

This closes #5083


Jens-G:
merged

PR #2007, opened 2020-02-14
THRIFT-5093: lib: cpp: test: clarify effect of MemoryPolicy on TMemoryBuffer

cfriedt:
My main use case with the original code change was that I wanted to be able to call `resetBuffer()` on `OBSERVE` buffers and reuse the contents for test cases over and over. While the default `resetBuffer()` cannot be used for that purpose, the same result can be achieved by calling `resetBuffer(ro_mem, ro_mem_size, OBSERVE)` as well, so the workaround is fairly trivial.

While the former implementation would be more convenient in some cases, it makes more sense to have consistent behaviour for all `resetBuffer()` implementations.

Added test cases to clarify the effect of `MemoryPolicy` on  `TMemoryBuffer` behaviour.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [x] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ ] If your change does not involve any code, add ` [skip ci]` at the end of your pull request to free up build resources.

PR #2030, opened 2020-02-24
THRIFT-5114: Simplified reallocation of TMemoryBuffer

emmenlau:
This PR makes a change to how the new buffer size of TMemoryBuffer is computed. In the current implementation, the new size of TMemoryBuffer is computed in a loop, which makes it less readable. This PR replaces the computation of the new size with an explicit formula.

- [x] Jira ticket at https://issues.apache.org/jira/browse/THRIFT-5114

### DIFF
--- a/lib/cpp/src/thrift/transport/TBufferTransports.cpp
+++ b/lib/cpp/src/thrift/transport/TBufferTransports.cpp
@@ -361,15 +361,17 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
   }
 
   // Grow the buffer as necessary.
-  uint64_t new_size = bufferSize_;
-  while (len > avail) {
-    new_size = new_size > 0 ? new_size * 2 : 1;
-    if (new_size > maxBufferSize_) {
-      throw TTransportException(TTransportException::BAD_ARGS,
-                                "Internal buffer size overflow");
-    }
-    avail = available_write() + (static_cast<uint32_t>(new_size) - bufferSize_);
+  const uint32_t current_used = bufferSize_ - avail;
+  const uint32_t required_buffer_size = len + current_used;
+  if (required_buffer_size > maxBufferSize_) {
+    throw TTransportException(TTransportException::BAD_ARGS,
+                              "Internal buffer size overflow when requesting a buffer of size " + std::to_string(required_buffer_size));
   }
+
+  // Always grow to the next bigger power of two:
+  const double suggested_buffer_size = std::exp2(std::ceil(std::log2(required_buffer_size)));
+  // Unless the power of two exceeds maxBufferSize_:
+  const uint64_t new_size = static_cast<uint64_t>((std::min)(suggested_buffer_size, static_cast<double>(maxBufferSize_)));
 
   // Allocate into a new pointer so we don't bork ours if it fails.
   auto* new_buffer = static_cast<uint8_t*>(std::realloc(buffer_, static_cast<std::size_t>(new_size)));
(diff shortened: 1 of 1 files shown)


Jens-G:
There's also #2007 ... may I ask to cross-review each others change, resolving any potential interferences?

emmenlau:
> There's also #2007 ... may I ask to cross-review each others change, resolving any potential interferences?

The two changes are unrelated. My change is foremost for better readability of the code. My change also introduces a difference in the allocation size of buffer enlargement. The current implementation computes this based on a relatively hard to understand formula. My change always allocates the next larger power of two, so essentially it doubles the buffer whenever the size is exceeded.

I can not say that my allocation policy is generally better. But it is hopefully easier to understand and maintain.

cfriedt:
I also agree #2007 is unrelated.

cfriedt:
Getting pushed back again, sorry.

lib/cpp/src/thrift/transport/TBufferTransports.cpp
```diff
@@ -360,15 +360,14 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
...
-  uint64_t new_size = bufferSize_;
-  while (len > avail) {
-    new_size = new_size > 0 ? new_size * 2 : 1;
-    if (new_size > maxBufferSize_) {
-      throw TTransportException(TTransportException::BAD_ARGS,
-                                "Internal buffer size overflow");
-    }
-    avail = available_write() + (static_cast<uint32_t>(new_size) - bufferSize_);
+  if (new_size > maxBufferSize_) {
+    throw TTransportException(TTransportException::BAD_ARGS,
```
Jens-G, line 369: Scenario: the caluclated new_size is above maxBufferSize, but requured_buffer_size would be in the limit. Now this error is thrown even though it should not, only because of the calculations above.

Questions:
 * is that a possible scenario or may I overlook sth?
 * if yes, would it make sense to simply limit the calculated new_size to maxBufferSize in such a case and move on?

Jens-G: ping

emmenlau: I agree, and I was slightly confused by the complexity and behavior of the old code. If you agree I will further simplify the code and only throw if the requested buffer can not be allocated.

emmenlau:
Pushed an update, and rebased on latest master.

Jens-G:
+1

lib/cpp/src/thrift/transport/TBufferTransports.cpp
```diff
@@ -361,16 +361,18 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
...
-  uint64_t new_size = bufferSize_;
-  while (len > avail) {
-    new_size = new_size > 0 ? new_size * 2 : 1;
-    if (new_size > maxBufferSize_) {
-      throw TTransportException(TTransportException::BAD_ARGS,
-                                "Internal buffer size overflow");
-    }
-    avail = available_write() + (static_cast<uint32_t>(new_size) - bufferSize_);
+  const uint32_t current_used = bufferSize_ - avail;
+  const uint32_t required_buffer_size = len + current_used;
```
Jens-G, line 365: const?

emmenlau: It's one of my clean-code-quirks: I use `const` wherever possible. I think it makes the code more robust, because none of these variables are expected to change in the remainder of the code.

Would you prefer to remove my strong `const`ing? Then I will not continue to use it in Thrift.

Jens-G: No, my fault, I was under the impression that since avail may change the value can't hardly be const but it is not exactly the way I assumed it is. const is fine.

PR #2035, opened 2020-02-26
THRIFT-5118: Fix memory leak when the handler method return a exception

wyjwang:
Client: c_glib
Patch: wangyunjian


wyjwang:
@Jens-G 
Is there any ideas on this bug? It will be closed.
Ping for review, thanks.

Jens-G:
Could you do a rebase?

PR #2037, opened 2020-02-27
THRIFT-5122: Fix memory leak in c_glib tutorial server

wyjwang:
Client: c_glib
Patch: wangyunjian

Signed-off-by: wangyunjian <wangyunjian@huawei.com>

### DIFF
--- a/tutorial/c_glib/c_glib_server.c
+++ b/tutorial/c_glib/c_glib_server.c
@@ -232,15 +232,9 @@ tutorial_calculator_handler_calculate (CalculatorIf      *iface,
          of its type and returning FALSE. */
       *ouch = g_object_new (TYPE_INVALID_OPERATION,
                             "whatOp", op,
-                            "why",  g_strdup ("Cannot divide by 0"),
+                            "why",   "Cannot divide by 0",
                             NULL);
       result = FALSE;
-
-      /* Note the call to g_strdup above: All the memory used by a
-         ThriftStruct's properties belongs to the object itself and
-         will be freed on destruction. Removing this call to g_strdup
-         will lead to a segmentation fault as the object tries to
-         release memory allocated statically to the program. */
     }
     else {
       *_return = num1 / num2;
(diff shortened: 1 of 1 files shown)


tutorial/c_glib/c_glib_server.c
```diff
@@ -232,15 +232,9 @@ tutorial_calculator_handler_calculate (CalculatorIf      *iface,
          of its type and returning FALSE. */
       *ouch = g_object_new (TYPE_INVALID_OPERATION,
                             "whatOp", op,
-                            "why",  g_strdup ("Cannot divide by 0"),
+                            "why",   "Cannot divide by 0",
                             NULL);
       result = FALSE;
-
-      /* Note the call to g_strdup above: All the memory used by a
```
allengeorge: @wyjwang Is this comment not valid any more? IOW, does no one ever free `*ouch`?

Is the actual problem that the processor code should free `*ouch` if it's defined, and that's the source of the leak?

allengeorge: I'm actually very confused about the memory model in the c_glib code: whose responsibility is it to free allocated memory? Is it the generated handling code?

wyjwang: Yes，the generated handling code will malloc memory in its set_property fuction and free memory in its finalize fuction.

static void
invalid_operation_set_property (GObject *object,
                                guint property_id,
                                const GValue *value,
                                GParamSpec *pspec)
{
  InvalidOperation *self = INVALID_OPERATION (object);

  switch (property_id)
  {
    case PROP_INVALID_OPERATION_WHAT_OP:
      self->whatOp = g_value_get_int (value);
      self->__isset_whatOp = TRUE;
      break;

    case PROP_INVALID_OPERATION_WHY:
      if (self->why != NULL)
        g_free (self->why);
      self->why = g_value_dup_string (value); 
      self->__isset_why = TRUE;
      break;

    default:
      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
      break;
  }
}

static void
invalid_operation_finalize (GObject *object)
{
  InvalidOperation *tobject = INVALID_OPERATION (object);

  /* satisfy -Wall in case we don't use tobject */
  THRIFT_UNUSED_VAR (tobject);
  if (tobject->why != NULL) 
  {
    g_free(tobject->why);
    tobject->why = NULL;
  }
}

ulidtko:
@stale undo

PR #2041, opened 2020-03-04
THRIFT-5130 Use Apcu instead of APC

sveneld:
Use Apcu instead of APC

https://pecl.php.net/package/APC does not maintained anymore, so it's better to change using it to  https://pecl.php.net/package/APCu


sveneld:
@Jens-G is it possible to recieve some reaction on pull-request?

PR #2051, opened 2020-03-07
THRIFT-5135: Compiler: Enable C++11 on OSX

janosvitok:
Fix Thrift compile CMAKE compilation by enabling C++11 standard.


janosvitok:
@Jens-G Does it make sense to add travis job to build compiler using cmake on OSX?

janosvitok:
Appveyor failure: timeout during concurrency_test -- not related.

emmenlau:
I think something else must be wrong. This PR should not be required.

See in https://github.com/apache/thrift/blob/dd63da19fad25ff8379f9505e626b16f45593e57/build/cmake/DefineCMakeDefaults.cmake#L66 there is c++11 enabled on all platforms. And its even respecting that you can overwrite the version from the command line, for example using:
```
cmake -DCMAKE_CXX_STANDARD=17 ... other options
```

Jens-G:
Any news here?

janosvitok:
I've checked this. The problem is that when running cmake on the whole tree, DefineCMakeDefaults.cmake is pulled in.
On the other hand, when running cmake just on compiler/cpp subtree, the relevant macro is not set.

See https://github.com/apache/thrift/blob/master/compiler/cpp/README.md#build-using-cmake that explicitly mentions running cmake from compiler/cpp

Summary: the issue here is lack of cmake documentation.

PR #2058, opened 2020-03-13
Updated known cross failures list

Jens-G:
get rid of the netstd SSL cross failure for now (to be fixed later)


janosvitok:
This should have been added to test/features/known_failures_Linux.json

PR #2060, opened 2020-03-14
THRIFT-5067: Invalid generated Cpp code from valid Thrift IDL with dots in file names for nested namespaces

zeshuai007:
This PR fix generated invalid Cpp code if .thrift file with dots in file names for nested namespaces. I replace dots with underscores and it works well.


Jens-G:
You will run into the same problem when you use anything else that gives a valid file name but not a valid identifier. I'll commit this, but it does not really solve the general problem.

PR #2061, opened 2020-03-14
THRIFT-5163 NodeJS: adds Q to exports for browserify

csordasmarton:
When using the *Node.js* libs in the browser (via `browser.js`) `Q` isn't
exposed the same way as it is in a *Node* environment. If I try to call
an API function without a callback function it will throw the following
exception: `TypeError: Cannot read property 'defer' of undefined`. This
just adds `Q` to the exports in `browser.js` to make consuming the libs
more consistent with how we do in an actual *Node* environment.


Jens-G:
Could you please rebase (or just push again) to trigger a new CI run?

csordasmarton:
@Jens-G Yes, sure, I rebased my branch.

PR #2075, opened 2020-03-27
THRIFT-5156: Fix library compilation on Windows and compiler warnings

penenin:
Client: d

  

### DIFF
--- a/lib/d/src/thrift/async/libevent.d
+++ b/lib/d/src/thrift/async/libevent.d
@@ -62,7 +62,7 @@ class TLibeventAsyncManager : TAsyncSocketManager {
     controlReceiveSocket_.blocking = false;
 
     // Register an event for receiving control messages.
-    controlReceiveEvent_ = event_new(eventBase_, cast(int)controlReceiveSocket_.handle,
+    controlReceiveEvent_ = event_new(eventBase_, cast(evutil_socket_t)controlReceiveSocket_.handle,
       EV_READ | EV_PERSIST | EV_ET, assumeNothrow(&controlMsgReceiveCallback),
       cast(void*)this);
     event_add(controlReceiveEvent_, null);
--- a/lib/d/src/thrift/async/libevent.d
+++ b/lib/d/src/thrift/async/libevent.d
@@ -62,7 +62,7 @@ class TLibeventAsyncManager : TAsyncSocketManager {
     controlReceiveSocket_.blocking = false;
 
     // Register an event for receiving control messages.
-    controlReceiveEvent_ = event_new(eventBase_, controlReceiveSocket_.handle,
+    controlReceiveEvent_ = event_new(eventBase_, cast(int)controlReceiveSocket_.handle,
       EV_READ | EV_PERSIST | EV_ET, assumeNothrow(&controlMsgReceiveCallback),
       cast(void*)this);
     event_add(controlReceiveEvent_, null);
@@ -188,7 +188,7 @@ private:
   void addOneshotListenerImpl(Socket socket, TAsyncEventType eventType,
      const(timeval)* timeout, TSocketEventListener listener
   ) {
-    registerOneshotEvent(socket.handle, libeventEventType(eventType),
+    registerOneshotEvent(cast(evutil_socket_t)socket.handle, libeventEventType(eventType),
       assumeNothrow(&socketCallback), timeout, listener);
   }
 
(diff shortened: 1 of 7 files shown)


lib/d/src/thrift/async/libevent.d
```diff
@@ -62,7 +62,7 @@ class TLibeventAsyncManager : TAsyncSocketManager {
     controlReceiveSocket_.blocking = false;
 
     // Register an event for receiving control messages.
-    controlReceiveEvent_ = event_new(eventBase_, controlReceiveSocket_.handle,
+    controlReceiveEvent_ = event_new(eventBase_, cast(int)controlReceiveSocket_.handle,
```
Jens-G: An int is 32 bits, but AFAIK a handle can be 64  bits when compiled into a 64 bt target, right? Is it therefore correct to assume that the cast may be unsafe?

https://stackoverflow.com/questions/1953639/is-it-safe-to-cast-socket-to-int-under-win64
and
https://docs.microsoft.com/de-de/windows/win32/winsock/socket-data-type-2?redirectedfrom=MSDN
> Windows Sockets handles have no restrictions, other than that the value INVALID_SOCKET is 
> not a valid socket. Socket handles may take any value in the range 0 to INVALID_SOCKET–1.

penenin: You are correct about the handle size on 64-bit Windows. For the line in question, I should have used the type alias instead of `int` directly.
```suggestion
    controlReceiveEvent_ = event_new(eventBase_, cast(evutil_socket_t)controlReceiveSocket_.handle,
```
However, I don't think that actually fixes the problem. It looks like the Deimos bindings that we depend on are broken. Looking at lines 168-172 of [util.d](https://github.com/D-Programming-Deimos/libevent/blob/master/deimos/event2/util.d), `evutil_socket_t` is only defined as `intptr_t` on 32-bit Windows. It will be incorrectly defined as `int` for 64-bit Windows. Line 168 should be `version (Windows)` instead of `version (Win32)`.

There is a similar problem with the Deimos OpenSSL binding definition of `SSL_get_fd` that we call in lib/d/src/thrift/transport/ssl.d. On line 1763 of (their) [ssl.d](https://github.com/D-Programming-Deimos/openssl/blob/master/deimos/openssl/ssl.d), `fd` is incorrectly defined on 64-bit Windows as an `int`.

I agree that we should use the correct type aliases where possible because things will be correct if the upstream bindings can be fixed. I am not sure of a better way of handling the `SSL_get_fd` call.

A couple of other questions:
1. What is the policy on supporting old compiler versions? It looks like `core.sys.windows.winsock2` replaced `std.c.windows.winsock` after 2.070.0 and Deimos accounts for this. Do we need to do the same?
2. In addition to the type alias fixes, I fixed an issue with the 32-bit Windows build. I verified that everything builds successfully on 32-bit and 64-bit Windows in addition to 32-bit and 64-bit Windows. Is it sufficient to push the changes up to my branch to update the PR? Can you squash the commits if and when you merge it or do I need to do something different?

Jens-G: (1) Good question. Compatibility to older compiler versions should be kept as long as it makes sense, i.e. there is some significant user base and the burden of keeping compatibility are not too overwhelming or making things unnecessarily complex. Of course, at some point one has to draw the line. That's the general idea, but when it gets down to specifics it may be helpful to check the general sentiment regarding a particular break. Another good indicator could be the "end of life status" of a particular target.

Jens-G: (2) push to the PR branch will suffice and trigger a CI run

Jens-G: (3) To squash or not ... well, I've seen PRs consisting of 4 commits changing 15 files, but all the changes suddenly vanished into thin air after doing a squash, except for 3 files that had the final changes. That's the good case and something the pull requester could have done as well. The bad and the ugly are PRs that are hard to review, because even after sqashing everything there's still 75 or more files with significant changes. So bottom line is twofold: (a) Squash would be nice unless you have the commits designed carefully to transport additional information for the reviewer, but the reviewer can do this as well. (b) Much more important is the set of changes in total vs. the efforts needed to review it. If the PR consists of three otherwise unrelated patches, split them up. If in contrast patch #2 builds upon patch #1, then it may be better to have them in one PR.

penenin: I have pushed up the fixes as discussed. I decided against doing the version check because this work will work for >= 2.070.0 which was released 1/27/2016.

PR #2083, opened 2020-04-02
Make named pipe security configurable by security descriptor string

Breakwell:
### Description
Named pipes on Windows are vulnerable to privilege escalation through named pipe impersonation, to protect against this in Thrift it should allow the security of a pipe to be configured.

### Implementation
Changed setting the pipe permissions to "everyone" to now be configured by security string this will default to "everyone" but can also by set in the constructor of a TPipeServer.

### DIFF
--- a/lib/cpp/src/thrift/transport/TPipeServer.cpp
+++ b/lib/cpp/src/thrift/transport/TPipeServer.cpp
@@ -17,17 +17,18 @@
  * under the License.
  */
 
-#include <thrift/thrift-config.h>
 #include <cstring>
+#include <thrift/thrift-config.h>
 
+#include <boost/noncopyable.hpp>
 #include <thrift/transport/TPipe.h>
 #include <thrift/transport/TPipeServer.h>
-#include <boost/noncopyable.hpp>
 
 #ifdef _WIN32
-#include <thrift/windows/OverlappedSubmissionThread.h>
 #include <AccCtrl.h>
 #include <Aclapi.h>
+#include <sddl.h>
+#include <thrift/windows/OverlappedSubmissionThread.h>
 #endif //_WIN32
 
 namespace apache {
@@ -95,9 +96,15 @@ private:
 
 class TNamedPipeServer : public TPipeServerImpl {
 public:
-  TNamedPipeServer(const std::string& pipename, uint32_t bufsize, uint32_t maxconnections)
-    : stopping_(false), pipename_(pipename), bufsize_(bufsize), maxconns_(maxconnections)
-  {
+  TNamedPipeServer(const std::string& pipename,
+                   uint32_t bufsize,
+                   uint32_t maxconnections,
+                   std::string securityDescriptor)
+    : stopping_(false),
+      pipename_(pipename),
+      bufsize_(bufsize),
+      maxconns_(maxconnections),
+      securityDescriptor_(securityDescriptor) {
     connectOverlap_.action = TOverlappedWorkItem::CONNECT;
     cancelOverlap_.action = TOverlappedWorkItem::CANCELIO;
     TAutoCrit lock(pipe_protect_);
@@ -125,8 +132,8 @@ public:
   virtual HANDLE getNativeWaitHandle() { return listen_event_.h; }
 
 private:
-  bool createNamedPipe(const TAutoCrit &lockProof);
-  void initiateNamedConnect(const TAutoCrit &lockProof);
+  bool createNamedPipe(const TAutoCrit& lockProof);
+  void initiateNamedConnect(const TAutoCrit& lockProof);
 
   TAutoOverlapThread thread_;
   TOverlappedWorkItem connectOverlap_;
@@ -134,6 +141,7 @@ private:
 
   bool stopping_;
   std::string pipename_;
+  std::string securityDescriptor_;
   uint32_t bufsize_;
   uint32_t maxconns_;
   TManualResetEvent listen_event_;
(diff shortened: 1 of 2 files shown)


kainjow:
you should probably make a Jira, this isn't a trivial change.

lib/cpp/src/thrift/transport/TPipeServer.cpp
```diff
@@ -323,73 +344,47 @@ void TPipeServer::close() {
...
+                                     bufsize_,                 // input buffer size
+                                     0,                        // client time-out
+                                     &sa));                    // security attributes
 
   auto lastError = GetLastError();
-  LocalFree(sd);
-  LocalFree(acl);
-  FreeSid(everyone_sid);
+  if (psd)
+    LocalFree(psd);
```
kainjow, line 381: can we use an RAII wrapper instead?

lib/cpp/src/thrift/transport/TPipeServer.h
```diff
@@ -76,6 +85,7 @@ class TPipeServer : public TServerTransport {
   bool getAnonymous();
   void setAnonymous(bool anon);
   void setMaxConnections(uint32_t maxconnections);
+  void setSecurityDescriptor(std::string securityDescriptor);
```
kainjow, line 88: make const reference.

kainjow:
I would suggest backing out the extra style changes to keep the diff minimal and relevant.

lib/cpp/src/thrift/transport/TPipeServer.cpp
```diff
@@ -95,9 +96,15 @@ class TAnonPipeServer : public TPipeServerImpl {
 
 class TNamedPipeServer : public TPipeServerImpl {
 public:
-  TNamedPipeServer(const std::string& pipename, uint32_t bufsize, uint32_t maxconnections)
-    : stopping_(false), pipename_(pipename), bufsize_(bufsize), maxconns_(maxconnections)
-  {
+  TNamedPipeServer(const std::string& pipename,
+                   uint32_t bufsize,
+                   uint32_t maxconnections,
+                   const std::string& securityDescriptor)
```
jeking3, line 102: > to protect against this in Thrift it should allow the security of a pipe to be configured

This forces the issue, and is a breaking change.  It would be better to introduce this functionality while still allowing existing code to compile, and deprecate the old constructors, and then remove them in the following release if we really think it is necessary to force a securityDescriptor in all use cases (I'm not convinced yet).

Breakwell: I agree that this shouldn't be a breaking a change and shouldn't enforce the use of a security descriptor. TNamedPipeServer is internal to TPipeServer. I have given TPipeServer an additional constructor that now takes a security descriptor as a parameter. Now when TPipeServer is constructed unless a security descriptor is provided it has the security descriptor set to "Everyone" to maintain the previous behavior . As TNamedPipeServer  is only in the cpp I don't see how this is a breaking change.

lib/cpp/src/thrift/transport/TPipeServer.cpp
```diff
@@ -400,13 +395,12 @@ bool TAnonPipeServer::createAnonPipe() {
...
-  {
-    GlobalOutput.perror("TPipeServer InitializeSecurityDescriptor (anon) failed, GLE=", GetLastError());
+  if (!InitializeSecurityDescriptor(&sd, SECURITY_DESCRIPTOR_REVISION)) {
+    GlobalOutput.perror("TPipeServer InitializeSecurityDescriptor (anon) failed, GLE=",
+                        GetLastError());
     return false;
   }
-  if (!SetSecurityDescriptorDacl(&sd, true, NULL, false))
-  {
+  if (!SetSecurityDescriptorDacl(&sd, true, NULL, false)) {
```
kainjow, line 403: nit: remove brace line change.

lib/cpp/src/thrift/transport/TPipeServer.cpp
```diff
@@ -400,13 +395,12 @@ bool TAnonPipeServer::createAnonPipe() {
   SECURITY_ATTRIBUTES sa;
   SECURITY_DESCRIPTOR sd; // security information for pipes
 
-  if (!InitializeSecurityDescriptor(&sd, SECURITY_DESCRIPTOR_REVISION))
-  {
-    GlobalOutput.perror("TPipeServer InitializeSecurityDescriptor (anon) failed, GLE=", GetLastError());
+  if (!InitializeSecurityDescriptor(&sd, SECURITY_DESCRIPTOR_REVISION)) {
```
kainjow, line 398: nit: remove unneeded formatting changes.

lib/cpp/src/thrift/transport/TPipeServer.cpp
```diff
@@ -323,73 +344,47 @@ void TPipeServer::close() {
...
+  if (psd)
+    LocalFree(psd);
 
   if (hPipe.h == INVALID_HANDLE_VALUE) {
     Pipe_.reset();
     GlobalOutput.perror("TPipeServer::TCreateNamedPipe() GLE=", lastError);
-    throw TTransportException(TTransportException::NOT_OPEN,
-                              "TCreateNamedPipe() failed",
-                lastError);
+    throw TTransportException(TTransportException::NOT_OPEN, "TCreateNamedPipe() failed",
```
kainjow, line 386: nit: remove unneeded formatting changes

Jens-G:
Not quite sure .... what is tghe status of this PR? Are we done or are there any pending change requests?

lib/cpp/src/thrift/transport/TPipeServer.h
Breakwell: Done.

lib/cpp/src/thrift/transport/TPipeServer.cpp
Breakwell: I feel it's a bit overkill to add a RAII considering we only free once.

lib/cpp/src/thrift/transport/TPipeServer.cpp
Breakwell: This is now inline with the clang format file in the repo. As it's a step in the right direction and doesn't make the PR unclear I think it's fine to leave the formatting change as is.

Breakwell:
> Not quite sure .... what is tghe status of this PR? Are we done or are there any pending change requests?

I have addressed the comments for clarity. I am happy with the changes unless anyone feels particularly strongly I think it's fine as is.

PR #2098, opened 2020-04-11
THRIFT-2245: Allow Custom Prefix in Include Guards

zeshuai007:


kainjow:
What about just using ‘pragma once;’ instead? I think app modern compilers support that now.

zeshuai007:
Although `pragma once` can simplify the code, `#ifndef` is more compatible and portable. 
@Jens-G what do you think.

kainjow:
Since Thrift is already using C++11, do you think there's a compiler that support that but not pragma once?

PR #2101, opened 2020-04-17
Add UnsetHeader to go library

fishy:
Client: go

We already have SetHeader and GetHeader helper functions in the go
library to deal with THeader injected into the context object. But we
didn't provide a way to unset/delete a key from the context object.

This will be useful with the TSimpleServer.SetForwardHeaders API. In the
scenario that a thrift server want to auto forward certain headers to
other upstream thrift servers as the fallback, but during the handling
of the request might decide to remove some of the auto forward headers.
This is also achievable through mutate the write header list, but since
that's a list, finding one key from the list and remove it is much more
hassle.

### DIFF
--- a/lib/go/thrift/header_context.go
+++ b/lib/go/thrift/header_context.go
@@ -44,6 +44,15 @@ func SetHeader(ctx context.Context, key, value string) context.Context {
 	)
 }
 
+// UnsetHeader unsets a previously set header in the context.
+func UnsetHeader(ctx context.Context, key string) context.Context {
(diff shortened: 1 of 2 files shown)


fishy:
@dcelasun I think this is a trivial change so didn't create a jira ticket first. If a jira ticket is still needed please let me know and I can create one.

dcelasun:
LGTM. I'll merge once CI is finished.

fishy:
The only failure in CI seems to be related to js

PR #2102, opened 2020-04-18
THRIFT-5177: getaddrinfo() should not be used for Unix sockets

dmaluka:
Client: cpp

TServerSocket::listen() should not call getaddrinfo() for Unix sockets. The results returned by getaddrinfo() are not used for Unix socket anyway.

Depending on implementation getaddrinfo() may not support Unix sockets properly. In particular, the glibc implementation treats the Unix socket as a network socket and tries to do a bunch of network related stuff with it.

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -261,7 +261,7 @@ void TServerSocket::listen() {
     throw TTransportException(TTransportException::BAD_ARGS, "Specified port is invalid");
   }
 
-  const struct addrinfo *res;
+  const struct addrinfo *res = nullptr;
   int error;
   char port[sizeof("65535")];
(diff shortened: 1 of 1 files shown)


kainjow:
@dmaluka looks like this is causing tests to fail?

dmaluka:
Yep. Looks like I need to change the scope of variables.

dmaluka:
Now it doesn't seem to cause tests to fail. The few remaining failures seem to be caused by some other reasons not related to my change. Could you check?

emmenlau:
@dmaluka This is a nice changeset. To promote your change, maybe just re-trigger the build so that hopefully tests work? With 1 out of 2 builds I'm lucky and get all green lights :-)

dmaluka:
Re-triggered via force-pushing the same change again.

Jens-G:
Thanks @emmenlau for reviewing

PR #2108, opened 2020-04-22
THRIFT-5211: Haskell: Handle incomplete reads correctly.

phile314:
The read functions for handles promise to return *up* to the requested number of bytes. This means in case we read less bytes, we should try again to read some more bytes. This issue caused invalid frame sizes which caused arbitrary decoding failures.

I believe that the bug has been introduced in https://github.com/apache/thrift/commit/3c420072ab5388c2c00d15ada72aec5b061c4d4d

  

### DIFF
--- a/lib/hs/src/Thrift/Transport/Framed.hs
+++ b/lib/hs/src/Thrift/Transport/Framed.hs
@@ -87,11 +87,11 @@ instance Transport t => Transport (FramedTransport t) where
 readFrame :: Transport t => FramedTransport t -> IO Int
 readFrame trans = do
   -- Read and decode the frame size.
-  szBs <- tRead (wrappedTrans trans) 4
+  szBs <- tReadAll (wrappedTrans trans) 4
   let sz = fromIntegral (B.decode szBs :: Int32)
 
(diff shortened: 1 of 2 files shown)


Jens-G:
@phile314 we are looking for people who want to get more involved with Thrift as languange maintainers. Interested?

phile314:
@Jens-G Thanks for reviewing my PR. Regarding becoming a thrift haskell maintainer, as far as I can tell, the Haskell bindings and thrift in general are quite mature and stable. So if it is only the occasional pull request to review, I am happy to do that. If it involves implementing big new features or going on long bug hunts for other peoples issues, then I am less inclined to do so.

zeshuai007:
Hi @Jens-G , I am a Thrift enthusiast, and I'm familiar with several programming languages.
I use Thrift a lot and it's awesome, cause we can flexibly choose the language we need.
During my use of Thrift, I found that it still has many features that could be optimized, such as
more detailed documentation, modern c++ compatibility, test case coverage, etc.
I am very interested in Thrift and hoping to be one of the committers if possible.

PR #2109, opened 2020-04-22
TOutput: Add further parsers for errno-to-string conversion

emmenlau:
This change has mostly an impact on Windows. Currently Thrift does not resolve error strings on Windows (for me). A typical error message may look like `Could not bind: errno = 10049`. With C++11 the newer formatting method `strerror_s()` is available on MSVC, however it fails to resolve many error messages. The most complete error handling on MSVC is possible only via Windows methods.

This PR adds support for decoding error codes on MSVC via `FormatMessageA()` into an ASCII string. In effect, above error id `10049` can be decoded as `Could not bind: The requested address is not valid in its context`.

Furthermore, on platforms that do not have `strerror_r` available (non-Posix systems), this PR turns to `strerror_s` from C++11 to decode the message.

### DIFF
--- a/lib/cpp/src/thrift/TOutput.cpp
+++ b/lib/cpp/src/thrift/TOutput.cpp
@@ -24,6 +24,10 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+#ifdef _WIN32
+#include <windows.h>
+#endif
+
 namespace apache {
 namespace thrift {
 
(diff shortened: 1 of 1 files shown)


emmenlau:
The single error in Travis CI is unrelated. Please review?

dcelasun:
Thanks for the PR @emmenlau!

I'm not familiar with Windows or MSVC but this is at least harmless for other platforms so if nobody objects, I'll go ahead and merge this in 24 hours.

PR #2118, opened 2020-04-24
Add cmake support for C# netstd

emmenlau:
This PR adds cmake support for the C# netstd implementation. The major changes are only in cmake and in the Visual Studio project files. The actual build is executed via `dotnet` (which has to be installed). cmake only orchestrates the build and can install the build artifacts.

A major change is that this PR fixes issue https://issues.apache.org/jira/browse/THRIFT-5143 for me. The required change is that the Thrift build for netstandard 2.0 can not be mixed and mingled with application builds for netstandard 2.0 or .net Framework, due to a missing dependency to reference assembly `System.Threading.Tasks.Extensions, Version=4.2.0.1`. This can be resolved by referencing the Thrift library artifact instead of the Thrift project file. See the Jira issue for a more detailed discussion.

I've tested the build on Windows, macOS and Linux successfully.

   - Full cmake support is in a single commit. There are two other, minor changes to C# code in two other commits in this same PR.


Jens-G:
+1 LGTM but not tested

emmenlau:
The problem with the CI is not easily fixed for me. The issue comes from the fact that the autotools build does not resolve the dependency between tests and the Thrift library in the same way as I encoded it in cmake. A solution would be to encode these inter-dependencies in Thrift.sln.

But sadly I'm not knowledgeable enough in Visual Studio or autotools to fix either of these. Can someone help? If not, I guess this PR will need to wait...

Jens-G:
.

Jens-G:
Rebase still needed

zeshuai007:
It looks like broken the automake for C# netstd,too.

emmenlau:
Updated to latest master, and various minor issues fixed.

Jens-G:
I'm generally positive about using CMake so I can only repeat my +1. 

PS: What does not make me happy is to maintain two build systems, but that's not really the topic here.

Jens-G:
Rebase needed, no actvity since 2022

PR #2124, opened 2020-04-27
THRIFT-5186: Don't pass AI_ADDRCONFIG to getaddrinfo()

ulidtko:
Hi!

Please see the [Jira issue][THRIFT-5186] for rationale. Basically, this patch enables Thrift clients & servers to continue working in "offline mode" — when only the loopback interace `lo` (127.0.0.1) is available, nothing more.

With Thrift 0.13.0 on POSIX, C++ and Python clients/servers **crash with exception** in those conditions.

This patch is used internally in our team, to compile a *vendored* version of Thrift. We'd like to contribute this patch upstream, to relieve ourselves from the maintenance burden of vendoring.

Code review, comments, suggestions welcome!

[THRIFT-5186]: https://issues.apache.org/jira/browse/THRIFT-5186

NB: no attempt was made to cover this patch with tests. Suggestions on how to simulate "no network" environment in the test suite would be highly appreciated.

------------------------------------------------------------------------------------------

- [x] Did you create an Apache Jira ticket?  (not required for trivial changes)

### DIFF
--- a/lib/cpp/src/thrift/transport/TNonblockingServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TNonblockingServerSocket.cpp
@@ -191,7 +191,7 @@ void TNonblockingServerSocket::listen() {
   std::memset(&hints, 0, sizeof(hints));
   hints.ai_family = PF_UNSPEC;
   hints.ai_socktype = SOCK_STREAM;
-  hints.ai_flags = AI_PASSIVE | AI_ADDRCONFIG;
+  hints.ai_flags = AI_PASSIVE;
 
   // If address is not specified use wildcard address (NULL)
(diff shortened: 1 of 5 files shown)


janosvitok:
I did a quick search and lib/delphi/src/Thrift.Socket.pas contains the last instance of AI_ADDRCONFIG. I suppose this should be removed as well. @Jens-G?

ulidtko:
Yea, indeed. Thanks! I've amended the patch.

ulidtko:
Oh well, Travis fell down this time.

The only indication of whats wrong in the logs:
> make[4]: *** [all-local] Segmentation fault (core dumped)

That's kind of interesting... But looks like a fluke. Let me just retry.

ulidtko:
Yep... I was right. It indeed was a fluke; all green :heavy_check_mark:  after restart.

PR #2125, opened 2020-04-28
THRIFT-5190: StringUtils haven't take `(offset + length) > bytes.length` into account

dugenkui03:
1. StringUtils haven't take `(offset + length) > bytes.length` into account;

2. `if ((offset | length | (offset + length) | (size - (offset + length))) < 0)` will faster than `>`

- [x]  [Apache Jira-THRIFT-5190](https://issues.apache.org/jira/browse/THRIFT-5190) 

### DIFF
--- a/lib/java/src/org/apache/thrift/utils/StringUtils.java
+++ b/lib/java/src/org/apache/thrift/utils/StringUtils.java
@@ -55,6 +55,9 @@ public final class StringUtils {
     if (offset < 0) {
       throw new IndexOutOfBoundsException("Negative start offset " + offset);
     }
+    if (length > bytes.length - offset) {
+      throw new IndexOutOfBoundsException("Invalid range, bytes.length: " + bytes.length + " offset: " + offset + " length: " + length);
+    }
     char[] chars = new char[length * 2];
     for (int i = 0; i < length; i++) {
       int unsignedInt = bytes[i + offset] & 0xFF;
--- a/lib/java/test/org/apache/thrift/utils/TestStringUtils.java
+++ b/lib/java/test/org/apache/thrift/utils/TestStringUtils.java
@@ -20,6 +20,7 @@
 package org.apache.thrift.utils;
 
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 
 public class TestStringUtils {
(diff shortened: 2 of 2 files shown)


lib/java/src/org/apache/thrift/utils/StringUtils.java
```diff
@@ -49,12 +49,10 @@ public static String bytesToHexString(byte[] bytes) {
    * @return hex string.
    */
   public static String bytesToHexString(byte[] bytes, int offset, int length) {
-    if (length < 0) {
-      throw new IllegalArgumentException("Negative length " + length);
-    }
-    if (offset < 0) {
-      throw new IndexOutOfBoundsException("Negative start offset " + offset);
+    if ((offset | length | (offset + length) | (bytes.length - (offset + length))) < 0) {
```
Jens-G, line 52: To be honest,I personally have a problem with code like this where you need to read it three times to get it right (and I'm still not fully sure I "groked" it). I leave it to someone else to give a final review statement. +0 from me stince I won't be standing in the way, but I'm not really in favor of it.

Jens-G: A testcase for this change would be a great addition.

joaopedroantonio:
```suggestion
    if (offset < 0 || bytes.length < (offset + length)) {
```

And maybe we should keep the IllegalArgumentException when length is less than 0?

ctubbsii: I'm responding to the request on the mailing list for another pair of eyes on this PR.

I agree with @Jens-G regarding the readability... readability is far more important than an every-so-slight performance benefit. Obfuscated code is very hard to maintain, even with good test cases.

@joaopedroantonio 's suggestions are a great improvement, and preserves existing behavior, while adding the bounds checks when using offset and length.

A unit test case that covers the edge cases would be a nice sanity check for this method.

dugenkui03: > if ((offset | length | (offset + length) | (bytes.length - (offset + length))) < 0)

when length is less than 0, it will throw IllegalArgumentException, such as `int offset=2,length=-1,byteLength=10;`

dugenkui03: it seem that there will always be readability problem when using bit operation.

I do agree with that the readability of code is important. The code will be changed and squashed to a single commit.

Thanks very much.

joaopedroantonio: Maybe I'm being picky here but I think we should keep the IllegalArgumentException when `length < 0`, for two reasons:
1. The semantics of IndexOutOfBoundsException and the IllegalArgumentException are different and I think `length < 0` is a case of the latter.
2. Neither of those exceptions is checked and changing the type of exception when `length < 0` will not be explicit in compile time. I'd consider this a breaking change in the API.

WYT?

dugenkui03: I haven't notice that I have removed IllegalArgumentException instead of IndexOutOfBoundsException, and I haven't notice that it will be a breaking change in the API.

great suggestion, thanks very much.

lib/java/test/org/apache/thrift/utils/TestStringUtils.java
```diff
@@ -31,4 +31,27 @@ public void testToHexString() {
     Assert.assertEquals("EFAB92", StringUtils.bytesToHexString(bytes, 2, 3));
     Assert.assertNull(StringUtils.bytesToHexString(null));
   }
+
+  @Test
+  public void testToHexStringIndexOutOfBoundsExceptionExamine() {
+    byte[] bytes = {1, 2, 3, 4, 5};
+    try {
+      StringUtils.bytesToHexString(bytes, 0, -1);
+    } catch (Exception e) {
```
Jens-G, line 40: The (hypothetical?) case when it would not throw is not covered. Same below.

PR #2127, opened 2020-04-30
THRIFT-5187: Added Win32 support for domain sockets (AF_UNIX)

emmenlau:
This PR adds support for domain sockets for Windows.

In order to implement this change I've changed a basic property of C++ build configuration for Windows. This would not be strictly required but is quite helpful. Domain sockets are only available on recent versions of the Windows platform SDK. Therefore I've enabled the automatic configuration of `config.h` with cmake. I've checked the results and on my platform(s) the generated config header is sensible and matches the hard-coded default. However this change is not easily portable to autotools.

During the implementation I found a number of minor issues in TServerSocket and TSocket that I believe are improved in this PR. None of these changes should be a breaking change. Mostly some methods should not be called on a domain socket and have been excluded. Also, the variable `listening_` was set to `true` at the beginning of the `listen()` method but I found that this opens a small time window where `listening_` is true but the socket is not actually listening (yet).

Please review and let me know if there is anything to discuss?

I've tested this implementation successfully on MSVC 2017 and 2019, on Ubuntu 18.04 x86_64 and on MacOS 14 with XCode 11.3.

- [x] I created Apache Jira Ticket https://issues.apache.org/jira/browse/THRIFT-5187

### DIFF
--- a/build/cmake/ConfigureChecks.cmake
+++ b/build/cmake/ConfigureChecks.cmake
@@ -21,6 +21,9 @@ include(CheckFunctionExists)
 include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckSymbolExists)
+include(CheckStructHasMember)
+include(CheckCSourceCompiles)
+include(CheckCXXSourceCompiles)
 
 if (Inttypes_FOUND)
   # This allows the inttypes.h and stdint.h checks to succeed on platforms that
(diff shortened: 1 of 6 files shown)


emmenlau:
I took special care to make implement TSocket.cpp and TServerSocket very consistent. Where possible I reduced inconsistencies rather than adding any.

build/cmake/ConfigureChecks.cmake
```diff
@@ -52,14 +55,24 @@ check_include_file(sched.h HAVE_SCHED_H)
...
+check_cxx_source_compiles(
+  "
+  #define WIN32_LEAN_AND_MEAN
+  #include <windows.h>
+  #include <winsock2.h>
+  #include <afunix.h>
+  int main(){(void)sizeof(((struct sockaddr_un *)0)->sun_path); return 0;}
+  "
+  HAVE_AF_UNIX_H)
+#check_struct_has_member("struct sockaddr_un" sun_path "afunix.h" HAVE_AF_UNIX_H LANGUAGE CXX)
```
kainjow, line 67: Is this commented-out line needed?

lib/cpp/src/thrift/transport/TSocket.cpp
```diff
@@ -700,6 +689,10 @@ int TSocket::getPort() {
   return port_;
 }
 
+std::string TSocket::getPath() {
```
kainjow, line 692: could this be `const`?

lib/cpp/src/thrift/transport/TSocket.cpp
```diff
@@ -792,6 +789,13 @@ void TSocket::setKeepAlive(bool keepAlive) {
     return;
   }
 
+#ifdef _WIN32
+  if (!path_.empty()) {
+      // Windows Domain sockets do not support SO_KEEPALIVE.
```
kainjow, line 794: should this be a config.h setting instead?

Jens-G:
.

emmenlau:
I'll rebase this on latest master. Dear stale bot, please reopen :-)

emmenlau:
I've rebased this issue on latest master. Could someone please re-open so the changes become visible?

build/cmake/ConfigureChecks.cmake
emmenlau: Thanks! Removed in latest PR.

lib/cpp/src/thrift/transport/TSocket.cpp
emmenlau: Thanks, fixed in latest PR.

PR #2134, opened 2020-05-08
THRIFT-5199: Fix infinite loop writing to closed TSocket in PHP

theilig:
Client: php

When you call fwrite in PHP with a handle that has been closed by peer the function will return 0 bytes written (rather than false to indicate an error).

This causes an infinite loop today because the code assumes 0 only means a temporary failure to write, rather than a permanent one. 

This change adds a check to for a closed socket when 0 bytes are written

- [ X] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [ X] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [ X] Did you squash your changes to a single commit?  (not required, but preferred)
- [ X] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?

### DIFF
--- a/lib/php/lib/Transport/TSocket.php
+++ b/lib/php/lib/Transport/TSocket.php
@@ -328,7 +328,8 @@ class TSocket extends TTransport
             if ($writable > 0) {
                 // write buffer to stream
                 $written = fwrite($this->handle_, $buf);
-                if ($written === -1 || $written === false) {
+                $closed_socket = $written === 0 && feof($this->handle_);
+                if ($written === -1 || $written === false || $closed_socket) {
                     throw new TTransportException(
                         'TSocket: Could not write ' . TStringFuncFactory::create()->strlen($buf) . ' bytes ' .
                         $this->host_ . ':' . $this->port_


lib/php/lib/Transport/TSocket.php
```diff
@@ -328,7 +328,8 @@ public function write($buf)
             if ($writable > 0) {
                 // write buffer to stream
                 $written = fwrite($this->handle_, $buf);
-                if ($written === -1 || $written === false) {
+                $closed_socket = $written === 0 && feof($this->handle_);
+                if ($written === -1 || $written === false || $closed_socket) {
                     throw new TTransportException(
```
Jens-G, line 333: $written === false

Not your code, but I always wonder if a language has no boolean operators when I see code like this.

Jens-G:
Plus, I found this statement in the comments at 

> After having **problems with fwrite() returning 0 in cases where one would fully expect a return value of false**, I took a look at the source code for php's fwrite() itself. The function will only return false if you pass in invalid arguments. **Any other error, just as a broken pipe or closed connection, will result in a return value of less than strlen($string), in most cases 0.**

If that is true then both the docs and all code relying on it are broken.

So why don't we just check for $written being less/equal 0 instead of `=== -1`) ? Wouldn't that suffice?

theilig:
I saw that comment on php.net and verified it was true (both by looking at the c implementation of php's fwrite, and by writing a simple test.

I believe that checking for written <= 0 would work and would be more robust.  If there is a closed duplex socket where data was sent from the peer before closing then fwrite will return 0 (the socket is closed) but feof will return false because there is still data to read.  

In the patch we are using locally I didn't go that route for two reasons:

1. I'm don't know enough about the internals of sockets to know if there are some cases where a write of 0 is temporary and a subsequent write would be successful. 
2. The implementation of TSocket::read uses the 0 length + feof check so I erred on the side of consistency.

`if ($readable > 0) {
            $data = fread($this->handle_, $len);
            if ($data === false) {
                throw new TTransportException('TSocket: Could not read ' . $len . ' bytes from ' .
                    $this->host_ . ':' . $this->port_);
            } elseif ($data == '' && feof($this->handle_)) {
                throw new TTransportException('TSocket read 0 bytes');
            }

            return $data;
        }`

I'm pretty sure that throwing an exception on 0 bytes in both cases without the feof check would be correct.  We just decided to be really conservative with our initial change as we make tens of thousands of thrift calls per second.

I can attest that adding the feof check has fixed our issue with stuck php processes.

PR #2141, opened 2020-05-12
CMake: Added better dependency tracking for external build tools

emmenlau:
When building components that use an external build tool (like Java gradle or Javascript grunt), cmake does currently not track if a rebuild is required. This PR creates a cmake-dependency between the sources and the target of these components "Java" and "Javascript". Assuming that all sources are fully contained in their respective component folder, cmake can infer if a require is required (i.e. the target is missing or one or more sources are more current). This drastically reduces build times.

However please be aware that if the dependency between sources and target should *not* be complete for some reason, then cmake may miss to rebuild a component. I could not observe this problem but it may be well advised to carefully check this PR.

### DIFF
--- a/lib/java/CMakeLists.txt
+++ b/lib/java/CMakeLists.txt
@@ -48,19 +48,25 @@ else()
         set(PRELEASE "false")
     endif ()
 
-    add_custom_target(ThriftJava ALL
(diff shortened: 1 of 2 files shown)


Jens-G:
.

emmenlau:
I've just rebased on latest master, please consider for merging.

emmenlau:
Rebased on latest master. If CI remains green I would merge this PR.

PR #2142, opened 2020-05-12
THRIFT-5127: Fix race conditions in TNonblockingServer

ajakubek:
Client: cpp

Several race conditions were present in TNonblockingServer when
TNonblockingServer::stop() method was called concurrently with initialization of
the server (which takes place in TNonblockingServer::serve()).

 1. Thrift server uses a bidirectional pipe to notify its IO threads that they
    should exit.
    The pipe was created in the TNonblockingIOThread::registerEvents() method,
    which is called by the IO thread itself.
    Since the pipe was created _after_ the thread which was supposed to receive
    the message, there was a short time window, when such request could be
    undelivered.
    In that case, the thread would never exit and Thrift could prevent the whole
    application from terminating cleanly.
    To fix this issue, the call to createNotificationPipe() has been moved to
    the constructor of TNonblockingIOThread.
    Breaking change: the TNonblockingIOThread constructor may now throw an
    exception if pipe creation fails. However, the only caller of that
    method (TNonblockingServer::registerEvents()) could already throw due to
    unrelated errors.

 2. Accesses to TNonblockingServer::ioThreads_ vector were unsynchronized.
    In most cases this was irrelevant, because the vector is modified only
    during initialization and destruction of the server.
    However, in one scenario, creation of new threads (which may result in
    reallocation of the vector) could race with iteration over the container in
    TNonblockingServer::stop() method.
    This issue has been fixed by wrapping the code in a critical section.

  

### DIFF
--- a/lib/cpp/src/thrift/server/TNonblockingServer.cpp
+++ b/lib/cpp/src/thrift/server/TNonblockingServer.cpp
@@ -1077,6 +1077,8 @@ void TNonblockingServer::expireClose(std::shared_ptr<Runnable> task) {
 }
 
 void TNonblockingServer::stop() {
+  Guard g(ioThreadsMutex_);
+
   // Breaks the event loop in all threads so that they end ASAP.
   for (auto & ioThread : ioThreads_) {
     ioThread->stop();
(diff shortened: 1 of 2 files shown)


Jens-G:
,

PR #2145, opened 2020-05-13
Make TTransportException unwrappable on go1.13+

fishy:
Client: go

Go 1.13 introduced a new, optional, hidden interface for error
implementations to make them unwrappable [1]. We currently already kind
of support that (via TTransportException.Err), so just add a new
function to make it Go 1.13+ compatible.

I know thrift go library itself hasn't dropped support for pre-go1.13
yet, but this enables users already on go1.13 to do error inspections
more easily.

I didn't implement it for PrependError, because git grep shows that it's
currently only used by test code.

[1] https://pkg.go.dev/errors@go1.13?tab=doc#pkg-overview

  

### DIFF
--- a/lib/go/thrift/transport_exception.go
+++ b/lib/go/thrift/transport_exception.go
@@ -60,6 +60,10 @@ func (p *tTransportException) Err() error {
 	return p.err
 }
 
+func (p *tTransportException) Unwrap() error {
+	return p.err
+}
+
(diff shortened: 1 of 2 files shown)


fishy:
@dcelasun I didn't create a JIRA ticket because I think this qualifies as a trivial change, but please let me know if you think a JIRA ticket is needed :)

dcelasun:
Looks good!

PR #2153, opened 2020-05-22
THRIFT-5214: Connectivity check on go's TSocket

fishy:
Client: go

Implement connectivity check on go's TSocket and TSSLSocket for
non-Windows systems.

The implementation is inspired by
https://github.blog/2020-05-20-three-bugs-in-the-go-mysql-driver//

  

### DIFF
--- a/lib/go/thrift/socket.go
+++ b/lib/go/thrift/socket.go
@@ -26,7 +26,7 @@ import (
 )
 
 type TSocket struct {
-	conn           net.Conn
+	conn           *socketConn
 	addr           net.Addr
 	connectTimeout time.Duration
 	socketTimeout  time.Duration
@@ -58,7 +58,7 @@ func NewTSocketFromAddrTimeout(addr net.Addr, connTimeout time.Duration, soTimeo
 
 // Creates a TSocket from an existing net.Conn
 func NewTSocketFromConnTimeout(conn net.Conn, connTimeout time.Duration) *TSocket {
-	return &TSocket{conn: conn, addr: conn.RemoteAddr(), connectTimeout: connTimeout, socketTimeout: connTimeout}
+	return &TSocket{conn: wrapSocketConn(conn), addr: conn.RemoteAddr(), connectTimeout: connTimeout, socketTimeout: connTimeout}
 }
 
 // Sets the connect timeout
@@ -89,7 +89,7 @@ func (p *TSocket) pushDeadline(read, write bool) {
 
 // Connects the socket, creating a new socket object if necessary.
 func (p *TSocket) Open() error {
-	if p.IsOpen() {
+	if p.conn.isValid() {
 		return NewTTransportException(ALREADY_OPEN, "Socket already connected.")
 	}
 	if p.addr == nil {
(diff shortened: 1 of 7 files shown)


fishy:
@dcelasun I did some tweak on the implementation and this version runs much better, and I think if you have time to take a look at the code, it's ready for review :)

I didn't mark it as ready for review because I still want to test it in our real prod first (currently it's running in our staging server). I'll mark it as "ready for review" once I'm satisfied with the prod test result. Which should happen before end of day Tuesday in the US.

fishy:
I lied. I didn't really test it in our prod environment (that's too risky), but I did manage to get two staging services, one running new code and one running old code, to run side-by-side for 24+ hours. In our staging setup all staging services are behind the same load balancer pool, so they are not handling exactly the same requests, but it should still be comparable:

1. Latency wise, sometimes the new code has slightly higher latency, and sometimes the old code has slightly higher latency, but overall they are quite similar (but new code seem to be tend to be the slightly slower one more often)
2. Error rate wise, it's similar to latency that it's not consistently one is better than the other, but new code does appear to be lower more often.

So I guess the conclusion we can get from there is that the new code might make latency slightly higher, but should help client pools a bit, and overall it doesn't make things worse?

Gonna mark it as "Ready for review".

fishy:
Another option we could consider is to make this feature optional, either by adding a new constructor and/or field to `TSocket`/`TSSLSocket`, or add another `TSocket` implementation (`TMonitoredSocket` maybe?)

dcelasun:
> new code seem to be tend to be the slightly slower one more often

I think a little overhead here is acceptable.

> overall it doesn't make things worse?

Agreed and generally LGTM but this PR [breaks](https://travis-ci.org/github/apache/thrift/jobs/690199684) cross [tests](https://travis-ci.org/github/apache/thrift/jobs/690199686) against C++:

```
Following 1 tests were expected to cleanly succeed but needed retry:
go-cpp                  binary            buffered-ip              flaky(1 retry)
*** Following 1 failures were unexpected ***:
go-cpp                  binary            framed-ip                failure(timeout)
Unexpected failures are logged to test/log/unexpected_failures.log
You can browse results at:
	file:///thrift/src/test/index.html
# If you use Chrome, run:
# 	cd /thrift/src
#	python -m http.server 8001
# then browse:
# 	http://localhost:8001/test/
Full log for each test is here:
	test/log/server_client_protocol_transport_client.log
	test/log/server_client_protocol_transport_server.log
1 failed of 2081 tests in total.
Test execution took 951.7 seconds.
Sun May 24 08:52:03 2020
Makefile:1130: recipe for target 'cross-'(binary)'' failed
make: *** [cross-'(binary)'] Error 1

RET=$?
if [ $RET -ne 0 ]; then
  if [ -f "test/features/log/unexpected_failures.log" ]; then 
    cat "test/features/log/unexpected_failures.log"
  fi
  if [ -f "test/log/unexpected_failures.log" ]; then
    cat "test/log/unexpected_failures.log"
  fi
fi
TEST: [go-cpp_binary_framed-ip]
*************************** server message ***************************
Sun May 24 08:39:03 2020
Executing: /thrift/src/test/go/bin/testserver --certPath=../../keys --protocol=binary --transport=framed --port=33477
Directory: /thrift/src/test/go/bin
config:delay: 5
config:timeout: 8
2020/05/24 08:39:03 error processing request: Connection not open
```

fishy:
@dcelasun the breaks you linked was from travis run of a previous version. I changed the code to check `isValid` instead of `IsOpen` before every Read/Write, and fixed that issue.

In the travis run of the current version, there's only one build failures and it's related to github unavailable: https://travis-ci.org/github/apache/thrift/jobs/690775140

dcelasun:
Sorry, my bad. I had Travis open on another tab and didn't realize it was an older build.

> Another option we could consider is to make this feature optional

I'm strongly against optional features like this, they just increase our maintenance burden. If there are any significant regressions (though I don't see how) we'll address them afterwards.

If there is nothing further to add, I'll go ahead and merge this once tests are green.

fishy:
In that case let me squash the two commits. I put the fixup one separately so it's easier for you to tell what is changed :)

PR #2157, opened 2020-05-26
THRIFT-4987: Fix TBinaryProtocol support in XHRClient

MAXruppert:
Client: nodejs

The XHRClient uses the XMLHttpRequest web api to do requests. To correctly interpret binary data, the `responseType` needs to be set to `arraybuffer`, and the `response` property should be used instead of `responseText` (`response` defaults to normal text when not specified). See [here](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest).

I'd like to add tests but am unsure what the best practice would be for this with regards to the project testing framework? It would involve bundling the `nodejs` lib into a browser for testing, or perhaps running a headless browser environment in nodejs? I'd be happy to add tests, but would greatly appreciate if someone could point me in the correct direction.

I'm succesfully using this fix with a C++ HTTP server using the binary protocol for my use case.

---
  
### DIFF
--- a/lib/nodejs/lib/thrift/xhr_connection.js
+++ b/lib/nodejs/lib/thrift/xhr_connection.js
@@ -22,6 +22,7 @@ var thrift = require('./thrift');
 
 var TBufferedTransport = require('./buffered_transport');
 var TJSONProtocol = require('./json_protocol');
+var TBinaryProtocol = require('./binary_protocol');
 var InputBufferUnderrunError = require('./input_buffer_underrun_error');
 
 var createClient = require('./create_client');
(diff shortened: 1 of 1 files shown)


emmenlau:
This seems like a valuable change. And while I'm not any expert in JS, it looks sufficiently small to raise at least some trust, i.e. it mostly adds new behavior for the binary use case that was not handled before. The critical change seems to be:
https://github.com/apache/thrift/pull/2157/files#diff-29759a12f6599c6aeca4c0ae88af5291R116

Could someone test it, or should me merge it on best trust and wait for user feedback?

nithinkdb:
Any chance this change could be merged? My team has a use-case that should be done via XHR, but we are needlessly using WebSocket because of this bug.

PR #2163, opened 2020-06-02
THRIFT-5224: Deprecated Nodejs Buffer()

zeshuai007:
This PR fix issue THRIFT-5224.


keptt:
That's a pity that this PR was closed. The deprecation warning is still a problem and haunts people who try to use libraries dependent on thrift. This leads to either having to turn the deprecation warnings off for everything, which is not ideal or to bundle the code and patch the bundle via something like custom esbuild plugins. For an example, see: https://github.com/databricks/databricks-sql-nodejs/issues/293

keptt:
and btw, I think plain Buffer constructor is also used in a deprecated [node-int64](https://github.com/broofa/node-int64) dependency of thrift - this might also need some attention

PR #2165, opened 2020-06-02
Deprecated WinXP and Server2003 methods

emmenlau:
This PR adds a very minor change: it deprecates handling of WinXP and Windows Server 2003 in `lib/cpp/src/thrift/windows/WinFcntl.{h,cpp}`. This simplifies code maintenance. I'm under the impression that support for WinXP should be dropped now.

Does this require an explicit mention in a change log? I would assume that WinXP and Server 2003 are not officially supported since a while now?


Jens-G:
> Does this require an explicit mention in a change log? 
> I would assume that WinXP and Server 2003 are not officially supported since a while now?

I'd still mention it. You're right but technically it is still a breaking change.

Jens-G:
+1 LGTM but needs a rebase

Jens-G:
>  I would assume that WinXP and Server 2003 are not officially supported since a while now?
There is no explicit policy in place but folllowing the support cycles given by Microsoft has proven to be a good working model. So I'd poropse we do the same here.

Jens-G:
BTW what's the state of the matter? You know what I mean ... ;-) Can we help somehow?

emmenlau:
I've rebased on latest master, and added the change to CHANGES.md. Does it look good like this?

emmenlau:
Checked, the build errors in Travis are all unrelated. Ready for review and merge...

emmenlau:
Thanks @Jens-G !

PR #2167, opened 2020-06-02
Thrift.csproj: Stop writing AssemblyAttributes.cs to /tmp

emmenlau:
The netstd build writes so-called `TargetFrameworkMonikerAssemblyAttributes` files to `/tmp`. These files can be left behind, and they do not feature a build-specific or user-specific name. This can be a problem in a multi-user environment, because the files block future builds by other users (that can not overwrite the files).

This PR adds a workaround by placing the files into the build directory rather than the `/tmp` directory. There should be no negative effects, other than a new, extra line in the `Thrift.csproj` settings.


Jens-G:
Could you force-push? There were a bunch of (probably unrelated) curl errors.

PR #2169, opened 2020-06-03
THRIFT-4731: Replace usage of deprecated network modules in haskell library

phile314:
Upgrade to network 3.0.0.0. Note that the `Network` module has been removed in version 3.0.0.0 and hence the thrift library no longer compiles with the current network package.

This is a breaking change, but it should be fairly easy to update call sites.

Related issues:
https://issues.apache.org/jira/browse/THRIFT-4731
https://issues.apache.org/jira/browse/THRIFT-5098
https://issues.apache.org/jira/browse/THRIFT-5193 

  

### DIFF
--- a/lib/hs/src/Thrift/Server.hs
+++ b/lib/hs/src/Thrift/Server.hs
@@ -27,7 +27,7 @@ import Control.Concurrent ( forkIO )
 import Control.Exception
 import Control.Monad ( forever, when )
 
-import Network
+import Network.Socket
 
 import System.IO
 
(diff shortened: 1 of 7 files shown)


lib/hs/src/Thrift/Server.hs
```diff
@@ -42,25 +42,41 @@ runThreadedServer :: (Protocol i, Protocol o)
                   => (Socket -> IO (i, o))
                   -> h
                   -> (h -> (i, o) -> IO Bool)
-                  -> PortID
+                  -> Int
                   -> IO a
 runThreadedServer accepter hand proc_ port = do
-    socket <- listenOn port
-    acceptLoop (accepter socket) (proc_ hand)
+    let hints = defaultHints {
```
phile314, line 48: This is basically copy-paste from the network library how-to on how to listen on a socket.

lib/hs/src/Thrift/Transport/Handle.hs
```diff
@@ -62,17 +63,12 @@ instance Transport Handle where
     tWrite = LBS.hPut
     tFlush = hFlush
 
-
--- | Type class for all types that can open a Handle. This class is used to
--- replace tOpen in the Transport type class.
-class HandleSource s where
```
phile314, line 68: This class does not look really useful. I have removed it, but kept a `hOpenSocket` function to make the common use case of connecting to a socket concise.

lib/hs/thrift.cabal
```diff
@@ -32,19 +32,12 @@ License-File:   LICENSE
...
-
 Library
   Hs-Source-Dirs:
     src
   Build-Depends:
-    base >= 4, base < 5, containers, ghc-prim, attoparsec, binary, bytestring >= 0.10, base64-bytestring, hashable, HTTP, text, hspec-core > 2.4.0, unordered-containers >= 0.2.6, vector >= 0.10.12.2, QuickCheck >= 2.8.2, split
-  if flag(network-uri)
-     build-depends: network-uri >= 2.6, network >= 2.6 && < 3.0
-  else
-     build-depends: network < 2.6
```
phile314, line 47: The network/network-uri split has already been a few years in the past, so I think it is safe to drop that compatibility code.

phile314:
I noticed now that this broke the cross-language haskell tests. I tried to get those tests running locally, but it looks like running them using a recent GHC/cabal version breaks a lot of things ....

Not sure what the best thing to do is ....

phile314:
Hm, I am completely confused now if I should use cmake or configure && make. There seems to be a cmake file inside `test/hs`, but the instructions in the top-level README refer to make. Can anybody help me?

zeshuai007:
Hi @phile314 ,both `cmake` and `make` can work, you can choose your favorite build method.But it seems that `cmake` is gradually becoming a mainstream build method.

phile314:
The tests are passing now, this is ready for review.

PR #2171, opened 2020-06-07
THRIFT-5230: Fix connection leak and CancelledKeyException when handl…

zengji:
please refer to https://issues.apache.org/jira/browse/THRIFT-5230


Jens-G:
Seems not needed then ...

PR #2173, opened 2020-06-07
Use io.CopyN to simplify THeaderTransport code

fishy:
Client: go

In THeaderTransport implementation, use io.CopyN instead of
io.Copy+io.LimitReader.

Underlying io.CopyN is actually implemented with io.Copy+io.LimitReader
[0], but it also does some extra checks we can take advantage of. It
also simplifies the code in thrift repo.

[0] https://github.com/golang/go/blob/83b181c68bf332ac7948f145f33d128377a09c42/src/io/io.go#L340

  


fishy:
@dcelasun :)

PR #2176, opened 2020-06-08
THRIFT-5214: Reset read deadline in socketConn

fishy:
Client: go

This is a slightly different, and less error-prone approach from the
fix in e382275.

The previous approach relies on passing the set socket timeout into the
underlying socketConn from TSocket and TSSLSocket. But since we have so
many different constructors for TSocket and TSSLSocket, some makes the
initial connection in the constructor and some does not, there are so
many different places we would need to remember to pass socketTimeout
into socketConn. In the future, when we add another constructor to them,
we could either forget to pass the socket timeout into socketConn, or
try to pass it while we haven't constructed socketConn yet (which will
cause panic), both are bad.

In this approach we just clear the read deadline in the connectivity
check read. Because that's a non-blocking read, it would work just fine
without a read deadline.

  

### DIFF
--- a/lib/go/thrift/socket.go
+++ b/lib/go/thrift/socket.go
@@ -58,9 +58,7 @@ func NewTSocketFromAddrTimeout(addr net.Addr, connTimeout time.Duration, soTimeo
 
 // Creates a TSocket from an existing net.Conn
 func NewTSocketFromConnTimeout(conn net.Conn, connTimeout time.Duration) *TSocket {
-	sock := &TSocket{conn: wrapSocketConn(conn), addr: conn.RemoteAddr(), connectTimeout: connTimeout, socketTimeout: connTimeout}
-	sock.conn.socketTimeout = connTimeout
-	return sock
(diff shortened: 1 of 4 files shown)


fishy:
@dcelasun

dcelasun:
This does seem cleaner. I'm not a fan of relying on comments for correct behaviour but that seems like a good trade-off here.

fishy:
I also added more comment on `socketConn.IsOpen` to explicitly call out the side-effect (I thought I pushed that last night, but apparently I forgot to do `git commit` before `git push` so push did nothing 😬)

In most cases there shouldn't be anything between `pushDeadline` and actual `Read`, as the whole purpose of doing `pushDeadline` is to set the deadline used by the `Read`. But I just added the comment there in case in the future someone tries to do something without realizing the side-effect. I agree that is not ideal but I don't know if there's any better solutions here.

dcelasun:
> I agree that is not ideal but I don't know if there's any better solutions here.

I can't think of anything either. Anyway, LGTM.

PR #2177, opened 2020-06-09
THRIFT-4926: Adding LOGGER.isDebugEnabled() conditions into TSaslTransport

em010272:
Adding LOGGER.isDebugEnabled() conditional statements into org.apache.thrift.transport.thrift.transport.TSaslTransport to fix the issue 4926

In org.apache.thrift.transport.TSaslTransport in the previous approach, sensitive information about Role is leaked. Thus, the LOGGER.isDebugEnabled() conditional statements should be added in the method public void open() throws TTransportException.


belugabehr:
No changes provided address any kind of "leakage".  If the logging level is not set to DEBUG, then a DEBUG statement will not be emitted.  The `LOGGER.debug()` method checks internally for the appropriate condition.  The method `isDebugEnabled` is available only for performance reasons.

Jens-G:
Patch rejected. See ticket for details.

PR #2194, opened 2020-07-04
remove unused import

dugenkui03:
remove unused import in `TBase`.


ctubbsii:
FWIW, the [impsort-maven-plugin](https://code.revelc.net/impsort-maven-plugin/) is great for sorting imports, and removing unused ones.

dugenkui03:
> FWIW, the [impsort-maven-plugin](https://code.revelc.net/impsort-maven-plugin/) is great for sorting imports, and removing unused ones.

yes, convenient tool. I just think that the source code will more readable.

Jens-G:
.

PR #2198, opened 2020-07-10
Create TMap/TList/TSet object after checkContainerReadLength

dugenkui03:
Create `TMap/TList/TSet` object when it is nessary.


Jens-G:
Could you force-push again?

dugenkui03:
> Could you force-push again?

There is just one single commit. How to force-push again?

PR #2199, opened 2020-07-13
set pos_ = 0 and endPos_ = 0 when invoke clear() in TMemoryInputTransport.java

dugenkui03:
set `pos_ = 0` and `endPos_ = 0` when invoke `clear()` in TMemoryInputTransport.java


Jens-G:
,

PR #2209, opened 2020-07-29
THRIFT-5257: Fix Go THeaderTransport endOfFrame handling

fishy:
Client: go

In the current implementation, we only call endOfFrame when we hit EOF
when reading from the frameReader. The problem is in go stdlib the Read
call finished reading the remaining data from frameReader will not
return EOF, the next Read will. This caused us in most cases only call
endOfFrame at the beginning of the next frame, which could cause
troubles because we didn't read the beginning of the frame properly.

  


fishy:
Added a test and confirmed that the test would fail without the fix and passes after the fix.

PR #2211, opened 2020-08-03
THRIFT-5259 Fix "Transport endpoint is not connected " error when run…

longzhiri:
…ning lua test server accepts a connection

Client: lua

the removed code  would recv on a socket  returned by listen() in a lua server and return a error, which i think it's useless.

  

### DIFF
--- a/lib/lua/src/usocket.c
+++ b/lib/lua/src/usocket.c
@@ -89,12 +89,6 @@ T_ERRCODE socket_wait(p_socket sock, int mode, int timeout) {
     return TIMEOUT;
   }
 
-  // Verify that we can actually read from the remote host
-  if (mode & WAIT_MODE_C && FD_ISSET(*sock, &rfds) &&
-      recv(*sock, (char*) &rfds, 0, 0) != 0) {
-    return errno;
-  }
-
   return SUCCESS;
 }
 


zeshuai007:
// Verify that we can actually read from the remote host
  if (mode & WAIT_MODE_C && FD_ISSET(*sock, &rfds) &&
      recv(*sock, (char*) &rfds, 0, 0) != 0) {
    return errno;
  }
After testing, it seems that only the socket_acceptfunction will fail in this judgment. Is it better to only let the socket_accept function not enter this judgment?

longzhiri:
The select() system call returns success already indicating read()/write()/accpet() is ready, there is no need to call recv() to verify. Non-blocking connect() is an exception, after checking for writability  with select(), you should also check for the socket errors. As man 2 connect says:
```bash
 EINPROGRESS
              The  socket is nonblocking and the connection cannot be completed immediately.  It is possible to select(2) or poll(2) for completion by selecting the socket for writing.
              After select(2) indicates writability, use getsockopt(2) to read the SO_ERROR option at level SOL_SOCKET to determine whether connect() completed  successfully  (SO_ERROR
              is zero) or unsuccessfully (SO_ERROR is one of the usual error codes listed here, explaining the reason for the failure).
```

zeshuai007:
Oh,I see.

zeshuai007 (approved):
LGTM.

PR #2222, opened 2020-08-29
THRIFT-5270: Fix wrong int to string conversions

fishy:
Client: go

Starting from go 1.15, go test starts to complain about wrongly used int
to string conversions:

    ./field.go:58:83: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
    ./numeric.go:72:12: conversion from int64 to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
    ./json_protocol_test.go:612:92: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
    ./simple_json_protocol_test.go:685:96: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)

json_protocol_test and simple_json_protocol_test usages are actually
already in format arg so just remove the conversion is good enough.
field is no longer used anywhere so just removed it (there's one line of
commented out code in compact_protocol so remove that line as well). The
one in numeric.go is actually a bug. We didn't set sValue correctly in
NewNumericFromI64 and NewNumericFromI32 functions.

  


dcelasun (approved):
LGTM

PR #2231, opened 2020-09-08
THRIFT-5274 - Thrift 0.13.0 does not work with JDK8

coheigea:
Thrift 0.13.0 does not work with JDK8, as it is compiled with JDK11. Even though the source version is JDK8, incorrect byte code is generated for various ByteBuffer methods, e.g.:

Exception in thread "TAsyncClientManager#SelectorThread 22" java.lang.NoSuchMethodError: java.nio.ByteBuffer.rewind()Ljava/nio/ByteBuffer;

The fix is to manually cast ByteBuffer to Buffer (or else switch to using the xenial docker image to create the release artifacts). See this task for more context: https://github.com/eclipse/jetty.project/issues/3244

### DIFF
--- a/lib/java/src/org/apache/thrift/TBaseHelper.java
+++ b/lib/java/src/org/apache/thrift/TBaseHelper.java
@@ -18,6 +18,7 @@
 package org.apache.thrift;
 
 import java.io.Serializable;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.util.Comparator;
 import java.util.Iterator;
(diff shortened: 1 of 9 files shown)


coheigea:
Hi @zeshuai007 ,

Yes you can reproduce as follows:

1. git clone https://github.com/apache/camel
2. Edit parent/pom.xml and change libthrift-version from 0.12.0 to 0.13.0
3. mvn clean install -f parent/pom.xml
4. mvn clean install -f components/camel-thrift/pom.xml

If you run the last command with JDK11 it works perfectly, otherwise you start seeing the errors as above, e.g.:
Exception in thread "TAsyncClientManager#SelectorThread 27" java.lang.NoSuchMethodError: java.nio.ByteBuffer.rewind()Ljava/nio/ByteBuffer;
	at org.apache.thrift.async.TAsyncMethodCall.doWritingRequestBody(TAsyncMethodCall.java:263)
	at org.apache.thrift.async.TAsyncMethodCall.transition(TAsyncMethodCall.java:195)

coheigea:
@zeshuai007 Will this PR be considered for 0.14.0? Or can you at least make sure that released artifacts are built with Java 8 to avoid this problem?

ctubbsii:
A much simpler fix would be to add `-release 8` to the `javac` command line when compiling with JDK9 and later.

PR #2232, opened 2020-09-09
TServerSocket.cpp: Ensure the server is really listening

emmenlau:
Currently the C++ server does not ensure that `listening_` is `true` in `isOpen()`. I think this is incorrect because in multi-threaded environments, the socket may not be open or listening yet. Also, `listening_` should only be set to `true` when the socket is fully set up and listening.

This PR adds a minor change: it respects `listening_` for `isOpen()`, and sets `listening_` to `true` only at the end of spcket initialization. In heavy stress tests, we found this makes the server method `isOpen()` more reliable.

Thanks for consideration.

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -167,7 +167,13 @@ TServerSocket::~TServerSocket() {
 }
 
 bool TServerSocket::isOpen() const {
-  return (serverSocket_ != THRIFT_INVALID_SOCKET);
+  if (serverSocket_ == THRIFT_INVALID_SOCKET)
+    return false;
+
+  if (!listening_)
+    return false;
+
+  return true;
 }
 
 void TServerSocket::setSendTimeout(int sendTimeout) {
@@ -324,7 +330,6 @@ void TServerSocket::_setup_tcp_sockopts() {
 }
 
 void TServerSocket::listen() {
-  listening_ = true;
 #ifdef _WIN32
   TWinsockSingleton::create();
 #endif // _WIN32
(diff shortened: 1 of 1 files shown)


ulidtko (approved):
LGTM — eventhough, I think, people should pick the `TNonBlockingServer` for multithreaded environments, as it has at least some synchronization and is in general more suitable for threadpools and the like.

emmenlau:
> LGTM — eventhough, I think, people should pick the `TNonBlockingServer` for multithreaded environments, as it has at least some synchronization and is in general more suitable for threadpools and the like.

I'm very curious about this! We did not try it yet, was it a simple drop-in replacement for you? It uses libevent, correct? Did you compare it with the new fbthrift c++ server?

emmenlau:
Checked, the build errors in Travis are all unrelated. Ready for review and merge...

ulidtko:
> was it a simple drop-in replacement for you?

Not sure about this...

> It uses libevent, correct?

Better to check the source, but I think not. It just expects to be used from multiple threads, and does appropriate locking.

> Did you compare it with the new fbthrift c++ server?

No; what's that?

emmenlau:
> > Did you compare it with the new fbthrift c++ server?
> 
> No; what's that?

Apparently Facebook has continued development on thrift in a separate project. I think it has less features and supports less languages, but it seems to have a cool new async c++ server. I've always wondered if this server could not be backported to Apache thrift, but that seems to involve complicated discussion of licenses. But for end-users, things may be simpler: possibly the server could just be used with Apache thrift if one would manage to take it out and build it against Apache thrift. No idea if this is possible, though :-) 
See https://github.com/facebook/fbthrift/blob/master/thrift/doc/Cpp2.md for more details.

PR #2234, opened 2020-09-09
Improve setting the domain socket path into the sockaddr_un struct in TSockets

emmenlau:
The code that sets a domain socket path in `TServerSocket.cpp` and `TSocket.cpp` is currently slightly more complex than required (at least that is my understanding). Also it seems to contain a potential out-of-bounds-access.

I simplified the code in the following way:
 - Ensure that the `sockaddr_un struct` is fully set to zero before using it
   - This removes the necessity to deal with zero-termination of the path. It also removes the need for the `len` variable
 - Do not use `path_.size() + 1` length when `memcpy` the path from `path_` to the `sockaddr_un struct`
   - This may be an out-of-bounds-access in the current code, depending on whether `path_` actually contains an additional zero-terminating character (which is unlikely)

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -398,8 +398,7 @@ void TServerSocket::listen() {
     _setup_unixdomain_sockopts();
 
 #ifndef _WIN32
-    size_t len = path_.size() + 1;
-    if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
+    if ((path_.size() + 1) > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
       errno_copy = THRIFT_GET_SOCKET_ERROR;
       GlobalOutput.perror("TSocket::listen() Unix Domain socket path too long", errno_copy);
       throw TTransportException(TTransportException::NOT_OPEN,
@@ -408,16 +407,13 @@ void TServerSocket::listen() {
     }
 
     struct sockaddr_un address;
+    // Store the zero-terminated path in address.sun_path
+    memset(&address, '\0', sizeof(address));
     address.sun_family = AF_UNIX;
-    memcpy(address.sun_path, path_.c_str(), len);
-
-    auto structlen = static_cast<socklen_t>(sizeof(address));
+    memcpy(address.sun_path, path_.c_str(), path_.size());
 
     if (!address.sun_path[0]) { // abstract namespace socket
-#ifdef __linux__
-      // sun_path is not null-terminated in this case and structlen determines its length
-      structlen -= sizeof(address.sun_path) - len;
-#else
+#ifndef __linux__
       GlobalOutput.perror("TSocket::open() Abstract Namespace Domain sockets only supported on linux: ", -99);
       throw TTransportException(TTransportException::NOT_OPEN,
                                 " Abstract Namespace Domain socket path not supported");
@@ -425,7 +421,7 @@ void TServerSocket::listen() {
     }
 
     do {
-      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, structlen)) {
+      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, static_cast<socklen_t>(sizeof(struct sockaddr_un)))) {
         break;
       }
       errno_copy = THRIFT_GET_SOCKET_ERROR;
--- a/lib/cpp/src/thrift/transport/TSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSocket.cpp
@@ -329,31 +329,27 @@ void TSocket::openConnection(struct addrinfo* res) {
   if (!path_.empty()) {
 
 #ifndef _WIN32
-    size_t len = path_.size() + 1;
-    if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
+    if ((path_.size() + 1) > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
       int errno_copy = THRIFT_GET_SOCKET_ERROR;
       GlobalOutput.perror("TSocket::open() Unix Domain socket path too long", errno_copy);
       throw TTransportException(TTransportException::NOT_OPEN, " Unix Domain socket path too long");
     }
 
     struct sockaddr_un address;
+    // Store the zero-terminated path in address.sun_path
+    memset(&address, '\0', sizeof(address));
     address.sun_family = AF_UNIX;
-    memcpy(address.sun_path, path_.c_str(), len);
-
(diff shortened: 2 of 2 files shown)


lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -408,24 +407,21 @@ void TServerSocket::listen() {
...
+#ifndef __linux__
       GlobalOutput.perror("TSocket::open() Abstract Namespace Domain sockets only supported on linux: ", -99);
       throw TTransportException(TTransportException::NOT_OPEN,
                                 " Abstract Namespace Domain socket path not supported");
 #endif
     }
 
     do {
-      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, structlen)) {
+      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, static_cast<socklen_t>(sizeof(address)))) {
```
ulidtko, line 424: I'm a bit wary about this `sizeof(address)` — because of how `struct sockaddr` can have more data following it. What type is `address` again?..

`man 7 unix` says this:
>    **Pathname sockets**
>       When binding a socket to a pathname, a few rules should be observed for maximum portability and ease of coding:
>
>   *  The pathname in sun_path should be null-terminated.
>   *  The length of the pathname, including the terminating null byte, should not exceed the size of sun_path.
>   *  The addrlen argument that describes the enclosing sockaddr_un structure should have a value of at least:
>
>             offsetof(struct sockaddr_un, sun_path)+strlen(addr.sun_path)+1
>
> or, more simply, addrlen can be specified as `sizeof(struct sockaddr_un)`.

Ah, so I see, line 409, it's exactly `sockaddr_un`. So all's good.

How about maybe:
```suggestion
      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, static_cast<socklen_t>(sizeof(struct sockaddr_un)))) {
```

? And similarly below

emmenlau: After re-considering #2235, I think this PR is wrong. I'm not handling the abstract socket paths correctly since they indeed may require `structlen` rather than `static_cast<socklen_t>(sizeof(struct sockaddr_un))`. What do you think?

emmenlau: Ok, refreshing my mind with https://stackoverflow.com/a/11640827/7200132, I take back the previous comment. It does work as I proposed, and zero-ing the full path is correct and necessary. So I'll consider your suggestion!

ulidtko: well I didn't check this in headers, but the manpage defines it as:
```
           struct sockaddr_un {
               sa_family_t sun_family;               /* AF_UNIX */
               char        sun_path[108];            /* Pathname */
           };
```

so it has fixed space for the socket path. I'm not sure the previous code was correct — but this PR seems to be OK. Also, you run tests with this patch, right?

lib/cpp/src/thrift/transport/TSocket.cpp
```diff
@@ -329,31 +329,27 @@ void TSocket::openConnection(struct addrinfo* res) {
...
-    memcpy(address.sun_path, path_.c_str(), len);
-
-    auto structlen = static_cast<socklen_t>(sizeof(address));
+    memcpy(address.sun_path, path_.c_str(), path_.size());
 
     if (!address.sun_path[0]) { // abstract namespace socket
-#ifdef __linux__
-      // sun_path is not null-terminated in this case and structlen determines its length
-      structlen -= sizeof(address.sun_path) - len;
-#else
```
ulidtko, line 349: I think I get it.

This code would "truncate" the unused bytes from `address.sun_path` (which'd include the null terminator BTW).

This may have worked — but `unix(7)` clearly says to pass `sizeof(struct sockaddr_un)`, the full thing, without any truncation business.

ulidtko (approved):
This addresses the null-termination concern nicely (via `memset`), better conforms to `unix(7)` and also obviates funny manipulations with lengths.

LGTM.

lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -408,24 +407,21 @@ void TServerSocket::listen() {
...
+#ifndef __linux__
       GlobalOutput.perror("TSocket::open() Abstract Namespace Domain sockets only supported on linux: ", -99);
       throw TTransportException(TTransportException::NOT_OPEN,
                                 " Abstract Namespace Domain socket path not supported");
 #endif
     }
 
     do {
-      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, structlen)) {
+      if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, static_cast<socklen_t>(sizeof(struct sockaddr_un)))) {
```
deiv, line 424: I think, this will break the abstract unix socket type name.

emmenlau: Can you elaborate further if you observed problems with this? I had the same problems as https://stackoverflow.com/a/11640827/7200132 and problems went away when switching to above solution. This was tested on Ubuntu 18.04 and CentOS 7.

deiv: As the manual page says:

> The socket's address in this namespace is given by the
> additional bytes in sun_path that are covered by the specified
> length of the address structure.  (Null bytes in the name have no
> special significance.)

For abstract socket you need to pass the length of the string (without the null termination), plus sockaddr_un.sun_family size.

`static_cast<socklen_t>(sizeof(struct sockaddr_un)))` I think that will give the entire struct size

ulidtko: Null termination, whether needed or not, is covered by the `memset`. Zero out the whole structure before use — that's the best way to go. So, null termination is *obviously correct* here.

Now, I think @deiv has a point too. The `addrlen` argument to `bind` may be incorrect. @emmenlau can you check if with this patch socket names in `/proc/net/unix` contain unexpected zero-bytes? Tests would still work this way I'm sure, but it's better to double-check this.

deiv: Yes, as I say, the problem is the size not the null terminating character. But if you prefer to fill it with zeros, without needing it, go ahead !!

ulidtko (changes requested):
As #2235 shows, I've LGTM'd this too soon.

The 3rd argument to `bind()`, the `addrlen`, should be computed differently. Depending on the requested socket type, `addrlen` should have a `+ 1` and null terminator for "pathname" sockets, and no `+1` neither terminator for "abstract" sockets.

Further, the `sun_path` space check also needs `+ 1` or `+0`, for pathname and abstract sockets respectively.

emmenlau:
This commit is now superseded by #2235

PR #2235, opened 2020-09-09
Fix abstract unix socket name

deiv:
For the abstract unix socket address type, the string in the
'sun_path' field of the 'sockaddr_un struct', is a not null-terminated
string (see unix(7)).

Fix the lentgh calculation of the 'sun_path' field to don't add
the termination null byte.

### DIFF
--- a/lib/cpp/CMakeLists.txt
+++ b/lib/cpp/CMakeLists.txt
@@ -59,6 +59,7 @@ set( thriftcpp_SOURCES
    src/thrift/transport/TBufferTransports.cpp
    src/thrift/transport/TWebSocketServer.h
    src/thrift/transport/TWebSocketServer.cpp
+   src/thrift/transport/SocketCommon.cpp
    src/thrift/server/TConnectedClient.cpp
    src/thrift/server/TServerFramework.cpp
    src/thrift/server/TSimpleServer.cpp
--- a/lib/cpp/Makefile.am
+++ b/lib/cpp/Makefile.am
@@ -88,6 +88,7 @@ libthrift_la_SOURCES = src/thrift/TApplicationException.cpp \
                        src/thrift/transport/TTransportUtils.cpp \
                        src/thrift/transport/TBufferTransports.cpp \
                        src/thrift/transport/TWebSocketServer.cpp \
+                       src/thrift/transport/SocketCommon.cpp \
                        src/thrift/server/TConnectedClient.cpp \
                        src/thrift/server/TServer.cpp \
                        src/thrift/server/TServerFramework.cpp \
@@ -206,7 +207,8 @@ include_transport_HEADERS = \
                          src/thrift/transport/TBufferTransports.h \
                          src/thrift/transport/TShortReadTransport.h \
                          src/thrift/transport/TZlibTransport.h \
-                         src/thrift/transport/TWebSocketServer.h
+                         src/thrift/transport/TWebSocketServer.h \
+                         src/thrift/transport/SocketCommon.h
 
 include_serverdir = $(include_thriftdir)/server
 include_server_HEADERS = \
--- a/lib/cpp/src/thrift/transport/SocketCommon.cpp
+++ b/lib/cpp/src/thrift/transport/SocketCommon.cpp
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ * @author: David Suárez <david.sephirot@gmail.com>
+ */
+
+#ifndef THRIFT_SOCKETCOMMON_H
+#define THRIFT_SOCKETCOMMON_H
+
+#ifndef _WIN32
+
+#include <thrift/thrift-config.h>
+
+#ifdef HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+
+#ifdef HAVE_SYS_UN_H
+#include <sys/un.h>
+#endif
+
+#include <string>
+
+#include <thrift/transport/PlatformSocket.h>
+#include <thrift/transport/TTransportException.h>
+#include <thrift/TOutput.h>
+
+namespace apache {
+namespace thrift {
+namespace transport {
+
+
+socklen_t fillUnixSocketAddr(struct sockaddr_un& address, std::string& path)
+{
+    // abstract namespace socket ?
+    bool isAbstractNamespace = path[0] == 0;
+
+#ifndef __linux__
+    if (isAbstractNamespace) {
+      GlobalOutput.perror("TSocket::open() Abstract Namespace Domain sockets only supported on linux: ", -99);
+      throw TTransportException(TTransportException::NOT_OPEN,
+                                " Abstract Namespace Domain socket path not supported");
+    }
+#endif
+
+    /*
+     * For abstract namespace sockets, the path string is not null-terminated (as opposite to path based), so we
+     *   rely in pass the string size, to the bind() call.
+     */
+    size_t addr_len = isAbstractNamespace ? path.size() : path.size() + 1;
+
+    if (addr_len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
+        int errno_copy = THRIFT_GET_SOCKET_ERROR;
+        GlobalOutput.perror("TSocket::open() Unix Domain socket path too long", errno_copy);
+        throw TTransportException(TTransportException::NOT_OPEN, " Unix Domain socket path too long");
+    }
+
+    address.sun_family = AF_UNIX;
+    memcpy(address.sun_path, path.c_str(), addr_len);
+
+    return static_cast<socklen_t>(sizeof((sockaddr_un*)nullptr)->sun_family + addr_len);
+}
+
+}
+}
+} // apache::thrift::transport
+
+#endif // _WIN32
+
+#endif //THRIFT_SOCKETCOMMON_H
--- a/lib/cpp/src/thrift/transport/SocketCommon.h
+++ b/lib/cpp/src/thrift/transport/SocketCommon.h
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ * @author: David Suárez <david.sephirot@gmail.com>
+ */
+
+#ifndef THRIFT_SOCKETCOMMON_H
+#define THRIFT_SOCKETCOMMON_H
+
+#ifndef _WIN32
+
+#include <thrift/thrift-config.h>
+
+#ifdef HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+
+#ifdef HAVE_SYS_UN_H
+#include <sys/un.h>
+#endif
+
+#include <string>
+
+namespace apache {
+namespace thrift {
+namespace transport {
+
+socklen_t fillUnixSocketAddr(struct sockaddr_un& address, std::string& path);
+
+}
+}
+} // apache::thrift::transport
+
+#endif // _WIN32
+
+#endif //THRIFT_SOCKETCOMMON_H
--- a/lib/cpp/src/thrift/transport/TNonblockingServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TNonblockingServerSocket.cpp
@@ -48,6 +48,7 @@
 #include <thrift/transport/TNonblockingServerSocket.h>
 #include <thrift/transport/TSocket.h>
 #include <thrift/transport/TSocketUtils.h>
+#include <thrift/transport/SocketCommon.h>
 
 #ifndef AF_LOCAL
 #define AF_LOCAL AF_UNIX
@@ -350,33 +351,14 @@ void TNonblockingServerSocket::listen() {
     _setup_sockopts();
     //_setup_unixdomain_sockopts();
 
+/*
+ * TODO: seems that windows now support unix sockets,
+ *       see: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/
+ */
 #ifndef _WIN32
 
-    size_t len = path_.size() + 1;
-    if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
-      errno_copy = THRIFT_GET_SOCKET_ERROR;
-      GlobalOutput.perror("TSocket::listen() Unix Domain socket path too long", errno_copy);
-      throw TTransportException(TTransportException::NOT_OPEN,
-                                "Unix Domain socket path too long",
-                                errno_copy);
-    }
-
     struct sockaddr_un address;
-    address.sun_family = AF_UNIX;
-    memcpy(address.sun_path, path_.c_str(), len);
-
-    auto structlen = static_cast<socklen_t>(sizeof(address));
-
-    if (!address.sun_path[0]) { // abstract namespace socket
-#ifdef __linux__
-      // sun_path is not null-terminated in this case and structlen determines its length
-      structlen -= sizeof(address.sun_path) - len;
-#else
-      GlobalOutput.perror("TSocket::open() Abstract Namespace Domain sockets only supported on linux: ", -99);
-      throw TTransportException(TTransportException::NOT_OPEN,
-                                " Abstract Namespace Domain socket path not supported");
-#endif
-    }
+    socklen_t structlen = fillUnixSocketAddr(address, path_);
 
     do {
       if (0 == ::bind(serverSocket_, (struct sockaddr*)&address, structlen)) {
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -48,6 +48,7 @@
 #include <thrift/transport/TServerSocket.h>
 #include <thrift/transport/TSocket.h>
 #include <thrift/transport/TSocketUtils.h>
+#include <thrift/transport/SocketCommon.h>
 
 #ifndef AF_LOCAL
 #define AF_LOCAL AF_UNIX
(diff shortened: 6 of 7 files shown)


emmenlau:
But `len` is already `path_.size() + 1`, does that not compensate for the null-termination?

ulidtko:
This is going to conflict with #2234 — @deiv would you help with review there?

deiv:
I think is the other way around:

The `path_.size() + 1` adds 1 to compesate for the null termination needed in the "pathname-based" unix socket type.

If for example, you create an abstract unix socket server in thtift:
```c++
...
new TServerSocket("\0abstract-socket");
...
```
Searching for it in '/proc' `cat /proc/net/unix  | grep abstract-', gives:
```bash
... @abstract-socket@
```

The last '@' is extra, due to the null char `path_.c_str(), len)`, return a null terminated string.

ulidtko:
That actually makes sense. Hmmm...

But regardless, I'd not be happy with the patch itself. ` structlen -= sizeof(address.sun_path) - len + 1;` — this approaches the comprehension limit for me (read: *this guy is stupid, doesn't grok the elite `-=-` operator*).

Could this be solved in a more clear way? Y'know, such that it'd be obviously-correct.

deiv:
@ulidtko 

The other way to do it, is calculating the len of the string based on the type of socket:

pathname socket: null terminated
abstract socket: not null terminated

This lead to an if in some point of the code. I prefer to do an aritmethic operation instead of adding a conditional jump (for a cpu point of view, the arithmethic way will not lead to control hazards).

ulidtko:
OK, so we have:
 * pathname socket — `'\0'` is last
 * abstract socket — `'\0'` is first

The `.sun_path` length is `path_.size() + 1` in both cases. This needs no `if`-s...

I mean: the `structlen -= …` line is bad, one needs to think really hard to understand it completely. While the PR needs to change it anyway — why not improve it? The `+ 1` you're adding does fix *the issue*, but makes *the code* worse.

deiv:
> OK, so we have:
> 
> * pathname socket — `'\0'` is last
> * abstract socket — `'\0'` is first
> 
> The `.sun_path` length is `path_.size() + 1` in both cases. This needs no `if`-s...
> 

* pathname socket — `'\0'` is last -> we need `path_.size() + 1`
* abstract socket — `'\0'` is first -> we don't need `path_.size() + 1` (we need only `path_.size() `

Take into account that, for the `memcpy(address.sun_path, path_.c_str(), len)` thing, we are using `c_str()' that returns an null terminated string 

> I mean: the `structlen -= …` line is bad, one needs to think really hard to understand it completely. While the PR needs to change it anyway — why not improve it? The `+ 1` you're adding does fix _the issue_, but makes _the code_ worse.

Maybe a comment could do the job ?

ulidtko:
> we need only path_.size() 

so you're saying, this would be correct call?
```
sockaddr_un addr;
addr.sun_family = AF_UNIX;
addr.sun_path[0] = '\0';
addr.sun_path[1] = 'A';
addr.sun_path[2] = 'B';
addr.sun_path[3] = 'C';

bind(sockfd, &addr, 3);
//                  ^ here, 3? not 4?
```

> Maybe a comment could do the job ?

As you can see, there's already one. Is it helpful?

ulidtko:
> we are using `c_str()' that returns an null terminated string

`c_str()` does return a null-terminated C string... but the `memcpy` is not copying it into the `sun_path`, due to `memcpy(address.sun_path, path_.c_str(), path_.size());` — `.size()` does not include the last null.

deiv:
> > we need only path_.size()
> 
> so you're saying, this would be correct call?
> 
> ```
> sockaddr_un addr;
> addr.sun_family = AF_UNIX;
> addr.sun_path[0] = '\0';
> addr.sun_path[1] = 'A';
> addr.sun_path[2] = 'B';
> addr.sun_path[3] = 'C';
> 
> bind(sockfd, &addr, 3);
> //                  ^ here, 3? not 4?
> ```

Not, is 4, but when you will do `"\0ABC".size() + 1`, it will gives 5.

> > Maybe a comment could do the job ?
> 
> As you can see, there's already one. Is it helpful?

The comment is:

>       // sun_path is not null-terminated in this case and structlen determines its length

Yes, is helpful, it's give the correct way to do.

But the code was not the correct, because it uses the variable `len` that is the length of the string plus the terminating caracter \0.

deiv:
> memcpy(address.sun_path, path_.c_str(), path_.size());

is `memcpy(address.sun_path, path_.c_str(), path_.size());`

emmenlau:
> > memcpy(address.sun_path, path_.c_str(), path_.size());
> 
> is `memcpy(address.sun_path, path_.c_str(), len)`

I think its better to copy only `path_size()` instead of `len` because it avoids copying a char that is not in the original string.

deiv:
> > we are using `c_str()' that returns an null terminated string
> 
> `c_str()` does return a null-terminated C string... but the `memcpy` is not copying it into the `sun_path`, due to `memcpy(address.sun_path, path_.c_str(), path_.size());` — `.size()` does not include the last null.
> 
> *edit: sorry I confused this from the other PR. code in master does
> 
> ```
>     size_t len = path_.size() + 1;
>     // ...
>     memcpy(address.sun_path, path_.c_str(), len);
> ```
> 
> so it's fine.

It's fine for the pathname socket (len includes the null). For abstract socket id not the correct, we don't need to count the null.But if you prefer to copy an additional char, go ahead. The problem is the size passed in `bind` call.

emmenlau:
Let me see if I get everything right:
 - From my side, it would be fine to pass the shorter struct length in case of abstract sockets. For me it worked to use the _full_ length, but I think it should be fine to use the shorter length.
 - The `memcpy` and `memset` are good for everybody? So we copy only the required bytes but ensure the full struct is zero'ed before use.

So the "only" challenge is to agree on a good style to compute `structlen`, correct? I personally would prefer something like:
```cpp
#ifdef __linux__
      // sun_path is not null-terminated in this case, and structlen must be shortened to the path size
      structlen = static_cast<socklen_t>(sizeof(address)) - sizeof(address.sun_path) + path_.size();
#else
```

deiv:
> Let me see if I get everything right:
> 
> * From my side, it would be fine to pass the shorter struct length in case of abstract sockets. For me it worked to use the _full_ length, but I think it should be fine to use the shorter length.
> * The `memcpy` and `memset` are good for everybody? So we copy only the required bytes but ensure the full struct is zero'ed before use.
> 

The `memcpy(address.sun_path, path_.c_str(), path_.size());` is not correct, think that you need always a null terminated string. path.size() return the bytes count in the string, not the null termination.

> So the "only" challenge is to agree on a good style to compute `structlen`, correct? I personally would prefer something like:
> 
> ```c++
> #ifdef __linux__
>       // sun_path is not null-terminated in this case, and structlen must be shortened to the path size
>       structlen = static_cast<socklen_t>(sizeof(address)) - sizeof(address.sun_path) + path_.size();
> #else
> ```

ulidtko:
@emmenlau all good as you said, per my understanding.

I'd update the comment too, maybe like this.

```C++
#ifdef __linux__
      // sun_path is not null-terminated for abstract sockets, truncate structlen to the path size
      structlen = static_cast<socklen_t>(sizeof(address) - sizeof(address.sun_path) + path_.size());
#else
```
Also, maybe move the `static_cast<socklen_t>` to the expr top-level, to not mess up integer overflows in the arithmetic.

deiv:
> > > memcpy(address.sun_path, path_.c_str(), path_.size());
> > 
> > 
> > is `memcpy(address.sun_path, path_.c_str(), len)`
> 
> I think its better to copy only `path_size()` instead of `len` because it avoids copying a char that is not in the original string.
> 
> Zero-termination is (better?) handled by the `memset`.

Not, if the string is the same size in bytes as the 'sockaddr_un.sun_path', you will lack the null terminated string, and posibly getting a memory error.

ulidtko:
> if the string is the same size in bytes as the 'sockaddr_un.sun_path', you will lack the null terminated string, and posibly getting a memory error.

This won't happen, due to
```
    size_t len = path_.size() + 1;
    if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
      ...
      throw TTransportException(...);
    }
```

deiv:
> > if the string is the same size in bytes as the 'sockaddr_un.sun_path', you will lack the null terminated string, and posibly getting a memory error.
> 
> This won't happen, due to
> 
> ```
>     size_t len = path_.size() + 1;
>     if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
>       ...
>       throw TTransportException(...);
>     }
> ```

If the path_ if for a abstract unix socket, the `size_t len = path_.size() + 1' is not ok, is legal to fill the entire address.sun_path field.

emmenlau:
> > > if the string is the same size in bytes as the 'sockaddr_un.sun_path', you will lack the null terminated string, and posibly getting a memory error.
> > 
> > 
> > This won't happen, due to
> > ```
> >     size_t len = path_.size() + 1;
> >     if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
> >       ...
> >       throw TTransportException(...);
> >     }
> > ```
> 
> If the path_is for an abstract unix socket, the `size_t len = path_.size() + 1' is not ok, is legal to fill the entire address.sun_path field.

But then copying the zero-termination would write beyond the end of `sockaddr_un.sun_path`. This is why I prefer _not_ to copy the zero-termination.

deiv:
> > > > if the string is the same size in bytes as the 'sockaddr_un.sun_path', you will lack the null terminated string, and posibly getting a memory error.
> > > 
> > > 
> > > This won't happen, due to
> > > ```
> > >     size_t len = path_.size() + 1;
> > >     if (len > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
> > >       ...
> > >       throw TTransportException(...);
> > >     }
> > > ```
> > 
> > 
> > If the path_is for an abstract unix socket, the `size_t len = path_.size() + 1' is not ok, is legal to fill the entire address.sun_path field.
> 
> But then copying the zero-termination would write beyond the end of `sockaddr_un.sun_path`. This is why I prefer _not_ to copy the zero-termination.

In the case of pathname socket yes, for abstract unix socket is legal, because you don't need to copy the null terminating.

Is simply, the problem is around checking correctly the size and the terminating null (there's no need for zeroing all the struct).

ulidtko:
> (there's no need for zeroing all the struct)

Disagreed here. Technically, you're right*, but maintenance-wise and for future-proofing (and not sending stack garbage directly into the kernel — for abstract sockets), it's best to zero it out.

﻿
﻿


﻿
﻿
﻿
﻿
﻿
﻿
﻿
﻿
﻿
﻿
﻿
﻿

*with conditions

deiv:
> > (there's no need for zeroing all the struct)
> 
> Disagreed here. Technically, you're right*, but maintenance-wise and for future-proofing (and not sending stack garbage directly into the kernel — for abstract sockets), it's best to zero it out.
> 
> ﻿
> ﻿
> 
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> ﻿
> 
> *with conditions

The zeros as the manual states, have no special significance, so they are garbage too, so no point here.

And for maintenance point, better to have the correct handling in code, instead of try to fix it, as zeoing something.

emmenlau:
> (there's no need for zeroing all the struct)

Actually, it sadly _is_ required. See the stackoverflow link from above (and also my own experience). The kernel _may_ use all characters in `sun_path` for the socket name, which can add funny characters if there is random data in the variable. This actually happened to me.

deiv:
> > (there's no need for zeroing all the struct)
> 
> Actually, it sadly _is_ required. See the stackoverflow link from above (and also my own experience). The kernel _may_ use all characters in `sun_path` for the socket name, which can add funny characters if there is random data in the variable. This actually happened to me.

With the correct code handling , (null terminating case ok for pahtname socket, and correct lenght for abstract unix socket) the special caracters will disapear...

For me I see zeroing them as an unneed operation, that mask the correct solution that is, better handling of null terminated string. (but c and c++ are always hard),

emmenlau:
> > But then copying the zero-termination would write beyond the end of `sockaddr_un.sun_path`. This is why I prefer _not_ to copy the zero-termination.
> 
> In the case of pathname socket yes, for abstract unix socket is legal, because you don't need to copy the null terminating. is the standard.

But this is more error-prone and complex than to just zero the full struct, and then copy _only_ `path_.size()` char's, isn't it? The latter proposed approach works for all cases (pathname sockets and abstract sockets), and it supports your suggested use case where the abstract socket uses the full `sun_path` without trailing zero-termination.

emmenlau:
> For me I see zeroing them as an unneed operation, that mask the correct solution that is, better handling of null terminated string. (but c and c++ are always hard),

Ok, lets settle for "would you be ok to leave the zeroing in place"? It does not hurt much, if nothing else? :-)

deiv:
> > > But then copying the zero-termination would write beyond the end of `sockaddr_un.sun_path`. This is why I prefer _not_ to copy the zero-termination.
> > 
> > 
> > In the case of pathname socket yes, for abstract unix socket is legal, because you don't need to copy the null terminating. is the standard.
> 
> But this is more error-prone and complex than to just zero the full struct, and then copy _only_ `path_.size()` char's, isn't it? The latter proposed approach works for all cases (pathname sockets and abstract sockets), and it supports your suggested use case where the abstract socket uses the full `sun_path` without trailing zero-termination.

Yes, more easy, this way, you forget about the terminating string. But the question is: If I go to a fountain because I need one liter of water, and I fill  five just in case (is the correct way o not?). Points of view.

But if the code need's to handling the null terminating and the size, and the two are correct. Removing the memset, will not make any difference.

ulidtko:
I think, the `if` for abstract/path is unavoidable — and arguments to both `memcpy` and `bind` will differ depending on that. Trying to unify the branches of that `if` leads straight into confusion.

```c++
    struct sockaddr_un address;
    memset(&address, '\0', sizeof(address));
    address.sun_family = AF_UNIX;

    if (!path_[0]) { // "abstract" namespace socket (Linux-specific)

        // space check
        if (path_.size() > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
             throw TTransportException(...);
        }

        // no null termination here
        memcpy(address.sun_path, path_.c_str(), path_.size());
        addrlen = path._size();

    } else { // "pathname" socket

        // space check, with additional byte for the terminator
        if (path_.size() + 1 > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
             throw TTransportException(...);
        }

        // including the null terminator
        strncpy(address.sun_path, path_.c_str(), path_.size() + 1);
        addrlen = path._size() + 1;
    }

    ::bind(/*…*/, /*…*/, addrlen);
```

-----------------------------

@deiv 
> ... So better to do a refactoring.

Yes exactly — that's why I'm saying I'm not totally happy with your patch!

deiv:
> I think, the `if` for abstract/path is unavoidable — and arguments to both `memcpy` and `bind` will differ depending on that. Trying to unify the branches of that `if` leads straight into confusion.
> 
> ```c++
>     struct sockaddr_un address;
>     memset(&address, '\0', sizeof(address));
>     address.sun_family = AF_UNIX;
> 
>     if (!path_[0]) { // "abstract" namespace socket (Linux-specific)
> 
>         // space check
>         if (path_.size() > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
>              throw TTransportException(...);
>         }
> 
>         // no null termination here
>         memcpy(address.sun_path, path_.c_str(), path_.size());
>         addrlen = path._size();
> 
>     } else { // "pathname" socket
> 
>         // space check, with additional byte for the terminator
>         if (path_.size() + 1 > sizeof(((sockaddr_un*)nullptr)->sun_path)) {
>              throw TTransportException(...);
>         }
> 
>         // including the null terminator
>         strncpy(address.sun_path, path_.c_str(), path_.size() + 1);
>         addrlen = path._size() + 1;
>     }
> 
>     ::bind(/*…*/, /*…*/, addrlen);
> ```
> 
> @deiv
> 
> > ... So better to do a refactoring.
> 
> Yes exactly — that's why I'm saying I'm not totally happy with your patch!

Make a unique function call for the 3 files, could be good too ... :)

And, the memset, for me, is not making nothing :)

ulidtko:
> Make a unique function call for the 3 files, could be good too ... :)

Well, not such a bad idea, haha. Will you do it?

> And, the memset, for me, is not making nothing :)

I understand; let's just agree to disagree on this, can we?

-----------------------

Also, as a reminder: we have 2 PR's fixing the same thing in different ways. #2234 and #2235.

It's best to focus on one and close the other.

deiv:
> > Make a unique function call for the 3 files, could be good too ... :)
> 
> Well, not such a bad idea, haha. Will you do it?

Can take a look tomorrow. If nobody want to do after.

> > And, the memset, for me, is not making nothing :)
> 
> I understand; let's just agree to disagree on this, can we?

uhm... xD

> Also, as a reminder: we have 2 PR's fixing the same thing in different ways. #2234 and #2235.
> 
> It's best to focus on one and close the other.

emmenlau:
> Make a unique function call for the 3 files, could be good too ... :)

This is actually a subset of a larger section of TServerSocket.cpp and TSocket.cpp that mostly overlaps. For maintenance this would much better be unified. I don't have too much spare time, anyone wants to give it a go?

emmenlau:
> I think, the `if` for abstract/path is unavoidable — and arguments to both `memcpy` and `bind` will differ depending on that. Trying to unify the branches of that `if` leads straight into confusion.

The longer I think about it, yes maybe the `if`is the cleanest solution. @deiv would you accept a solution based on `if`? It would be awesome if you can propose something...

deiv:
@ulidtko there goes

emmenlau:
> @ulidtko there goes

This newest version looks very good to me. Nice work.

ulidtko (approved):
Awesome, this looks pretty good!

deiv:
@ulidtko @emmenlau 

thanks (the twoo). Interesting discussion hehe :)

deiv:
@ulidtko 

do you known when we get a new release ?

emmenlau:
> @ulidtko
> 
> do you known when we get a new release ?

There is no real schedule, and no release plan. So I would not hold my breath for it. If you are very curious or urgently need a release, I would suggest to bring the topic to the mailing list.

Jens-G:
> do you known when we get a new release ?

The de-facto truth is: When somebody<sup>*)</sup> manages some time to start working on one.

<sub>*) out of a group of certain people, not just anybody</sub>

zeshuai007:
It build fail in CI. see [this](https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/35128105/job/5alf601ondssa2vy)

AppVeryor log show like this.
```
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(52): error C2065: 'is_abstract_namespace': undeclared identifier [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(65): error C2027: use of undefined type 'apache::thrift::transport::sockaddr_un' [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
  C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(46): note: see declaration of 'apache::thrift::transport::sockaddr_un'
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(65): error C2227: left of '->sun_path' must point to class/struct/union/generic type [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(71): error C2027: use of undefined type 'apache::thrift::transport::sockaddr_un' [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
  C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(46): note: see declaration of 'apache::thrift::transport::sockaddr_un'
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(71): error C2228: left of '.sun_family' must have class/struct/union [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(72): error C2027: use of undefined type 'apache::thrift::transport::sockaddr_un' [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
  C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(46): note: see declaration of 'apache::thrift::transport::sockaddr_un'
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(72): error C2228: left of '.sun_path' must have class/struct/union [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(74): error C2027: use of undefined type 'apache::thrift::transport::sockaddr_un' [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
  C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(46): note: see declaration of 'apache::thrift::transport::sockaddr_un'
C:\projects\thrift\lib\cpp\src\thrift\transport\SocketCommon.cpp(74): error C2227: left of '->sun_family' must point to class/struct/union/generic type [C:\projects\build\MSVC2015\x86\lib\cpp\thrift.vcxproj]
```

Could you build success in your local environment ?

deiv:
@zeshuai007 

Fixed, seems that thrift is not ready for compiling unix sockets under windows. I just added some conditional macros.

I fixed too, a variable that does not has the correct name (`is_abstract_namespace`)

I added a TODO, to take into account, that windows has support to unix sockets, but the code is not prepared for that. At least it needs to include the windows headers for unix sockets. See [af_unix-comes-to-windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)

I don't have a windows development enviroment, so I can't take a look.

emmenlau:
> I added a TODO, to take into account, that windows has support to unix sockets, but the code is not prepared for that. At least it needs to include the windows headers for unix sockets. See [af_unix-comes-to-windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)

I have PR #2127 that adds support for Unix sockets under Windows. Generally its positive to have separate PRs for separate functionality, but in this case its not necessarily better that you need to guard *against* supporting Unix sockets only for me to later add the support in a subsequent PR. So please feel free to cannibalize #2127 and take everything you need to support the Unix sockets on MSVC?

deiv:
> > I added a TODO, to take into account, that windows has support to unix sockets, but the code is not prepared for that. At least it needs to include the windows headers for unix sockets. See [af_unix-comes-to-windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)
> 
> I have PR #2127 that adds support for Unix sockets under Windows. Generally its positive to have separate PRs for separate functionality, but in this case its not necessarily better that you need to guard _against_ supporting Unix sockets only for me to later add the support in a subsequent PR. So please feel free to cannibalize #2127 and take everything you need to support the Unix sockets on MSVC?

Hi @emmenlau,

I don't have a windows development enviroment, and because, to add support for Unix sockets, is needed a new header (and the checks for it), better to have it in another separate PR.

And is needed to remove other windows conditional compilation in sockets sources files, too. Eg:

```
#ifndef _WIN32
    socklen_t structlen = fillUnixSocketAddr(address, path_);
```

This one only fix abstract namespace, so better if you can do another PR removing the conditional compilation for windows. Seems more logical to have the new windows unix socket support separated.

emmenlau:
Good, I agree! Then @zeshuai007 please review!

PR #1859, opened 2019-08-28
THRIFT-4942: Generated fields privatized; add getters/setters

2DSharp:
Fixes THRIFT-4942: Set PHP struct generated field values as private with getters and setters

  Classes generated from structs create public fields; The access specifiers to them have been made private and they can be accessed/modified via generated getters and setters

- [X] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [X] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [X] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [X] If your change does not involve any code, add ` [skip ci]` at the end of your pull request to free up build resources.

PR #2238, opened 2020-09-19
THRIFT-4942: Add getters setters, improves on #1859

2DSharp:
Relates to THRIFT-4942

  A non-breaking version of #1859. Adds getters and setters to objects to access fields.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_php_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_php_generator.cc
@@ -411,6 +411,11 @@ private:
    * Global namespace for PHP 5.3
    */
   std::string nsglobal_;
+
+  /**
+   * Whether to generate getters and setters
+   */
+  bool getters_setters_;
 };
 
 bool t_php_generator::is_valid_namespace(const std::string& sub_namespace) {
--- a/compiler/cpp/src/thrift/generate/t_php_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_php_generator.cc
@@ -89,8 +89,8 @@ public:
           pwarning(0, "psr4 is default option! needn't add psr4 option!\n");
         }
       } else if (iter->first.compare("getters_setters") == 0) {
-	getters_setters_ = true;
-      }else {
+        getters_setters_ = true;
+      } else {
         throw "unknown option php:" + iter->first;
       }
     }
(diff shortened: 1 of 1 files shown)


Jens-G:
My idea was that you could add an option to the php generator that controls the new way using getters/setters and private fields, and if not set it would just generate the same code as before. Sorry for being not clear enough.

If I get it right, what you do here is to add the getters/setters but leave the fields publicly accessible?

2DSharp:
On it, I assume I will need to create a `getters_setters_` variable and check for the correct flag in the passed options iterator and set that variable to true.

Jens-G:
:\projects\thrift\compiler\cpp\src\thrift\generate\t_php_generator.cc:64:5: error: 'getters_setters_' was not declared in this scope
   64 |     getters_setters_ = false;
      |     ^~~~~~~~~~~~~~~~
compilation terminated due to -fmax-errors=1.

2DSharp:
@Jens-G Fixed!

Jens-G:
CI still complains (about unused parameters) https://travis-ci.org/github/apache/thrift/jobs/729280303

2DSharp:
@Jens-G The CI tests passed, however, two of them seem to fail because of failing EUNIT tests, possibly because modules couldn't be found for the Erlang library.

2DSharp:
A thing to note is that I've reverted the public/private thing to only public to not break previously working code. Adding the flag however will be a conscious decision and the user will have to migrate to using the getters/setters for accessing the values if the fields are set to private. 

Should I let it stay?

Jens-G:
IMHO when there are getters/setters chosen by setting the flag the dev should be aware that it might break code and so we should do it right. That was the idea, that we make it an opt-in decision, instead of forcing everybody into it.  Feel free to send another PR, since I already committed this just minutes ago.

2DSharp:
Right, I was a bit late to write it and saw it got merged right after I wrote it and so deleted the comment. I was thinking the same, since the developer is opting for it, they are already aware that it can break code.

2DSharp:
@Jens-G I was wondering if I should create another `private_fields` flag that creates private fields with getters and setters, and leave the `getters_setters` as it is right now, this way, one of them will be breaking (`private_fields`), and developers unwilling to break existing code can use getters and setters. Please let me know what your thoughts are on this.

Jens-G:
No private_fields flag, just connect it to the existing. Don't overcomplicate it.

2DSharp:
Updated with: https://github.com/apache/thrift/issues/2245

PR #2241, opened 2020-09-22
Compiler: ["markdown"]

kashirin-alex:
**Added a Markdown Generator:** 
generate documentations in 'markdown' format
_The main reason is that some builds break with markdown files that include html._

The ```thrift --help``` is now with a generator 'markdown' option:
```bash
  markdown (MARKDOWN):
    suffix:          Create files/links with/out 'md|html' default None
    noescape:        Do not escape with html-entities in doc text.
```

**Some Markdown Generator Results Available at:**
* a direct view of a 'md' files
    - https://github.com/kashirin-alex/swc-db/blob/master/src/thrift/swcdb/README.md
    - https://github.com/kashirin-alex/swc-db/tree/master/docs/use/thriftclient#the-swc-db-thrift-modules
* a view of docs-generators:
    - mkdocs build - https://swc-db.readthedocs.io/en/latest/use/thriftclient/
    - GitHub Jekyll theme build - https://kashirin-alex.github.io/swc-db/use/thriftclient/#the-swc-db-thrift-modules

Note:
_-- the t_markdown_generator.cc is mostly based on the t_html_generator.cc, while both can be improved by using structured template for populating data by the specific-structure_


compiler/cpp/src/thrift/generate/t_markdown_generator.cc
```diff
@@ -0,0 +1,1085 @@
...
+
+//using std::endl;
+static const string endl = "\n"; // avoid ostream << std::endl flushes
+
+std::string str_to_id(const std::string& s) {
+  std::string id;
+  for(auto chr=s.begin();chr<=s.end(); ++chr) {
+    if(*chr == '.' || *chr == 0)
+      continue;
+    id += std::tolower(*chr);
```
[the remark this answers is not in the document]
kashirin-alex: The corresponding [Build Error](https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/35332170/job/kr58i5k958daewpl#L1204)

While the meaning it be using the C `tolower` whereas Microsoft Visual Studio 14.0 has only the C type.
It will be OK(confirming include tree has <ctype.h>),

compiler/cpp/src/thrift/generate/t_markdown_generator.cc
```diff
@@ -0,0 +1,1085 @@
...
+  if (!tprog->get_services().empty()) {
+    vector<t_service*> services = tprog->get_services();
+    vector<t_service*>::iterator sv_iter;
+    for (sv_iter = services.begin(); sv_iter != services.end(); ++sv_iter) {
+      if(sv_iter != services.begin()) {
+        filling.emplace_back();
+        fill = &filling.back();
+      }
+      string name = get_service_name(*sv_iter);
+      (*fill)[1] = "[" + name + "](" + make_file_link(fname) + "#service" + str_to_id(name) + ")";
```
kashirin-alex, line 186: Missing Hyphen after `#service`
`#service` > `#service-`

zeshuai007 (approved):

compiler/cpp/src/thrift/generate/t_markdown_generator.cc
```diff
@@ -0,0 +1,1090 @@
...
+using std::pair;
+using std::string;
+using std::stringstream;
+using std::vector;
+
+//using std::endl;
+static const string endl = "\n"; // avoid ostream << std::endl flushes
+
+std::string str_to_id(const std::string& s) {
+  std::string id;
```
Jens-G, line 45: Why does it have to be a function outside of the class?

kashirin-alex: True, in such case all the definitions out of class be set to inner class static & private

PR #2246, opened 2020-09-29
go: Use sync.Pool for gzip in HTTP transport

dcelasun:
b67cad4 introduced transparent gzip support for the HTTP transport but
calling gzip.NewWriter() with every request causes a large number of
memory allocations [1] and can create GC pressure.

Avoid this issue by using a sync.Pool for gzip writers.

[1] https://old.reddit.com/r/golang/comments/9uejp4/usage_of_syncpool_for_gzipwriter_in_http_handlers/e94jh8c/

### DIFF
--- a/lib/go/thrift/http_transport.go
+++ b/lib/go/thrift/http_transport.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync"
 )
 
 // NewThriftHandlerFunc is a function that create a ready to use Apache Thrift Handler function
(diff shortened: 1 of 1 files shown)


lib/go/thrift/http_transport.go
```diff
@@ -40,14 +41,23 @@ func NewThriftHandlerFunc(processor TProcessor,
 
 // gz transparently compresses the HTTP response if the client supports it.
 func gz(handler http.HandlerFunc) http.HandlerFunc {
+	sp := &sync.Pool{
+		New: func() interface{} {
+			return gzip.NewWriter(nil)
+		}}
```
fishy, line 47: nit: style wise I'd slightly prefer `},\n}`, but we are not very strictly on style things anyways so I'm not feeling strongly.

lib/go/thrift/http_transport.go
```diff
@@ -40,14 +41,23 @@ func NewThriftHandlerFunc(processor TProcessor,
...
 			handler(w, r)
 			return
 		}
 		w.Header().Set("Content-Encoding", "gzip")
-		gz := gzip.NewWriter(w)
-		defer gz.Close()
+		gz := sp.Get().(*gzip.Writer)
+		gz.Reset(w)
+		defer func() {
+			_ = gz.Close()
```
fishy, line 58: also nit: For writes `Close` usually also includes `Flush` so ignore the error could be bad. But we were ignoring the error before this change, so this at least won't make things worse. We don't have a way to return error here anyways.

fishy (approved):
I have a few nitpicks

lib/go/thrift/http_transport.go
dcelasun: No worries, will change this before merging.

lib/go/thrift/http_transport.go
dcelasun: Yeah I thought about it but couldn't think of a clean way of handling the error. I'm open to suggestions :)

lib/go/thrift/http_transport.go
dcelasun, line 47:
```suggestion
		},
	}
```

lib/go/thrift/http_transport.go
fishy: Reading code I think the best we could do is probably pass in a logger through `NewThriftHandlerFunc` and log the error if it's non-nil, but that's not very useful and it's also a breaking change so probably not worth doing.

fishy (approved):

PR #2255, opened 2020-10-06
Throw exception when there is no data in curl transport

thedebugger:
My team has been running a fork of php thrift (super old) and we had an issue parsing never terminates when body is empty. After looking open source code, current master is not vulnerable to our specific use case because curl throws an error when response is empty. However, I feel that explicitly throwing an error when there is no more data available in curl transport is beneficial in the case if caller has bugs. Let me know what you folks think. And Happy to add test and jira if that makes sense. Thxs

- [] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
Happy to create it if we my problem/changes make sense.
- [] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [] Did you squash your changes to a single commit?  (not required, but preferred)

### DIFF
--- a/lib/php/lib/Transport/TCurlClient.php
+++ b/lib/php/lib/Transport/TCurlClient.php
@@ -159,14 +159,24 @@ class TCurlClient extends TTransport
      */
     public function read($len)
     {
+        $len_avail = strlen($this->response_);
+        if ($len_avail === 0) {
+            throw new TTransportException(
+                'TCurlClient: Could not read ' . $len . ' bytes from ' .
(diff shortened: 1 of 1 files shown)


Jens-G:
>  And Happy to add test and jira if that makes sense. Thxs
I was about to ask for it, so yes, please do.

thedebugger:
@Jens-G thanks for the feedback. I'll start doing that today. But the patch makes sense...right?

PR #2261, opened 2020-10-14
Add DuplicateTo protocol to go's TDebugProtocol

fishy:
This also comes from the discussion of
https://github.com/apache/thrift/pull/1992#issuecomment-705903922.
I think TDebugProtocol is a better fit for this feature than creating a
new TProtocol implementation.

The DuplicateTo field is not added to TDebugProtocolFactory because I
don't think it makes sense from the factory setup. In vast majority
cases users would need direct access to the underlying TMemoryBuffer to
make it useful, which is easier this way than an additional
TTransportFactory plus TProtocolFactory to make TDebugProtocolFactory
way too complicated.

### DIFF
--- a/lib/go/thrift/debug_protocol.go
+++ b/lib/go/thrift/debug_protocol.go
@@ -25,9 +25,27 @@ import (
 )
 
 type TDebugProtocol struct {
-	Delegate  TProtocol
-	LogPrefix string
+	// Required. The actual TProtocol to do the read/write.
+	Delegate TProtocol
+
+	// Optional. The logger and prefix to log all the args/return values
+	// from Delegate TProtocol calls.
+	//
+	// If Logger is nil, StdLogger using stdlib log package with os.Stderr
+	// will be used. If disable logging is desired, set Logger to NopLogger
+	// explicitly instead of leaving it as nil/unset.
 	Logger    Logger
+	LogPrefix string
+
+	// Optional. An TProtocol to duplicate everything read/written from Delegate.
+	//
+	// A typical use case of this is to use TSimpleJSONProtocol wrapping
+	// TMemoryBuffer in a middleware to json logging requests/responses.
+	//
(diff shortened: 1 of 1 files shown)


fishy:
cc: @voodoo-dn

lib/go/thrift/debug_protocol.go
```diff
@@ -25,9 +25,27 @@ import (
...
+
+	// Optional. An TProtocol to dump everything read/wrote from Delegate.
+	//
+	// A typical use case of this is to use TSimpleJSONProtocol wrapping
+	// TMemoryBuffer in a middleware to json logging requests/responses.
+	//
+	// This feature is not available from TDebugProtocolFactory. In order to
+	// use it you have to construct TDebugProtocol directly, or set Dump
+	// field after getting a TDebugProtocol from the factory.
+	Dump TProtocol
```
fishy, line 48: I guess "Dump" is not the best name, and I'm open to suggestions :)

dcelasun: Hmm, "Export"? "Intercept"?

dcelasun (approved):
LGTM, except the naming thing :)

lib/go/thrift/debug_protocol.go
fishy: Those names don't sound right to me either 😕 What do you think about "WriteTo", "Writer", "DuplicateTo"?

dcelasun: Of those, DuplicateTo sounds best :)

dcelasun: Of those, DuplicateTo sounds best :)

fishy: OK renamed to `DuplicateTo`. commit message also updated accordingly.

dcelasun (approved):

PR #2274, opened 2020-11-13
fix Out Of Memory when use TNonblockingServer

namaiwa:
the TNonblockingServer may run out of memory when a TBufferedTransport client requests it, the server go into a endless loop and it will do nothing until  the somebody kills it.


zeshuai007 (approved):
LGTM.

emmenlau (approved):
Looks good and is relatively self-explaining.

emmenlau:
LGTM, but I could not quickly find where the return argument if `struct.unpack()` is documented. Is `mlen` the size of the structure?

PR #2281, opened 2020-11-19
Make ~TException() virtual

emmenlau:
This is a trivial change. It makes the destructor for the class `TException` virtual because its a base class for other exceptions. Base classes generally should have virtual destructors (or am I missing something very specific to this use case here?).

### DIFF
--- a/lib/cpp/src/thrift/Thrift.h
+++ b/lib/cpp/src/thrift/Thrift.h
@@ -79,7 +79,7 @@ public:
 
   TException(const std::string& message) : message_(message) {}
 
-  ~TException() noexcept override = default;
+  virtual ~TException() noexcept override = default;
 
   const char* what() const noexcept override {
     if (message_.empty()) {


ulidtko (changes requested):
> Base classes generally should have virtual destructors

... whenever you're using runtime polymorphism — not in general.

An example would be:
```cpp
  TApplicationException* specific_derived_instance = [...];
  TException* upcasted_ptr = specific_derived_instance;
  delete upcasted_ptr;
```
For this code to work correctly, yes, `TException` dtor must be virtual.

If such usage isn't the case — then the extra cost of making *the whole class* virtual (which entails vtable creation and associated runtime lookups) may not be justified.

Put another way: if the class in question *has already* virtual methods -- you pay nothing, may as well slap that `virtual` onto the dtor and forget about it. If the class isn't virtual yet -- you pay the cost and need to consider if that's really needed.

... Does this solve a specific problem you have? Just asking.

emmenlau:
> > Base classes generally should have virtual destructors
> 
> ... whenever you're using runtime polymorphism — not in general.
> 
> For this code to work correctly, yes, `TException` dtor must be virtual.
> 
> ... Does this solve a specific problem you have? Just asking.

No, it was found only by chance because we have some exceptions that inherit from exceptions that inherit from `TException`. The static analyzer warned that the base class was not virtual and resources may be leaked.

emmenlau:
... and thanks for the nice review! You are perfectly correct that virtualization comes at a cost and is not required in all scenarios.

ulidtko (approved):
Oh well, `std::exception` is already virtual class and [has virtual dtor](https://en.cppreference.com/w/cpp/error/exception/~exception).

So this one should be marked as such too. LGTM

emmenlau:
The Travis errors seem to be unrelated, so this should be good to go!

emmenlau:
Thanks @Jens-G !

PR #2284, opened 2020-11-21
netstd: Log errors in processor via standard logger 

cspwizard:
netstd: Log errors in processor via standard logger 

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_netstd_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_netstd_generator.cc
@@ -361,6 +361,7 @@ string t_netstd_generator::netstd_type_usings() const
         "using System.Linq;\n"
         "using System.Threading;\n"
         "using System.Threading.Tasks;\n"
+        "using Microsoft.Extensions.Logging;\n"
         "using Thrift;\n"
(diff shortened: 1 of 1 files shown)


Jens-G:
I made the logger optional to a) not break existing code (including the Test suite) and b) not to force people into using it.

cspwizard:
I may miss the idea, but how is it optional when there is an output directly to console?

Jens-G:
Main reason for the change was that you added the logger argument in a way that broke existing code. You could have easily recognised this yourself just by trying to build the code we have. Since logging is a non-mission critical task to Thrift the very obvious solution was to make the argument option by adding `= default` and of course take care of that case when the logger is used. That way you are free to provide a logger but are not forced to do it, in which case it simply behaves as before. I'd consider this as the optimum solution. Whether or not someone is using docker or not is of no relevance here.

cspwizard:
Ah, ok. I'll change the code. 
My point was that the logger/console in current implementation is not optional in any way. It just prints unformatted text to a console and that brings some problems in use

Jens-G:
I already merged you PR, in case you did'nt notice. Further PRs welcome of course.

cspwizard:
Great, thanks

PR #2292, opened 2020-12-10
THRIFT-5322: Guard against large string/binary lengths in Go

fishy:
Client: go

In TBinaryProtocol.ReadString, TBinaryProtocol.ReadBinary,
TCompactProtocol.ReadString, and TCompactProtocol.ReadBinary, use
safeReadBytes to prevent from large allocation on malformed sizes.

    $ go test -bench=SafeReadBytes -benchmem
    goos: linux
    goarch: amd64
    BenchmarkSafeReadBytes/normal-12                 2664687               424 ns/op            2176 B/op          5 allocs/op
    BenchmarkSafeReadBytes/max-size-12                462532              2494 ns/op           14464 B/op          7 allocs/op
    PASS
    ok      _/home/fishy/work/thrift/lib/go/thrift  2.901s

### DIFF
--- a/lib/go/thrift/binary_protocol.go
+++ b/lib/go/thrift/binary_protocol.go
@@ -432,6 +432,15 @@ func (p *TBinaryProtocol) ReadString(ctx context.Context) (value string, err err
 		err = invalidDataLength
 		return
 	}
+	if size == 0 {
+		return "", nil
+	}
+	if size < int32(len(p.buffer)) {
+		// Avoid allocation on small reads
+		buf := p.buffer[:size]
+		read, e := io.ReadFull(p.trans, buf)
+		return string(buf[:read]), NewTProtocolException(e)
+	}
 
 	return p.readStringBody(size)
 }
@@ -445,9 +454,7 @@ func (p *TBinaryProtocol) ReadBinary(ctx context.Context) ([]byte, error) {
 		return nil, invalidDataLength
 	}
 
-	isize := int(size)
-	buf := make([]byte, isize)
-	_, err := io.ReadFull(p.trans, buf)
+	buf, err := safeReadBytes(size, p.trans)
 	return buf, NewTProtocolException(err)
 }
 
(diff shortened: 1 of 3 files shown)


lib/go/thrift/binary_protocol.go
```diff
@@ -482,35 +480,37 @@ func (p *TBinaryProtocol) readAll(ctx context.Context, buf []byte) (err error) {
 const readLimit = 32768
 
 func (p *TBinaryProtocol) readStringBody(size int32) (value string, err error) {
+	buf, err := safeReadBytes(size, p.trans, p.buffer[:], readLimit)
+	return buf.String(), NewTProtocolException(err)
+}
+
+// This function is shared between TBinaryProtocol and TCompactProtocol.
+func safeReadBytes(size int32, trans io.Reader, buffer []byte, singleReadLimit int) (*bytes.Buffer, error) {
```
yurishkuro, line 488: I would rather return just the slice `[]byte`, which would be heap-allocated anyway, and keep the `bytes.Buffer` hidden and allocated on the stack, as it was previously.

Also the semantics of this function is a bit confusing - what is the role of `buffer` argument?

fishy: Good point. Actually we don't need to specify the limit at all. Combining bytes.Buffer with io.CopyN is enough.

lib/go/thrift/binary_protocol.go
```diff
@@ -432,6 +432,15 @@ func (p *TBinaryProtocol) ReadString(ctx context.Context) (value string, err err
...
 	}
+	if size == 0 {
+		return "", nil
+	}
+	if size < int32(len(p.buffer)) {
+		// Avoid allocation on small reads
+		buf := p.buffer[:size]
+		read, e := io.ReadFull(p.trans, buf)
+		return string(buf[:read]), NewTProtocolException(e)
+	}
```
yurishkuro, line 443: why not move this logic inside `readStringBody` and use that from both protocols?

lib/go/thrift/binary_protocol.go
```diff
@@ -479,38 +486,21 @@ func (p *TBinaryProtocol) readAll(ctx context.Context, buf []byte) (err error) {
...
-		if e != nil {
-			break
-		}
-		size -= readLimit
-		if size < readLimit && size > 0 {
-			b = b[:size]
-		}
-	}
-	return buf.String(), NewTProtocolException(e)
+	buf := new(bytes.Buffer)
```
yurishkuro, line 503: This is still two allocations, one for Buffer, another for Buffer.buf slice. Why not do just one?
```go
buf := make([]byte, size)
_, err := io.ReadFull(trans, buf)
return buf, err
```

lib/go/thrift/binary_protocol.go
```diff
@@ -479,38 +486,21 @@ func (p *TBinaryProtocol) readAll(ctx context.Context, buf []byte) (err error) {
...
-	case int(size) <= len(p.buffer):
-		b = p.buffer[:size] // avoids allocation for small reads
-	case int(size) < readLimit:
-		b = make([]byte, size)
-	default:
-		b = make([]byte, readLimit)
+// This function is shared between TBinaryProtocol and TCompactProtocol.
+//
+// It tries to read size bytes from trans, in a way that prevents large
+// allocations when size is insanely large (mostly caused by malformed message).
```
yurishkuro, line 497: How does it actually prevent large allocations? Previously there was readLimit

lib/go/thrift/binary_protocol.go
fishy: What you suggested IS the large allocation when size is insanely large.

lib/go/thrift/binary_protocol.go
fishy: It's in the details of the implementation of `io.CopyN` and `bytes.Buffer`. what they did is similar to the previous version, that they will do the copy chunk by chunk (and grow the buffer when necessary), instead of allocate the whole size up front.

lib/go/thrift/binary_protocol.go
fishy: This avoids the extra allocation (we reused `p.buffer`)

`readStringBody` is a function attached to type `TBinaryProtocol`, so it cannot be used by `TCompactProtocol`. we cannot move it to top level function either as it's also used by other parts of `TBinaryProtocol`. I mean we can do that refactor, but consider it's just 4 lines it's not really worth it.

lib/go/thrift/binary_protocol.go
yurishkuro: if `size` is large you end up with a huge allocated array no matter _how you copy data into it_. The main issue in the ticket is that the value of `size` could be corrupted and be greater than the total message size.

fishy: No. With `io.CopyN` + `bytes.Buffer` it stops when there's no more data to read. See this line in the benchmark result I attached to the commit message:

```
BenchmarkSafeReadBytes/max-size-12                462532              2494 ns/op           14464 B/op          7 allocs/op
```

The size in that call is max int32, but it only allocated 14KB of memory.

yurishkuro: Got it. Thanks!

jpkrohling (approved):
I monkey patched it here and re-run the reproducer, and can confirm it does fix the issue. Thanks for the speedy resolution!

PR #2293, opened 2020-12-14
THRIFT-5324: reset http client buffer after flush

NIHERASE:
Client: go
THttpClient did not reset it's internal buffer when http client returned
an error, leaving the whole or partially read message in the buffer.
Now we reset the buffer in defer.

Details: https://issues.apache.org/jira/browse/THRIFT-5324

### DIFF
--- a/lib/go/thrift/http_client.go
+++ b/lib/go/thrift/http_client.go
@@ -197,6 +197,14 @@ func (p *THttpClient) Flush(ctx context.Context) error {
 	// Close any previous response body to avoid leaking connections.
 	p.closeResponse()
 
+	// Request might not have been fully read by http client.
+	// Reset so we don't send the remains on next call.
+	defer func() {
+		if p.requestBuffer != nil {
+			p.requestBuffer.Reset()
+		}
+	}()
+
 	req, err := http.NewRequest("POST", p.url.String(), p.requestBuffer)
 	if err != nil {
 		return NewTTransportExceptionFromError(err)
--- a/lib/go/thrift/http_client_test.go
+++ b/lib/go/thrift/http_client_test.go
@@ -20,6 +20,8 @@
 package thrift
 
 import (
+	"bytes"
+	"context"
 	"net/http"
 	"testing"
 )
@@ -32,14 +34,14 @@ func TestHttpClient(t *testing.T) {
 	trans, err := NewTHttpPostClient("http://" + addr.String())
 	if err != nil {
 		l.Close()
-		t.Fatalf("Unable to connect to %s: %s", addr.String(), err)
+		t.Fatalf("Unable to connect to %s: %v", addr.String(), err)
 	}
 	TransportTest(t, trans, trans)
 
 	t.Run("nilBuffer", func(t *testing.T) {
 		_ = trans.Close()
 		if _, err = trans.Write([]byte{1, 2, 3, 4}); err == nil {
-			t.Fatalf("writing to a closed transport did not result in an error")
+			t.Fatal("writing to a closed transport did not result in an error")
 		}
 	})
 }
@@ -52,7 +54,7 @@ func TestHttpClientHeaders(t *testing.T) {
 	trans, err := NewTHttpPostClient("http://" + addr.String())
 	if err != nil {
 		l.Close()
-		t.Fatalf("Unable to connect to %s: %s", addr.String(), err)
+		t.Fatalf("Unable to connect to %s: %v", addr.String(), err)
 	}
 	TransportHeaderTest(t, trans, trans)
 }
@@ -72,7 +74,7 @@ func TestHttpCustomClient(t *testing.T) {
 	})
 	if err != nil {
 		l.Close()
-		t.Fatalf("Unable to connect to %s: %s", addr.String(), err)
+		t.Fatalf("Unable to connect to %s: %v", addr.String(), err)
 	}
 	TransportHeaderTest(t, trans, trans)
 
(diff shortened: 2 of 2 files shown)


dcelasun:
I've just discovered the same bug today and fixed it in [my fork](https://github.com/dcelasun/thrift/commit/86478dc) :) Your commit includes a test, so I'll merge yours.

lib/go/thrift/http_client.go
```diff
@@ -197,6 +197,10 @@ func (p *THttpClient) Flush(ctx context.Context) error {
 	// Close any previous response body to avoid leaking connections.
 	p.closeResponse()
 
+	// Request might not have been fully read by http client.
+	// Reset so we don't send the remains on next call.
+	defer p.requestBuffer.Reset()
```
dcelasun, line 202: Needs a nil check.

NIHERASE: Added, thanks!

lib/go/thrift/http_client_test.go
```diff
@@ -103,6 +105,54 @@ func TestHttpCustomClientPackageScope(t *testing.T) {
...
+		write2 = []byte("write 2")
+	)
+
+	l, addr := HttpClientSetupForTest(t)
+	if l != nil {
+		defer l.Close()
+	}
+	trans, err := NewTHttpPostClient("http://" + addr.String())
+	if err != nil {
+		t.Fatalf("Unable to connect to %s: %s", addr.String(), err)
```
fishy, line 120: This is nitpicking and also a problem already exists in the old test code, but using `%s` on errors is the same as using `%v` on errors except that `%s` handles it much more poorer when error is nil (https://play.golang.org/p/yHovIp3m37B). I'd rather just use `%v` with errors instead.

lib/go/thrift/http_client_test.go
```diff
@@ -103,6 +105,54 @@ func TestHttpCustomClientPackageScope(t *testing.T) {
...
+	}
+
+	_, err = trans.Write(write1)
+	if err != nil {
+		t.Fatalf("Failed to write to transport: %s", err.Error())
```
fishy, line 125: Similar to the comment above, instead of using `%s` with `err.Error()`, just use `%v` with `err`.

lib/go/thrift/http_client_test.go
```diff
@@ -103,6 +105,54 @@ func TestHttpCustomClientPackageScope(t *testing.T) {
...
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	err = trans.Flush(ctx)
+	if err == nil {
+		t.Fatalf("Expected flush error")
```
fishy, line 131: There's no formatting verbs so this can be `t.Fatal` instead of `t.Fatalf`.

lib/go/thrift/http_client_test.go
```diff
@@ -103,6 +105,54 @@ func TestHttpCustomClientPackageScope(t *testing.T) {
...
+	n, err := trans.Read(data)
+	if err != nil {
+		t.Fatalf("Failed to read: %s", err.Error())
+	}
+
+	data = data[:n]
+	if !bytes.Equal(data, write2) {
+		t.Fatalf("Received unexpected data: %s", data)
+	}
+	trans.Close()
```
fishy, line 153: Please move this to right after you checked for error on Line 121 as `defer trans.Close()`. You have a lot of `t.Fatalf` that could exit the test early without closing the opened trans (this is test code so this is nitpicking).

lib/go/thrift/http_client_test.go
```diff
@@ -103,6 +105,54 @@ func TestHttpCustomClientPackageScope(t *testing.T) {
...
+
+	data := make([]byte, 1024)
+	n, err := trans.Read(data)
+	if err != nil {
+		t.Fatalf("Failed to read: %s", err.Error())
+	}
+
+	data = data[:n]
+	if !bytes.Equal(data, write2) {
+		t.Fatalf("Received unexpected data: %s", data)
```
fishy, line 151: My suggestions here are:

1. Use `%q` instead of `%s` (sometimes when the test failed it's because you read zero data, with `%s` the error message will show as `Received unexpected data: `, while with `%q` it shows as `Received unexpected data: ""` instead, which makes it much more obvious that the failure is due to empty string).
2. Also print out what's the expected data in the failure message (e.g. `t.Fatalf("Received data expected %q, got %q", write2, data)`)

fishy (commented):
I left some nitpicking comments in the test code. None of them is blocking.

dcelasun:
Merged, thanks for the PR!

fishy:
@NIHERASE I think the added test is flaky. If you scroll to the bottom of https://travis-ci.org/github/apache/thrift/jobs/749896929, there are a lot of data races caused by this added test, and also the test failed (looks like the failure is caused by the data race, but I'm not 100% sure),

fishy:
Saw the same test failure on one of the previous travis run on this PR (not on the latest version) as well: https://travis-ci.org/github/apache/thrift/jobs/749705982

fishy:
OK I cannot reproduce this issue with go1.15.6 locally. Looking through travis log we were using go 1.10.8 to run that test, and I do can reproduce this issue with go1.10.8 almost 100% locally.

My current understanding is that in the test we are relying on canceled context to make the http request fail: https://github.com/apache/thrift/blob/e872b350ab0cfd2402340b8ca82b02dc299db69e/lib/go/thrift/http_client_test.go#L129

This works fine in go1.15.6. In go1.10.8 although it did make the `Flush` call next line fail, it doesn't really cause the write loop of the http client to stop correctly. So after you reset the `requestBuffer` inside `Flush`, the write loop of the http client still tries to read from `requestBuffer` in order to write the request, and that caused both data race and the test failure.

I think the big, underlying problem is that when making the http request inside `Flush`, we passed the `requestBuffer`  to the http request (https://github.com/apache/thrift/blob/e872b350ab0cfd2402340b8ca82b02dc299db69e/lib/go/thrift/http_client.go#L208). This doesn't make a copy of the buffer anywhere, so the ownership of the buffer becomes muddy: `THttpClient` thinks it owns the data and can call its `Reset`, while the http request also thinks it owns the data and it's the only one reading/writing from it.

I think the proper fix would be that we need to make a copy of the buffer before sending it to http request. It will hurt performance, though. @dcelasun what do you think?

dcelasun:
Replied on the other PR.

PR #2296, opened 2020-12-17
THRIFT-5322: Implement TConfiguration in Go library

fishy:
Client: go

Define TConfiguration following the spec, and also move the following
configurations scattered around different TTransport/TProtocol into it:

- connect and socket timeouts for TSocket and TSSLSocket
- tls config for TSSLSocket
- max frame size for TFramedTransport
- strict read and strict write for TBinaryProtocol
- proto id for THeaderTransport

Also add TConfiguration support for the following and their factories:

- THeaderTransport and THeaderProtocol
- TBinaryProtocol
- TCompactProtocol
- TFramedTransport
- TSocket
- TSSLSocket

Also define TConfigurationSetter interface for easier TConfiguration
propagation between wrapped TTransports/TProtocols , and add
implementations to the following for propagation
(they don't use anything from TConfiguration themselves):

- StreamTransport
- TBufferedTransport
- TDebugProtocol
- TJSONProtocol
- TSimpleJSONProtocol
- TZlibTransport

TConfigurationSetter are not implemented by the factories of the
"propagation only" TTransports/TProtocols, if they have a factory. For
those use cases, TTransportFactoryConf and TProtocolFactoryConf are
provided to wrap a factory with the ability to propagate TConfiguration.

Also add simple sanity check for TBinaryProtocol and TCompactProtocol's
ReadString and ReadBinary functions. Currently it only report error if
the header length is larger than MaxMessageSize configured in
TConfiguration, for simplicity.

### DIFF
--- a/lib/go/thrift/framed_transport.go
+++ b/lib/go/thrift/framed_transport.go
@@ -28,11 +28,13 @@ import (
 	"io"
 )
 
+// Deprecated: Use DEFAULT_MAX_FRAME_SIZE instead.
 const DEFAULT_MAX_LENGTH = 16384000
 
 type TFramedTransport struct {
 	transport TTransport
-	maxLength uint32
+
+	cfg *TConfiguration
 
 	writeBuf bytes.Buffer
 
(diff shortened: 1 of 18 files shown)


lib/go/thrift/header_protocol.go
```diff
@@ -45,65 +47,49 @@ type THeaderProtocol struct {
...
 func NewTHeaderProtocolFactory() TProtocolFactory {
-	return tHeaderProtocolFactory{
-		protoID: THeaderProtocolDefault,
-	}
+	return NewTHeaderProtocolFactoryConf(nil)
 }
 
-// NewTHeaderProtocolFactoryWithProtocolID creates a factory for THeader with
-// given protocol ID.
-func NewTHeaderProtocolFactoryWithProtocolID(protoID THeaderProtocolID) (TProtocolFactory, error) {
```
fishy, line 100: note that this is a breaking change from a previous version on master branch. however it is not a breaking change from 0.13.0.

dcelasun:
LGTM from a quick read but I won't be able to test until the new year.

lib/go/thrift/framed_transport.go
```diff
@@ -155,7 +189,7 @@ func (p *TFramedTransport) readFrame() error {
 		return err
 	}
 	size := binary.BigEndian.Uint32(buf)
-	if size < 0 || size > p.maxLength {
+	if size < 0 || size > uint32(p.cfg.GetMaxFrameSize()) {
```
Jens-G, line 192: The code was there before but I really wonder when size will become negative on an uint32?

lib/go/thrift/header_protocol.go
```diff
@@ -346,6 +329,13 @@ func (p *THeaderProtocol) Skip(ctx context.Context, fieldType TType) error {
 	return p.protocol.Skip(ctx, fieldType)
 }
 
+// SetTConfiguration implemtns TConfigurationSetter.
```
Jens-G, line 332: ... implem**e**nts ... (there is more than one instance of this typo)

Jens-G (commented):
Good work!

LGTM overall. I only did a code review, did not run any tests on it. Only minor/cosmetic issues found.

lib/go/thrift/framed_transport.go
fishy: Before `p.maxLength` actually has the type of `uint32` instead of `int32`.

To become negative when converting to `int32` the `uint32` value needs to be at least 2GB (MaxInt32 is `1<<31-1`). I'm not sure how much do we really care about that use case. If we do care about it I can change the types of `MaxMessageSize` and `MaxFrameSize` both from `int32` to `uint32`.

lib/go/thrift/header_protocol.go
fishy: Thanks. Fixed.

fishy:
I moved timeouts from `TSocket`/`TSSLSocket` to `TConfiguration`, and added propagation support to more `TTransport`/`TProtocol` implementations, and also updated the commit message/PR description. I believe now `TMemoryBuffer` is the only one not supporting either `TConfiguration` or `TConfigurationSetter`, as it doesn't need any configurations itself and also it's always the terminal `TTransport` that it has nothing underneath to propagate into.

Please take another look.

lib/go/thrift/framed_transport.go
dcelasun: I very much doubt this is a legitimate possibility, `int32` seems fine to me.

lib/go/thrift/header_protocol.go
```diff
@@ -34,76 +34,65 @@ type THeaderProtocol struct {
...
+func NewTHeaderProtocol(trans TTransport) *THeaderProtocol {
+	return newTHeaderProtocolConf(trans, &TConfiguration{
+		noPropagation: true,
+	})
 }
 
-// NewTHeaderProtocol creates a new THeaderProtocol from the underlying
-// transport with default protocol ID.
+// NewTHeaderProtocolConf creates a new THeaderProtocol from the underlying
+// transport with gien TConfiguration.
```
dcelasun, line 49:
```suggestion
// transport with given TConfiguration.
```

dcelasun (approved):
Finally had a chance to play around with this. I've tested framed transport and binary protocol, works as expected.

Thanks for working on this :)

PR #2298, opened 2020-12-22
THRIFT-5326: Expand TException interface in go library

fishy:
Client: go

Add TExceptionType enum type, and add

    TExceptionType() TExceptionType

function to TException definition.


compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -2700,8 +2707,8 @@ void t_go_generator::generate_service_server(t_service* tservice) {
     f_types_ << indent() << "func (p *" << serviceName
                << "Processor) Process(ctx context.Context, iprot, oprot thrift.TProtocol) (success bool, err "
                   "thrift.TException) {" << endl;
-    f_types_ << indent() << "  name, _, seqId, err := iprot.ReadMessageBegin(ctx)" << endl;
-    f_types_ << indent() << "  if err != nil { return false, err }" << endl;
+    f_types_ << indent() << "  name, _, seqId, err2 := iprot.ReadMessageBegin(ctx)" << endl;
```
fishy, line 2710: Note: all the compiler changes starting from this line are because in Process function we specified the returned error type to be `TException` instead of just `error`. I'm not sure why is that, if we can make another breaking change to `TProcessor` and `TProcessorFunction` these changes can be avoided.

lib/go/thrift/exception.go
```diff
@@ -26,19 +26,86 @@ import (
...
+	return errors.New(msg)
+}
+
+// TExceptionType is an enum type to categorize different "subclasses" of TExceptions.
+type TExceptionType byte
+
+// TExceptionType values
+const (
+	TExceptionTypeUnknown     TExceptionType = iota
+	TExceptionTypeCompiled                   // TExceptions defined in thrift files and generated by thrift compiler
```
fishy, line 68: I'm open to suggestions on this name :)

dcelasun: I think it's fine :)

dcelasun (approved):
LGTM. Apologies for the review delay!

lib/go/thrift/exception.go
```diff
@@ -26,19 +26,86 @@ import (
...
+				return NewTProtocolExceptionWithType(t.TypeId(), errors.New(msg))
+			}
+		case TExceptionTypeApplication:
+			if t, ok := err.(TApplicationException); ok {
+				return NewTApplicationException(t.TypeId(), msg)
+			}
+		}
+
+		return wrappedTException{
+			err:            errors.New(msg),
```
tte, line 54: @fishy Hi, thank you for expanding thrift errors and Unwrap support!

At this line we are losing here underline error and seems it's OK by design. I'm just curious why we are setting `msg` instead of storing the original err + msg at wrappedTException? 

 I'm working on error handler for thrift clients and the most common case when `context.Canceled` error occurred. The problem is when we perform read\write operations at thrift fields - we can get `context.Canceled` error, but at the moment I can't unwrap thrift result error because of concatenation of pretending msg.

Saving underlying error would be a simplified error handling process. For example:
```golang
// updated wrappedTException
type wrappedTException  struct {
  err            error
  pretendMsg     string
  tExceptionType TExceptionType
}

func (w wrappedTException) Error() string {
  return w.pretendMsg + w.err.Error()
}

// error occursed during thrift ReadField1 and return context.Canceled
func (p *ServiceResult)  ReadField1(iprot thrift.TProtocol) error {
  if err := p.Field.Read(iprot); err != nil {
    return thrift.PrependError(fmt.Sprintf("%T error reading struct: ", p.Field), err)
  }

  return nil
}

// app error helper fn
func isContextCancelledError(err error) bool {
  if err == nil {
    return false
  }

  switch e := err.(type) {
  case thrift.TException:
   return errors.Is(e, context.Canceled)
  default:
    return false
  }
}
```

fishy: @tte

1. the old `PrependError` doesn't keep the original error either, so this is not a regression :)
2. that being said, this is a reasonable proposal and trivial to implement so I'll just do it
3. I'm curious, which TProtocol/TTransport combination do you use to actually get `context.Canceled` to be wrapped under those `ReadFieldX` functions? I don't think any of the TProtocol implementation would return `context.Canceled` as the error

fishy: Oh I think this is the one coming from `THeaderTransport`: https://github.com/apache/thrift/blob/7f9abb1cc0f4b2793a48f45ddfcf0d2b287cc50c/lib/go/thrift/header_transport.go#L689, as that's the only place in the codebase we return `ctx.Err()` in any way.

So this is actually `TTransportException` wrapped `context.Canceled`, which means doing it inside `wrappedTException` won't fix the issue. We need to implement that in `tTransportException` as well.

fishy: but really in the old code (0.13.0) this won't work with prepended `TTransportException` either: https://github.com/apache/thrift/blob/v0.13.0/lib/go/thrift/exception.go#L33-L35

fishy: @tte See https://github.com/apache/thrift/pull/2308

PR #2305, opened 2021-01-09
THRIFT-5355 Do not rely on compiler and check boundaries

EnigmaTriton:
C++ considers the overflow of a signed integer to be an undefined behavior (even with the "Signed Integers are Two’s Complement" update: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0907r1.html ).

Instead of relying on tests(1) to discover if a compiler does not handle the signed integer overflow as we expect it to, we should add an explicit check before incrementing.

(1) See https://github.com/apache/thrift/blob/c4e899a6d64aa97430ec9f7608d38db2095f6159/lib/cpp/src/thrift/async/TConcurrentClientSyncInfo.cpp#L33-L34

Quick check on https://godbolt.org/z/d7W8ds shows that Clang is able to optimize and just do the +1 so modern compilers should be able to keep the same performances where hardware permits.

In addition, some company rules might require code to do just that:
https://wiki.sei.cmu.edu/confluence/display/c/INT32-C.+Ensure+that+operations+on+signed+integers+do+not+result+in+overflow

### DIFF
--- a/lib/cpp/src/thrift/async/TConcurrentClientSyncInfo.cpp
+++ b/lib/cpp/src/thrift/async/TConcurrentClientSyncInfo.cpp
@@ -184,7 +184,11 @@ int32_t TConcurrentClientSyncInfo::generateSeqId()
       throw apache::thrift::TApplicationException(
         TApplicationException::BAD_SEQUENCE_ID,
         "about to repeat a seqid");
-  int32_t newSeqId = nextseqid_++;
+  int32_t newSeqId = nextseqid_;
+  if (nextseqid_ == (std::numeric_limits<int32_t>::max)())
+    nextseqid_ = (std::numeric_limits<int32_t>::min)();
+  else
+    ++nextseqid_;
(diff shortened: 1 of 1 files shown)


emmenlau (commented):
Looks good to me, albeit I can not judge if this is not what compilers (or processors) anyways would do.

EnigmaTriton:
Regarding the compilers behavior, some of them do have flags to enforce some behavior or another.

In particular GCC has -fwrapv and -ftrapv (which seem to be valid for Clang as well but I did not check the documentation).
-fwrapv explicitly requires the behavior we expect
-ftrapv explicitly forbids overflow (and if company policy enforce this flag, we're doomed at the 10th or 11th command sent).

Some compilers (MSVC) don't have any flag and just let the processor do what it wants (undefined = I don't care what happens, which is a perfectly acceptable behavior for the compiler).

That's why I think it's better to do the check ourselves.

PR #2307, opened 2021-01-19
THRIFT-5337 Go set fields write improvement

BytedanceRPC:
There is a duplicate elements check for set in `writeFields*` function,  and it compares elements using `reflect.DeepEqual` which is expensive.

It's much faster that generates a "DeepEqual" function for set elements and apply it in duplicate elements check, especially for nested struct element.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -152,6 +152,7 @@ public:
                                  const string& tstruct_name,
                                  bool is_result = false,
                                  bool uses_countsetfields = false);
+  void generate_go_struct_equals(std::ostream& out, t_struct* tstruct, const string& tstruct_name);
   void generate_go_function_helpers(t_function* tfunction);
   void get_publicized_name_and_def_value(t_field* tfield,
                                          string* OUT_pub_name,
@@ -228,6 +229,12 @@ public:
   void generate_serialize_set_element(std::ostream& out, t_set* tmap, std::string iter);
 
   void generate_serialize_list_element(std::ostream& out, t_list* tlist, std::string iter);
+
+  void generate_go_equals(std::ostream& out, t_type* ttype, string tgt, string src);
+
+  void generate_go_equals_struct(std::ostream& out, t_type* ttype, string tgt, string src);
+
+  void generate_go_equals_container(std::ostream& out, t_type* ttype, string tgt, string src);
 
   void generate_go_docstring(std::ostream& out, t_struct* tstruct);
 
@@ -307,6 +314,7 @@ private:
   std::set<std::string> package_identifiers_set_;
   std::string read_method_name_;
   std::string write_method_name_;
+  std::string equals_method_name_;
 
   std::set<std::string> commonInitialisms;
 
@@ -724,6 +732,7 @@ void t_go_generator::init_generator() {
     read_method_name_ = "Read";
     write_method_name_ = "Write";
   }
+  equals_method_name_ = "Equals";
 
   while (true) {
     // TODO: Do better error checking here.
@@ -912,7 +921,6 @@ string t_go_generator::go_imports_begin(bool consts) {
   std::vector<string> system_packages;
   system_packages.push_back("bytes");
   system_packages.push_back("context");
-  system_packages.push_back("reflect");
   // If not writing constants, and there are enums, need extra imports.
   if (!consts && get_program()->get_enums().size() > 0) {
     system_packages.push_back("database/sql/driver");
@@ -937,7 +945,6 @@ string t_go_generator::go_imports_end() {
       "var _ = thrift.ZERO\n"
       "var _ = fmt.Printf\n"
       "var _ = context.Background\n"
-      "var _ = reflect.DeepEqual\n"
       "var _ = time.Now\n"
       "var _ = bytes.Equal\n\n");
 }
@@ -1482,6 +1489,9 @@ void t_go_generator::generate_go_struct_definition(ostream& out,
   generate_isset_helpers(out, tstruct, tstruct_name, is_result);
   generate_go_struct_reader(out, tstruct, tstruct_name, is_result);
   generate_go_struct_writer(out, tstruct, tstruct_name, is_result, num_setable > 0);
+  if (!is_result && !is_args) {
+    generate_go_struct_equals(out, tstruct, tstruct_name);
+  }
 
   out << indent() << "func (p *" << tstruct_name << ") String() string {" << endl;
   out << indent() << "  if p == nil {" << endl;
@@ -1849,6 +1859,61 @@ void t_go_generator::generate_go_struct_writer(ostream& out,
     out << indent() << "  return err" << endl;
     out << indent() << "}" << endl << endl;
   }
+}
+
+void t_go_generator::generate_go_struct_equals(ostream& out,
+                                               t_struct* tstruct,
+                                               const string& tstruct_name) {
+  string name(tstruct->get_name());
+  const vector<t_field*>& fields = tstruct->get_sorted_members();
+  vector<t_field*>::const_iterator f_iter;
+  indent(out) << "func (p *" << tstruct_name << ") " << equals_method_name_ << "(other *"
+              << tstruct_name << ") bool {" << endl;
+  indent_up();
+
+  string field_name;
+  string publicize_field_name;
+  out << indent() << "if p == other {" << endl;
+  indent_up();
+  out << indent() << "return true" << endl;
+  indent_down();
+  out << indent() << "} else if p == nil || other == nil {" << endl;
+  indent_up();
+  out << indent() << "return false" << endl;
+  indent_down();
+  out << indent() << "}" << endl;
+
+  for (f_iter = fields.begin(); f_iter != fields.end(); ++f_iter) {
+    field_name = (*f_iter)->get_name();
+    t_type* field_type = (*f_iter)->get_type();
+    publicize_field_name = publicize(field_name);
+    string goType = type_to_go_type_with_opt(field_type, is_pointer_field(*f_iter));
+
+    string tgt = "p." + publicize_field_name;
+    string src = "other." + publicize_field_name;
+    t_type* ttype = field_type->get_true_type();
+    // Compare field contents
+    if (is_pointer_field(*f_iter)
+        && (ttype->is_base_type() || ttype->is_enum() || ttype->is_container())) {
+      string tgtv = "(*" + tgt + ")";
+      string srcv = "(*" + src + ")";
+      out << indent() << "if " << tgt << " != " << src << " {" << endl;
+      indent_up();
+      out << indent() << "if " << tgt << " == nil || " << src << " == nil {" << endl;
+      indent_up();
+      out << indent() << "return false" << endl;
+      indent_down();
+      out << indent() << "}" << endl;
+      generate_go_equals(out, field_type, tgtv, srcv);
+      indent_down();
+      out << indent() << "}" << endl;
+    } else {
+      generate_go_equals(out, field_type, tgt, src);
+    }
+  }
+  out << indent() << "return true" << endl;
+  indent_down();
+  out << indent() << "}" << endl << endl;
 }
 
 /**
@@ -3389,15 +3454,30 @@ void t_go_generator::generate_serialize_container(ostream& out,
   } else if (ttype->is_set()) {
     t_set* tset = (t_set*)ttype;
     out << indent() << "for i := 0; i<len(" << prefix << "); i++ {" << endl;
-    out << indent() << "  for j := i+1; j<len(" << prefix << "); j++ {" << endl;
+    indent_up();
+    out << indent() << "for j := i+1; j<len(" << prefix << "); j++ {" << endl;
+    indent_up();
     string wrapped_prefix = prefix;
     if (pointer_field) {
       wrapped_prefix = "(" + prefix + ")";
     }
-    out << indent() << "    if reflect.DeepEqual(" << wrapped_prefix << "[i]," << wrapped_prefix << "[j]) { " << endl;
-    out << indent() << "      return thrift.PrependError(\"\", fmt.Errorf(\"%T error writing set field: slice is not unique\", " << wrapped_prefix << "[i]))" << endl;
-    out << indent() << "    }" << endl;
-    out << indent() << "  }" << endl;
+    string goType = type_to_go_type(tset->get_elem_type());
+    out << indent() << "if func(tgt, src " << goType << ") bool {" << endl;
+    indent_up();
+    generate_go_equals(out, tset->get_elem_type(), "tgt", "src");
+    out << indent() << "return true" << endl;
+    indent_down();
+    out << indent() << "}(" << wrapped_prefix << "[i], " << wrapped_prefix << "[j]) {" << endl;
+    indent_up();
+    out << indent()
+        << "return thrift.PrependError(\"\", fmt.Errorf(\"%T error writing set field: slice is not "
+           "unique\", "
+        << wrapped_prefix << "))" << endl;
+    indent_down();
+    out << indent() << "}" << endl;
+    indent_down();
+    out << indent() << "}" << endl;
+    indent_down();
     out << indent() << "}" << endl;
     out << indent() << "for _, v := range " << prefix << " {" << endl;
     indent_up();
@@ -3461,6 +3541,111 @@ void t_go_generator::generate_serialize_list_element(ostream& out, t_list* tlist
   t_field efield(tlist->get_elem_type(), "");
   efield.set_req(t_field::T_OPT_IN_REQ_OUT);
   generate_serialize_field(out, &efield, prefix);
+}
+
+/**
+ * Compares any type
+ */
+void t_go_generator::generate_go_equals(ostream& out, t_type* ori_type, string tgt, string src) {
+
+  t_type* ttype = get_true_type(ori_type);
+  // Do nothing for void types
+  if (ttype->is_void()) {
+    throw "compiler error: cannot generate equals for void type: " + tgt;
+  }
+
+  if (ttype->is_struct() || ttype->is_xception()) {
+    generate_go_equals_struct(out, ttype, tgt, src);
+  } else if (ttype->is_container()) {
+    generate_go_equals_container(out, ttype, tgt, src);
+  } else if (ttype->is_base_type() || ttype->is_enum()) {
+    out << indent() << "if ";
+    if (ttype->is_base_type()) {
+      t_base_type::t_base tbase = ((t_base_type*)ttype)->get_base();
+      switch (tbase) {
+      case t_base_type::TYPE_VOID:
+        throw "compiler error: cannot equals void: " + tgt;
+        break;
+
+      case t_base_type::TYPE_STRING:
+        if (ttype->is_binary()) {
+          out << "bytes.Compare(" << tgt << ", " << src << ") != 0";
+        } else {
+          out << tgt << " != " << src;
+        }
+        break;
+
+      case t_base_type::TYPE_BOOL:
+      case t_base_type::TYPE_I8:
+      case t_base_type::TYPE_I16:
+      case t_base_type::TYPE_I32:
+      case t_base_type::TYPE_I64:
+      case t_base_type::TYPE_DOUBLE:
+        out << tgt << " != " << src;
+        break;
+
+      default:
+        throw "compiler error: no Go name for base type " + t_base_type::t_base_name(tbase);
+      }
+    } else if (ttype->is_enum()) {
+      out << tgt << " != " << src;
+    }
+
+    out << " { return false }" << endl;
+  } else {
+    throw "compiler error: Invalid type in generate_go_equals '" + ttype->get_name() + "' for '"
+        + tgt + "'";
+  }
+}
+
+/**
+ * Compares the members of a struct
+ */
+void t_go_generator::generate_go_equals_struct(ostream& out,
+                                               t_type* ttype,
+                                               string tgt,
+                                               string src) {
+  (void)ttype;
+  out << indent() << "if !" << tgt << "." << equals_method_name_ << "(" << src
+      << ") { return false }" << endl;
+}
+
+/**
+ * Compares any container type
+ */
+void t_go_generator::generate_go_equals_container(ostream& out,
+                                                  t_type* ttype,
+                                                  string tgt,
+                                                  string src) {
+  out << indent() << "if len(" << tgt << ") != len(" << src << ") { return false }" << endl;
+  if (ttype->is_map()) {
+    t_map* tmap = (t_map*)ttype;
+    out << indent() << "for k, _tgt := range " << tgt << " {" << endl;
+    indent_up();
+    string element_source = tmp("_src");
+    out << indent() << element_source << " := " << src << "[k]" << endl;
+    generate_go_equals(out, tmap->get_val_type(), "_tgt", element_source);
+    indent_down();
+    indent(out) << "}" << endl;
+  } else if (ttype->is_list() || ttype->is_set()) {
+    t_type* elem;
+    if (ttype->is_list()) {
+      t_list* temp = (t_list*)ttype;
+      elem = temp->get_elem_type();
+    } else {
+      t_set* temp = (t_set*)ttype;
+      elem = temp->get_elem_type();
+    }
+    out << indent() << "for i, _tgt := range " << tgt << " {" << endl;
+    indent_up();
+    string element_source = tmp("_src");
+    out << indent() << element_source << " := " << src << "[i]" << endl;
+    generate_go_equals(out, elem, "_tgt", element_source);
+    indent_down();
+    indent(out) << "}" << endl;
+  } else {
+    throw "INVALID TYPE IN generate_go_equals_container '" + ttype->get_name();
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -152,6 +152,10 @@ class t_go_generator : public t_generator {
                                  const string& tstruct_name,
                                  bool is_result = false,
                                  bool uses_countsetfields = false);
+  void generate_go_struct_deepequal(std::ostream& out,
```
fishy, line 155: Can we rename all `*deepequal*` into `*equals*`? I don't think it's accurate to call them "deepequal".

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1847,6 +1860,56 @@ void t_go_generator::generate_go_struct_writer(ostream& out,
   }
 }
 
+void t_go_generator::generate_go_struct_deepequal(ostream& out,
+                                               t_struct* tstruct,
```
fishy, line 1864: nit: I'm not sure what's the coding style for compiler c++ code but this indentation looks weird.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1847,6 +1860,56 @@ void t_go_generator::generate_go_struct_writer(ostream& out,
...
+  string name(tstruct->get_name());
+  const vector<t_field*>& fields = tstruct->get_sorted_members();
+  vector<t_field*>::const_iterator f_iter;
+  indent(out) << "func (p *" << tstruct_name << ") " << deepequal_method_name_ << "(other *"
+              << tstruct_name << ") bool {" << endl;
+  indent_up();
+
+  string field_name;
+  string publicize_field_name;
+  int32_t field_id = -1;
```
fishy, line 1879: this is unused and it will cause certain compilers to fail:

```
src/thrift/generate/t_go_generator.cc:1879:11: error: unused variable ‘field_id’ [-Werror=unused-variable]
 1879 |   int32_t field_id = -1;
      |           ^~~~~~~~
cc1plus: all warnings being treated as errors
```

fishy (commented):
I haven't gone through all the changes yet, but wanted to give you some early feedback.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1847,6 +1860,56 @@ void t_go_generator::generate_go_struct_writer(ostream& out,
...
+    publicize_field_name = publicize(field_name);
+    string goType = type_to_go_type_with_opt(field_type, is_pointer_field(*f_iter));
+
+    string tgt = "p." + publicize_field_name;
+    string src = "other." + publicize_field_name;
+    t_type* ttype = field_type->get_true_type();
+    if (is_pointer_field(*f_iter)
+        && (ttype->is_base_type() || ttype->is_enum() || ttype->is_container())) {
+    tgt = "(*" + tgt + ")";
+    src = "(*" + src + ")";
```
fishy, line 1902: 1. wrong indentation
2. This generates code like this for optional fields:
```go
if (*p.FieldN) != (*other.FieldN) { return false }
```
this will cause panic when the field is not set.

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: Thanks for review, *equals* sounds better.

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: Well, I use clang-format according to https://github.com/apache/thrift/blob/master/.clang-format

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: removed latest commit

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: Thanks, I add a pointer check to avoid panic last commit

dcelasun:
Looks reasonable at a first glance, will do a proper review later.

I wonder if there are any cases covered by `reflect.DeepEqual` that aren't covered by this method. If there are, that would be a BC break and we'd have to mention it in the changelog.

BytedanceRPC:
@dcelasun The process of `Equals` is pretty like `Write` (deserialize). Since `write` can cover all cases, I believe so can `Equals`.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1847,6 +1862,48 @@ void t_go_generator::generate_go_struct_writer(ostream& out,
...
+    field_name = (*f_iter)->get_name();
+    t_type* field_type = (*f_iter)->get_type();
+    publicize_field_name = publicize(field_name);
+    string goType = type_to_go_type_with_opt(field_type, is_pointer_field(*f_iter));
+
+    string tgt = "p." + publicize_field_name;
+    string src = "other." + publicize_field_name;
+    t_type* ttype = field_type->get_true_type();
+    if (is_pointer_field(*f_iter)
+        && (ttype->is_base_type() || ttype->is_enum() || ttype->is_container())) {
```
fishy, line 1893: I'm not sure why the condition after `&&` is needed, can you explain?

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -3439,6 +3511,123 @@ void t_go_generator::generate_serialize_list_element(ostream& out, t_list* tlist
...
+                                                  string src) {
+  (void)ttype;
+  out << indent() << "if !" << tgt << "." << equals_method_name_ << "(" << src
+      << ") { return false }" << endl;
+}
+
+/**
+ * Compares any container type
+ */
+void t_go_generator::generate_go_equals_container(std::ostream& out,
```
fishy, line 3599: For containers, another optimization you can do is to compare the sizes first. If the sizes doesn't match you can return `false` directly.

fishy (commented):
I like where this is going, but left some small comments.

Also I believe with this change there's no other way the `reflect` import is used, so we can totally remove it from compiled code now.

fishy:
I believe currently there's some bug in your implementation. The result is different from `reflect.DeepEqual`, Here's the code I used:

`test.thrift`:
```thrift
namespace go test

enum Enum {
  One = 1,
  Two = 2,
}

struct Struct {
  1: optional string Str,
  2: optional i32 I32,
  3: optional set<Enum> Enums,
}
```

`bench_test.go`:
```go
package thriftsetbench

import (
        "reflect"
        "testing"

        "github.com/apache/thrift/lib/go/thrift"
        "github.com/fishy/thriftsetbench/gen-go/test"
)

func BenchmarkEqual(b *testing.B) {
        equalGetter := func() (*test.Struct, *test.Struct) {
                s1 := &test.Struct{
                        Str: thrift.StringPtr("foobar"),
                        Enums: []test.Enum{
                                test.Enum_One,
                                test.Enum_Two,
                        },
                }
                s2 := &test.Struct{
                        Str: thrift.StringPtr("foobar"),
                        Enums: []test.Enum{
                                test.Enum_One,
                                test.Enum_Two,
                        },
                }
                return s1, s2
        }
        nonEqualGetter := func() (*test.Struct, *test.Struct) {
                s1 := &test.Struct{
                        Str: thrift.StringPtr("foobar"),
                        Enums: []test.Enum{
                                test.Enum_One,
                        },
                }
                s2 := &test.Struct{
                        Str: thrift.StringPtr("foobar"),
                        Enums: []test.Enum{
                                test.Enum_Two,
                        },
                }
                return s1, s2
        }

        for label, getter := range map[string]func() (*test.Struct, *test.Struct){
                "equal":    equalGetter,
                "nonequal": nonEqualGetter,
        } {
                b.Run(label, func(b *testing.B) {
                        s1, s2 := getter()
                        reflectResult := reflect.DeepEqual(s1, s2)
                        equalsResult := s1.Equals(s2)
                        if reflectResult != equalsResult {
                                b.Errorf("Reflect result: %v, Equals result: %v", reflectResult, equalsResult)
                        }

                        b.Run("reflect", func(b *testing.B) {
                                b.RunParallel(func(pb *testing.PB) {
                                        for pb.Next() {
                                                reflect.DeepEqual(s1, s2)
                                        }
                                })
                        })

                        b.Run("equals", func(b *testing.B) {
                                b.RunParallel(func(pb *testing.PB) {
                                        for pb.Next() {
                                                s1.Equals(s2)
                                        }
                                })
                        })
                })
        }
}
```

result:
```
$ go test -bench . -benchmem
goos: linux
goarch: amd64
pkg: github.com/fishy/thriftsetbench
BenchmarkEqual/equal/reflect-12         10597189               113 ns/op              64 B/op          6 allocs/op
BenchmarkEqual/equal/equals-12          1000000000               0.680 ns/op           0 B/op          0 allocs/op
BenchmarkEqual/nonequal/reflect-12              12201334                99.9 ns/op            48 B/op          4 allocs/op
BenchmarkEqual/nonequal/equals-12               1000000000               0.712 ns/op           0 B/op          0 allocs/op
--- FAIL: BenchmarkEqual/nonequal
    bench_test.go:54: Reflect result: false, Equals result: true
--- FAIL: BenchmarkEqual
FAIL
exit status 1
FAIL    github.com/fishy/thriftsetbench 4.183s
```

We likely would need to have some thorough test for the `Equals` implementation with this PR.

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: for struct and xception, since they have `Equals` method, there is no need to get their value

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: Good idea, will do

BytedanceRPC:
@fishy It's a bug in `generate_go_equals_guard`, I fixed it. 
Yes, some thorough test is pretty necessary. I will also do more test.

fishy:
@BytedanceRPC there's another go compiler change merged, so can you rebase your PR on latest master branch?

Also please add some tests for the `Equals` implementation. This change looks promising, but we need the confidence that it's implemented correctly.

BytedanceRPC:
@fishy rebased PR on latest master branch, and added `Equals` test under lib/go/test

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1851,6 +1864,65 @@ void t_go_generator::generate_go_struct_writer(ostream& out,
   }
 }
 
+void t_go_generator::generate_go_struct_equals(ostream& out,
+                                               t_struct* tstruct,
+                                               const string& tstruct_name,
+                                               bool is_args_or_result) {
+  if (is_args_or_result) {
+    return;
+  }
```
fishy, line 1873: nit: I think we should do this check on the call site instead, and currently there's only one caller of it (Line 1497), so there should be:

```
if !is_result && !is_args {
  generate_go_struct_equals(out, tstruct, tstruct_name);
}
```

Semantically this function is called "generate_*", but when one of its arg is passed in true it doesn't generate anything, which feels wrong to me.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -3463,6 +3550,112 @@ void t_go_generator::generate_serialize_list_element(ostream& out, t_list* tlist
   generate_serialize_field(out, &efield, prefix);
 }
 
+/**
+ * Compares any type
+ */
+void t_go_generator::generate_go_equals(std::ostream& out,
+                                           t_type* ori_type,
```
fishy, line 3557: nit: this indentation looks weird. usually you either align with `(`, or use fixed 2/4/6 spaces. this is neither.

I don't believe clang-format is used to generate this indentation either, at least it's not enforced by travis. when I run `clang-format -i --style=file compiler/cpp/src/thrift/generate/t_go_generator.cc` it made a ton of changes in this file.

lib/go/test/EqualsTest.thrift
```diff
@@ -0,0 +1,81 @@
...
+  10: optional i64 OptI64Foo,
+  11: double DoubleFoo,
+  12: optional double OptDoubleFoo,
+  13: string StrFoo,
+  14: optional string OptStrFoo,
+  15: binary BinFoo,
+  16: optional binary OptBinFoo,
+  17: EnumFoo EnumFoo,
+  18: optional EnumFoo OptEnumFoo,
+}
```
fishy, line 25: I would suggest also add a few typedef'd primitive types into this (and also into `ListEqualsFoo`, `SetEqualsFoo`, `MapEqualsFoo`).

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"equalstest"
+	"strconv"
+	"testing"
+)
+
+func TestEquals(t *testing.T) {
+	basicTgt, basicSrc := genBasicFoo(), genBasicFoo()
+	if !basicTgt.Equals(basicSrc) {
+		t.Fatal("BasicEqualsFoo.Equals() test failed")
```
fishy, line 12: These test failures should be `t.Error` instead of `t.Fatal`. Fatal stops the test from running, so if someone made a mistake in the future and failed multiple cases inside this test, they will only see the first failure in the output.

Fatal should only be used when the following tests are depending on this one (e.g. used for nil checks).

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	mapTgt, mapSrc := genMapFoo(), genMapFoo()
+	if !mapTgt.Equals(mapSrc) {
+		t.Fatal("MapEqualsFoo.Equals() test failed")
+	}
+	mapSrc.OptI64I64StringMapMapFoo[1][1] = "0"
+	if mapTgt.Equals(mapSrc) {
+		t.Fatal("MapEqualsFoo.Equals() test failed")
+	}
+}
```
fishy, line 50: Another issue with the test is that all the optional fields are set, so if someone missed the nil check in the compiler (like in the first version of this PR) this test would not catch that error. I would suggest add a few more cases to each `gen*` call, while changing some of the optional fields to `nil` (for example, you should add a test to change one of the optional field to nil and make sure `Equals` return false, then change the other also to nil and make sure `Equals` return true).

fishy (commented):
Overall lgtm, but there are a few things can be improved in the test code.

Also you can remove https://github.com/apache/thrift/blob/c843abbe20427f07143266ca565a820945ff8d23/compiler/cpp/src/thrift/generate/t_go_generator.cc#L927 and https://github.com/apache/thrift/blob/c843abbe20427f07143266ca565a820945ff8d23/compiler/cpp/src/thrift/generate/t_go_generator.cc#L952. Since this change eliminated the usage of `reflect.DeepEqual`, the reflect package is not used any more in compiled go code.

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: looks reasonable, thanks for suggestion.

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: not sure is the `std::` making indentation looks weird. Indentations looks proper after I remove `std::` and apply `clang-format`.
To avoid unnecessary change, I only apply the output of clang-format on my changes in `compiler/cpp/src/thrift/generate/t_go_generator.cc`.

lib/go/test/EqualsTest.thrift
BytedanceRPC: good suggestion

lib/go/test/tests/equals_test.go
BytedanceRPC: Thanks for suggestion

lib/go/test/tests/equals_test.go
BytedanceRPC: Yes, I missed `optional field`'s pointer check test.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -3463,6 +3543,109 @@ void t_go_generator::generate_serialize_list_element(ostream& out, t_list* tlist
...
+}
+
+/**
+ * Compares the members of a struct
+ */
+void t_go_generator::generate_go_equals_struct(ostream& out,
+                                               t_type* ttype,
+                                               string tgt,
+                                               string src) {
+  (void)ttype;
```
fishy, line 3608: what does this line do?

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -3463,6 +3543,109 @@ void t_go_generator::generate_serialize_list_element(ostream& out, t_list* tlist
...
+  if (ttype->is_map()) {
+    t_map* tmap = (t_map*)ttype;
+    out << indent() << "for k, _tgt := range " << tgt << " {" << endl;
+    indent_up();
+    string element_source = tmp("_src");
+    out << indent() << element_source << " := " << src << "[k]" << endl;
+    generate_go_equals(out, tmap->get_val_type(), "_tgt", element_source);
+    indent_down();
+    indent(out) << "}" << endl;
+  } else if (ttype->is_list() || ttype->is_set()) {
```
fishy, line 3630: should this just be `} else {`, or maybe add another `} else { throw "unsupported container" }` at the end? currently if this is a container but it's not a map/list/set (I'm not sure if we have other container types) this function only checks for length and then does nothing.

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,219 @@
...
+		OptI64Foo:    &(&struct{ x int64 }{4}).x,
+		DoubleFoo:    5,
+		OptDoubleFoo: &(&struct{ x float64 }{5}).x,
+		StrFoo:       "6",
+		OptStrFoo:    &(&struct{ x string }{"6"}).x,
+		BinFoo:       []byte("7"),
+		OptBinFoo:    []byte("7"),
+		EnumFoo:      equalstest.EnumFoo_e1,
+		OptEnumFoo:   &(&struct{ x equalstest.EnumFoo }{equalstest.EnumFoo_e1}).x,
+	}
```
fishy, line 117: the newly added typedef'd fields are not added here (and other `gen*` functions)

test/go/src/bin/testequals/main.go
```diff
@@ -0,0 +1,179 @@
+package main
```
fishy, line 1: hmm why do we need to add this directory/file?

BytedanceRPC: sry, wrong push

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: try to keep `generate_go_equals*` with the same arguments, but `ttype` not really used here.

compiler/cpp/src/thrift/generate/t_go_generator.cc
BytedanceRPC: throw exception in else condition should work well according to `generate_deserialize_container`

lib/go/test/tests/equals_test.go
BytedanceRPC: sry, wrong push

fishy (approved):
Thanks. Looks good. I'll merge it once the tests pass.

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,243 @@
...
+		BinFoo:       []byte("7"),
+		OptBinFoo:    []byte("7"),
+		EnumFoo:      equalstest.EnumFoo_e1,
+		OptEnumFoo:   &(&struct{ x equalstest.EnumFoo }{equalstest.EnumFoo_e1}).x,
+		MyByteFoo:    equalstest.Mybyte(8),
+		OptMyByteFoo: &(&struct{ x equalstest.Mybyte }{8}).x,
+		MyStrFoo:     equalstest.Mystr("9"),
+		OptMyStrFoo:  &(&struct{ x equalstest.Mystr }{"9"}).x,
+		MyBinFoo:     equalstest.Mybin("10"),
+		OptMyBinFoo:  &(&struct{ x equalstest.Mybin }{equalstest.Mybin("10")}).x,
```
fishy, line 122: This one does not compile, see https://travis-ci.org/github/apache/thrift/jobs/757426362

fishy (changes requested):

lib/go/test/tests/equals_test.go
fishy: Oh I think you just misses another level of `&()` around the whole thing.

But this is getting really hard to read so please just use some local variables instead.

fishy: You can just write them this way, which is way simpler:

```go
                OptEnumFoo:   equalstest.EnumFooPtr(equalstest.EnumFoo_e1),
                MyByteFoo:    equalstest.Mybyte(8),
                OptMyByteFoo: equalstest.MybytePtr(8),
                MyStrFoo:     equalstest.Mystr("9"),
                OptMyStrFoo:  equalstest.Mystr("9"),
                MyBinFoo:     equalstest.Mybin("10"),
                OptMyBinFoo:  equalstest.MybinPtr(equalstest.Mybin("10")),
```

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,243 @@
...
+		StrFoo:       "6",
+		OptStrFoo:    &(&struct{ x string }{"6"}).x,
+		BinFoo:       []byte("7"),
+		OptBinFoo:    []byte("7"),
+		EnumFoo:      equalstest.EnumFoo_e1,
+		OptEnumFoo:   &(&struct{ x equalstest.EnumFoo }{equalstest.EnumFoo_e1}).x,
+		MyByteFoo:    equalstest.Mybyte(8),
+		OptMyByteFoo: &(&struct{ x equalstest.Mybyte }{8}).x,
+		MyStrFoo:     equalstest.Mystr("9"),
+		OptMyStrFoo:  &(&struct{ x equalstest.Mystr }{"9"}).x,
```
fishy, line 120:
```suggestion
		OptMyStrFoo:  equalstest.Mystr("9"),
```

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,243 @@
...
+		DoubleFoo:    5,
+		OptDoubleFoo: &(&struct{ x float64 }{5}).x,
+		StrFoo:       "6",
+		OptStrFoo:    &(&struct{ x string }{"6"}).x,
+		BinFoo:       []byte("7"),
+		OptBinFoo:    []byte("7"),
+		EnumFoo:      equalstest.EnumFoo_e1,
+		OptEnumFoo:   &(&struct{ x equalstest.EnumFoo }{equalstest.EnumFoo_e1}).x,
+		MyByteFoo:    equalstest.Mybyte(8),
+		OptMyByteFoo: &(&struct{ x equalstest.Mybyte }{8}).x,
```
fishy, line 118:
```suggestion
		OptMyByteFoo: equalstest.MybytePtr(8),
```

lib/go/test/tests/equals_test.go
```diff
@@ -0,0 +1,243 @@
...
+		I64Foo:       4,
+		OptI64Foo:    &(&struct{ x int64 }{4}).x,
+		DoubleFoo:    5,
+		OptDoubleFoo: &(&struct{ x float64 }{5}).x,
+		StrFoo:       "6",
+		OptStrFoo:    &(&struct{ x string }{"6"}).x,
+		BinFoo:       []byte("7"),
+		OptBinFoo:    []byte("7"),
+		EnumFoo:      equalstest.EnumFoo_e1,
+		OptEnumFoo:   &(&struct{ x equalstest.EnumFoo }{equalstest.EnumFoo_e1}).x,
```
fishy, line 116:
```suggestion
		OptEnumFoo:   equalstest.EnumFooPtr(equalstest.EnumFoo_e1),
```

lib/go/test/tests/equals_test.go
fishy: Oh right it's because for binary type optional are the same as required (because slices are already pointers), so this should be:

```suggestion
		OptMyBinFoo:  equalstest.MybinPtr(equalstest.Mybin("10")),
```

fishy: This change passes the tests: https://travis-ci.org/github/apache/thrift/builds/757454619

https://github.com/apache/thrift/pull/2321/commits/8b820c3dd40a00c7f3243dff496eb59730e2d727 are all the changes I made but I believe only the change on this line is required (but I'd still suggest you to make other changes as well, and run `gofmt -s` on this file)

fishy:
Can you squash your PR to a single commit, with proper commit message (what you used in the PR title and description)?

lib/go/test/tests/equals_test.go
BytedanceRPC: I think it shoud be just `OptMyBinFoo:  equalstest.Mybin("10"),`

fishy (approved):
Travis should finish in 20min or so and I'll merge it after that.

PR #2310, opened 2021-01-20
THRIFT-5339: Add `TObject` trait to cover sync read & write methods of Struct/Union/Enum

JasonShui:
Each Struct/Union/Enum generated by thrift-compiler will get two methods: `read_from_in_protocol` and `write_to_out_protocol`. 
Firstly, I think these two methods are representing a certain ability of serialization, so they should be put in a trait;
Moreover, if this trait is provided, we can use trait bounding in our user codes, such as
```rust
// pseudo code
fn serialize_struct<T>(s: T) -> Vec<u8> where T: TObject {
    s.write_to_out_protocol(protocol);
    // do other things..
}
```
It could also be a pretty useful marker trait as you can see.
The name of `TObject` is only a superficial opinion, but I believe that there are enough reasons to make the trait.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -565,7 +565,7 @@ void t_rs_generator::render_attributes_and_includes() {
   f_gen_ << endl;
   f_gen_ << "use thrift::OrderedFloat;" << endl;
   f_gen_ << "use thrift::{ApplicationError, ApplicationErrorKind, ProtocolError, ProtocolErrorKind, TThriftClient};" << endl;
-  f_gen_ << "use thrift::protocol::{TFieldIdentifier, TListIdentifier, TMapIdentifier, TMessageIdentifier, TMessageType, TInputProtocol, TOutputProtocol, TSetIdentifier, TStructIdentifier, TType};" << endl;
+  f_gen_ << "use thrift::protocol::{TFieldIdentifier, TListIdentifier, TMapIdentifier, TMessageIdentifier, TMessageType, TInputProtocol, TOutputProtocol, TSetIdentifier, TStructIdentifier, TType, TObject};" << endl;
   f_gen_ << "use thrift::protocol::field_id;" << endl;
   f_gen_ << "use thrift::protocol::verify_expected_message_type;" << endl;
   f_gen_ << "use thrift::protocol::verify_expected_sequence_number;" << endl;
@@ -898,7 +898,7 @@ void t_rs_generator::render_enum_definition(t_enum* tenum, const string& enum_na
 }
 
 void t_rs_generator::render_enum_impl(const string& enum_name) {
-  f_gen_ << "impl " << enum_name << " {" << endl;
+  f_gen_ << "impl TObject for " << enum_name << " {" << endl;
   indent_up();
 
   // taking enum as 'self' here because Thrift enums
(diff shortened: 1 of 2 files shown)


JasonShui:
my trait bounding demo: https://github.com/JasonShui/thrift-rs-sample/blob/master/src/main.rs

allengeorge:
Thank you for your PR! I'll take a look at thisi @JasonShui.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -1141,13 +1141,18 @@ void t_rs_generator::render_struct_impl(
...
   }
 
-  render_struct_sync_read(struct_name, tstruct, struct_type);
-  render_struct_sync_write(tstruct, struct_type);
-
   if (struct_type == t_rs_generator::T_RESULT) {
     render_result_struct_to_result_method(tstruct);
   }
 
+  f_gen_ << "}" << endl;
```
allengeorge, line 1148: Please fix indentation here. You need to add an `indent_down()` call before closing the `impl` block..

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -1141,13 +1141,18 @@ void t_rs_generator::render_struct_impl(
...
+  f_gen_ << endl;
+
+  f_gen_ << "impl TObject for " << struct_name << " {" << endl;
+
```
allengeorge, line 1152: Please fix indentation here. You need to add an `indent_up` before rendering the read/write methods.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -907,16 +907,16 @@ void t_rs_generator::render_enum_impl(const string& enum_name) {
   // taking a reference to the enum
   f_gen_
     << indent()
-    << "pub fn write_to_out_protocol(self, o_prot: &mut dyn TOutputProtocol) -> thrift::Result<()> {"
+    << "fn write_to_out_protocol(&self, o_prot: &mut dyn TOutputProtocol) -> thrift::Result<()> {"
```
allengeorge, line 910: Please remove the comment on line 984. As a result of your change the comment is no longer valid.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -1406,7 +1411,6 @@ void t_rs_generator::render_struct_sync_write(
 ) {
   f_gen_
     << indent()
-    << visibility_qualifier(struct_type)
```
allengeorge, line 1409: I assume the reason why this can be removed is because the methods on the trait are already in the public scope?

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -898,7 +898,7 @@ void t_rs_generator::render_enum_definition(t_enum* tenum, const string& enum_na
 }
 
 void t_rs_generator::render_enum_impl(const string& enum_name) {
-  f_gen_ << "impl " << enum_name << " {" << endl;
+  f_gen_ << "impl TObject for " << enum_name << " {" << endl;
```
allengeorge, line 901: Please rename to `TSerializable`.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
allengeorge: Also, please note that as a result of this _all_ structs (args and results included) suddenly have public read/write methods. This was not intended originally, and it's not clear we want it as a result of this PR as well.

allengeorge (changes requested):
Hi @JasonShui - thank you for making this change, and sorry for the delay in reviewing this.

I believe [THRIFT-5339](https://issues.apache.org/jira/browse/THRIFT-5399) duplicates [THRIFT-4100](https://issues.apache.org/jira/browse/THRIFT-4100). The trait name suggested there after much earlier discussion was `TSerializable`.

Some other comments:

1. It appears that _all_ structs (including args and results structs) now have `TObject` defined on them. I think this trait should _only_ be defined for IDL-defined types.
2. Please check indentation of generated code. You can do so by running `make` in the `test` or `tutorial` directory and looking at the output code.
3. Please run the cross-tests after your change to ensure that tests before/after your change still function.

JasonShui:
Thanks for your comments @allengeorge , I will look through them and feed back later :)

WANG-lp:
This change really helps in some situations. Hope we can move forward on this.

JasonShui:
I will try to fix them this month. Sorry for the late update :(

PR #2311, opened 2021-01-20
THRIFT-5338: Remove lib/go/thrift/go.mod

fishy:
Client: go

Having it under a subdirectory has some unexpected consequences, so
remove it for now. Another PR will be open up later to add it back to
the root directory.

### DIFF
--- a/lib/go/thrift/go.mod
+++ b/lib/go/thrift/go.mod
@@ -1,3 +1,0 @@
-module github.com/apache/thrift/lib/go/thrift
-
-go 1.14


dcelasun:
Is this related to these issues?

https://github.com/golang/go/issues/34055
https://github.com/golang/go/issues/33562

fishy:
Yea kinda.

With `go.mod` in a subdirectory it still works, but go cannot get the correct version from git tag and this will be treated as version `v0.0.0-timestamp-commitsha1`

fishy:
uh it caused some issues in the CI process. looking.

fishy:
@dcelasun To unblock things I'm gonna change this PR to "remove go.mod" instead. I'll create another PR to add `go.mod` back to the root directory, but that will take longer and it does not necessarily need to be a 0.14.0 blocker.

dcelasun:
Sounds good. It's probably worth delaying 0.14 for this, who knows when 0.15 will happen.

fishy:
@dcelasun Adding `go.mod` back to root directory comes with some bigger issues I described in https://github.com/golang/go/issues/34055#issuecomment-763880343. It will likely take a lot longer so we probably don't want to wait for it for 0.14.0 :)

Travis just finished. Gonna merge this one first.

dcelasun:
I'll follow that thread then, thanks!

PR #2312, opened 2021-01-20
[skip ci] Clarify Compact Protocol var int encoding definition

jviotti:
I'm having problems following the var int explanation from the Compact
Protocol spec. Here is an attempt to clarify it with more precise
encoding steps and with an example.

I'm also mentioning, for completeness, that the formal name of such
variable-length integer encoding is Unsigned LEB128 (Unsigned Little
Endian Base-128).

Signed-off-by: Juan Cruz Viotti <jv@jviotti.com>

  

### DIFF
--- a/doc/specs/thrift-compact-protocol.md
+++ b/doc/specs/thrift-compact-protocol.md
@@ -61,9 +61,21 @@ def longToZigZag(n: Long): Long = (n << 1) ^ (n >> 63)
 def zigzagToLong(n: Long): Long = (n >>> 1) ^ - (n & 1)
 ```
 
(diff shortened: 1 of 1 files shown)


jviotti:
Also, related to this: The specification says that:

> But while the binary protocol encodes the int64 in 8 bytes in big endian order, the compact protocol encodes it in little endian order - this is due to an early implementation bug that finally became the de-facto standard.

But I'm not sure if this is 100% accurate. The var int encoding scheme that Thrift is using takes a Big Endian integer, but the result is technically Little Endian. What do you think?

PR #2314, opened 2021-01-22
Make GetResponseHeadersFromClient work with client middlewares

fishy:
Client: go

When GetResponseHeadersFromClient was added we only have TStandardClient
and it rely on the TClient is actually a *TStandardClient. With client
middlewares and WrappedTClient added later, that's no longer true.

Define an optional interface, TStandardClientUnwrapper, to return the
underlying *TStandardClient, so that GetResponseHeadersFromClient can
still work with wrapped clients.

### DIFF
--- a/lib/go/thrift/client.go
+++ b/lib/go/thrift/client.go
@@ -9,6 +9,17 @@ type TClient interface {
 	Call(ctx context.Context, method string, args, result TStruct) error
 }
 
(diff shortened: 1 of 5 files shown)


dcelasun (approved):

fishy:
Closed in favor of https://github.com/apache/thrift/pull/2315.

PR #2315, opened 2021-01-22
THRIFT-4914: Make TClient.Call to return the response meta

fishy:
Client: go

Make a breaking change so that TClient.Call returns the response
meta, currently only contains headers but could be expanded in the
future, and make a compiler change to compiler generated clients to take
advantage of that and provide access to response metadata to users.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -1493,11 +1493,15 @@ void t_go_generator::generate_go_struct_definition(ostream& out,
 
   if (is_exception) {
     out << indent() << "func (p *" << tstruct_name << ") Error() string {" << endl;
-    out << indent() << indent() << "return p.String()" << endl;
+    indent_up();
+    out << indent() << "return p.String()" << endl;
+    indent_down();
     out << indent() << "}" << endl << endl;
 
     out << indent() << "func (" << tstruct_name << ") TExceptionType() thrift.TExceptionType {" << endl;
-    out << indent() << indent() << "return thrift.TExceptionTypeCompiled" << endl;
+    indent_up();
+    out << indent() << "return thrift.TExceptionTypeCompiled" << endl;
+    indent_down();
     out << indent() << "}" << endl << endl;
 
     out << indent() << "var _ thrift.TException = (*" << tstruct_name << ")(nil)"
@@ -1990,6 +1994,7 @@ void t_go_generator::generate_service_client(t_service* tservice) {
     f_types_ << indent() << "*" << extends_client << endl;
   } else {
     f_types_ << indent() << "c thrift.TClient" << endl;
+    f_types_ << indent() << "meta thrift.ResponseMeta" << endl;
   }
 
   indent_down();
@@ -2059,7 +2064,19 @@ void t_go_generator::generate_service_client(t_service* tservice) {
     indent_up();
     f_types_ << indent() << "return p.c" << endl;
     indent_down();
-    f_types_ << indent() << "}" << endl;
+    f_types_ << indent() << "}" << endl << endl;
+
+    f_types_ << indent() << "func (p *" << serviceName << "Client) LastResponseMeta_() thrift.ResponseMeta {" << endl;
+    indent_up();
+    f_types_ << indent() << "return p.meta" << endl;
+    indent_down();
+    f_types_ << indent() << "}" << endl << endl;
+
+    f_types_ << indent() << "func (p *" << serviceName << "Client) SetLastResponseMeta_(meta thrift.ResponseMeta) {" << endl;
+    indent_up();
+    f_types_ << indent() << "p.meta = meta" << endl;
+    indent_down();
+    f_types_ << indent() << "}" << endl << endl;
   }
 
   // Generate client method implementations
@@ -2091,8 +2108,11 @@ void t_go_generator::generate_service_client(t_service* tservice) {
       std::string resultName = tmp("_result");
       std::string resultType = publicize(method + "_result", true);
       f_types_ << indent() << "var " << resultName << " " << resultType << endl;
-      f_types_ << indent() << "if err = p.Client_().Call(ctx, \""
-        << method << "\", &" << argsName << ", &" << resultName << "); err != nil {" << endl;
+      f_types_ << indent() << "var meta thrift.ResponseMeta" << endl;
+      f_types_ << indent() << "meta, err = p.Client_().Call(ctx, \""
+        << method << "\", &" << argsName << ", &" << resultName << ")" << endl;
+      f_types_ << indent() << "p.SetLastResponseMeta_(meta)" << endl;
+      f_types_ << indent() << "if err != nil {" << endl;
 
       indent_up();
       f_types_ << indent() << "return" << endl;
@@ -2131,9 +2151,13 @@ void t_go_generator::generate_service_client(t_service* tservice) {
         f_types_ << indent() << "return nil" << endl;
       }
     } else {
+      // Since we don't have response meta for oneway calls, overwrite it with
+      // an empty one to avoid users getting the meta from last call and
+      // mistaken it as from the oneway call.
+      f_types_ << indent() << "p.SetLastResponseMeta_(thrift.ResponseMeta{})" << endl;
       // TODO: would be nice to not to duplicate the call generation
-      f_types_ << indent() << "if err := p.Client_().Call(ctx, \""
-      << method << "\", &"<< argsName << ", nil); err != nil {" << endl;
+      f_types_ << indent() << "if _, err := p.Client_().Call(ctx, \""
+        << method << "\", &"<< argsName << ", nil); err != nil {" << endl;
 
       indent_up();
       f_types_ << indent() << "return err" << endl;
--- a/lib/go/thrift/header_protocol.go
+++ b/lib/go/thrift/header_protocol.go
@@ -345,20 +345,6 @@ func (p *THeaderProtocol) SetTConfiguration(cfg *TConfiguration) {
 	p.cfg = cfg
 }
 
-// GetResponseHeadersFromClient is a helper function to get the read THeaderMap
-// from the last response received from the given client.
-//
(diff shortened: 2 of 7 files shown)


lib/go/thrift/header_protocol.go
```diff
@@ -345,20 +345,6 @@ func (p *THeaderProtocol) SetTConfiguration(cfg *TConfiguration) {
 	p.cfg = cfg
 }
 
-// GetResponseHeadersFromClient is a helper function to get the read THeaderMap
-// from the last response received from the given client.
-//
-// If the last response was not sent over THeader protocol,
-// a nil map will be returned.
-func GetResponseHeadersFromClient(c TClient) THeaderMap {
```
dcelasun, line 353: This function was in a tagged release. Maybe mark it as deprecated instead of removing?

fishy: It was added after 0.13.0 so I don't believe it's in any tagged release :)

fishy: At least it's not here: https://pkg.go.dev/github.com/apache/thrift@v0.13.0/lib/go/thrift :)

dcelasun: Ah, I was confused because the commit date was earlier than the 0.13 tag :)

dcelasun (approved):

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1493,11 +1493,15 @@ void t_go_generator::generate_go_struct_definition(ostream& out,
...
+    indent_up();
+    out << indent() << "return p.String()" << endl;
+    indent_down();
     out << indent() << "}" << endl << endl;
 
     out << indent() << "func (" << tstruct_name << ") TExceptionType() thrift.TExceptionType {" << endl;
-    out << indent() << indent() << "return thrift.TExceptionTypeCompiled" << endl;
+    indent_up();
+    out << indent() << "return thrift.TExceptionTypeCompiled" << endl;
+    indent_down();
```
fishy, line 1504: These are cosmetic fix of a wrongly did indent from my previous PR :$  not really related to this PR.

fishy:
@dcelasun On second thought, I changed it from `THeaderMap` to a struct `ResponseMeta` containing `THeaderMap`, so it can be expanded in the future to add more stuff and that's no longer a breaking change.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -2091,8 +2108,11 @@ void t_go_generator::generate_service_client(t_service* tservice) {
       std::string resultName = tmp("_result");
       std::string resultType = publicize(method + "_result", true);
       f_types_ << indent() << "var " << resultName << " " << resultType << endl;
-      f_types_ << indent() << "if err = p.Client_().Call(ctx, \""
-        << method << "\", &" << argsName << ", &" << resultName << "); err != nil {" << endl;
+      f_types_ << indent() << "var meta thrift.ResponseMeta" << endl;
+      f_types_ << indent() << "meta, err = p.Client_().Call(ctx, \""
+        << method << "\", &" << argsName << ", &" << resultName << ")" << endl;
+      f_types_ << indent() << "p.SetLastResponseMeta_(meta)" << endl;
+      f_types_ << indent() << "if err != nil {" << endl;
```
yhlee-tw, line 2115: should we use `tmp("_meta")` or something? I'm seeing `meta` redeclaration errors when `meta` is already defined in args, for example:
https://github.com/apache/storm/blob/v2.2.0/storm-client/src/storm.thrift#L752
produces the snippet below which throws `meta` redeclared error
```go
// Parameters:
//  - Key
//  - Meta
func (p *NimbusClient) BeginCreateBlob(ctx context.Context, key string, meta *SettableBlobMeta) (r string, err error) {
  var _args293 NimbusBeginCreateBlobArgs
  _args293.Key = key
  _args293.Meta = meta
  var _result294 NimbusBeginCreateBlobResult
  var meta thrift.ResponseMeta
  meta, err = p.Client_().Call(ctx, "beginCreateBlob", &_args293, &_result294)
  p.SetLastResponseMeta_(meta)
  if err != nil {
    return
  }
  switch {
  case _result294.Aze!= nil:
    return r, _result294.Aze
  case _result294.Kae!= nil:
    return r, _result294.Kae
  }

  return _result294.GetSuccess(), nil
}
```

fishy: Yes. Thanks for reporting. Preparing a PR to fix it.

fishy: https://github.com/apache/thrift/pull/2332

litao3rd:
hello, in the latest version of thrift golang library, when I compile gen-go code with go build, I got error like this
```text
gen-go/package/XXXX.go:1353:11: cannot use temp (type Text) as type string in assignment
gen-go/package/XXXX.go:3028:8: undefined: thrift.ResponseMeta
gen-go/package/XXXX.go:3053:43: undefined: thrift.ResponseMeta
gen-go/package/XXXX.go:3057:49: undefined: thrift.ResponseMeta
gen-go/package/XXXX.go:3069:14: undefined: thrift.ResponseMeta
gen-go/package/XXXX.go:3070:16: assignment mismatch: 2 variables but p.Client_().Call returns 1 values
gen-go/package/XXXX.go:3092:15: undefined: thrift.ResponseMeta
gen-go/package/XXXX.go:3093:17: assignment mismatch: 2 variables but p.Client_().Call returns 1 values
gen-go/package/XXXX.go:3114:15: undefined: thrift.ResponseMeta
gen-go/package/XXXX.go:3115:17: assignment mismatch: 2 variables but p.Client_().Call returns 1 values
gen-go/package/XXXX.go:3115:17: too many errors
```

It seems that when I install the library with command `github.com/apache/thrift/lib/go/thrift/...`, it installed with an old version code but not latest. The gen-codes were generated by thrift-0.14.2 which is latest version.

fishy:
@litao3rd you are not using latest thrift go library, please check your go.mod file to make sure you are using the correct version of the library.

litao3rd:
@fishy  thanks for your reply. I'm sure I was using the latest thrift go library.

I tested under a pure new environment built by these steps.
```
1, download the latest go executable and installed it with the version go1.16.6
2, create a new go project by command `go mod init demo`
3, generate thrift codes by command `thrift -gen go file.thrift`
4, installed the latest go thrift library by command `go get github.com/apache/thrift/lib/go/thrift/...`
5, run command `go build`
```
the generated go.mod file showed as below
```go.mod
module demo

go 1.16

require github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4 // indirect

```

the generated go.sum file showed as below
```go.sum
github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4 h1:orNYqmQGnSjgOauLWjHEp9/qIDT98xv/0Aa4Zet3/Y8=
github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4/go.mod h1:V/LzksIyqd3KZuQ2SunvReTG/UkArhII1dAWY5U1sCE=
```


It's weird that when I install the thrift library it prompts I'm installing v0.14.2 which is the latest version. But in go.mod file there is no version information just v0.0.0.

```log
go: downloading github.com/apache/thrift v0.14.2
go: downloading github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4
go get: added github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4
```

fishy:
That is the wrong version. On step 4, do this instead: `go get github.com/apache/thrift`.

That will give you `v0.14.2` instead of the wrong version.

litao3rd:
Hi, thanks. I did as follow commands.

```shell
1, remove the directory $HOME/go/pkg/mod as I did not set $GOROOT and $GOPATH explicitly.
2, run the command `go clean -cache`
3, go get github.com/apache/thrift
4, go mod init demo2
5, go build
```
On step3, I do get the right version v0.14.2 as I get the message 
```
go: downloading github.com/apache/thrift v0.14.2
```

and the directory
```
$HOME/go/pkg/mod/github.com/apache/thrift@v0.14.2
```

But on step4, I get the prompt
```
go: creating new go.mod: module demo2
go: to add module requirements and sums:
	go mod tidy
```
If I do as it requires to run the command `go mod tidy`, I will get the follow messages.
```
go: finding module for package github.com/apache/thrift/lib/go/thrift
go: found github.com/apache/thrift/lib/go/thrift in github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4
```
Yep, I still get the wrong version of the library.

If I do NOT run the command `go mod tidy`, but run `go build` directly. I will get the follow messages
```
gen-go/${pkg_name}/${filename}-consts.go:10:2: no required module provides package github.com/apache/thrift/lib/go/thrift; to add it:
	go get github.com/apache/thrift/lib/go/thrift
```

That meas I must run the command go get before `go build`. But when I run the `go get` command as it prompts. I get the follow messages
```
go: downloading github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4
go get: added github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4
```

I still get the wrong version. 

I do not know does it any relation with my go executable version as I use the latest version of go 1.16.6

Thanks very much.

fishy:
That's because you also swapped the steps. Don't do that. You need to do `go get` **after** `go mod init`.

Can you go back to the exact steps of https://github.com/apache/thrift/pull/2315#issuecomment-886030562, and **only** change Step 4 into `go get github.com/apache/thrift`?

litao3rd:
You are right. The problem is solved as your instructions. Thanks very much.

litao3rd:
By the way, I installed the go thrift library as the instruction introduced in [README ](https://github.com/apache/thrift/blob/master/lib/go/README.md) file. Maybe it should correct ?

fishy:
Thanks, I'll update the readme file.

PR #2319, opened 2021-01-28
THRIFT-5344 follow-up fix to THRIFT 5237 Implement MAX_MESSAGE_SIZE and consolidate limits ...

Jens-G:
Client: cpp
Patch: Jens Geyer

Thanks to @taras-protsiv who pointed this out.


Jens-G:
@zeshuai007 could you please have a look at this? I'm pretty sure about the (rather trivial) change but it breaks one of the tests. Can we solve that before 0.14.0 please?

emmenlau:
I'll check this out today.

PR #2327, opened 2021-02-12
THRIFT-5187: Added Win32 support for domain sockets (AF_UNIX)

emmenlau:
This PR adds support for domain sockets for Windows.

It is a follow-up of #2127 because the former ran into a timeout and can no be reopened (due to https://github.com/isaacs/github/issues/361).

In order to implement this change I've changed a basic property of C++ build configuration for Windows. This would not be strictly required but is quite helpful. Domain sockets are only available on recent versions of the Windows platform SDK. Therefore I've enabled the automatic configuration of `config.h` with cmake. I've checked the results and on my platform(s) the generated config header is sensible and matches the hard-coded default. However this change is not easily portable to autotools.

During the implementation I found a number of minor issues in TServerSocket and TSocket that I believe are improved in this PR. None of these changes should be a breaking change. Mostly some methods should not be called on a domain socket and have been excluded. Also, the variable `listening_` was set to `true` at the beginning of the `listen()` method but I found that this opens a small time window where `listening_` is true but the socket is not actually listening (yet).

Please review and let me know if there is anything to discuss?

I've tested this implementation successfully on MSVC 2017 and 2019, on Ubuntu 18.04 x86_64 and on MacOS 14 with XCode 11.3.

- [x] I created Apache Jira Ticket https://issues.apache.org/jira/browse/THRIFT-5187

### DIFF
--- a/lib/cpp/src/thrift/transport/SocketCommon.cpp
+++ b/lib/cpp/src/thrift/transport/SocketCommon.cpp
@@ -19,31 +19,16 @@
  * @author: David Suárez <david.sephirot@gmail.com>
  */
 
-#ifndef THRIFT_SOCKETCOMMON_H
-#define THRIFT_SOCKETCOMMON_H
-
-#ifndef _WIN32
-
-#include <thrift/thrift-config.h>
-
-#ifdef HAVE_UNISTD_H
-#include <unistd.h>
-#endif
-
-#ifdef HAVE_SYS_UN_H
-#include <sys/un.h>
-#endif
-
-#include <string>
-
+#include <thrift/transport/SocketCommon.h>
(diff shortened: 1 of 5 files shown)


emmenlau:
Rebased on latest master.

emmenlau:
Would someone be willing to review? I think this PR should be relatively self-contained, so it should not harm the existing functionality. It adds support for Un*x domain sockets on Windows, which can be an interesting use case.

For the review, I recommend reviewing the two commits individually. The first one (adding new method `isUnixDomainSocket()`) is a trivial change. The second commit has the functional improvements but is comparably smaller when reviewed individually.

emmenlau:
Remaining build errors are unrelated. Merging.

PR #2330, opened 2021-02-16
THRIFT-5352: Fix construction of Py exceptions with no fields

spladug:
Client: py

When no fields are present, we don't get the special constructor that
uses `__setattr__` to avoid these checks. So the default constructor sets
message normally and triggers the anti-mutation tripwires.

Before:

```console
$ thrift --version
Thrift version 0.14.0

$ pip freeze | grep thrift
thrift==0.14.0

$ cat test.thrift
exception LookMaNoFields {}

$ thrift -out . -gen py test.thrift

$ python
>>> from test.ttypes import LookMaNoFields
>>> LookMaNoFields()
  File "<stdin>", line 1, in <module>
  File "/src/thrift/lib/py/thrift/Thrift.py", line 93, in __init__
    super(TException, self).__setattr__("message", message)
  File "/src/example/test/ttypes.py", line 23, in __setattr__
```

After:

```
$ python
>>> from test.ttypes import LookMaNoFields
>>> LookMaNoFields()
LookMaNoFields(message=None)
```

### DIFF
--- a/lib/py/src/Thrift.py
+++ b/lib/py/src/Thrift.py
@@ -90,7 +90,7 @@ class TException(Exception):
 
     def __init__(self, message=None):
         Exception.__init__(self, message)
-        self.message = message
+        super(TException, self).__setattr__("message", message)
 
 
 class TApplicationException(TException):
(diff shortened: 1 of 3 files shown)


fishy (approved):

fishy:
The only failure on travis is about java, which seems unrelated to this PR.

PR #2335, opened 2021-02-18
THRIFT-5353: Fix import dedup without explicit go namespace

fishy:
Client: go

When a thrift file includes 2 or more other thrift files, and those
included thrift files do not have explicit go namespaces defined, the
current import dedup logic would wrongly use their empty namespace and
skip the second one, while the real import namespace should be inferred
from the filename.


dcelasun (approved):

dcelasun:
Nice catch! In f4475ff I didn't think about creating a test with an empty namespace :/

PR #2336, opened 2021-02-20
THRIFT-5299: Encode sequence numbers as non-zigzag varint

allengeorge:
Client: rs


allengeorge:
AFAICT, looking at the cross tests this looks like it works as expected.

allengeorge:
Ah, dammit - super sorry :/ Forgot to "squash and merge".

PR #2346, opened 2021-03-10
THRIFT-5367 Ruby library crashes when using GC.compact

Jens-G:
Client: rb
Patch: Stan Hu

### DIFF
--- a/lib/rb/ext/binary_protocol_accelerated.c
+++ b/lib/rb/ext/binary_protocol_accelerated.c
@@ -457,4 +457,5 @@ void Init_binary_protocol_accelerated() {
   rb_define_method(bpa_class, "read_set_end", rb_thrift_binary_proto_read_set_end, 0);
 
   rbuf_ivar_id = rb_intern("@rbuf");
+  rb_global_variable(&rbuf_ivar_id);
 }
--- a/lib/rb/thrift.gemspec
+++ b/lib/rb/thrift.gemspec
@@ -27,7 +27,8 @@ Gem::Specification.new do |s|
 
   s.require_paths = %w[lib ext]
 
-  s.add_development_dependency 'bundler',            '~> 1.11'
+  s.add_development_dependency 'bundler',            '>= 1.0'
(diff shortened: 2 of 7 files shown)


lib/rb/ext/binary_protocol_accelerated.c
```diff
@@ -457,4 +457,5 @@ void Init_binary_protocol_accelerated() {
   rb_define_method(bpa_class, "read_set_end", rb_thrift_binary_proto_read_set_end, 0);
 
   rbuf_ivar_id = rb_intern("@rbuf");
+  rb_global_variable(&rbuf_ivar_id);
```
stanhu, line 460: I suspect this isn't needed because I think `rb_intern` gets marked already.

lib/rb/thrift.gemspec
```diff
@@ -27,7 +27,8 @@ Gem::Specification.new do |s|
 
   s.require_paths = %w[lib ext]
 
-  s.add_development_dependency 'bundler',            '~> 1.11'
+  s.add_development_dependency 'bundler',            '>= 1.0'
```
stanhu, line 30: Ruby 2.7 uses Bundler v2 by default, so maybe we should do this?

```suggestion
  s.add_development_dependency 'bundler',            '>= 1.11'
```

lib/rb/thrift.gemspec
```diff
@@ -27,7 +27,8 @@ Gem::Specification.new do |s|
...
+  s.add_development_dependency 'srv',                '~> 1.0'
```
stanhu, line 31: We might also want to sort this line.

Jens-G:
@stanhu Feel free to improve, it's your patch.

stanhu:
@Jens-G How would you like me to tweak this patch? I don't have write access to this pull request.

Jens-G:
Create a new one?

stanhu:
Ok, I thought this repository was just a mirror: https://github.com/apache/thrift/pull/2347.

PR #2353, opened 2021-03-22
THRIFT-5358: Add go.mod file to root directory

fishy:
Client: go

This unblocks the development under go 1.16+, which starts to complain
when there's no go.mod file in any of the directories.

The current approach is certainly not the best solution ever, for
example it does not run the tests under lib/go/test/tests but copy them
into lib/go/test/gopath/src/sometest and run them there instead, but
those improvements can be done in the future in follow up PRs and this
should be a good enough first step to unblock developments.

  


fishy:
It's ready for review now.

Note that I choose to add a single go.mod file at the root directory instead. As a result doing prefixed tagging (e.g. `lib/go/thrift/v0.15.0`) is no longer needed and it continue to be no extra steps for go releases.

dcelasun (approved):
LGTM. Looks so much cleaner.

lib/go/test/fuzz/fuzz.go
```diff
@@ -24,10 +24,10 @@ package fuzz
 import (
 	"context"
 	"fmt"
-	"shared"
 	"strconv"
-	"tutorial"
 
+	"github.com/apache/thrift/lib/go/test/fuzz/gen-go/shared"
+	"github.com/apache/thrift/lib/go/test/fuzz/gen-go/tutorial"
```
fishy, line 30: cc @catenacyber : I changed the import paths of these 2 so that we do not need to do `go mod edit --replace` inside thrift's CI. I believe in oss-fuzz you can still do `go mod edit --replace` to replace these 2 to something local, but please give that a try and let me know if that doesn't work.

catenacyber: Thanks @fishy 
Ok, let me know when you merge this go module, and I will go back to adapt oss-fuzz PR

fishy: can you try it from my branch instead? I'd rather not to merge it if it doesn't work :)

catenacyber: Well, I guess this branch is about supporting go modules for all of thrift.
When this will me merged, I do not doubt I can manage to adapt the oss-fuzz PR

If you want, I can test it with this branch, but that may take a few days

fishy: OK that makes sense. I'll merge this soon and if it turns out to break oss-fuzz, we can make follow up changes to fix them.

catenacyber: You were right, I do not get it to work with this change
As the package is named `shared` elsewhere and not `github.com/apache/thrift/lib/go/test/fuzz/gen-go/shared`
Could you revert this ?
Sorry for missing it in the first place

fishy: @catenacyber https://github.com/apache/thrift/pull/2358

PR #2356, opened 2021-03-25
THRIFT-5378: Amend Go String() generation for enums

maguro:


Jens-G:
Could you force push again? CI seems not related but its largely red, so let us just try again.

maguro:
> Could you force push again? CI seems not related but its largely red, so let us just try again.

Still have errors that don't seem to have anything to do w/ the PR.  Is there a ticket to fix the CI checks?  I can try to help.

fishy:
Copying the comment I left on the JIRA ticket here:

>The PR you made generated essentially the same go code, so you must have some other issue caused the problem. Can you provide a minimal example with detailed steps (for example, which version of which debugger are you using and on which platform) to reproduce the issue?

maguro:
Unable to reproduce original problem.

PR #2358, opened 2021-03-25
Fix lib/go/test/fuzz for oss-fuzz project

fishy:


fishy:
@catenacyber I'm also wondering why do we really need the code currently in `lib/go/test/fuzz` to live inside thrift project. I mean, those code can totally live inside oss-fuzz project instead, right?

catenacyber:
That is a good question
Here is the oss-fuzz answer https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#fuzz-target

TL;DR this way, if you somehow change the API of thrift, your CI will make you think about changing the fuzz target to use the new API

catenacyber:
Tested this with https://github.com/google/oss-fuzz/pull/5264 and it works

dcelasun (approved):

PR #2359, opened 2021-03-26
THRIFT-5369: Use MaxMessageSize to check container sizes

fishy:
Client: go

~Also remove size arg from make() calls in compiler.~

  

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 - [THRIFT-5347](https://issues.apache.org/jira/browse/THRIFT-5347) - Haskell support dropped
 - [THRIFT-5381](https://issues.apache.org/jira/browse/THRIFT-5381) - possible collisions at VOID type with some 3rd-party libraries on Haxe cpp targets
 
+### Go
+
+- [THRIFT-5369](https://issues.apache.org/jira/browse/THRIFT-5369) - No longer pre-allocating the whole container (map/set/list) in compiled go code to avoid huge allocations on malformed messages
+
 
 ## 0.14.1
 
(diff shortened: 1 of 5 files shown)


dcelasun:
Seems fine from a quick a look, but it'd be great if someone from jaeger can verify it as well.

fishy:
@Jens-G I updated this PR to remove MaxContainerSize from TConfiguration and use MaxMessageSize to check for container sizes.

fishy:
@jpkrohling can you try this version to see if you can still reproduce the bug? (note that you need to set a smaller, sane MaxMessageSize to TConfiguration to actual limit it. the default limit is 100M)

CHANGES.md
```diff
@@ -1,5 +1,17 @@
...
+
+### Deprecated Languages
+
+### Removed Languages
+
+### Breaking Changes
+
+### Go
+
+- [THRIFT-5369](https://issues.apache.org/jira/browse/THRIFT-5369) - No longer pre-allocation the whole container (map/set/list) in compiled go code to avoid huge allocations on malformed messages.
```
dcelasun, line 13: nit
```suggestion
- [THRIFT-5369](https://issues.apache.org/jira/browse/THRIFT-5369) - No longer pre-allocating the whole container (map/set/list) in compiled go code to avoid huge allocations on malformed messages.
```

jpkrohling:
@fishy sorry for being not very responsive here. I'm on PTO until next Tuesday, so I probably can't test this before then but I am interested in checking whether this solves our problem!

jpkrohling:
I still wasn't able to test this, but a Jaeger community member (@forestsword) was able to test it and confirmed that this indeed fixes the memory leak.

fishy:
As mentioned in https://github.com/jaegertracing/jaeger/issues/2638#issuecomment-815859394, I reverted the compiler changes from this PR as they have negative performance impacts for everyone and it doesn't seem they are really necessary to fix the memory issue. We can have a follow up PR to add them back if they are proven to be necessary.

PR #2360, opened 2021-03-26
THRIFT-5236 THttpTransport.cs still has bad timeout code

thomasbruggink:
Client: netstd

Allow connect timeout to be configurable.

### DIFF
--- a/lib/netstd/Tests/Thrift.Tests/Transports/THttpTransportTests.cs
+++ b/lib/netstd/Tests/Thrift.Tests/Transports/THttpTransportTests.cs
@@ -1,0 +1,40 @@
+// Licensed to the Apache Software Foundation(ASF) under one
+// or more contributor license agreements.See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.The ASF licenses this file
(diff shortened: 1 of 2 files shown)


Jens-G:
Still? Wonder why it has not been done "properly" at the first time then.

thomasbruggink:
It seems the implementation was lost when it was migrated from csharp, netcore to netstd. Since there is now only one implementation hopefully it will stay here.

Jens-G:
Since https://dotnet.microsoft.com/download/dotnet does no longer list 2.0 as supported version I think it is safe to remove all netcore 2.0 specifics.

thomasbruggink:
> Since https://dotnet.microsoft.com/download/dotnet does no longer list 2.0 as supported version I think it is safe to remove all netcore 2.0 specifics.

My original comment was written against an older master that didn't have net5.0 targets yet. Sorry for the confusion :bowing_man:

Jens-G:
No confusion, 2.0 is still in the list of supported platforms. And there are several special cases to deal with it. I really meant it. Maybe a different PR, that's fine.

Jens-G:
Thank you!

PR #2361, opened 2021-03-26
Fix netstd test lib build on non windows hosts

thomasbruggink:
Some general fixes for netstd test build.
* Non windows builds did not search for thrift binary on path
* Fail if a thrift binary cannot be found instead of silently continuing
* Simplified thrift generation
* Check if thrift binary used is the same version as the local projects version
  
Tested on Linux and Windows.


Jens-G:
Modified the version check to properly deal with custom build compiler versions, merged. Thanks!

PR #2369, opened 2021-04-08
Abnormal Branch Memory Leak

unfetteredwind:

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_c_glib_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_c_glib_generator.cc
@@ -4061,8 +4061,17 @@ void t_c_glib_generator::generate_deserialize_field(ostream& out,
       throw "compiler error: no C reader for base type " + t_base_type::t_base_name(tbase) + name;
     }
     out << ", error)) < 0)" << endl;
-    out << indent() << "  return " << error_ret << ";" << endl << indent() << "xfer += ret;"
-        << endl;
+    if ((tbase != t_base_type::TYPE_STRING) && allocate) {
+      scope_up(out);
+      out << indent() << "g_free (&" << name << ");" << endl;
+      out << indent() << "return " << error_ret << ";" << endl;
+      scope_down(out);
+      out << indent() << "xfer += ret;" << endl;
+    }
+    else{
+      out << indent() << "  return " << error_ret << ";" << endl << indent() << "xfer += ret;"
+         << endl;
+    }
 
     // load the byte array with the data
     if (tbase == t_base_type::TYPE_STRING && type->is_binary()) {


unfetteredwind:
image

compiler/cpp/src/thrift/generate/t_c_glib_generator.cc
```diff
@@ -4061,8 +4061,17 @@ void t_c_glib_generator::generate_deserialize_field(ostream& out,
       throw "compiler error: no C reader for base type " + t_base_type::t_base_name(tbase) + name;
     }
     out << ", error)) < 0)" << endl;
-    out << indent() << "  return " << error_ret << ";" << endl << indent() << "xfer += ret;"
-        << endl;
+    if ((tbase != t_base_type::TYPE_STRING) && allocate) {
```
Jens-G, line 4064: Why is that test different to line 4022 (allocate is not tested there)? Shouldn't that be consistent?

unfetteredwind: 4022 is :
image
If it's not equal t_base_type::TYPE_STRING , I've increased the processing。The two are different,and I think it's an omission.

Jens-G:
Could you provide a test case that shows the issue?

ulidtko (changes requested):
`g_new` returns owned memory (malloc'd pointer) that must be `g_free`'d on every path. It's not `g_free`'d in the error-path with `return -1` — thus creating a memory leak [in the codegen output for `c_glib`]. That's indeed an issue.

I think @zeshuai007 can review `glib` stuff better than me.

I don't think the check for `t_base_type::TYPE_STRING` belongs there as proposed. The fix doesn't look correct to me.

@unfetteredwind could you please explain the `if` condition for the additional processing you added? In your original screenshot, an integer `gint32` leaks from `g_new()`; how are *strings* related? Creating a little test as @Jens-G says would also help, both to illustrate the problem and to verify its solution.

Jens-G:
> That's indeed an issue.

I was not able to trigger the case that would invoke the new code with my test case. So even if it is a problem, the solution seems not to work at all. Essentially I got the same code generated before and after. That's why I asked for a test case in the first place.

Jens-G:
Is this related to THRIFT-5399 maybe?

ulidtko:
> Is this related to THRIFT-5399 maybe? A bit more information could really help to move on with this.

No, that's different AFAICS. There, a socket leaks on `listen` failure; here, a memory allocation leaks on deserialization failure.

I suppose, repro of this issue involves running a `c_glib` client on intentionally malformed buffers & under a leak detector (asan/valgrind). Then, the leak would trigger whenever we expect to read an i32 but get something else instead.

unfetteredwind:
OK, first of all, let me briefly explain why this judgment condition is written.

There are two writing conditions:
1、The value of allocate is true.
2、The type is not TYPE_STRING type in the base type.

1、In this document, the function generate_deserialize_field is invoked in the following four locations:
struct:generate_struct_reader                      allocate：default value: false    3579  3664
map: generate_deserialize_map_element   allocate：default value: true      4271  4294
set: generate_deserialize_set_element        allocate：default value: true      4305  4316
list: generate_deserialize_list_element        allocate：default value: true      4325  4338

Check the function invoked by each. If the value of allocate is false, new is not triggered(declare_local_variable is not invoked in the struct ). In this case, free is not required.
When the value of allocate is true, the declare_local_variable function is invoked in the map, set, and list to trigger new. In this case, you need to perform the free operation when an exception occurs.
So that's one of my judgements.

2、And going back to this function : generate_deserialize_list_element.
The judgment condition is in the base_type branch. 4020
As you can see, if the type is TYPE_STRING, there is no leakage problem in this part,。Because he did special treatment., 4023-4028.
For the remaining types, if they enter the exception branch after new, memory leakage will be triggered.
That's another criterion for me. If the condition is met, it needs to be processed.

That's why I chose the condition. And, perhaps the judgment is not rigorous enough, and my intention is to raise this issue so that people can see this issue and push this issue to be resolved.


As for the test..., it's working on it.And....I'm lazy. Excuse me.

PR #2378, opened 2021-04-18
THRIFT-5400: Add the .annotation package to Export-Package

bluezio:
This PR adds the .annotation package to the Export-Package clause of the MANIFEST.MF file in the Java Thrift library.


Jens-G:
> This PR adds the .annotation package to the Export-Package clause of the MANIFEST.MF file in the Java Thrift library.

This comment comments.

Honestly, this is the obvious part. More interesting would be, WHY is the change necessary or helpful and/or what problem does it solve? Could you add that information?

bluezio:
This change is needed to allow Eclipse to consume the Java library for Thrift as an OSGi bundle. Without that directive, Eclipse will give an error on the Java code generated by the Thrift compiler, saying that the `@Nullable` annotation is not API (as the package which contains it is not exported). This also fails Eclipse Tycho builds of Thrift-generated code (Tycho allows for building Eclipse plugins from a CI environment using Maven).

In other words, the OSGi metadata for the Java Thrift library was not updated when the .annotation package and its `@Nullable` was added in a prior release, and Eclipse refuses to compile code using the Java Thrift library because of this.

bluezio:
This information was included in the original JIRA issue:

````
We use the Java Thrift library as an OSGi bundle, taking advantage of the existing Export-Package and Import-Package annotations.

After moving to the Thrift 0.13.0 library, we tried using the Thrift 0.13.0 compiler to produce up-to-date Java code, and our builds are now failing with this error message:

Access restriction: The type 'Nullable' is not API (restriction on required library '/home/antonio/workspaces/eclipse-hawk/.metadata/.plugins/org.eclipse.pde.core/.bundle_pool/plugins/org.apache.thrift_0.13.0.jar')

This appears to be due to the fact that the Java library does not list the .annotation package that contains the Nullable annotation in the Export-Package line of the JAR manifest, but still uses it all over the place in the generated code.
```

PR #2386, opened 2021-05-06
THRIFT-3508: Map optional fields from thrift idl to optional fields i…

axeljaeger-arculus:
…n TypeScript.

Client: node,js

Add ? behind optional members in classes if the field is marked as optional in thrift-idl.

Closes THRIFT-3508 and makes THRIFT-3227 probably obsolete as optional class members are supported since TypeScript 2.0

No, I am hijacking the existing Ticket of THRIFT-3508 as it is exactly what I implemented.

Yes

Yes

Yes, I am not aware of braking changes, but was unable to to do the Test drive development approach from the provided documentation.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_js_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_js_generator.cc
@@ -909,13 +909,9 @@ void t_js_generator::generate_js_struct_definition(ostream& out,
       out << indent() << dval << ";" << endl;
     }
     if (gen_ts_) {
-      if (gen_node_) {
-        f_types_ts_ << ts_indent() << "public " << (*m_iter)->get_name() << ": "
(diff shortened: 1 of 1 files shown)


emmenlau (approved):
Looks good to me.

axeljaeger-arculus:
So what are the next steps? Looks like CI is failing but not due to my code.

Jens-G:
> makes THRIFT-3227 probably obsolete 

What do we do with that one?

axeljaeger-arculus:
After reading THRIFT-3227 again, I see that there might be actual usecases for the proposed split between classes and interfaces, although the original motivation is invalid now. So my suggestion is:
- Add a comment there. I see this issue was already added as reference.
- Leave it open. Maybe someone wants to work on it.

PR #2390, opened 2021-05-11
c++: fix MaxMessageSize reached

Irelian:
c++: TMemoryBuffer: consume remainin MessageSize but never reset，we should reset remainin MessageSize after readEnd.

 RemainingMessageSize_ reduce in TBufferBase::consume, but nerver reset remainingMessageSize_ to MaxMessageSize.
Which result in  TTransportException:"MaxMessageSize reached" after to many connection between client and server.
After TMemoryBuffer::readEnd(), we should reset remainingMessageSize_ by calling resetConsumedMessageSize().

### DIFF
--- a/lib/cpp/src/thrift/transport/TBufferTransports.h
+++ b/lib/cpp/src/thrift/transport/TBufferTransports.h
@@ -658,6 +658,7 @@ public:
     if (rBase_ == wBase_) {
       resetBuffer();
     }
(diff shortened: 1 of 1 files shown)


Jens-G:
All transports do this in flush(). Why do we need to change that for this particular transport? What makes it so special? Could you provide a test case for the underlying issue you are trying to fix?

PR #2393, opened 2021-05-17
THRIFT-5416: Allow UDP Sockets

bforbis:
Allows perl Socket.pm constructor to take an optional "proto" parameter which would let user's specify the socket to be of type "UDP".

### DIFF
--- a/lib/perl/lib/Thrift/Socket.pm
+++ b/lib/perl/lib/Thrift/Socket.pm
@@ -57,7 +57,7 @@ sub new
         port         => 9090,
         recvTimeout  => 10000,
         sendTimeout  => 10000,
-
+        proto        => 'tcp',
         handle       => undef
     };
 
@@ -260,7 +260,7 @@ sub __open
     my $self = shift;
     return IO::Socket::INET->new(PeerAddr => $self->{host},
                                  PeerPort => $self->{port},
-                                 Proto    => 'tcp',
+                                 Proto    => $self->{proto},
                                  Timeout  => $self->{sendTimeout} / 1000);
 }
 


emmenlau:
The change itself seems sufficiently clear to be acceptable. But I worry about the implications. AFAIK Thrift is not UDP-aware and has no mechanism to re-transmit or cope with lost packages. Is this in the interest of the user? Are more actions required to make this a useful change?

bforbis:
If a user is not tolerant to data loss, they should not use UDP. This is an option to increase the efficiency of data transfer and avoid blocking the main application. A specific example of a use case for this is instrumenting a perl application for sending spans to [Jaeger agent](https://www.jaegertracing.io/docs/1.22/deployment/#agent), which only accepts thrift void-style messages over UDP.

emmenlau:
> If a user is not tolerant to data loss, they should not use UDP. This is an option to increase the efficiency of data transfer and avoid blocking the main application. A specific example of a use case for this is instrumenting a perl application for sending spans to [Jaeger agent](https://www.jaegertracing.io/docs/1.22/deployment/#agent), which only accepts thrift void-style messages over UDP.

I can very much relate to your use case. But I still worry that from the point of view of an unsuspecting Thrift user, its not easily understandable that Thrift will allow UDP and then silently drop lost messages. Furthermore, for non-void-messages its very likely that this will mess with protocols and servers. It seems to me this hands the user a running chainsaw without any guidance. But I may not see all the details here, since AFAICS the change is "only" in the client? Maybe its less harmful there because users are not able to create a corresponding UDP server(?). Opinions?

bforbis:
Yes, this would not work for anyone who is using non-void style RPC calls, however considering that UDP is not the default protocol, I find it unlikely this will actually cause any issues for users. Using UDP with thrift should be considered an advanced option and someone who intentionally sets up their client to use UDP should be responsible for understanding the tradeoffs between UDP and TCP and know what they are getting into.

As for setting up a UDP server, I specifically don't have a use case for it in perl so did not implement any server side code. In my example, Jaeger Agent is already setup to be a thrift UDP server (presumably that means there is already thrift UDP server support in golang, though I haven't looked closely at the Jaeger codebase to see if they wrote their own custom thrift code).

Perhaps I could just leave some documentation in the source file about the use cases for tcp vs udp.

jeking3 (approved):
Performance logging the Jaeger over UDP should be fine.  That use case is good for UDP.  There's nothing to test it against, which kind of stinks, but it is a very simple change.

PR #2395, opened 2021-05-25
THRIFT-5419: Update TThreadPoolAsyncServer.Execute to be async/await

phxnsharp:
Client: netstd

Do not put long-lived tasks on the thread pool, which causes deadlock and extreme slow performance.

Changes Execute() to be ExecuteAsync() using async/await. It still runs on the thread pool, but now freely among the threads without blocking.

I attempted to follow https://github.com/phxnsharp/thrift/blob/master/build/docker/README.md to build using docker on Linux but ran into errors. My changes do not add additional errors.

I did not see where the test results get put, so I do not know how to verify the tests.

@Jens-G

### DIFF
--- a/lib/netstd/Thrift/Server/TThreadPoolAsyncServer.cs
+++ b/lib/netstd/Thrift/Server/TThreadPoolAsyncServer.cs
@@ -177,7 +177,7 @@ namespace Thrift.Server
                     try
                     {
                         TTransport client = await ServerTransport.AcceptAsync(cancellationToken);
-                        ThreadPool.QueueUserWorkItem(this.Execute, client);
+                        var ignore = Task.Run(() => ExecuteAsync(client));
                     }
                     catch (TaskCanceledException)
                     {
@@ -218,11 +218,11 @@ namespace Thrift.Server
         /// threadContext will be a TTransport instance
         /// </summary>
         /// <param name="threadContext"></param>
-        private void Execute(object threadContext)
+        private async void ExecuteAsync(TTransport client)
         {
             var cancellationToken = ServerCancellationToken;
 
-            using (TTransport client = (TTransport)threadContext)
+            using (client)
             {
                 ITAsyncProcessor processor = ProcessorFactory.GetAsyncProcessor(client, this);
                 TTransport inputTransport = null;
@@ -241,12 +241,12 @@ namespace Thrift.Server
 
                         //Recover event handler (if any) and fire createContext server event when a client connects
                         if (ServerEventHandler != null)
-                            connectionContext = ServerEventHandler.CreateContextAsync(inputProtocol, outputProtocol, cancellationToken).Result;
+                            connectionContext = await ServerEventHandler.CreateContextAsync(inputProtocol, outputProtocol, cancellationToken);
 
                         //Process client requests until client disconnects
                         while (!stop)
                         {
-                            if (! inputTransport.PeekAsync(cancellationToken).Result)
+                            if (! await inputTransport.PeekAsync(cancellationToken))
                                 break;
 
                             //Fire processContext server event
@@ -254,9 +254,9 @@ namespace Thrift.Server
                             //That is to say it may be many minutes between the event firing and the client request
                             //actually arriving or the client may hang up without ever makeing a request.
                             if (ServerEventHandler != null)
-                                ServerEventHandler.ProcessContextAsync(connectionContext, inputTransport, cancellationToken).Wait();
+                                await ServerEventHandler.ProcessContextAsync(connectionContext, inputTransport, cancellationToken);
                             //Process client request (blocks until transport is readable)
-                            if (!processor.ProcessAsync(inputProtocol, outputProtocol, cancellationToken).Result)
+                            if (! await processor.ProcessAsync(inputProtocol, outputProtocol, cancellationToken))
                                 break;
                         }
                     }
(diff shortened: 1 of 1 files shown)


Jens-G:
> Do not put long-lived tasks on the thread pool, which causes deadlock and extreme slow performance.

Right, but shouldn't that be in the responsibility of the implementor?

Jens-G:
I'm not fully convinced. The patch is good, that's not the point. 

What I am missing is a test case that clearly shows the issue you are reporting. I just added the threadpool server impl to the Test Suite (PR currently building) and ran both the old code and the patched code with a varying set of parameters and clients but I still struggle to reproduce the "dramatic poor performance". Could you elaborate more on that? How can I test that in a possibly simple way?

Also, I slightly changed the proposed patch, please find [my version here](https://github.com/Jens-G/thrift/tree/THRIFT-5419).

lib/netstd/Thrift/Server/TThreadPoolAsyncServer.cs
```diff
@@ -218,11 +218,11 @@ public override async Task ServeAsync(CancellationToken cancellationToken)
         /// threadContext will be a TTransport instance
         /// </summary>
         /// <param name="threadContext"></param>
-        private void Execute(object threadContext)
+        private async void ExecuteAsync(TTransport client)
         {
```
Jens-G, line 222: `async void` is usually a bad combination that should be avoided (pun intended)

phxnsharp: We use it only in specific circumstances like this one where it matches with the reality of what you are creating. In this case we are effectively starting a new "thread" (though it isn't an actual thick thread) and nothing will be listening to the return value. Having the signature makes it obvious to everyone working on the code that nothing should be returned and no exceptions should be let to exit the function. 

However, there is also value in writing it the way you have in your version. It lets the caller decide if they want to know when the task/thread is done and could be helpful in unit testing. I fully support the change as this code should match your style.

phxnsharp:
> 
> 
> > Do not put long-lived tasks on the thread pool, which causes deadlock and extreme slow performance.
> 
> Right, but shouldn't that be in the responsibility of the implementor?

I'm not sure I follow this question. The implementor of what? If you are putting long lived tasks on the thread pool without using the [TaskCreationOptions.LongRunning](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcreationoptions?view=net-5.0) option, you are at risk of deadlocking (at least for 1 second) everything in the system, even completely unrelated code.

phxnsharp:
> What I am missing is a test case that clearly shows the issue you are reporting.

Note that I am sensitive to this issue due to a [previous experience](https://github.com/jaegertracing/jaeger-client-csharp/issues/181) I had with a similar problem in an unrelated tool. The problem is that since the thread pool is a shared resource with complex logic that depends on the hardware being run on, the expression of this type of bug is often very intermittent, and usually far away from the actual bug. It took me forever to debug that problem. Since then I am always on the lookout for improper long lived tasks on the thread pool.

I found this bug in some test cases that we have of our own code. In this case I am standing up mocks of the in and out of process sides of a Thrift communication channel all inside the test and testing the channel itself. It was intermittently failing and when I investigated I found these long lived threads sitting in the thread pool. 

Note that in this case, my tests are not useful to you because they are testing our code a layer up. While I was debugging, I also found that NamedPipeClientStream.ConnectAsync is also implemented incorrectly (I have not submitted that issue to Microsoft yet). Putting a workaround in for that one (I use the synchronous Connect() call on a TaskCreationOption.LongLived task) actually mostly (but not totally) fixes this specific unit test problem for us. Note that there is no way to put a workaround for TThreadPoolAsyncServer in (short of creating a cut-n-paste clone of it) because it is a private internal method.

I openly admit that I have no proof that these long lived tasks are causing active problems. I still submit that if they don't now, they will someday in a subtle and difficult to debug way.

The way I would expect to be able to reproduce this would be to start a TThreadPoolAsyncServer and then start many connections to it without closing the previous connections. My expectations would be that the first few would be fast, but when you exhausted the thread pool, the next connection would take 1 second to connect while the thread pool deadlocks.

Note that the exact behavior will likely depend on whether you are using .NET Framework or .NET Core, the specific version you are using, and the hardware you are running on. It is entirely possible that Microsoft has or will change the nature of how the thread pool deals with this situation. All the more reason in my mind to watch out for this situation.

Sorry for the diatribe! Thank you for the thoughtful and timely dialog, as well as for the great project! We use Thrift for a lot of things these days and it has been a valuable tool to us!

lib/netstd/Thrift/Server/TThreadPoolAsyncServer.cs
```diff
@@ -177,7 +177,7 @@ public override async Task ServeAsync(CancellationToken cancellationToken)
                     try
                     {
                         TTransport client = await ServerTransport.AcceptAsync(cancellationToken);
-                        ThreadPool.QueueUserWorkItem(this.Execute, client);
+                        var ignore = Task.Run(() => ExecuteAsync(client));
```
phxnsharp, line 180: Note, I had meant to put in a comment about this but see I forgot. Task.Run() is actually not entirely required here. As ExecuteAsync is async, but we ignore the return value, it would be almost equivalent to just call it without the Task.Run(). The reason I put it here is that this code is executed in the SynchronizationContext of the caller. If that is a restricted UI thread, that would put unneeded burden on ExecuteAsync. Task.Run frees the execution to run on any thread pool thread. 

An alternate solution would be to put .ConfigureAwait(false) on every async call per [Microsoft's recommendation](https://devblogs.microsoft.com/dotnet/configureawait-faq/). Arguably, looking at it now, I probably should have done that anyhow out of best practice. 

As an aside, I considered putting those into the rest of this class but because this class has callbacks into the client code (PreServeAsync), that might introduce subtle bugs in client code as those would now be called on the Thread Pool. You may want to consider that in the long run, though, as it frees your code from being tied to an unknown SynchronizationContext.

phxnsharp:
Thank you!

PR #2400, opened 2021-05-28
THRIFT-5425 Throw an exception when reading TSimpleJson in Java

thomasbruggink:
Client java

Throw an exception when reading TSimpleJson and update the comment
explain why.

I'm wondering if just throwing is acceptable since this will break backwards compatibility with clients that actually try to read. Let me know if just updating the comment would be a better solution.


Jens-G:
+1 Looks reasonable to me, I'm for it.

PR #2401, opened 2021-05-31
THRIFT-3840: C++ TJSONProtocol still using locale dependent formatting

aaronmjones:
Add unit test for system locale with thousands separator comma
Install en_US.UTF-8 locale in Dockerfile

The global locale affects integer formatting. Some LC_NUMERIC locales specify a thousands separator "," which would break integer formatting. Add a unit test that sets the global locale to "en_US.UTF-8" (which uses a comma thousands separator). Fix the integer formatting by imbue'ing the "C" locale in ostringstream.

### DIFF
--- a/build/docker/ubuntu-bionic/Dockerfile
+++ b/build/docker/ubuntu-bionic/Dockerfile
@@ -253,6 +253,13 @@ RUN cd / && \
     rm swift-5.1.4-RELEASE-ubuntu18.04.tar.gz && \
     swift --version
 
+# Locale(s) for cpp unit tests
+RUN apt-get install -y --no-install-recommends \
+`# Locale dependencies` \
+      locales && \
+    locale-gen en_US.UTF-8 && \
+    update-locale
+
 # cppcheck-1.82 has a nasty cpp parser bug, so we're using something newer
 RUN apt-get install -y --no-install-recommends \
 `# Static Code Analysis dependencies` \
--- a/build/docker/ubuntu-bionic/Dockerfile
+++ b/build/docker/ubuntu-bionic/Dockerfile
@@ -258,6 +258,7 @@ RUN apt-get install -y --no-install-recommends \
 `# Locale dependencies` \
       locales && \
     locale-gen en_US.UTF-8 && \
+    locale-gen de_DE.UTF-8 && \
     update-locale
 
 # cppcheck-1.82 has a nasty cpp parser bug, so we're using something newer
--- a/build/docker/ubuntu-disco/Dockerfile
+++ b/build/docker/ubuntu-disco/Dockerfile
@@ -255,6 +255,13 @@ ENV PATH /root/.cargo/bin:$PATH
 #     rm swift-4.2.1-RELEASE-ubuntu18.04.tar.gz && \
 #     swift --version
 
+# Locale(s) for cpp unit tests
+RUN apt-get install -y --no-install-recommends \
+`# Locale dependencies` \
+      locales && \
+    locale-gen en_US.UTF-8 && \
+    update-locale
+
 # cppcheck-1.82 has a nasty cpp parser bug, so we're using something newer
 # don't need this on disco, nobody uses it
 # RUN apt-get install -y --no-install-recommends \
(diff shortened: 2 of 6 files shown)


emmenlau (changes requested):
Would it be possible to add also a small test directly to `lib/cpp/test/ToStringTest.cpp` to see more transparently what case will fail? Is this only about JSON or are all number converters affected (I assume the latter is true)?

emmenlau:
Other than the additional test for `to_string()`, this looks good and valuable to me.

aaronmjones:
> Would it be possible to add also a small test directly to `lib/cpp/test/ToStringTest.cpp` to see more transparently what case will fail? Is this only about JSON or are all number converters affected (I assume the latter is true)?

Yeah, ToStringTest.cpp looks like a better place for this than JSONProtoTest.cpp. The problem is about number formatting and not JSON in particular. I'll add unit tests for integer and float string formatting to ToStringTest.cpp and remove the changes to JSONProtocolTest. The integer string formatting test will set the global locale to "en_US.UTF-8" which uses comma for a thousands separator. The float string formatting test will set the global locale to "de_DE.UTF-8" which uses a decimal comma -- not a decimal point.

Before changes to TToString.h, the tests would fail like:

```
# ./UnitTests 
Running 57 test cases...
/thrift/src/lib/cpp/test/ToStringTest.cpp(52): error: in "ToStringTest/locale_en_US_int_to_string": check to_string(1000000) == "1000000" has failed [1,000,000 != 1000000]
/thrift/src/lib/cpp/test/ToStringTest.cpp(62): error: in "ToStringTest/locale_de_DE_float_to_string": check to_string(1.2) == "1.2" has failed [1,2 != 1.2]
```

I'll have the changes ready soon.

emmenlau (approved):
Very nice work.

emmenlau:
Build errors are all unrelated, merging.

emmenlau:
Dear @aaronmjones,
our CI was silently not running the tests previously, so your Windows test was never executed. In a newer PR I have fixed that, but it seems that this breaks your test. Could you kindly take a look at https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/40271635/job/3dser86k51a8eco2?fullLog=true where it says
```
Test command: C:\projects\build\MSVC2015\x86\bin\Debug\UnitTests.exe
Test timeout computed to be: 300
Running 57 test cases...
unknown location(0): fatal error: in "ToStringTest/locale_en_US_int_to_string": f:\dd\vctools\crt\crtw32\stdcpp\xmbtowc.c(89) : Assertion failed: ploc->_Mbcurmax == 1 || ploc->_Mbcurmax == 2

C:\projects\thrift\lib\cpp\test\ToStringTest.cpp(44): last checkpoint: "locale_en_US_int_to_string" entry.
unknown location(0): fatal error: in "ToStringTest/locale_de_DE_floating_point_to_string": f:\dd\vctools\crt\crtw32\stdcpp\xmbtowc.c(89) : Assertion failed: ploc->_Mbcurmax == 1 || ploc->_Mbcurmax == 2

C:\projects\thrift\lib\cpp\test\ToStringTest.cpp(53): last checkpoint: "locale_de_DE_floating_point_to_string" entry.
An error message from getaddrinfo on the console is expected:
Thrift: Thu Aug  5 19:53:55 2021 getaddrinfo() -> 11001; No such host is known.


*** 2 failures are detected in the test module "thrift"
Test #381: UnitTests ........................***Failed    3.59 sec
```

aaronmjones:
That assertion happens in the std::locale() constructor and was also reported at https://developercommunity.visualstudio.com/t/stdlocale-l2-en-usutf-8-broken-crashes-now/330322. That link says it was fixed in VS2019 16.9.

Maybe there's a way to avoid the assertion with the version of VS the Windows CI is using (is it VS2015?). But reading the documentation of std::locale(), it can throw:

>std::runtime_error if the operating system has no locale named std_name or if std_name is a null pointer.

I modified the ubuntu Dockerfiles to install the locales used by the unit tests ("en-US.UTF-8" and de-DE.UTF-8"). I don't know how to do the same on Windows in order to guarantee std::locale() does not throw.

How do you feel about enabling those 2 unit tests only on Linux platforms?

emmenlau:
Yes, maybe you are right and this is the way to go.

PR #2407, opened 2021-06-11
fix reconnect issue for nodejs

wujjpp:
CHANGES: 

1.  remove [/lib/nodejs/lib/thrift/connection.js#L233-L236](https://github.com/apache/thrift/blob/master/lib/nodejs/lib/thrift/connection.js#L233-L236), it cause thrift cannot do reconnect
2. add `forceClose` variable for indicating `close` action caseued by manual, and give up reconnecting


Jens-G:
Anybody who could doublecheck this patch?

wujjpp:
> Anybody who could doublecheck this patch?

We have used this patch in production

lib/nodejs/lib/thrift/connection.js
```diff
@@ -198,6 +200,16 @@ Connection.prototype.connection_gone = function () {
   var self = this;
   this.connected = false;
 
+  // If closed by manual, emit close event and cancel reconnect process
+  if(this.forceClose) {
+    self.emit("close");
```
emmenlau, line 205: This may be overly pedantic, but would it be better to *first* clear the timer and *then* emit close? In other words, move the line `self.emit("close");` down below the `if (this.retry_timer) {` block?

emmenlau (commented):
Generally looks good, but I'm no expert in this. Also, the same change may be needed in other Javascript-implementations?

lib/nodejs/lib/thrift/connection.js
wujjpp: > 

Correct, `close callback` maybe cause pedantic.
I have updated this change

emmenlau:
Generally looks good to me. Could you kindly rebase on latest master and push again?

wujjpp:
@emmenlau rebased on latest master and pushed

PR #2408, opened 2021-06-11
THRIFT-5369: Use MaxMessageSize to check container sizes

fishy:
As described in the maillist, this PR is to cherry pick 6583f4e52 into 0.14.2 release branch.


Jens-G:
Every single time. If I had placed a bet I had won it without a blink.

Jens-G:
> As described in the maillist
Which one? Can't find it.

PR #2412, opened 2021-07-01
fix wrong thrift strict header comparing

ihciah:
Is there something wrong with the thrift strict header comparing? 

The strict header highest bit should be 1, and as u8 in rust, maybe we should compare it to 0x80 which is 0x10000000.

I didn't submit the issue on apache jira because I was unable to reset my password. Sorry for that.


Jens-G:
Good catch. What about `if ((first_bytes[0] & 0x80) != 0) { ...` ?:

ihciah:
> Good catch. What about `if ((first_bytes[0] & 0x80) != 0) { ...` ?

Indeed this way is better. I've rebased the code and rewritten the commit.

PR #2413, opened 2021-07-05
THRIFT-5439: Lua Generator: not support const i64

hkb1990:


Jens-G:
Some more information about the problem and why the patch solves it would be super awesome.

hkb1990:
> Some more information about the problem and why the patch solves it would be super awesome.

described in jira ticket 5439.
`const i64` is not generated correctly to lua. it is always generated to `lualongnumber.new('')` , no matter what value it is. Because t_base_type is i64, `get_string` returns empty.

PR #2416, opened 2021-07-09
Small bug in Rust generated code for writing container types

AlephAlpha:
The methods `write_container_end` should be call after the for loop, not within it.

Currently this does not affect anything, because these methods are no-ops for all the implemented protocols. But it may cause an error when more protocols are implemented.

By the way, could anyone please release version 0.14.0, 0.14.1, and 0.14.2 to crates.io?


Jens-G:
> could anyone please release version 0.14.0, 0.14.1, and 0.14.2 to crates.io?

@allengeorge could you do that?

allengeorge:
Yeap - I can do that.

PR #2424, opened 2021-07-31
THRIFT-5444

Jens-G:
THRIFT-5444 Netstd generator produces uncompileable code for enums ending with "_result" or "_args"


EnigmaTriton (approved):
I understand what was done in this patch (thanks for the explanations on the history) and it works as expected. Looks good to me.

Jens-G:
THRIFT-5445 still one missing, going to post separate PR and exclude it from this one.

PR #2431, opened 2021-08-04
THRIFT-5447: Update supported Go versions

fishy:
Client: go

Update go versions used in travis to 1.16.7 and 1.17, update
LANGUAGES.md, and update go's README to clarify on support policy.

This change will be cherry-picked into 0.15.0 branch after merged.

### DIFF
--- a/test/go/genmock.sh
+++ b/test/go/genmock.sh
@@ -4,7 +4,6 @@ set -e
 
 export GOPATH=$(mktemp -d -t gopath-XXXXXXXXXX)
 
-# TODO: Once we dropped support to go 1.15, add "@v1.5.0" suffix to go install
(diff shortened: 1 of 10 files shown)


test/go/genmock.sh
```diff
@@ -4,7 +4,6 @@ set -e
 
 export GOPATH=$(mktemp -d -t gopath-XXXXXXXXXX)
 
-# TODO: Once we dropped support to go 1.15, add "@v1.5.0" suffix to go install
```
fishy, line 7: NOTE: I removed this TODO because I realized that without "@v1.5.0", `go install` will actually pick the version according to `go.mod`, which is better because we don't need to keep the two versions in sync.

dcelasun (approved):

PR #2432, opened 2021-08-05
THRIFT-5449: Use poll instead of select in Python TNonblockingServer if available

frankiechang123:
Use poll instead of select if possible

Currently, Python TNonblockingServer uses `select.select` to monitor the IO of all existing connections. The value of file descriptors monitored by `select()` is limited to be lower than FD_SETSIZE (1024). Any file descriptor higher than 1024 will trigger the `fd out of range` error in `select.select`

Switching to `select.poll` can solve this problem. However, as noted in the [official documentation](https://docs.python.org/3/library/select.html#select.poll), `select.poll` is not supported in all operating systems. We should also check whether the current system has `poll` in the select module, and use `poll` if it's available. 

Another advantage of `poll` came up while I was debugging my original draft: `poll` is able to explicitly catch the case where the client hangs up by returning `POLLHUP` and we can close the connection gracefully. As a result, the [logging message](https://github.com/apache/thrift/blob/82a9c67405602ff053146a8df6bd566d90a9bf03/lib/py/src/server/TNonblockingServer.py#L149) `DEBUG:thrift.server.TNonblockingServer:read zero length. client might have disconnected` will not appear in the output in the testing script.

### DIFF
--- a/lib/py/src/server/TNonblockingServer.py
+++ b/lib/py/src/server/TNonblockingServer.py
@@ -253,6 +253,7 @@ class TNonblockingServer(object):
         self._read, self._write = socket.socketpair()
         self.prepared = False
         self._stop = False
+        self.poll = select.poll() if hasattr(select, 'poll') else None
 
     def setNumThreads(self, num):
         """Set the number of worker threads that should be created."""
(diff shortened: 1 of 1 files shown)


frankiechang123:
I think the Travis CI is falling for unrelated reasons. Will this PR be reviewed soon? :)

Jens-G:
@frankiechang123 could you rebase & force-push to trigger another CI run please?

frankiechang123:
Done :)

PR #2435, opened 2021-08-09
THRIFT-5453: Defer DNS from NewTSocketConf to TSocket.Open

fishy:
Client: go

We used to do DNS lookups in NewTSocketConf, without any timeout checks.
Stop doing that and do DNS lookups in TSocket.Open instead, which
already checks for ConnectTimeout set in TConfiguration.

This will be cherry-picked into 0.15.0 branch.


dcelasun (approved):

fishy:
The test looks good, but I realized that since `NewTSocketConf` can never return an error now, I'd remove that from the function signature while I'm here.

Also updated some of the tests to stop using deprecated functions and switch to use the -Conf versions.

@dcelasun please take another look :)

dcelasun (approved):
Looks good!

PR #2439, opened 2021-08-16
THRIFT-5443: add support for partial Thrift deserialization

bhalchandrap:
Client: java

Adds support for partial Thrift deserialization in Java.
More information is available at these locations:
blog: https://medium.com/pinterest-engineering/improving-data-processing-efficiency-using-partial-deserialization-of-thrift-16bc3a4a38b4
in source: lib/java/src/org/apache/thrift/partial/README.md  


Jens-G (commented):
Nice! I admit that I did not run any tests, just from what I saw from looking at the code and the README.

That would be awesome:
> "TBD: add additional information on defining map keys and values."
I have a strong feeling this will be among the first questions people are going to ask ;-)

Also, did you run tests against "fuzzy" input data? That's becoming kind of standard these days, given an extension like this that should probably be done before we merge.

bhalchandrap:
Updated README.md file to add clarification on map field path syntax.

I have not done fuzzy testing. However, the code includes in depth testing. In addition, the code is in active use in production jobs at Pinterest.

lib/java/src/org/apache/thrift/partial/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Partial deserialization
+
+Partial deserialization involves deserializing only a subset of the fields of a serialized Thrift object while efficiently skipping over the rest. One very important benefit of partial deserialization is that the output of the deserialization process is not limited to a `TBase` derived object. It can deserialize a serialized blob into any type by using an appropriate `ThriftFieldValueProcessor`.
```
fishy, line 20: I see this is where you explained the API difference from `TDeserialize`.

Is there a real use case of deserializing a non-`TBase`, though? My main concern is that to map this API to other languages, `Object` would be `interface{}` in go, and we really want to avoid `interface{}` in go as much as possible.

lib/java/src/org/apache/thrift/partial/README.md
```diff
@@ -0,0 +1,112 @@
...
+At this point, we have an efficient internal representation of the fields that need to get deserialized.
+
+### Partial Thrift Protocol
+
+Source files:
+- PartialThriftProtocol.java
+- PartialThriftBinaryProtocol.java
+- PartialThriftCompactProtocol.java
+
+This component implements efficient skipping over fields that need not be deserialized. Note that this skipping is more efficient compared to that achieved by using `TProtocolUtil.skip()`. The latter calls the corresponding `read()`, allocates and initializes certain values (for example, strings) and then discards the returned value. In comparison, `PartialThriftProtocol` skips a field by incrementing internal offset into the transport buffer.
```
fishy, line 88: While this is true that the existing `skip` implementations all need to read the field wholly then discard it, they don't necessarily have to be this way. I would prefer to replace the existing `TProtocolUtil.skip` with the more efficient implementation, than creating a new, separated API for them.

fishy (commented):
Java is also not my bread and butter, so my feedback here are more focused on the API than the implementation itself.

I think overall the design makes sense. To map it to go (my bread and butter), some heavy `reflect` usage will be needed (we _might_ even need some compiler changes to support it), but go's `reflect` implementation is in general efficient so I'm not too worried about that. An alternative to the string version of fully qualified field definition is to use the numeric ids, but that doesn't really make the implementation easier (at least not for go), and string version is more readable.

lib/java/src/org/apache/thrift/partial/PartialThriftCompactProtocol.java
```diff
@@ -0,0 +1,93 @@
...
+
+  // -----------------------------------------------------------------
+  // Additional methods to improve performance.
+
+  @Override
+  public int readFieldBeginData() throws TException {
+    // Having to call readFieldBegin() to compute TFieldData really results in lower
+    // performance. However, readFieldBegin() accesses some private vars that this method
+    // does not have access to. We could make it more performant when contributing to
+    // origianl source code.
```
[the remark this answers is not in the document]
bhalchandrap: added a TODO comment for now. That way we can better separate original contribution from subsequent performance improvement changes.

lib/java/src/org/apache/thrift/partial/PartialThriftDeserializer.java
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /**
+   * Deserializes the given serialized blob.
+   *
+   * @param bytes the serialized blob.
+   * @return deserialized instance.
+   * @throws TException if an error is encountered during deserialization.
+   */
+  public Object deserialize(byte[] bytes) throws TException {
```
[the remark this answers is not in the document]
bhalchandrap: It returns Object on purpose. See the class header comment as well as the README.md file for more info.
This arrangement allows non-TBase values to be directly instantiated from serialized Thrift objects. One example of its usefulness is Spark engine reading SequenceFile of Thrift objects. In that case, this arrangement allows SparkThriftRowProcessor (not a part of this change) to directly deserialize a serialized Thrift blob into Spark's InternalRow representation. That enables significant performance boost.

lib/java/src/org/apache/thrift/partial/PartialThriftDeserializer.java
```diff
@@ -0,0 +1,337 @@
...
+      throws TException {
+    int ordinal = tprot.readI32();
+    Class<? extends TEnum> enumClass = ((EnumMetaData) data.data.valueMetaData).enumClass;
+    return this.processor.prepareEnum(enumClass, ordinal);
+  }
+
+  private TBase fullDeserialize(PartialThriftProtocol tprot, ThriftMetadata.ThriftStruct data)
+      throws TException {
+    Validate.checkState(
+        data.fields.size() == 0, "Cannot fully deserialize when some fields specified");
```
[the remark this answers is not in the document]
bhalchandrap: good catch. fixed.

lib/java/src/org/apache/thrift/partial/README.md
bhalchandrap: Yes, it is very useful in certain cases (see earlier comment). ThriftStructProcessor enables direct deserialization into TBase.

lib/java/src/org/apache/thrift/partial/README.md
bhalchandrap: That would have been great. However, TProtocolUtil is protocol agnostic and the specifics of certain types of skip differ between compact and binary protocol. That is why it needs to happen at protocol level.

bhalchandrap:
addressed review comments

bhalchandrap:
@fishy @Jens-G do you have additional comments? do you know a Java dev who can review this PR?

lib/java/src/org/apache/thrift/partial/PartialThriftDeserializer.java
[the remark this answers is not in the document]
fishy: did you? the error message looks still the same as before?

lib/java/src/org/apache/thrift/partial/PartialThriftDeserializer.java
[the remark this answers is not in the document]
fishy: I'm still not convinced that this API need to be so dramatically different from the existing TDeserializer API (https://www.javadoc.io/doc/org.apache.thrift/libthrift/latest/org/apache/thrift/TDeserializer.html), can you try to match that as much as possible? The current TDeserializer API in Java already provided `partiallyDeserialize*` functions.

I'm really worried about how this API will map to other languages, even if it makes sense for the Java case.

lib/java/src/org/apache/thrift/partial/README.md
fishy: The current `TProtocolUtil.skip` requires a `TProtocol` arg (https://www.javadoc.io/static/org.apache.thrift/libthrift/0.11.0/org/apache/thrift/protocol/TProtocolUtil.html), so you can totally then delegate to the concrete TProtocol to decide how to skip that field in the most efficient way. (for example, in Go we already have `Skip` function defined in the `TProtocol` interface, so it can be different from `read*`. we can probably do something similar in java.)

lib/java/src/org/apache/thrift/partial/PartialThriftDeserializer.java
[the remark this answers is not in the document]
bhalchandrap: Hi @fishy, 
Happy to make changes if I get clarity on the concerns you have and on the type of changes you are looking for. 

Based on your comments I believe you prefer adding functionality to existing interfaces instead of adding new ones (please confirm since this is my guess). If so, how does the following work for you?

1. move common skip*() methods from PartialThriftProtocol.java to TProtocol.java.
2. move encoding specific skip*() methods from PartialThrift*Protocol.java to T*Protocol (for example, from PartialThriftBinaryProtocol.java to TBinaryProtocol)
3. get rid of PartialThriftDeserializer.java and add its methods to TDeserializer

Would something like the above address your concerns?

lib/java/src/org/apache/thrift/partial/README.md
bhalchandrap: got it. basically you are recommending moving skip*() methods directly into T*Protocol. can you please confirm?

bhalchandrap:
@fishy Thanks for reviewing! Can you please provide clarification on the questions?

fishy:
@bhalchandrap for skip: yes please make `TProtocolUtil.skip` choose the most efficient way to skip a field in a TProtocol, instead of adding a new API for "efficient skip" and keep `TProtocolUtil.skip` as the "slow skip".

For partial deserialization, I don't know which makes more sense in Java API. If you can totally use the existing API to implement it that would be awesome (I imagine you at least need to add new constructors to `TDeserializer` to take the field maps you want to partially deserialize). If you can't, I think it's fine to add new APIs for it as long as the new API design makes sense, but the new API should be more similar to the existing `TDeserializer` API, unless we have a good reason to make it quite different (e.g. the current TDeserializer API is bad/inefficient/etc. because of X Y Z).

bhalchandrap:
@fishy thanks for the earlier suggestions. I applied all of them. Can you please re-review?

lib/java/src/org/apache/thrift/TDeserializer.java
```diff
@@ -353,4 +418,265 @@ private TField locateField(byte[] bytes, TFieldIdEnum fieldIdPathFirst, TFieldId
...
+   * Deserializes the given serialized blob.
+   *
+   * @param bytes the serialized blob.
+   * @param offset the blob is read starting at this offset.
+   * @param length the size of blob read (in number of bytes).
+   * @return deserialized instance.
+   * @throws TException if an error is encountered during deserialization.
+   */
+  public Object deserialize(byte[] bytes, int offset, int length) throws TException {
+    ensurePartialDeserializationMode();
```
fishy, line 446: We already have a `deserialize` function with very similar signature:
```
public void deserialize​(TBase base, byte[] bytes, int offset, int length) throws TException
```
adding this function will make it really confusing to users on what to call. my suggestions are:
1. Maybe rename this one to `deserializeObject` to emphasize that it returns `Object` instead of writes to `TBase`.
2. Why does this only support partial mode (it throws a runtime exception when the `TDeserializer` was not consutrcted with partial mode), while there's nothing in its name or even javadoc mentioning that? that sounds like a huge footgun. Is there even a reason this should call `ensurePartialDeserializationMode`?

lib/java/src/org/apache/thrift/TDeserializer.java
```diff
@@ -61,6 +78,54 @@ public TDeserializer(TProtocolFactory protocolFactory) throws TTransportExceptio
...
+  /**
+   * Construct a new TDeserializer that supports partial deserialization
+   * that outputs instances of type controlled by the given {@code processor}.
+   *
+   * @param thriftClass a TBase derived class.
+   * @param fieldNames list of fields to deserialize.
+   * @param processor the Processor that handles deserialized field values.
+   * @param protocolFactory the Factory to create a protocol.
+   */
+  public TDeserializer(
```
fishy, line 90: (maybe it just works, I'm not familiar with the java codebase and this is not obvious to me)

I don't see you changing any of the existing public functions of `TDeserializer`, for example:
```
public void deserialize​(TBase base, byte[] bytes) throws TException
public void deserialize​(TBase base, byte[] bytes, int offset, int length) throws TException
public void deserialize​(TBase base, java.lang.String data, java.lang.String charset) throws TException
```
Since you are adding new constructors to `TDeserializer`, you need to ensure that those existing public functions still work as intended. When someone uses this constructor to create a `TDeserializer`, then calls `deserialize​(TBase base, byte[] bytes)`, they would expect that `deserialize` function does partial deserialization, not the full one.

lib/java/src/org/apache/thrift/protocol/TProtocol.java
```diff
@@ -180,4 +181,145 @@ public void reset() {}
...
+    this.skipBytes(4);
+  }
+
+  protected void skipI64() throws TException {
+    this.skipBytes(8);
+  }
+
+  protected void skipDouble() throws TException {
+    this.skipBytes(8);
+  }
```
fishy, line 300: I don't think the default implementations should use hardcoded number of bytes. the default implementations should just call `read*` and discard the result. this guarantees correctness, and leaves individual protocols the room to override the default implementation to improve performance. Using hardcoded number of bytes cannot guarantee the correctness.

fishy (changes requested):

bhalchandrap:
@fishy thanks for the careful code review and the suggestions. I applied your suggestions. Can you please re-review? I will add tests corresponding to the new code once the overall structure looks ok to you.

bhalchandrap:
@Jens-G can you please let me know if I can do anything from my side to make progress on this pull request?

Jens-G:
> This branch cannot be rebased due to conflicts

If there are no objections and the CI is green again I would merge this in the next days.

Jens-G:
Next time please either squash the PR or at least rebase it onto master every now and then.

bhalchandrap:
Certainly. This was the first time I contributed to OSS. It was a good
learning experience for me.

lib/java/src/org/apache/thrift/partial/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+At this point, we have an efficient internal representation of the fields that need to get deserialized.
+
+### Partial Thrift Protocol
+
+Source files:
+- PartialThriftProtocol.java
+- PartialThriftBinaryProtocol.java
+- PartialThriftCompactProtocol.java
```
dulong, line 86: If these source files have been removed，the document should be updated。

bhalchandrap: Sorry that was an oversight. I will update the document asap and send another PR.

PR #2440, opened 2021-08-16
THRIFT-5456: [c++] Disable SIGPIPE on TServerSocket

kainjow:
TSocketServer sets up many socket options, but unlike TSocket, it doesn't disable SIGPIPE.

One example for how this can be triggered is the server receives a request, but takes too long to process it. The sender then times out, and potentially closes its sockets. The server then completes the request, and then attempts to write back the result. Since the socket is now closed, a SIGPIPE signal is sent to the process, and if that signal isn't ignored, the process would crash, with a potential -13 exit code.

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -311,6 +311,17 @@ void TServerSocket::_setup_sockopts() {
     throw TTransportException(TTransportException::NOT_OPEN, "Could not set SO_LINGER", errno_copy);
   }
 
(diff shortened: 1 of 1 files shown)


Jens-G:
Could you trigger the build again, e.g. by force-pushing the branch?

emmenlau:
Is this also missing in `TNonblockingServerSocket`?

Jens-G:
@kainjow any news here?

kainjow:
> @kainjow any news here?

Will take a look soon.

kainjow:
> TNonblockingServerSocket

Done!

PR #2452, opened 2021-09-01
THRIFT-5459: Fix breaking issue when adding a new exception

fishy:
Client: go

Currently in the compiler generated go code, adding a new exception to
an existing endpoint can cause unexpected behaviors when the client
isn't updated. Fix the issue.

Will be cherry-picked into 0.15.0 after merged.


dcelasun (approved):

fishy:
Sorry @dcelasun I forgot to write change log before. Just added, please take another look (code part is unchanged so you only need to look at the change log part :) )

dcelasun (approved):
Looks good!

PR #2453, opened 2021-09-08
THRIFT-4868: Golang: Fix compilation for optional set<binary> with default values

vernor1:

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -1309,7 +1309,7 @@ string t_go_generator::render_const_value(t_type* type, t_const_value* value, co
     t_type* ktype = ((t_map*)type)->get_key_type();
     t_type* vtype = ((t_map*)type)->get_val_type();
     const map<t_const_value*, t_const_value*, t_const_value::value_compare>& val = value->get_map();
-    out << "map[" << type_to_go_type(ktype) << "]" << type_to_go_type(vtype) << "{" << endl;
+    out << "map[" << type_to_go_key_type(ktype) << "]" << type_to_go_type(vtype) << "{" << endl;
     indent_up();
     map<t_const_value*, t_const_value*, t_const_value::value_compare>::const_iterator v_iter;
 
@@ -1336,7 +1336,7 @@ string t_go_generator::render_const_value(t_type* type, t_const_value* value, co
   } else if (type->is_set()) {
     t_type* etype = ((t_set*)type)->get_elem_type();
     const vector<t_const_value*>& val = value->get_list();
-    out << "[]" << type_to_go_key_type(etype) << "{" << endl;
+    out << "[]" << type_to_go_type(etype) << "{" << endl;
     indent_up();
     vector<t_const_value*>::const_iterator v_iter;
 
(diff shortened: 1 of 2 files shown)


compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -4021,7 +4021,7 @@ string t_go_generator::type_to_go_key_type(t_type* type) {
   }
 
   if (resolved_type->is_binary())
-    return "string";
+    return "[]byte";
```
fishy, line 4024: this is not the correct fix.

this function is called `type_to_go_key_type`, it's used to generate go type for keys in maps, and you cannot use `[]byte` as a map key in go. this will break all the maps using binary as the key type.

the correct fix is to call `type_to_go_type` instead in `render_const_value` for sets (because in go we don't actually render sets as maps, we render them as slices instead).

fishy (changes requested):

compiler/cpp/src/thrift/generate/t_go_generator.cc
vernor1: @fishy Thank you for the clarification. Could you check the updated solution?
Both types are now properly handled: `set<binary>` and `map<binary, T>`:
```
root@2dda16a0c248:/thrift/src# make -Ctest/go check
make: Entering directory '/thrift/src/test/go'
Makefile:657: warning: overriding recipe for target 'check'
Makefile:506: warning: ignoring old recipe for target 'check'
mkdir -p src/gen
/thrift/src/compiler/cpp/thrift -out src/gen --gen go:thrift_import=github.com/apache/thrift/lib/go/thrift,package_prefix=github.com/apache/thrift/test/go/src/gen/ ThriftTest.thrift
[WARNING:/thrift/src/test/go/ThriftTest.thrift:43] No generator named 'noexist' could be found!
[WARNING:/thrift/src/test/go/ThriftTest.thrift:45] cpp generator does not accept 'noexist' as sub-namespace!
/thrift/src/compiler/cpp/thrift -out src/gen --gen go:thrift_import=github.com/apache/thrift/lib/go/thrift,package_prefix=github.com/apache/thrift/test/go/src/gen/ ../StressTest.thrift
[WARNING:/thrift/src/test/StressTest.thrift:31] Consider using the more efficient "binary" type instead of "list<byte>".
[WARNING:/thrift/src/test/StressTest.thrift:31] Consider using the more efficient "binary" type instead of "list<byte>".
touch gopath
sh genmock.sh
go: downloading github.com/golang/mock v1.5.0
go: downloading golang.org/x/mod v0.3.0
go: downloading golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e
go: downloading golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898
/usr/local/bin/go test -mod=mod -v ./src/common/...
=== RUN   TestAllConnection
--- PASS: TestAllConnection (0.02s)
=== RUN   TestHttpContextTimeout
--- PASS: TestHttpContextTimeout (0.05s)
PASS
ok  	github.com/apache/thrift/test/go/src/common	0.088s
make: Leaving directory '/thrift/src/test/go'
```

fishy: looks good to me, will approve and merge it after the travis run

fishy (approved):

PR #2454, opened 2021-09-09
THRIFT-5457 Pin bitflags to 1.2 to maintain MSRV

allengeorge:
Client: rs


allengeorge:
Tested locally, and confirmed that this works.

PR #2455, opened 2021-09-19
Modified build tags to allow js/wasm target

balazsgrill:
I tried the golang implementation to be used for wasm platform, and found that I can do so by disabling the unix (not windows) specific `checkConn` and `read0` implementations.

### DIFF
--- a/lib/go/thrift/socket_non_unix_conn.go
+++ b/lib/go/thrift/socket_non_unix_conn.go
@@ -1,4 +1,5 @@
-// +build windows
+//go:build windows || wasm
+// +build windows wasm
 
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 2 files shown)


lib/go/thrift/socket_windows_conn.go
```diff
@@ -1,4 +1,4 @@
-// +build windows
+// +build windows wasm
```
fishy, line 1: with this change it's better to rename the file as well. maybe `socket_non_unix_conn.go`?

balazsgrill: Done. Do you need me to manually sqash the commits?

fishy: yes please.

while you are here, could you please also run `gofmt` from 1.17+ on these 2 files? go 1.17 introduced new build constraint format, and `gofmt` 1.17+ would auto generate the new build constraint from the old one. (the new one starts with `//go:build` instead of `// +build`, after running `gofmt` both should be there).

fishy (approved):
thank you!

lib/go/thrift/socket_windows_conn.go
balazsgrill: I ran gofmt on the affected files. It also kept the original build tags, which I guess is okay to be compatible with go 1.16 (as stated by the [go.mod](https://github.com/apache/thrift/blob/master/go.mod) file)

fishy:
Looks like we disabled travis, but I verified this locally with `make check` under `lib/go` and `test/go`

PR #2469, opened 2021-10-18
THRIFT-5423: Support go parameter validation in IDL

BytedanceRPC:
Support parameter validation, only go validation generator is supported now.
Users can define validation rule for struct in Thrift file, then the generator will generate validation code, for example, go validation generator will generate IsValid() method.

Below is an example:
```Thrift
enum MapKey {
    A, B, C, D, E, F
}

struct Example {
    1: string Message (vt.min_size = "30") // length of Message should be greater than or equal to 30
    2: i32 ID (vt.ge = "10000") // ID must be greater than or euqal to 10000
    3: list<double> Values (vt.elem.gt = "0.25") // element of Values must be greater than 0.25
    4: map<MapKey, string> KeyValues (vt.key.defined_only = "true") // value of KeyValues'key must be defined in MapKey
}
```

Feature Matrix:
prefix `vt`, short for "validation"
Numeric(i8/i16/i32/i64/double):

| Rule      |                                          |
| --------- | ---------------------------------------- |
| vt.const  | must be specified value                  |
| vt.lt     | less than the specified value            |
| vt.le     | less than or equal to specified value    |
| vt.gt     | greater than the specified value         |
| vt.ge     | greater than or equal to specified value |
| vt.in     | must be in specified values              |
| vt.not_in | must not be in specified values          |

Bool:

| Rule     |                         |
| -------- | ----------------------- |
| vt.const | must be specified value |

String/Binary:

| Rule            |                                  |
| --------------- | -------------------------------- |
| vt.const        | must be specified value          |
| vt.pattern      | regexp pattern                   |
| vt.prefix       | prefix must be specified value   |
| vt.suffix       | suffix must be specified value   |
| vt.contains     | must contain specified value     |
| vt.not_contains | must not contain specified value |
| vt.min_size     | min size                         |
| vt.max_size     | max size                         |

Bool:

| Rule     |                         |
| -------- | ----------------------- |
| vt.const | must be specified value |

Enum:

| Rule            |                         |
| --------------- | ----------------------- |
| vt.const        | must be specified value |
| vt.defined_only | must be defined value   |

Set/List:

| Rule            |                       |
| --------------- | --------------------- |
| vt.min_size     | min size              |
| vt.max_size     | max size              |
| vt.elem         | rule for list element |

Map:

| Rule            |                                   |
| --------------- | --------------------------------- |
| vt.min_size     | min size                          |
| vt.max_size     | max size                          |
| vt.key          | rule for map key                  |
| vt.value        | rule for map value                |

Struct

| Rule    |                                  |
| ------- | -------------------------------- |
| vt.skip | skip struct recursive validation |

Special Value:
1. Field Reference. We can use another field as a validation value.
2. Validation Function. We can use those functions to provide extensive validation ability.

Field Reference Example:
```Thrift
struct Example {
    1: string StringFoo (vt.max_size = "$MaxStringSize")
    2: i32 MaxStringSize
}
```

Validation Function:
```Thrift
struct Example {
    1: string MaxString
    2: list<string> StringList (vt.elem.max_size = "@len($MaxString)")
}
```
for now, only `len` function is available.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/browse/THRIFT-5423) ticket?  (not required for trivial changes)

### DIFF
--- a/compiler/cpp/CMakeLists.txt
+++ b/compiler/cpp/CMakeLists.txt
@@ -46,6 +46,8 @@ add_library(parse STATIC ${parse_SOURCES})
 set(compiler_core
     src/thrift/common.cc
     src/thrift/generate/t_generator.cc
+    src/thrift/generate/validator_parser.cc
+    src/thrift/generate/validator_parser.h
     src/thrift/parse/t_typedef.cc
     src/thrift/parse/parse.cc
     src/thrift/version.h
@@ -71,36 +73,51 @@ macro(THRIFT_ADD_COMPILER name description initial)
     endif()
 endmacro()
 
+# This macro adds an option THRIFT_VALIDATOR_COMPILER_${NAME}
+# that allows enabling or disabling certain languages' validator
+macro(THRIFT_ADD_VALIDATOR_COMPILER name description initial)
+    string(TOUPPER "THRIFT_COMPILER_${name}" enabler)
+    set(src "src/thrift/generate/${name}_validator_generator.cc")
+    list(APPEND "src/thrift/generate/${name}_validator_generator.h")
+    option(${enabler} ${description} ${initial})
+    if(${enabler})
+        list(APPEND thrift-compiler_SOURCES ${src})
+    endif()
+endmacro()
+
 # The following compiler can be enabled or disabled
-THRIFT_ADD_COMPILER(c_glib  "Enable compiler for C with Glib" ON)
-THRIFT_ADD_COMPILER(cl      "Enable compiler for Common LISP" ON)
-THRIFT_ADD_COMPILER(cpp     "Enable compiler for C++" ON)
-THRIFT_ADD_COMPILER(d       "Enable compiler for D" ON)
-THRIFT_ADD_COMPILER(dart    "Enable compiler for Dart" ON)
-THRIFT_ADD_COMPILER(delphi  "Enable compiler for Delphi" ON)
-THRIFT_ADD_COMPILER(erl     "Enable compiler for Erlang" ON)
-THRIFT_ADD_COMPILER(go      "Enable compiler for Go" ON)
-THRIFT_ADD_COMPILER(gv      "Enable compiler for GraphViz" ON)
-THRIFT_ADD_COMPILER(haxe    "Enable compiler for Haxe" ON)
-THRIFT_ADD_COMPILER(html    "Enable compiler for HTML Documentation" ON)
-THRIFT_ADD_COMPILER(markdown "Enable compiler for Markdown Documentation" ON)
-THRIFT_ADD_COMPILER(java    "Enable compiler for Java"   ON)
-THRIFT_ADD_COMPILER(javame  "Enable compiler for Java ME" ON)
-THRIFT_ADD_COMPILER(js      "Enable compiler for JavaScript" ON)
-THRIFT_ADD_COMPILER(json    "Enable compiler for JSON" ON)
-THRIFT_ADD_COMPILER(kotlin  "Enable compiler for Kotlin" ON)
-THRIFT_ADD_COMPILER(lua     "Enable compiler for Lua" ON)
-THRIFT_ADD_COMPILER(netstd  "Enable compiler for .NET Standard" ON)
-THRIFT_ADD_COMPILER(ocaml   "Enable compiler for OCaml" ON)
-THRIFT_ADD_COMPILER(perl    "Enable compiler for Perl" ON)
-THRIFT_ADD_COMPILER(php     "Enable compiler for PHP" ON)
-THRIFT_ADD_COMPILER(py      "Enable compiler for Python 2.0" ON)
-THRIFT_ADD_COMPILER(rb      "Enable compiler for Ruby" ON)
-THRIFT_ADD_COMPILER(rs      "Enable compiler for Rust" ON)
-THRIFT_ADD_COMPILER(st      "Enable compiler for Smalltalk" ON)
-THRIFT_ADD_COMPILER(swift   "Enable compiler for Cocoa Swift" ON)
-THRIFT_ADD_COMPILER(xml     "Enable compiler for XML" ON)
-THRIFT_ADD_COMPILER(xsd     "Enable compiler for XSD" ON)
+THRIFT_ADD_COMPILER(c_glib        "Enable compiler for C with Glib" ON)
+THRIFT_ADD_COMPILER(cl            "Enable compiler for Common LISP" ON)
+THRIFT_ADD_COMPILER(cpp           "Enable compiler for C++" ON)
+THRIFT_ADD_COMPILER(d             "Enable compiler for D" ON)
+THRIFT_ADD_COMPILER(dart          "Enable compiler for Dart" ON)
+THRIFT_ADD_COMPILER(delphi        "Enable compiler for Delphi" ON)
+THRIFT_ADD_COMPILER(erl           "Enable compiler for Erlang" ON)
+THRIFT_ADD_COMPILER(go            "Enable compiler for Go" ON)
+THRIFT_ADD_COMPILER(gv            "Enable compiler for GraphViz" ON)
+THRIFT_ADD_COMPILER(haxe          "Enable compiler for Haxe" ON)
+THRIFT_ADD_COMPILER(html          "Enable compiler for HTML Documentation" ON)
+THRIFT_ADD_COMPILER(markdown      "Enable compiler for Markdown Documentation" ON)
+THRIFT_ADD_COMPILER(java          "Enable compiler for Java"   ON)
+THRIFT_ADD_COMPILER(javame        "Enable compiler for Java ME" ON)
+THRIFT_ADD_COMPILER(js            "Enable compiler for JavaScript" ON)
+THRIFT_ADD_COMPILER(json          "Enable compiler for JSON" ON)
+THRIFT_ADD_COMPILER(kotlin        "Enable compiler for Kotlin" ON)
+THRIFT_ADD_COMPILER(lua           "Enable compiler for Lua" ON)
+THRIFT_ADD_COMPILER(netstd        "Enable compiler for .NET Standard" ON)
+THRIFT_ADD_COMPILER(ocaml         "Enable compiler for OCaml" ON)
+THRIFT_ADD_COMPILER(perl          "Enable compiler for Perl" ON)
+THRIFT_ADD_COMPILER(php           "Enable compiler for PHP" ON)
+THRIFT_ADD_COMPILER(py            "Enable compiler for Python 2.0" ON)
+THRIFT_ADD_COMPILER(rb            "Enable compiler for Ruby" ON)
+THRIFT_ADD_COMPILER(rs            "Enable compiler for Rust" ON)
+THRIFT_ADD_COMPILER(st            "Enable compiler for Smalltalk" ON)
+THRIFT_ADD_COMPILER(swift         "Enable compiler for Cocoa Swift" ON)
+THRIFT_ADD_COMPILER(xml           "Enable compiler for XML" ON)
+THRIFT_ADD_COMPILER(xsd           "Enable compiler for XSD" ON)
+
+# The following compiler can be enabled or disabled by enabling or disabling certain languages
+THRIFT_ADD_VALIDATOR_COMPILER(go           "Enable validator compiler for Go" ON)
 
 # Thrift is looking for include files in the src directory
 # we also add the current binary directory for generated files
--- a/compiler/cpp/src/thrift/generate/go_validator_generator.h
+++ b/compiler/cpp/src/thrift/generate/go_validator_generator.h
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef T_GO_VALIDATOR_GENERATOR_H
+#define T_GO_VALIDATOR_GENERATOR_H
+
+#include "thrift/generate/t_generator.h"
+#include "thrift/generate/t_go_generator.h"
+#include "thrift/generate/validator_parser.h"
+#include <fstream>
+#include <iostream>
+#include <limits>
+#include <string>
+#include <vector>
+
+class go_validator_generator {
+public:
+  go_validator_generator(t_go_generator* gg) : go_generator(gg){};
+  void generate_struct_validator(std::ostream& out, t_struct* tstruct);
+
+  struct generator_context {
+    std::string field_symbol;
+    std::string src;
+    std::string tgt;
+    bool opt;
+    t_type* type;
+    std::vector<validation_rule*> rules;
+  };
+
+private:
+  void generate_field_validator(std::ostream& out, const generator_context& context);
+  void generate_enum_field_validator(std::ostream& out, const generator_context& context);
+  void generate_bool_field_validator(std::ostream& out, const generator_context& context);
+  void generate_integer_field_validator(std::ostream& out, const generator_context& context);
+  void generate_double_field_validator(std::ostream& out, const generator_context& context);
+  void generate_string_field_validator(std::ostream& out, const generator_context& context);
+  void generate_list_field_validator(std::ostream& out, const generator_context& context);
+  void generate_set_field_validator(std::ostream& out, const generator_context& context);
+  void generate_map_field_validator(std::ostream& out, const generator_context& context);
+  void generate_struct_field_validator(std::ostream& out, const generator_context& context);
+
+  void indent_up() { go_generator->indent_up(); }
+  void indent_down() { go_generator->indent_down(); }
+  std::string indent() { return go_generator->indent(); }
+
+  std::string get_field_name(t_field* field);
+  std::string get_field_reference_name(t_field* field);
+
+  std::string GenID(std::string id) { return id + std::to_string(tmp_[id]++); };
+
+  t_go_generator* go_generator;
+
+  std::map<std::string, int> tmp_;
+};
+
+#endif
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -41,6 +41,8 @@
 #include "thrift/platform.h"
 #include "thrift/version.h"
 #include "thrift/generate/t_generator.h"
+#include "thrift/generate/t_go_generator.h"
+#include "thrift/generate/go_validator_generator.h"
 
 using std::map;
 using std::ostream;
@@ -48,8 +50,6 @@ using std::ostringstream;
 using std::string;
 using std::stringstream;
 using std::vector;
-
-static const string endl = "\n"; // avoid ostream << std::endl flushes
 
 /**
  * A helper for automatically formatting the emitted Go code from the Thrift
@@ -61,277 +61,6 @@ static const string endl = "\n"; // avoid ostream << std::endl flushes
  *           still generated.
  */
 bool format_go_output(const string& file_path);
-
-const string DEFAULT_THRIFT_IMPORT = "github.com/apache/thrift/lib/go/thrift";
-static std::string package_flag;
-
-/**
- * Go code generator.
- */
-class t_go_generator : public t_generator {
-public:
-  t_go_generator(t_program* program,
-                 const std::map<std::string, std::string>& parsed_options,
-                 const std::string& option_string)
-    : t_generator(program) {
-    (void)option_string;
-    std::map<std::string, std::string>::const_iterator iter;
-
-
-    gen_thrift_import_ = DEFAULT_THRIFT_IMPORT;
-    gen_package_prefix_ = "";
-    package_flag = "";
-    read_write_private_ = false;
-    ignore_initialisms_ = false;
-    skip_remote_ = false;
-    for( iter = parsed_options.begin(); iter != parsed_options.end(); ++iter) {
-      if( iter->first.compare("package_prefix") == 0) {
-        gen_package_prefix_ = (iter->second);
-      } else if( iter->first.compare("thrift_import") == 0) {
-        gen_thrift_import_ = (iter->second);
-      } else if( iter->first.compare("package") == 0) {
-        package_flag = (iter->second);
-      } else if( iter->first.compare("read_write_private") == 0) {
-        read_write_private_ = true;
-      } else if( iter->first.compare("ignore_initialisms") == 0) {
-        ignore_initialisms_ =  true;
-      } else if( iter->first.compare("skip_remote") == 0) {
-        skip_remote_ =  true;
-      } else {
-        throw "unknown option go:" + iter->first;
-      }
-    }
-
-    out_dir_base_ = "gen-go";
-  }
-
-  /**
-   * Init and close methods
-   */
-
-  void init_generator() override;
-  void close_generator() override;
-
-  /**
-   * Program-level generation functions
-   */
-
-  void generate_typedef(t_typedef* ttypedef) override;
-  void generate_enum(t_enum* tenum) override;
-  void generate_const(t_const* tconst) override;
-  void generate_struct(t_struct* tstruct) override;
-  void generate_xception(t_struct* txception) override;
-  void generate_service(t_service* tservice) override;
-
-  std::string render_const_value(t_type* type, t_const_value* value, const string& name, bool opt = false);
-
-  /**
-   * Struct generation code
-   */
-
-  void generate_go_struct(t_struct* tstruct, bool is_exception);
-  void generate_go_struct_definition(std::ostream& out,
-                                     t_struct* tstruct,
-                                     bool is_xception = false,
-                                     bool is_result = false,
-                                     bool is_args = false);
-  void generate_go_struct_initializer(std::ostream& out,
-                                      t_struct* tstruct,
-                                      bool is_args_or_result = false);
-  void generate_isset_helpers(std::ostream& out,
-                              t_struct* tstruct,
-                              const string& tstruct_name,
-                              bool is_result = false);
-  void generate_countsetfields_helper(std::ostream& out,
-                                      t_struct* tstruct,
-                                      const string& tstruct_name,
-                                      bool is_result = false);
-  void generate_go_struct_reader(std::ostream& out,
-                                 t_struct* tstruct,
-                                 const string& tstruct_name,
-                                 bool is_result = false);
-  void generate_go_struct_writer(std::ostream& out,
-                                 t_struct* tstruct,
-                                 const string& tstruct_name,
-                                 bool is_result = false,
-                                 bool uses_countsetfields = false);
-  void generate_go_struct_equals(std::ostream& out, t_struct* tstruct, const string& tstruct_name);
-  void generate_go_function_helpers(t_function* tfunction);
-  void get_publicized_name_and_def_value(t_field* tfield,
-                                         string* OUT_pub_name,
-                                         t_const_value** OUT_def_value) const;
-
-  /**
-   * Service-level generation functions
-   */
-
-  void generate_service_helpers(t_service* tservice);
-  void generate_service_interface(t_service* tservice);
-  void generate_service_client(t_service* tservice);
-  void generate_service_remote(t_service* tservice);
-  void generate_service_server(t_service* tservice);
-  void generate_process_function(t_service* tservice, t_function* tfunction);
-
-  /**
-   * Serialization constructs
-   */
-
-  void generate_deserialize_field(std::ostream& out,
-                                  t_field* tfield,
-                                  bool declare,
-                                  std::string prefix = "",
-                                  bool inclass = false,
-                                  bool coerceData = false,
-                                  bool inkey = false,
-                                  bool in_container = false);
-
-  void generate_deserialize_struct(std::ostream& out,
-                                   t_struct* tstruct,
-                                   bool is_pointer_field,
-                                   bool declare,
-                                   std::string prefix = "");
-
-  void generate_deserialize_container(std::ostream& out,
-                                      t_type* ttype,
-                                      bool pointer_field,
-                                      bool declare,
-                                      std::string prefix = "");
-
-  void generate_deserialize_set_element(std::ostream& out,
-                                        t_set* tset,
-                                        bool declare,
-                                        std::string prefix = "");
-
-  void generate_deserialize_map_element(std::ostream& out,
-                                        t_map* tmap,
-                                        bool declare,
-                                        std::string prefix = "");
-
-  void generate_deserialize_list_element(std::ostream& out,
-                                         t_list* tlist,
-                                         bool declare,
-                                         std::string prefix = "");
-
-  void generate_serialize_field(std::ostream& out,
-                                t_field* tfield,
-                                std::string prefix = "",
-                                bool inkey = false);
-
-  void generate_serialize_struct(std::ostream& out, t_struct* tstruct, std::string prefix = "");
-
-  void generate_serialize_container(std::ostream& out,
-                                    t_type* ttype,
-                                    bool pointer_field,
-                                    std::string prefix = "");
-
-  void generate_serialize_map_element(std::ostream& out,
-                                      t_map* tmap,
-                                      std::string kiter,
-                                      std::string viter);
-
-  void generate_serialize_set_element(std::ostream& out, t_set* tmap, std::string iter);
-
-  void generate_serialize_list_element(std::ostream& out, t_list* tlist, std::string iter);
-
-  void generate_go_equals(std::ostream& out, t_type* ttype, string tgt, string src);
-
-  void generate_go_equals_struct(std::ostream& out, t_type* ttype, string tgt, string src);
-
-  void generate_go_equals_container(std::ostream& out, t_type* ttype, string tgt, string src);
-
-  void generate_go_docstring(std::ostream& out, t_struct* tstruct);
-
-  void generate_go_docstring(std::ostream& out, t_function* tfunction);
-
-  void generate_go_docstring(std::ostream& out,
-                             t_doc* tdoc,
-                             t_struct* tstruct,
-                             const char* subheader);
-
-  void generate_go_docstring(std::ostream& out, t_doc* tdoc);
-
-  void parse_go_tags(map<string,string>* tags, const string in);
-
-  /**
-   * Helper rendering functions
-   */
-
-  std::string go_autogen_comment();
-  std::string go_package();
-  std::string go_imports_begin(bool consts);
-  std::string go_imports_end();
-  std::string render_includes(bool consts);
-  std::string render_included_programs(string& unused_protection);
-  std::string render_program_import(const t_program* program, string& unused_protection);
-  std::string render_system_packages(std::vector<string> &system_packages);
-  std::string render_import_protection();
-  std::string render_fastbinary_includes();
-  std::string declare_argument(t_field* tfield);
-  std::string render_field_initial_value(t_field* tfield, const string& name, bool optional_field);
-  std::string type_name(t_type* ttype);
-  std::string module_name(t_type* ttype);
-  std::string function_signature(t_function* tfunction, std::string prefix = "");
-  std::string function_signature_if(t_function* tfunction,
-                                    std::string prefix = "",
-                                    bool addError = false);
-  std::string argument_list(t_struct* tstruct);
-  std::string type_to_enum(t_type* ttype);
-  std::string type_to_go_type(t_type* ttype);
-  std::string type_to_go_type_with_opt(t_type* ttype,
-                                       bool optional_field);
-  std::string type_to_go_key_type(t_type* ttype);
-  std::string type_to_spec_args(t_type* ttype);
-
-  static std::string get_real_go_module(const t_program* program) {
-
-    if (!package_flag.empty()) {
-      return package_flag;
-    }
-    std::string real_module = program->get_namespace("go");
-    if (!real_module.empty()) {
-      return real_module;
-    }
-
-    return lowercase(program->get_name());
-  }
-
-private:
-  std::string gen_package_prefix_;
-  std::string gen_thrift_import_;
-  bool read_write_private_;
-  bool ignore_initialisms_;
-  bool skip_remote_;
-
-  /**
-   * File streams
-   */
-
-  ofstream_with_content_based_conditional_update f_types_;
-  std::string f_types_name_;
-  ofstream_with_content_based_conditional_update f_consts_;
-  std::string f_consts_name_;
-  std::stringstream f_const_values_;
-
-  std::string package_name_;
-  std::string package_dir_;
-  std::unordered_map<std::string, std::string> package_identifiers_;
-  std::set<std::string> package_identifiers_set_;
-  std::string read_method_name_;
-  std::string write_method_name_;
-  std::string equals_method_name_;
-
-  std::set<std::string> commonInitialisms;
-
-  std::string camelcase(const std::string& value) const;
-  void fix_common_initialism(std::string& value, int i) const;
-  std::string publicize(const std::string& value, bool is_args_or_result = false) const;
-  std::string publicize(const std::string& value, bool is_args_or_result, const std::string& service_name) const;
-  std::string privatize(const std::string& value) const;
-  std::string new_prefix(const std::string& value) const;
-  static std::string variable_name_to_go_name(const std::string& value);
-  static bool is_pointer_field(t_field* tfield, bool in_container = false);
-  static bool omit_initialization(t_field* tfield);
-};
 
 // returns true if field initialization can be omitted since it has corresponding go type zero value
 // or default value is not set
@@ -973,6 +702,10 @@ string t_go_generator::go_imports_begin(bool consts) {
   system_packages.push_back("time");
   // For the thrift import, always do rename import to make sure it's called thrift.
   system_packages.push_back("thrift \"" + gen_thrift_import_ + "\"");
+
+  // validator import
+  system_packages.push_back("strings");
+  system_packages.push_back("regexp");
   return "import (\n" + render_system_packages(system_packages);
 }
 
@@ -983,7 +716,7 @@ string t_go_generator::go_imports_begin(bool consts) {
  * This will have to do in lieu of more intelligent import statement construction
  */
 string t_go_generator::go_imports_end() {
-  return string(
+  string import_end = string(
       ")\n\n"
       "// (needed to ensure safety because of naive import list construction.)\n"
       "var _ = thrift.ZERO\n"
@@ -991,7 +724,11 @@ string t_go_generator::go_imports_end() {
       "var _ = errors.New\n"
       "var _ = context.Background\n"
       "var _ = time.Now\n"
-      "var _ = bytes.Equal\n\n");
+      "var _ = bytes.Equal\n"
+      "// (needed by validator.)\n"
+      "var _ = strings.Contains\n"
+      "var _ = regexp.MatchString\n\n");
+  return import_end;
 }
 
 /**
@@ -1384,6 +1121,14 @@ void t_go_generator::generate_xception(t_struct* txception) {
  */
 void t_go_generator::generate_go_struct(t_struct* tstruct, bool is_exception) {
   generate_go_struct_definition(f_types_, tstruct, is_exception);
+  // generate Validate function
+  std::string tstruct_name(publicize(tstruct->get_name(), false));
+  f_types_ << "func (p *" << tstruct_name << ") Validate() error {" << endl;
+  indent_up();
+  go_validator_generator(this).generate_struct_validator(f_types_, tstruct);
+  f_types_ << indent() << "return nil" << endl;
+  indent_down();
+  f_types_ << "}" << endl;
 }
 
 void t_go_generator::get_publicized_name_and_def_value(t_field* tfield,
@@ -1498,9 +1243,9 @@ void t_go_generator::generate_go_struct_definition(ostream& out,
 
       // Check for user defined tags and them if there are any. User defined tags
       // can override the above db and json tags.
-      std::map<string, string>::iterator it = (*m_iter)->annotations_.find("go.tag");
+      std::map<string, std::vector<string>>::iterator it = (*m_iter)->annotations_.find("go.tag");
       if (it != (*m_iter)->annotations_.end()) {
-        parse_go_tags(&tags, it->second);
+        parse_go_tags(&tags, it->second.back());
       }
 
       string gotag;
--- a/compiler/cpp/src/thrift/generate/t_go_generator.h
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.h
@@ -1,0 +1,329 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef T_GO_GENERATOR_H
+#define T_GO_GENERATOR_H
+
+#include <fstream>
+#include <iostream>
+#include <limits>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "thrift/generate/t_generator.h"
+#include "thrift/platform.h"
+#include "thrift/version.h"
+#include <algorithm>
+#include <clocale>
+#include <sstream>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+using std::map;
+using std::ostream;
+using std::ostringstream;
+using std::string;
+using std::stringstream;
+using std::vector;
+
+static const string endl = "\n"; // avoid ostream << std::endl flushes
+
+const string DEFAULT_THRIFT_IMPORT = "github.com/apache/thrift/lib/go/thrift";
+static std::string package_flag;
+
+/**
+ * Go code generator.
+ */
+class t_go_generator : public t_generator {
+public:
+  t_go_generator(t_program* program,
+                 const std::map<std::string, std::string>& parsed_options,
+                 const std::string& option_string)
+    : t_generator(program) {
+    (void)option_string;
+    std::map<std::string, std::string>::const_iterator iter;
+
+    gen_thrift_import_ = DEFAULT_THRIFT_IMPORT;
+    gen_package_prefix_ = "";
+    package_flag = "";
+    read_write_private_ = false;
+    ignore_initialisms_ = false;
+    skip_remote_ = false;
+    for (iter = parsed_options.begin(); iter != parsed_options.end(); ++iter) {
+      if (iter->first.compare("package_prefix") == 0) {
+        gen_package_prefix_ = (iter->second);
+      } else if (iter->first.compare("thrift_import") == 0) {
+        gen_thrift_import_ = (iter->second);
+      } else if (iter->first.compare("package") == 0) {
+        package_flag = (iter->second);
+      } else if (iter->first.compare("read_write_private") == 0) {
+        read_write_private_ = true;
+      } else if (iter->first.compare("ignore_initialisms") == 0) {
+        ignore_initialisms_ = true;
+      } else if( iter->first.compare("skip_remote") == 0) {
+        skip_remote_ =  true;
+      } else {
+        throw "unknown option go:" + iter->first;
+      }
+    }
+
+    out_dir_base_ = "gen-go";
+  }
+
+  /**
+   * Init and close methods
+   */
+
+  void init_generator() override;
+  void close_generator() override;
+
+  /**
+   * Program-level generation functions
+   */
+
+  void generate_typedef(t_typedef* ttypedef) override;
+  void generate_enum(t_enum* tenum) override;
+  void generate_const(t_const* tconst) override;
+  void generate_struct(t_struct* tstruct) override;
+  void generate_xception(t_struct* txception) override;
+  void generate_service(t_service* tservice) override;
+
+  std::string render_const_value(t_type* type,
+                                 t_const_value* value,
+                                 const string& name,
+                                 bool opt = false);
+
+  /**
+   * Struct generation code
+   */
+
+  void generate_go_struct(t_struct* tstruct, bool is_exception);
+  void generate_go_struct_definition(std::ostream& out,
+                                     t_struct* tstruct,
+                                     bool is_xception = false,
+                                     bool is_result = false,
+                                     bool is_args = false);
+  void generate_go_struct_initializer(std::ostream& out,
+                                      t_struct* tstruct,
+                                      bool is_args_or_result = false);
+  void generate_isset_helpers(std::ostream& out,
+                              t_struct* tstruct,
+                              const string& tstruct_name,
+                              bool is_result = false);
+  void generate_countsetfields_helper(std::ostream& out,
+                                      t_struct* tstruct,
+                                      const string& tstruct_name,
+                                      bool is_result = false);
+  void generate_go_struct_reader(std::ostream& out,
+                                 t_struct* tstruct,
+                                 const string& tstruct_name,
+                                 bool is_result = false);
+  void generate_go_struct_writer(std::ostream& out,
+                                 t_struct* tstruct,
+                                 const string& tstruct_name,
+                                 bool is_result = false,
+                                 bool uses_countsetfields = false);
+  void generate_go_struct_equals(std::ostream& out, t_struct* tstruct, const string& tstruct_name);
+  void generate_go_function_helpers(t_function* tfunction);
+  void get_publicized_name_and_def_value(t_field* tfield,
+                                         string* OUT_pub_name,
+                                         t_const_value** OUT_def_value) const;
+
+  /**
+   * Service-level generation functions
+   */
+
+  void generate_service_helpers(t_service* tservice);
+  void generate_service_interface(t_service* tservice);
+  void generate_service_client(t_service* tservice);
+  void generate_service_remote(t_service* tservice);
+  void generate_service_server(t_service* tservice);
+  void generate_process_function(t_service* tservice, t_function* tfunction);
+
+  /**
+   * Serialization constructs
+   */
+
+  void generate_deserialize_field(std::ostream& out,
+                                  t_field* tfield,
+                                  bool declare,
+                                  std::string prefix = "",
+                                  bool inclass = false,
+                                  bool coerceData = false,
+                                  bool inkey = false,
+                                  bool in_container = false);
+
+  void generate_deserialize_struct(std::ostream& out,
+                                   t_struct* tstruct,
+                                   bool is_pointer_field,
+                                   bool declare,
+                                   std::string prefix = "");
+
+  void generate_deserialize_container(std::ostream& out,
+                                      t_type* ttype,
+                                      bool pointer_field,
+                                      bool declare,
+                                      std::string prefix = "");
+
+  void generate_deserialize_set_element(std::ostream& out,
+                                        t_set* tset,
+                                        bool declare,
+                                        std::string prefix = "");
+
+  void generate_deserialize_map_element(std::ostream& out,
+                                        t_map* tmap,
+                                        bool declare,
+                                        std::string prefix = "");
+
+  void generate_deserialize_list_element(std::ostream& out,
+                                         t_list* tlist,
+                                         bool declare,
+                                         std::string prefix = "");
+
+  void generate_serialize_field(std::ostream& out,
+                                t_field* tfield,
+                                std::string prefix = "",
+                                bool inkey = false);
+
+  void generate_serialize_struct(std::ostream& out, t_struct* tstruct, std::string prefix = "");
+
+  void generate_serialize_container(std::ostream& out,
+                                    t_type* ttype,
+                                    bool pointer_field,
+                                    std::string prefix = "");
+
+  void generate_serialize_map_element(std::ostream& out,
+                                      t_map* tmap,
+                                      std::string kiter,
+                                      std::string viter);
+
+  void generate_serialize_set_element(std::ostream& out, t_set* tmap, std::string iter);
+
+  void generate_serialize_list_element(std::ostream& out, t_list* tlist, std::string iter);
+
+  void generate_go_equals(std::ostream& out, t_type* ttype, string tgt, string src);
+
+  void generate_go_equals_struct(std::ostream& out, t_type* ttype, string tgt, string src);
+
+  void generate_go_equals_container(std::ostream& out, t_type* ttype, string tgt, string src);
+
+  void generate_go_docstring(std::ostream& out, t_struct* tstruct);
+
+  void generate_go_docstring(std::ostream& out, t_function* tfunction);
+
+  void generate_go_docstring(std::ostream& out,
+                             t_doc* tdoc,
+                             t_struct* tstruct,
+                             const char* subheader);
+
+  void generate_go_docstring(std::ostream& out, t_doc* tdoc);
+
+  void parse_go_tags(map<string, string>* tags, const string in);
+
+  /**
+   * Helper rendering functions
+   */
+
+  std::string go_autogen_comment();
+  std::string go_package();
+  std::string go_imports_begin(bool consts);
+  std::string go_imports_end();
+  std::string render_includes(bool consts);
+  std::string render_included_programs(string& unused_protection);
+  std::string render_program_import(const t_program* program, string& unused_protection);
+  std::string render_system_packages(std::vector<string>& system_packages);
+  std::string render_import_protection();
+  std::string render_fastbinary_includes();
+  std::string declare_argument(t_field* tfield);
+  std::string render_field_initial_value(t_field* tfield, const string& name, bool optional_field);
+  std::string type_name(t_type* ttype);
+  std::string module_name(t_type* ttype);
+  std::string function_signature(t_function* tfunction, std::string prefix = "");
+  std::string function_signature_if(t_function* tfunction,
+                                    std::string prefix = "",
+                                    bool addError = false);
+  std::string argument_list(t_struct* tstruct);
+  std::string type_to_enum(t_type* ttype);
+  std::string type_to_go_type(t_type* ttype);
+  std::string type_to_go_type_with_opt(t_type* ttype, bool optional_field);
+  std::string type_to_go_key_type(t_type* ttype);
+  std::string type_to_spec_args(t_type* ttype);
+
+  void indent_up() { t_generator::indent_up(); }
+  void indent_down() { t_generator::indent_down(); }
+  std::string indent() { return t_generator::indent(); }
+  std::ostream& indent(std::ostream& os) { return t_generator::indent(os); }
+
+  static std::string get_real_go_module(const t_program* program) {
+
+    if (!package_flag.empty()) {
+      return package_flag;
+    }
+    std::string real_module = program->get_namespace("go");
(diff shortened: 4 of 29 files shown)


Jens-G:
PS: How do I use it?

BytedanceRPC:
> Re "sparse": It was not about the feature, but about terminology. A sparse array is an array where most values are unused (e.g. NULL/nil), and the most significant problem about it is to store it with a minimum amount of memory, e.g. imagine an 2-sided matrix with 5 mio by 5 mio possible combinations, where only a few thousand are really used. The challenge here is to get it small AND still fast. With that in mind, a map or dictionary is by definition an efficient, sparse data structure with something in the range of O(1) for reads. Hence, "sparse map" makes not much sense, plus what you really mean is sth completely different: The entry is used, only the value happens to be null. An existing null entry and a non-existing entry are two different things.

Thanks for explanation.

BytedanceRPC:
> PS: Is that IsValid() called on reads and writes or only at one of these? Or do I have to invoke it explicitly by calling IsValid()?

For now, go generator only provides 'generate_validator' flag to generate IsValid() method, so we have to call IsValid() to validate the structure date.

BytedanceRPC:
@Jens-G Could you help review the code or assign it to someone else for review?

fishy:
> vt.in: must be in specified values
> vt.not_in: must not be in specified values

it's not clear to me whether that is an enumeration (e.g. `vt.in = "[1, 2, 4, 8]"` or a range (e.g. `vt.in = "[1, 4]"`), and if it's a range, are ends inclusive or exclusive?

I think it would help if you can include the markdown file you attached to the JIRA ticket into this PR so I can review the spec first (and it will be easier to give inline feedback to the spec there), before I review the implementation code.

BytedanceRPC:
> > vt.in: must be in specified values
> > vt.not_in: must not be in specified values
> 
> it's not clear to me whether that is an enumeration (e.g. `vt.in = "[1, 2, 4, 8]"` or a range (e.g. `vt.in = "[1, 4]"`), and if it's a range, are ends inclusive or exclusive?
> 
> I think it would help if you can include the markdown file you attached to the JIRA ticket into this PR so I can review the spec first (and it will be easier to give inline feedback to the spec there), before I review the implementation code.

Sorry for not getting back to you sooner. Value of 'vt.in' must be specified，it works like below
```thrift
3: i8 Byte0 = 1 (vt.in = "0", vt.in = "1", vt.in = "2", vt.not_in = "3", vt.not_in = "4", vt.not_in = "5")
```
We update thrift-parameter-validation-proposal, and put it in [doc/proposal/thrift-parameter-validation-proposal.md](https://github.com/apache/thrift/pull/2469/files?file-filters%5B%5D=.md). Thanks for your advice, and welcome to comment.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,182 @@
...
+```
+- String/Binary
+``` 
+struct StringDemo{
+    1: string Uninitialized (vt.const = "abc")
+    2: string Name (vt.min_size = "6", vt.max_size = "12")
+    3: string SomeStuffs (vt.pattern = "[0-9A-Za-z]+")
+    4: string DebugInfo (vt.prefix = "[Debug]")
+    5: string PanicInfo (vt.contains = "panic")
+    6: string Editor (vt.in = "vscode", vt.in = "vim", vt.in = "goland")
```
fishy, line 118: in the example on field 2 (`vt.min_size = "6", vt.max_size = "12"`), `,` between rules means **and**, otherwise min 6 or max 12 just mean everything and doesn't make any sense.

but in the example on field 6 (`vt.in = "vscode", vt.in = "vim", vt.in = "goland"`), `,` between rules means **or**, if it means "and" then this rule won't match anything.

so as a first thing, we need a consistent way to define the logical relation between multiple rules.

fishy (changes requested):

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,182 @@
...
+- Bool
+```
+struct BoolDemo {
+    1: bool AMD (vt.const = "true")
+}
+```
+- Enum
+```
+struct EnumDemo {
+    1: Type AddressType (vt.const = "String")
```
fishy, line 130: with the above example of `vt.in`, I don't see how `vt.const` is different from `vt.in`? can you explain the differences between them?

simon0-o: "vt.const" and "vt.in" both are used to limit the range of values. The difference is that "vt.const" only specifies one value, "vt.in" multiple values can be specified

doc/proposal/thrift-parameter-validation-proposal.md
simon0-o: The logical relation of all rules is "and", " vt.in " and "vt.not _in" are two very special rules, all " vt.in " will be merged into one rule, so is "vt.not _in"

fishy: This is too subtle to be in a spec. At the very least, it should be explicitly defined/discussed in the spec. But I would rather we have a spec that has clear definitions of and/or logical relations in the rules (e.g. use `and`/`or` or `&&`/`||` explicitly over `,`)

BytedanceRPC: Great suggestion, I updated the proposal and modified the code according to the proposal, please review it again

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,180 @@
...
+| contain      | must contain                     | string                               | string                 | -                  |
+| not_contain  | must not contain                 | string                               | string                 | -                  |
+
+- Secodary validator is a successive validator, usually used at container-type field. See below Set/List/Map examples.
+
+#### 3.2.2 IDL example
+- Number
+```
+struct NumericDemo{
+    1: double Value (validator.gte = "1000.1", validator.lte = "10000.1")
```
fishy, line 105: in the table above they are defined as `le`/`gt`, not `lte`/`gte`.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,180 @@
...
+struct SetListDemo {
+    1: list<string> Persons (vt.min_size = "5", vt.max_size = "10")
+    2: set<double> HealthPoints (vt.elem.gt = "0")
+}
+```
+- Map
+```
+struct MapDemo {
+    1: map<i32, string> IdName (vt.min_size = "5", vt.max_size = "10")
+    2: map<i32, DemoTestRequest> Requests (vt.no_sparse = "true")
```
fishy, line 143: what does `vt.no_sparse` mean? I don't see it defined anywhere in the proposal.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,180 @@
...
+- Bool
+```
+struct BoolDemo {
+    1: bool AMD (vt.const = "true")
+}
+```
+- Enum
+```
+struct EnumDemo {
+    1: Type AddressType (vt.const = "String")
```
fishy, line 128: I guess here in this example `String` is one of the values of enum `Type`? it helps to add that into the example (add the definition of `Type` here).

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,180 @@
...
+    1: list<string> Persons (vt.min_size = "5", vt.max_size = "10")
+    2: set<double> HealthPoints (vt.elem.gt = "0")
+}
+```
+- Map
+```
+struct MapDemo {
+    1: map<i32, string> IdName (vt.min_size = "5", vt.max_size = "10")
+    2: map<i32, DemoTestRequest> Requests (vt.no_sparse = "true")
+    3: map<i32, double> Some, (vt.key.gt = "0", vt.value.lt = "1000")
```
fishy, line 144: this seems to be the only line in all the examples that has the trailing comma. without validator rules thrift is very lenient on trailing comma, so that's fine, but when we actually have trailing comma, it likely makes more sense to have it after the validator rules? so this line should be this instead:
```suggestion
    3: map<i32, double> Some (vt.key.gt = "0", vt.value.lt = "1000"),
```

doc/proposal/thrift-parameter-validation-proposal.md
fishy: also according to the table `const` can only be used on string and bool types, while an enum is technically an int type. I also don't see why `in` and `not_in` can only be used for int types but not string/bool types. all in all I still don't see anything `const` can do that `in` cannot so I'd suggest just use `in` and remove `const`.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,180 @@
...
+    1: double Value (validator.gte = "1000.1", validator.lte = "10000.1")
+    2: i8 Type (validator.in = "[1, 2, 4]")
+}
+```
+- String/Binary
+``` 
+struct StringDemo{
+    1: string Uninitialized (vt.const = "abc")
+    2: string Name (vt.min_size = "6", vt.max_size = "12")
+    3: string SomeStuffs (vt.pattern = "[0-9A-Za-z]+")
```
fishy, line 114: I don't see `pattern` defined in the table. for a regexp there are so many flavors, I'd strongly suggest to either define the flavor here, or just remove it altogether.

fishy (changes requested):
I see that in the new version of the proposal, `,` always means logical *and* and we don't support logical or. That is probably fine (we probably don't really have a need for logical or), but it should be stated explicitly in the proposal.

doc/proposal/thrift-parameter-validation-proposal.md
BytedanceRPC: Considering that it will be implemented in many languages, it is more appropriate to use BRE (basic regular expression) here. I will add it to the table.

doc/proposal/thrift-parameter-validation-proposal.md
BytedanceRPC: `vt.int` and `vt.not` will be implemented in the form of an array, and the value of annotation is a string, which means that we need to escape special characters. In `vt.in`, we use `,` as separators which means we must use `"` or `'` to identify the value of the string, which means that we have to escape again, which will make `vt.in` and `vt.not_in` are very complex. That's why in previous versions we defined different values by defining multiple `vt.in` instead of defining multiple values in one `vt.in`.
To sum up, we think it is inappropriate to use `vt.in` or `vt.not_in` in string type. If there is such a requirement, it can be implemented by `pattern`.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,187 @@
...
+```
+- Validator
+
+    Build-in validating logics. See [Supported Validator](#321-supported-validator) part.
+```apache
+Validator = '.' Identifier
+```
+- ValidatingValue
+```apache
+ValidatingValue  :=  (ToolFunction '(' )? Arguments ‘)’?
```
fishy, line 44: disclaimer: I'm not an expert in BNF nor ABNF. but looking at wikipedia, in ABNF an optional sequence is using `[]` instead of `()?`: https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form#Optional_sequence

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,187 @@
...
+```
+- ToolFunction
+
+    Build-in or user-defined tool functions. See [Tool Function](#325-tool-function) part.
+```apache
+ToolFunction  :=  '@' Identifier
+```
+- Arguments
+```apache
+Arguments  :=  (DynamicValue ListSeparator?)*
```
fishy, line 54: and for this it should be `0*()`: https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form#Variable_repetition

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,187 @@
...
+| min_size     | minimal length                   | i8, i16, i32, i64                    | string, list, set, map | -                  |
+| max_size     | maximal length                   | i8, i16, i32, i64                    | string, list, set, map | -                  |
+| prefix       | field prefix must be             | string                               | string                 | -                  |
+| suffix       | suffix must be                   | string                               | string                 | -                  |
+| contains     | must contain                     | string                               | string                 | -                  |
+| not_contains | must not contain                 | string                               | string                 | -                  |
+| pattern      | basic regular expression         | string                               | string                 | -                  |
+
+- Secodary validator is a successive validator, usually used at container-type field. See below Set/List/Map examples.
```
fishy, line 100: I would prefer to explicitly define `elem` used in examples here as well.

fishy (commented):
I think the proposal itself looks mostly good to me, with a few caveats listed below. I'll review the actual code changes later.

* maybe I missed it but I still did not see "comma means and" explicitly defined in the proposal
* I see you explicitly called out "basic regular expression" in the table, but it's probably still worth adding a note after the table to highlight it (with link to wikipedia, etc.) as people looking at "basic regular expression" might not realize you are talking about the regexp flavor here.
* I also echo @Jens-G 's comment regarding using "vt" as the short version of "validate" is a weird choice, but I don't feel strongly about it.

doc/proposal/thrift-parameter-validation-proposal.md
BytedanceRPC: > 

I add an explicit definition for commas between rules and also a reference for basic regular expression.

BytedanceRPC:
> 

I add an explicit definition for commas between rules and also a reference for basic regular expression.

GuangmingLuo:
@fishy Please take a look at this pr, we are looking forward to any progress related to it.  : )
We, together with @simon0-o , are very glad to contribute to Thrift community continuously.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,189 @@
...
+- Example
+```
+struct FieldReferenceExample {
+    1: string A (vt.eq = "$B") //field A must equal to field B
+    2: string B (vt.gt = "$C[0]") //field B must be greater than field C
+    3: list<string> C 
+}
+```
+
+#### 3.2.5 Tool Function
```
fishy, line 182: I find this section quite ambiguous. for example, if a validation rule is `vt.prefix = "@foo(bar)"`, does that mean the prefix must match a function return, or does that mean the prefix must be "@foo(bar)" literal?

I would rather add a validator rule that's just an "escape hatch" (name tbd, but I just use `escape_hatch` here as an example):
```thrift
struct Foo{
1: i8 A (vt.escape_hatch = "len($A) > len($B)")
2: i8 B
}
```

so basically whatever is in that rule is up to the actual validator implementation to decide.

fishy:
please also resolve the merge conflicts.

it would also be helpful (since the review on the spec/proposal part is mostly done) to paste the compiler generated go code into the pr so it's easier to review the code.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,189 @@
...
+| max_size     | maximal length                   | i8, i16, i32, i64                    | string, list, set, map | -                  |
+| prefix       | field prefix must be             | string                               | string                 | -                  |
+| suffix       | suffix must be                   | string                               | string                 | -                  |
+| contains     | must contain                     | string                               | string                 | -                  |
+| not_contains | must not contain                 | string                               | string                 | -                  |
+| pattern      | basic regular expression         | string                               | string                 | -                  |
+
+- Basic Regular Expression (BRE), the syntax of BRE can be found in [manual](https://www.gnu.org/software/sed/manual/html_node/BRE-syntax.html) of GNU sed
+- Secodary validator (`elem`, `key` and `value`) is a successive validator, usually used at container-type field. See below Set/List/Map examples.
```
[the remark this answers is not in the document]
simon0-o: Good point, I will add `case-sensitive` notice to `prefix`, `suffix`, `contains` and `not_contains`

doc/proposal/thrift-parameter-validation-proposal.md
simon0-o: I think we can add a suffix like `_escape` to validators to indicate that the value behind will use the literal string rather than the function result. eg: `"vt.eq_escape" = "@len(A)"` will match literal `@len(A)` rather than the length of filed A.

doc/proposal/thrift-parameter-validation-proposal.md
```diff
@@ -0,0 +1,189 @@
...
+    2: string B (vt.gt = "$C[0]") //field B must be greater than field C
+    3: list<string> C 
+}
+```
+
+#### 3.2.5 Tool Function
+Tool Function is use to enhance the operating of Validating Value. For example, if we want to ensure one field is larger than the length of string field A, we can use `len()` function: `vd.gt = "@len($A)"`. The arguments can be either literals or variables, and no size limit. However, we won't suggest any build-in function here, because the categary is too big and always language-related. Instead, we only propose one mechanism for thrift developers to extends their implementation according to used language.
+
+### 3.3 Feedback
+The generated validating codes should be included in struct's `IsValid() error` method. If all validate rule declared by one struct get passed, the struct's `IsValid() error` method returns nil; Otherwise it returns false and report Feedback Message indicating failure reason. Due to language function implementations are different, we won't constrain the interface of feedback messages. However, by practice we suggest developers to give below three detail information:
```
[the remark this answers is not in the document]
simon0-o: I think `TApplicationException` is fair enough since `TProtocolException` to me is more likely to cause a communication failure.

GuangmingLuo:
@fishy hi, if you have any progress or new comments related to this pr, please let us know.

fishy:
@GuangmingLuo none of the travis test passed, it doesn't look the compiler code compiles, so maybe fix that first?

also in my previous comment I said that:

>it would also be helpful (since the review on the spec/proposal part is mostly done) to paste the compiler generated go code into the pr so it's easier to review the code.

and I don't see anyone pasting any generated validator go code into the comments.

simon0-o:
Generated Code Examples:
ValidateTest.thrift
```thrift
namespace go validatetest

enum EnumFoo {
  e1
  e2
}

struct Foo {
        1: bool Bool
}

struct BasicTest {
        1: bool Bool0 (vt.const = "true")
        2: optional bool Bool1 (vt.const = "true")
        3: i8 Byte0 = 1 (vt.lt = "2", vt.le = "2", vt.gt = "0", vt.ge = "0", vt.in = "[0, 1, 2]", vt.not_in = "[3, 4, 5]")
        4: optional i8 Byte1 (vt.lt = "1", vt.le = "1", vt.gt = "-1", vt.ge = "-1", vt.in = "[-1, 0, 1]", vt.not_in = "[1, 2, 3]")
        5: double Double0 = 1.0 (vt.lt = "2.0", vt.le = "2.0", vt.gt = "0", vt.ge = "0", vt.in = "[0, 1.0, 2.0]", vt.not_in = "[3.0, 4.0, 5.0]")
        6: optional double Double1 (vt.lt = "2.0", vt.le = "2.0", vt.gt = "0", vt.ge = "0", vt.in = "[0, 1.0, 2.0]", vt.not_in = "[3.0, 4.0, 5.0]")
        7: string String0 = "my const string" (vt.const = "my const string", vt.min_size = "0", vt.max_size = "100", vt.pattern = ".*", vt.prefix = "my", vt.suffix = "string", vt.contains = "const", vt.not_contains = "oh")
        8: optional string String1 (vt.const = "my const string", vt.min_size = "0", vt.max_size = "100", vt.pattern = ".*", vt.prefix = "my", vt.suffix = "string", vt.contains = "const", vt.not_contains = "oh")
        9: binary Binary0 = "my const string" (vt.const = "my const string", vt.min_size = "0", vt.max_size = "100", vt.pattern = ".*", vt.prefix = "my", vt.suffix = "string", vt.contains = "const", vt.not_contains = "oh")
        10: optional binary Binary1 = "my const string" (vt.const = "my const string", vt.min_size = "0", vt.max_size = "100", vt.pattern = ".*", vt.prefix = "my", vt.suffix = "string", vt.contains = "const", vt.not_contains = "oh")
        11: map<string, string> Map0 (vt.min_size = "0", vt.max_size = "10", vt.key.min_size = "0", vt.key.max_size = "10", vt.value.min_size = "0", vt.value.max_size = "10")
        12: optional map<string, string> Map1 (vt.min_size = "0", vt.max_size = "10", vt.key.min_size = "0", vt.key.max_size = "10", vt.value.min_size = "0", vt.value.max_size = "10")
        13: set<string> Set0 (vt.min_size = "0", vt.max_size = "10", vt.elem.min_size = "0")
        14: optional set<string> Set1 (vt.min_size = "0", vt.max_size = "10", vt.elem.min_size = "0")
        15: EnumFoo Enum0 (vt.const = "EnumFoo.e1", vt.defined_only = "true")
        16: optional EnumFoo Enum1 (vt.const = "EnumFoo.e1", vt.defined_only = "true")
        17: map<string, Foo> Map2 (vt.no_sparse = "true")
        18: optional map<string, Foo> Map3 (vt.no_sparse = "true")
        19: Foo Struct0 (vt.skip = "true")
        20: optional Foo Struct1 (vt.skip = "true")
}

struct FieldReferenceTest {
        1: bool Bool0 (vt.const = "$Bool2")
        2: optional bool Bool1 (vt.const = "$Bool2")
        3: i8 Byte0 = 10 (vt.lt = "$Byte4", vt.le = "$Byte4", vt.gt = "$Byte2", vt.ge = "$Byte2", vt.in = "[$Byte2, $Byte3, $Byte4]", vt.not_in = "[$Byte2, $Byte4]")
        4: optional i8 Byte1 (vt.lt = "$Byte4", vt.le = "$Byte4", vt.gt = "$Byte2", vt.ge = "$Byte2", vt.in = "[$Byte2, $Byte3, $Byte4]", vt.not_in = "[$Byte2, $Byte4]")
        5: double Double0 = 10.0 (vt.lt = "$Double4", vt.le = "$Double4", vt.gt = "$Double2", vt.ge = "$Double2", vt.in = "[$Double2, $Double3, $Double4]", vt.not_in = "[$Double2, $Double4]")
        6: optional double Double1 (vt.lt = "$Double4", vt.le = "$Double4", vt.gt = "$Double2", vt.ge = "$Double2", vt.in = "[$Double2, $Double3, $Double4]", vt.not_in = "[$Double2, $Double4]")
        7: string String0 = "my string" (vt.const = "$String2", vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.pattern = "$String4", vt.prefix = "$String2", vt.suffix = "$String2", vt.contains = "$String2", vt.not_contains = "$String3")
        8: optional string String1 (vt.const = "$String2", vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.pattern = "$String4", vt.prefix = "$String2", vt.suffix = "$String2", vt.contains = "$String2", vt.not_contains = "$String3")
        9: binary Binary0 = "my binary" (vt.const = "$Binary2", vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.pattern = "$Binary4", vt.prefix = "$Binary2", vt.suffix = "$Binary2", vt.contains = "$Binary2", vt.not_contains = "$Binary3")
        10: optional binary Binary1 = "my binary" (vt.const = "$Binary2", vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.pattern = "$Binary4", vt.prefix = "$Binary2", vt.suffix = "$Binary2", vt.contains = "$Binary2", vt.not_contains = "$Binary3")
        11: map<string, string> Map0 (vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.key.min_size = "$Byte2", vt.key.max_size = "$Byte3", vt.value.min_size = "$Byte2", vt.value.max_size = "$Byte3")
        12: optional map<string, string> Map1 (vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.key.min_size = "$Byte2", vt.key.max_size = "$Byte3", vt.value.min_size = "$Byte2", vt.value.max_size = "$Byte3")
        13: list<string> List0 (vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.elem.min_size = "$Byte2", vt.elem.max_size = "$Byte3")
        14: optional list<string> List1 (vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.elem.min_size = "$Byte2", vt.elem.max_size = "$Byte3")
        15: set<string> Set0 (vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.elem.min_size = "$Byte2", vt.elem.max_size = "$Byte3")
        16: optional set<string> Set1 (vt.min_size = "$Byte2", vt.max_size = "$Byte3", vt.elem.min_size = "$Byte2", vt.elem.max_size = "$Byte3")
        17: bool Bool2 = false
        18: i8 Byte2 = 0
        19: i8 Byte3 = 10
        20: i8 Byte4 = 20
        21: double Double2 = 0
        22: double Double3 = 10.0
        23: double Double4 = 20.0
        24: string String2 = "my string"
        25: string String3 = "other string"
        26: string String4 = ".*"
        27: binary Binary2 = "my binary"
        28: binary Binary3 = "other binary"
        29: binary Binary4 = ".*"
}

struct ValidationFunctionTest {
        1: string StringFoo
        2: i64 StringLength (vt.in = "[@len($StringFoo)]")
}
```
Related Generated Code:
```go
func (p *Foo) IsValid() error {
	return nil
}

func (p *BasicTest) IsValid() error {
	if p.Bool0 != false {
		return errors.New("p.Bool0 not valid, rule vt.const check failed")
	}
	if p.Bool1 != nil {
		if *p.Bool1 != false {
			return errors.New("*p.Bool1 not valid, rule vt.const check failed")
		}
	}
	if p.Byte0 >= 2 {
		return errors.New("p.Byte0 not valid, rule vt.lt check failed")
	}
	if p.Byte0 > 2 {
		return errors.New("p.Byte0 not valid, rule vt.le check failed")
	}
	if p.Byte0 <= 0 {
		return errors.New("p.Byte0 not valid, rule vt.gt check failed")
	}
	if p.Byte0 < 0 {
		return errors.New("p.Byte0 not valid, rule vt.ge check failed")
	}
	var _exist0 bool
	_src0 := []int8{0, 1, 2}
	for _, src := range _src0 {
		if p.Byte0 == src {
			_exist0 = true
		}
	}
	if _exist0 == false {
		return errors.New("p.Byte0 not valid, rule vt.in check failed")
	}
	_src1 := []int8{3, 4, 5}
	for _, src := range _src1 {
		if p.Byte0 == src {
			{
				return errors.New("p.Byte0 not valid, rule vt.not_in check failed")
			}
		}
	}
	if p.Byte1 != nil {
		if *p.Byte1 >= 1 {
			return errors.New("*p.Byte1 not valid, rule vt.lt check failed")
		}
		if *p.Byte1 > 1 {
			return errors.New("*p.Byte1 not valid, rule vt.le check failed")
		}
		if *p.Byte1 <= -1 {
			return errors.New("*p.Byte1 not valid, rule vt.gt check failed")
		}
		if *p.Byte1 < -1 {
			return errors.New("*p.Byte1 not valid, rule vt.ge check failed")
		}
		var _exist1 bool
		_src2 := []int8{-1, 0, 1}
		for _, src := range _src2 {
			if *p.Byte1 == src {
				_exist1 = true
			}
		}
		if _exist1 == false {
			return errors.New("*p.Byte1 not valid, rule vt.in check failed")
		}
		_src3 := []int8{1, 2, 3}
		for _, src := range _src3 {
			if *p.Byte1 == src {
				{
					return errors.New("*p.Byte1 not valid, rule vt.not_in check failed")
				}
			}
		}
	}
	if p.Double0 >= 2 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	if p.Double0 > 2 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	if p.Double0 <= 0 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	if p.Double0 < 0 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	var _exist2 bool
	_src4 := []float64{0, 1, 2}
	for _, src := range _src4 {
		if p.Double0 == src {
			_exist2 = true
		}
	}
	if _exist2 == false {
		return errors.New("p.Double0 not valid, rule vt.in check failed")
	}
	_src5 := []float64{3, 4, 5}
	for _, src := range _src5 {
		if p.Double0 == src {
			return errors.New("p.Double0 not valid, rule vt.not_in check failed")
		}
	}
	if p.Double1 != nil {
		if *p.Double1 >= 2 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		if *p.Double1 > 2 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		if *p.Double1 <= 0 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		if *p.Double1 < 0 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		var _exist3 bool
		_src6 := []float64{0, 1, 2}
		for _, src := range _src6 {
			if *p.Double1 == src {
				_exist3 = true
			}
		}
		if _exist3 == false {
			return errors.New("*p.Double1 not valid, rule vt.in check failed")
		}
		_src7 := []float64{3, 4, 5}
		for _, src := range _src7 {
			if *p.Double1 == src {
				return errors.New("*p.Double1 not valid, rule vt.not_in check failed")
			}
		}
	}
	if p.String0 != "my const string" {
		return errors.New("p.String0 not valid, rule vt.const check failed")
	}
	if len(p.String0) < int(0) {
		return errors.New("p.String0 not valid, rule vt.min_size check failed")
	}
	if len(p.String0) > int(100) {
		return errors.New("p.String0 not valid, rule vt.max_size check failed")
	}
	if ok, _ := regexp.MatchString(p.String0, ".*"); ok {
		return errors.New("p.String0 not valid, rule vt.pattern check failed")
	}
	if !strings.HasPrefix(p.String0, "my") {
		return errors.New("p.String0 not valid, rule vt.prefix check failed")
	}
	if !strings.HasSuffix(p.String0, "string") {
		return errors.New("p.String0 not valid, rule vt.suffix check failed")
	}
	if !strings.Contains(p.String0, "const") {
		return errors.New("p.String0 not valid, rule vt.contains check failed")
	}
	if strings.Contains(p.String0, "oh") {
		return errors.New("p.String0 not valid, rule vt.not_contains check failed")
	}
	if p.String1 != nil {
		if *p.String1 != "my const string" {
			return errors.New("*p.String1 not valid, rule vt.const check failed")
		}
		if len(*p.String1) < int(0) {
			return errors.New("*p.String1 not valid, rule vt.min_size check failed")
		}
		if len(*p.String1) > int(100) {
			return errors.New("*p.String1 not valid, rule vt.max_size check failed")
		}
		if ok, _ := regexp.MatchString(*p.String1, ".*"); ok {
			return errors.New("*p.String1 not valid, rule vt.pattern check failed")
		}
		if !strings.HasPrefix(*p.String1, "my") {
			return errors.New("*p.String1 not valid, rule vt.prefix check failed")
		}
		if !strings.HasSuffix(*p.String1, "string") {
			return errors.New("*p.String1 not valid, rule vt.suffix check failed")
		}
		if !strings.Contains(*p.String1, "const") {
			return errors.New("*p.String1 not valid, rule vt.contains check failed")
		}
		if strings.Contains(*p.String1, "oh") {
			return errors.New("*p.String1 not valid, rule vt.not_contains check failed")
		}
	}
	_tgt0 := string(p.Binary0)
	if _tgt0 != "my const string" {
		return errors.New("p.Binary0 not valid, rule vt.const check failed")
	}
	if len(_tgt0) < int(0) {
		return errors.New("p.Binary0 not valid, rule vt.min_size check failed")
	}
	if len(_tgt0) > int(100) {
		return errors.New("p.Binary0 not valid, rule vt.max_size check failed")
	}
	if ok, _ := regexp.MatchString(_tgt0, ".*"); ok {
		return errors.New("p.Binary0 not valid, rule vt.pattern check failed")
	}
	if !strings.HasPrefix(_tgt0, "my") {
		return errors.New("p.Binary0 not valid, rule vt.prefix check failed")
	}
	if !strings.HasSuffix(_tgt0, "string") {
		return errors.New("p.Binary0 not valid, rule vt.suffix check failed")
	}
	if !strings.Contains(_tgt0, "const") {
		return errors.New("p.Binary0 not valid, rule vt.contains check failed")
	}
	if strings.Contains(_tgt0, "oh") {
		return errors.New("p.Binary0 not valid, rule vt.not_contains check failed")
	}
	_tgt1 := string(p.Binary1)
	if _tgt1 != "my const string" {
		return errors.New("p.Binary1 not valid, rule vt.const check failed")
	}
	if len(_tgt1) < int(0) {
		return errors.New("p.Binary1 not valid, rule vt.min_size check failed")
	}
	if len(_tgt1) > int(100) {
		return errors.New("p.Binary1 not valid, rule vt.max_size check failed")
	}
	if ok, _ := regexp.MatchString(_tgt1, ".*"); ok {
		return errors.New("p.Binary1 not valid, rule vt.pattern check failed")
	}
	if !strings.HasPrefix(_tgt1, "my") {
		return errors.New("p.Binary1 not valid, rule vt.prefix check failed")
	}
	if !strings.HasSuffix(_tgt1, "string") {
		return errors.New("p.Binary1 not valid, rule vt.suffix check failed")
	}
	if !strings.Contains(_tgt1, "const") {
		return errors.New("p.Binary1 not valid, rule vt.contains check failed")
	}
	if strings.Contains(_tgt1, "oh") {
		return errors.New("p.Binary1 not valid, rule vt.not_contains check failed")
	}
	if len(p.Map0) < 0 {
		return errors.New("p.Map0 not valid, rule vt.min_size check failed")
	}
	if len(p.Map0) > 10 {
		return errors.New("p.Map0 not valid, rule vt.max_size check failed")
	}
	if len(p.Map1) < 0 {
		return errors.New("p.Map1 not valid, rule vt.min_size check failed")
	}
	if len(p.Map1) > 10 {
		return errors.New("p.Map1 not valid, rule vt.max_size check failed")
	}
	if len(p.Set0) < 0 {
		return errors.New("p.Set0 not valid, rule vt.min_size check failed")
	}
	if len(p.Set0) > 10 {
		return errors.New("p.Set0 not valid, rule vt.max_size check failed")
	}
	if len(p.Set1) < 0 {
		return errors.New("p.Set1 not valid, rule vt.min_size check failed")
	}
	if len(p.Set1) > 10 {
		return errors.New("p.Set1 not valid, rule vt.max_size check failed")
	}
	if p.Enum0 != 0 {
		return errors.New("p.Enum0 not valid, rule vt.const check failed")
	}
	if p.Enum1 != nil {
		if *p.Enum1 != *EnumFooPtr(0) {
			return errors.New("*p.Enum1 not valid, rule vt.const check failed")
		}
	}
	if err := p.Struct0.IsValid(); err != nil {
		return err
	}
	if err := p.Struct1.IsValid(); err != nil {
		return err
	}
	return nil
}

func (p *FieldReferenceTest) IsValid() error {
	if p.Bool0 != p.Bool2 {
		return errors.New("p.Bool0 not valid, rule vt.const check failed")
	}
	if p.Bool1 != nil {
		if *p.Bool1 != p.Bool2 {
			return errors.New("*p.Bool1 not valid, rule vt.const check failed")
		}
	}
	if p.Byte0 >= p.Byte4 {
		return errors.New("p.Byte0 not valid, rule vt.lt check failed")
	}
	if p.Byte0 > p.Byte4 {
		return errors.New("p.Byte0 not valid, rule vt.le check failed")
	}
	if p.Byte0 <= p.Byte2 {
		return errors.New("p.Byte0 not valid, rule vt.gt check failed")
	}
	if p.Byte0 < p.Byte2 {
		return errors.New("p.Byte0 not valid, rule vt.ge check failed")
	}
	var _exist0 bool
	_src0 := []int8{p.Byte2, p.Byte3, p.Byte4}
	for _, src := range _src0 {
		if p.Byte0 == src {
			_exist0 = true
		}
	}
	if _exist0 == false {
		return errors.New("p.Byte0 not valid, rule vt.in check failed")
	}
	_src1 := []int8{p.Byte2, p.Byte4}
	for _, src := range _src1 {
		if p.Byte0 == src {
			{
				return errors.New("p.Byte0 not valid, rule vt.not_in check failed")
			}
		}
	}
	if p.Byte1 != nil {
		if *p.Byte1 >= p.Byte4 {
			return errors.New("*p.Byte1 not valid, rule vt.lt check failed")
		}
		if *p.Byte1 > p.Byte4 {
			return errors.New("*p.Byte1 not valid, rule vt.le check failed")
		}
		if *p.Byte1 <= p.Byte2 {
			return errors.New("*p.Byte1 not valid, rule vt.gt check failed")
		}
		if *p.Byte1 < p.Byte2 {
			return errors.New("*p.Byte1 not valid, rule vt.ge check failed")
		}
		var _exist1 bool
		_src2 := []int8{p.Byte2, p.Byte3, p.Byte4}
		for _, src := range _src2 {
			if *p.Byte1 == src {
				_exist1 = true
			}
		}
		if _exist1 == false {
			return errors.New("*p.Byte1 not valid, rule vt.in check failed")
		}
		_src3 := []int8{p.Byte2, p.Byte4}
		for _, src := range _src3 {
			if *p.Byte1 == src {
				{
					return errors.New("*p.Byte1 not valid, rule vt.not_in check failed")
				}
			}
		}
	}
	if p.Double0 >= p.Double4 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	if p.Double0 > p.Double4 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	if p.Double0 <= p.Double2 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	if p.Double0 < p.Double2 {
		return errors.New("p.Double0 not valid, rule vt.const check failed")
	}
	var _exist2 bool
	_src4 := []float64{p.Double2, p.Double3, p.Double4}
	for _, src := range _src4 {
		if p.Double0 == src {
			_exist2 = true
		}
	}
	if _exist2 == false {
		return errors.New("p.Double0 not valid, rule vt.in check failed")
	}
	_src5 := []float64{p.Double2, p.Double4}
	for _, src := range _src5 {
		if p.Double0 == src {
			return errors.New("p.Double0 not valid, rule vt.not_in check failed")
		}
	}
	if p.Double1 != nil {
		if *p.Double1 >= p.Double4 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		if *p.Double1 > p.Double4 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		if *p.Double1 <= p.Double2 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		if *p.Double1 < p.Double2 {
			return errors.New("*p.Double1 not valid, rule vt.const check failed")
		}
		var _exist3 bool
		_src6 := []float64{p.Double2, p.Double3, p.Double4}
		for _, src := range _src6 {
			if *p.Double1 == src {
				_exist3 = true
			}
		}
		if _exist3 == false {
			return errors.New("*p.Double1 not valid, rule vt.in check failed")
		}
		_src7 := []float64{p.Double2, p.Double4}
		for _, src := range _src7 {
			if *p.Double1 == src {
				return errors.New("*p.Double1 not valid, rule vt.not_in check failed")
			}
		}
	}
	if p.String0 != string(p.String2) {
		return errors.New("p.String0 not valid, rule vt.const check failed")
	}
	if len(p.String0) < int(p.Byte2) {
		return errors.New("p.String0 not valid, rule vt.min_size check failed")
	}
	if len(p.String0) > int(p.Byte3) {
		return errors.New("p.String0 not valid, rule vt.max_size check failed")
	}
	if ok, _ := regexp.MatchString(p.String0, string(p.String4)); ok {
		return errors.New("p.String0 not valid, rule vt.pattern check failed")
	}
	if !strings.HasPrefix(p.String0, string(p.String2)) {
		return errors.New("p.String0 not valid, rule vt.prefix check failed")
	}
	if !strings.HasSuffix(p.String0, string(p.String2)) {
		return errors.New("p.String0 not valid, rule vt.suffix check failed")
	}
	if !strings.Contains(p.String0, string(p.String2)) {
		return errors.New("p.String0 not valid, rule vt.contains check failed")
	}
	if strings.Contains(p.String0, string(p.String3)) {
		return errors.New("p.String0 not valid, rule vt.not_contains check failed")
	}
	if p.String1 != nil {
		if *p.String1 != string(p.String2) {
			return errors.New("*p.String1 not valid, rule vt.const check failed")
		}
		if len(*p.String1) < int(p.Byte2) {
			return errors.New("*p.String1 not valid, rule vt.min_size check failed")
		}
		if len(*p.String1) > int(p.Byte3) {
			return errors.New("*p.String1 not valid, rule vt.max_size check failed")
		}
		if ok, _ := regexp.MatchString(*p.String1, string(p.String4)); ok {
			return errors.New("*p.String1 not valid, rule vt.pattern check failed")
		}
		if !strings.HasPrefix(*p.String1, string(p.String2)) {
			return errors.New("*p.String1 not valid, rule vt.prefix check failed")
		}
		if !strings.HasSuffix(*p.String1, string(p.String2)) {
			return errors.New("*p.String1 not valid, rule vt.suffix check failed")
		}
		if !strings.Contains(*p.String1, string(p.String2)) {
			return errors.New("*p.String1 not valid, rule vt.contains check failed")
		}
		if strings.Contains(*p.String1, string(p.String3)) {
			return errors.New("*p.String1 not valid, rule vt.not_contains check failed")
		}
	}
	_tgt0 := string(p.Binary0)
	if _tgt0 != string(p.Binary2) {
		return errors.New("p.Binary0 not valid, rule vt.const check failed")
	}
	if len(_tgt0) < int(p.Byte2) {
		return errors.New("p.Binary0 not valid, rule vt.min_size check failed")
	}
	if len(_tgt0) > int(p.Byte3) {
		return errors.New("p.Binary0 not valid, rule vt.max_size check failed")
	}
	if ok, _ := regexp.MatchString(_tgt0, string(p.Binary4)); ok {
		return errors.New("p.Binary0 not valid, rule vt.pattern check failed")
	}
	if !strings.HasPrefix(_tgt0, string(p.Binary2)) {
		return errors.New("p.Binary0 not valid, rule vt.prefix check failed")
	}
	if !strings.HasSuffix(_tgt0, string(p.Binary2)) {
		return errors.New("p.Binary0 not valid, rule vt.suffix check failed")
	}
	if !strings.Contains(_tgt0, string(p.Binary2)) {
		return errors.New("p.Binary0 not valid, rule vt.contains check failed")
	}
	if strings.Contains(_tgt0, string(p.Binary3)) {
		return errors.New("p.Binary0 not valid, rule vt.not_contains check failed")
	}
	_tgt1 := string(p.Binary1)
	if _tgt1 != string(p.Binary2) {
		return errors.New("p.Binary1 not valid, rule vt.const check failed")
	}
	if len(_tgt1) < int(p.Byte2) {
		return errors.New("p.Binary1 not valid, rule vt.min_size check failed")
	}
	if len(_tgt1) > int(p.Byte3) {
		return errors.New("p.Binary1 not valid, rule vt.max_size check failed")
	}
	if ok, _ := regexp.MatchString(_tgt1, string(p.Binary4)); ok {
		return errors.New("p.Binary1 not valid, rule vt.pattern check failed")
	}
	if !strings.HasPrefix(_tgt1, string(p.Binary2)) {
		return errors.New("p.Binary1 not valid, rule vt.prefix check failed")
	}
	if !strings.HasSuffix(_tgt1, string(p.Binary2)) {
		return errors.New("p.Binary1 not valid, rule vt.suffix check failed")
	}
	if !strings.Contains(_tgt1, string(p.Binary2)) {
		return errors.New("p.Binary1 not valid, rule vt.contains check failed")
	}
	if strings.Contains(_tgt1, string(p.Binary3)) {
		return errors.New("p.Binary1 not valid, rule vt.not_contains check failed")
	}
	if len(p.Map0) < int(p.Byte2) {
		return errors.New("p.Map0 not valid, rule vt.min_size check failed")
	}
	if len(p.Map0) > int(p.Byte3) {
		return errors.New("p.Map0 not valid, rule vt.max_size check failed")
	}
	if len(p.Map1) < int(p.Byte2) {
		return errors.New("p.Map1 not valid, rule vt.min_size check failed")
	}
	if len(p.Map1) > int(p.Byte3) {
		return errors.New("p.Map1 not valid, rule vt.max_size check failed")
	}
	if len(p.List0) < int(p.Byte2) {
		return errors.New("p.List0 not valid, rule vt.min_size check failed")
	}
	if len(p.List0) > int(p.Byte3) {
		return errors.New("p.List0 not valid, rule vt.max_size check failed")
	}
	if len(p.List1) < int(p.Byte2) {
		return errors.New("p.List1 not valid, rule vt.min_size check failed")
	}
	if len(p.List1) > int(p.Byte3) {
		return errors.New("p.List1 not valid, rule vt.max_size check failed")
	}
	if len(p.Set0) < int(p.Byte2) {
		return errors.New("p.Set0 not valid, rule vt.min_size check failed")
	}
	if len(p.Set0) > int(p.Byte3) {
		return errors.New("p.Set0 not valid, rule vt.max_size check failed")
	}
	if len(p.Set1) < int(p.Byte2) {
		return errors.New("p.Set1 not valid, rule vt.min_size check failed")
	}
	if len(p.Set1) > int(p.Byte3) {
		return errors.New("p.Set1 not valid, rule vt.max_size check failed")
	}
	return nil
}

func (p *ValidationFunctionTest) IsValid() error {
	var _exist0 bool
	_src0 := []int64{int64(len(p.StringFoo))}
	for _, src := range _src0 {
		if p.StringLength == src {
			_exist0 = true
		}
	}
	if _exist0 == false {
		return errors.New("p.StringLength not valid, rule vt.in check failed")
	}
	return nil
}
```

simon0-o:
> @GuangmingLuo none of the travis test passed, it doesn't look the compiler code compiles, so maybe fix that first?
> 
> also in my previous comment I said that:
> 
> > it would also be helpful (since the review on the spec/proposal part is mostly done) to paste the compiler generated go code into the pr so it's easier to review the code.
> 
> and I don't see anyone pasting any generated validator go code into the comments.

@fishy Sorry for missing attaching the generated code example. I have pasted the ValidateTest.thrift and related generated code above. And I have fixed the cmake compile error and rebased the master branch, please take another look.

fishy:
https://app.travis-ci.com/github/apache/thrift/jobs/578599367 is a new failure introduced by this PR

fishy:
Feedback on the compiler generated go code:
1. I thought we agreed on https://github.com/apache/thrift/pull/2469#discussion_r858950006 to use `TApplicationException` for validation errors, but the example is still using `errors.New`
2. In error messages the `p.` prefix doesn't really bring any extra value, it would be more useful to use the type name instead of `p` (example: "BasicTest.Byte0 not valid, ...")
3. (minor) For `vt.in` check you can break the for loop early after first find
4. (minor) For `vt.pattern` check `regexp.MatchString` calls would be a performance nightmare because it needs to compile the pattern on every check. It's probably OK for the first version but we should consider declare global variables to compile the patterns defined in thrift files.
5. Is `IsValid` called by `Read`/`Write` or does it need to be called explicitly (e.g. manual check only)?

simon0-o:
> Feedback on the compiler generated go code:
> 
> 1. I thought we agreed on [THRIFT-5423: Support go parameter validation in IDL #2469 (comment)](https://github.com/apache/thrift/pull/2469#discussion_r858950006) to use `TApplicationException` for validation errors, but the example is still using `errors.New`
> 2. In error messages the `p.` prefix doesn't really bring any extra value, it would be more useful to use the type name instead of `p` (example: "BasicTest.Byte0 not valid, ...")
> 3. (minor) For `vt.in` check you can break the for loop early after first find
> 4. (minor) For `vt.pattern` check `regexp.MatchString` calls would be a performance nightmare because it needs to compile the pattern on every check. It's probably OK for the first version but we should consider declare global variables to compile the patterns defined in thrift files.
> 5. Is `IsValid` called by `Read`/`Write` or does it need to be called explicitly (e.g. manual check only)?

Thanks for your quick reply. I have fixed the runtime error that leads to travis ci fail.
I have added a new exception type `VALIDATOR_CHECK_FAILED` to application exceptions, and use it as an exception when a validation rule check fails.
For 2 && 3, thanks for your suggestion, now we will use `{struct name}.{field name}` in validator exception and validator will break early in `vt.in`.
About 4, I will try to optimize this part, or we can do this in the later version.
About 5, in my view it should be called explicitly. If it is called by `Read`/`Write`, it could cause serialization /deserialization to fail.

fishy:
>it could cause serialization /deserialization to fail.

I thought that's the point :)

But I think it's fine to make it manual checks only first then iterate on top of that later. But I think the function name `IsValid` implies that it would return a `bool` but it actually returns an `error`. `Validate() error` is probably a better name.

simon0-o:
> `Validate() error` is probably a better name.

Sounds a better practice, fix done.

compiler/cpp/src/thrift/generate/t_go_generator.h
```diff
@@ -0,0 +1,328 @@
...
+      } else if (iter->first.compare("package") == 0) {
+        package_flag = (iter->second);
+      } else if (iter->first.compare("read_write_private") == 0) {
+        read_write_private_ = true;
+      } else if (iter->first.compare("ignore_initialisms") == 0) {
+        ignore_initialisms_ = true;
+      } else if( iter->first.compare("skip_remote") == 0) {
+        skip_remote_ =  true;
+      } else if (iter->first.compare("generate_validator") == 0) {
+        generate_validator = true;
```
fishy, line 85: so this is actually opt-in, e.g. users need to add `generate_validator` to the thrift compiler command line to enable this feature, right? I don't see this documented anywhere (my impression from the proposal is that this would be opt-out instead)

also if we have this as a command line arg then I'm not sure why would we add it as an option to CMake, on CMake level it should always be enabled or disabled together with the go compiler itself?

lib/go/thrift/application_exception.go
```diff
@@ -49,6 +50,7 @@ var defaultApplicationExceptionMessage = map[int32]string{
 	INVALID_TRANSFORM:              "Invalid transform",
 	INVALID_PROTOCOL:               "Invalid protocol",
 	UNSUPPORTED_CLIENT_TYPE:        "Unsupported client type",
+	VALIDATOR_CHECK_FAILED:         "Validator check failed",
```
fishy, line 53: "validator check failed" sounds weird to me, maybe it should be "validator failed" or "validation failed"?

lib/go/test/tests/validate_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"testing"
+
+	"github.com/apache/thrift/lib/go/test/gopath/src/validatetest"
+)
+
+func TestValidator(t *testing.T) {
+	bt := validatetest.NewBasicTest()
+	if err := bt.Validate(); err != nil {
+		t.Fatal(err)
```
fishy, line 31: 1. this should be `t.Error` not `t.Fatal`
2. what this tests is that all zero/default values of the thrift struct passes the validation, which both requires the validation definition written in the thrift file to allow all zero/default values, and also is not very useful. we need to actually fill some data that would fail the validator and test that `Validate` does indeed return an error, and also valid data passes them.

compiler/cpp/src/thrift/generate/t_json_generator.cc
```diff
@@ -268,7 +268,9 @@ void t_json_generator::write_type_spec(t_type* ttype) {
     write_key_and("annotations");
     start_object();
     for (auto & annotation : ttype->annotations_) {
-      write_key_and_string(annotation.first, annotation.second);
+      for (auto& annotation_value : annotation.second) {
+        write_key_and_string(annotation.first, annotation_value);
+      }
```
fishy, line 273: we should also add a test to make sure that the validator annotations can co-exist with other annotations.

compiler/cpp/CMakeLists.txt
```diff
@@ -71,35 +73,50 @@ macro(THRIFT_ADD_COMPILER name description initial)
...
 
+# This macro adds an option THRIFT_VALIDATOR_COMPILER_${NAME}
+# that allows enabling or disabling certain languages
+macro(THRIFT_ADD_VALIDATOR_COMPILER name description initial)
+    string(TOUPPER "THRIFT_VALIDATOR_COMPILER_${name}" enabler)
+    set(src "src/thrift/generate/${name}_validator_generator.cc")
+    list(APPEND "src/thrift/generate/${name}_validator_generator.h")
+    option(${enabler} ${description} ${initial})
+    if(${enabler})
+        list(APPEND thrift-compiler_SOURCES ${src})
```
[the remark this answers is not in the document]
simon0-o: > it seems to me that if someone tries to disable `THRIFT_VALIDATOR_COMPILER_go` it will also disable `THRIFT_COMPILER_go`

Actually, it won't enable/disable `THRIFT_COMPILER_go`, `enabler` is a local variable, and also the `src`. `THRIFT_ADD_VALIDATOR_COMPILER` will check if `initial` is `ON`, if it is `THRIFT_ADD_VALIDATOR_COMPILER` will add "src/thrift/generate/${name}_validator_generator.cc" and "src/thrift/generate/${name}_validator_generator.h" to compiler source files.

compiler/cpp/src/thrift/generate/t_go_generator.h
simon0-o: Seems `generate_validator` option is not appropriate, I will remove it. And also I removed `THRIFT_VALIDATOR_COMPILER_go` from CMakeLists.txt, now `THRIFT_COMPILER_go` will also enable/disable go validator generator.

lib/go/test/tests/validate_test.go
simon0-o: fixed and add more test

compiler/cpp/src/thrift/generate/t_json_generator.cc
simon0-o: I added a `AnnotationCompatiableTest`, will this do?
```thrift
struct AnnotationCompatiableTest {
        1: bool Bool0 = true (vt.const = "true", go.tag = 'json:"bool0"')
        2: i8 Byte0 = 1 (vt.lt = "2", go.tag = 'json:"byte0"')
        3: double Double0 = 1.0 (vt.lt = "2.0", go.tag = 'json:"double0"')
        4: string String0 = "my const string" (vt.const = "my const string", go.tag = 'json:"string0"')
        5: binary Binary0 = "my const string" (vt.const = "my const string", go.tag = 'json:"binary0"')
        6: map<string, string> Map0 (vt.min_size = "0", go.tag = 'json:"map0"')
        7: set<string> Set0 (vt.min_size = "0", go.tag = 'json:"set0"')
        8: EnumFoo Enum0 = EnumFoo.e2 (vt.in = "[EnumFoo.e2]", go.tag = 'json:"enum0"')
        9: Foo Struct0 (vt.skip = "true", go.tag = 'json:"struct0"')
}
```

simon0-o:
@fishy There is a `concurrency_test` in AppVeyor, but I can't reproduce it locally. Is that an occasional case or does it fails due to my changes? And all fixes mentioned above have been done, please take another look

lib/go/test/tests/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/apache/thrift/lib/go/test/gopath/src/validatetest"
+)
+
+func TestBasicValidator(t *testing.T) {
+	bt := validatetest.NewBasicTest()
+	if err := bt.Validate(); err != nil {
+		t.Error(err)
+	}
+	bt = validatetest.NewBasicTest()
+	bt.Bool1 = &(&struct{ x bool }{x: false}).x
```
fishy, line 36: you can use the new generic version of `thrift.Pointer(false)` or the old `thrift.BoolPtr(false)`, no need to make such an effort to get a pointer to `false`.

lib/go/test/tests/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := bt.Validate(); err == nil {
+		t.Error("Expected vt.const error for Bool1")
+	} else if !strings.Contains(err.Error(), "vt.const") {
+		t.Errorf("Expected vt.const error for Bool1, but got %s\n", err.Error())
```
fishy, line 40: don't use `%s` with `err.Error()`, use `%v` with `err` instead.

fishy (changes requested):
@Jens-G and @dcelasun : could you please take a look? the code is mostly good to me now (there are still a few issues with unit tests).

Things to note for review:
* We are adding a new `VALIDATION_FAILED` to `TApplicationException`
* The proposal itself

dcelasun:
The code is generally OK, minor issues aside. 

> We are adding a new `VALIDATION_FAILED` to `TApplicationException`

This should be fine, I can't think of a realistic scenario where it'd be a problem.

More generally, I'm not sure about "tool functions". It can easily cause further drift between supported languages. I think we should define and document which functions we support (`len()` is fine). We can add more in future versions if necessary.

simon0-o:
> The code is generally OK, minor issues aside.
> 
> > We are adding a new `VALIDATION_FAILED` to `TApplicationException`
> 
> This should be fine, I can't think of a realistic scenario where it'd be a problem.
> 
> More generally, I'm not sure about "tool functions". It can easily cause further drift between supported languages. I think we should define and document which functions we support (`len()` is fine). We can add more in future versions if necessary.

I add a supported functions table to the validator proposal, which can be modified when more functions or language supports are available.

dcelasun:
> I add a supported functions table to the validator proposal, which can be modified when more functions or language supports are available.

Looks good. Before we merge, let's move the doc from `doc/proposal` to `docs/specs`. You should also add a line to `CHANGES.md` for 0.17.

simon0-o:
> > I add a supported functions table to the validator proposal, which can be modified when more functions or language supports are available.
> 
> Looks good. Before we merge, let's move the doc from `doc/proposal` to `docs/specs`. You should also add a line to `CHANGES.md` for 0.17.

I also rebased the master branch, if commits need to be squashed please let me know.

dcelasun:
Yes, please squash them. Commit message can be `THRIFT-5423: IDL parameter validation for Go`.

lib/go/test/ValidateTest.thrift
```diff
@@ -0,0 +1,103 @@
...
+        30: i64 MaxSize = 10
+}
+
+struct ValidationFunctionTest {
+        1: string StringFoo
+        2: i64 StringLength (vt.in = "[@len($StringFoo)]")
+}
+
+struct AnnotationCompatibleTest {
+        1: bool Bool0 = true (vt.const = "true", go.tag = 'json:"bool0"')
```
fishy, line 94: hmm I thought I left a comment here but maybe I forgot to submit it.

so here `go.tag = 'json:"bool0"'`  is the same as the default json tag the compiler would generate, so even if there's a bug that `go.tag` was not applied correctly, we won't be able to catch the bug.

a meaningful test would be that we specify a non-default json tag here, and in `_test.go` we construct a json blob using the non-default json tag, and verify that we can still correctly deserialize the json blob into the struct (which verifies that the `go.tag` part is handled correctly), and also verify the validator.

lib/go/test/tests/validate_test.go
```diff
@@ -0,0 +1,202 @@
...
+	bt := validatetest.NewBasicTest()
+	if err := bt.Validate(); err != nil {
+		t.Error(err)
+	}
+	bt = validatetest.NewBasicTest()
+	bt.Bool1 = thrift.BoolPtr(false)
+	if err := bt.Validate(); err == nil {
+		t.Error("Expected vt.const error for Bool1")
+	} else if !strings.Contains(err.Error(), "vt.const") {
+		t.Errorf("Expected vt.const error for Bool1, but got %v\n", err)
```
fishy, line 41: there's really no reason to add `\n` to the end of `t.Errorf`.

lib/go/test/ValidateTest.thrift
simon0-o: I think the default json tag of field `Bool0` is `json:"Bool0"`, but specified value is `json:"bool0"` starting with lower case 'b'. It might look a little bit confusing.
I change it to `json:"bool1"`, and also others, and add more test for `AnnotationCompatibleTest`.

lib/go/test/tests/validate_test.go
simon0-o: fix done.

fishy:
Another suggestion, which can be done in a follow-up PR instead, but doing it in this PR will help a lot of the test code:

Define a `ValidationError` type in go library, and implement `Unwrap` on `tApplicationException` so that when a `TApplicationException` is validation error it can also unwrap to that, that way when checking for `ValidationError` you can use its fields to determine what's the check (e.g. `vt.const`) and what's the field that failed validation.

simon0-o:
> Another suggestion, which can be done in a follow-up PR instead, but doing it in this PR will help a lot of the test code:
> 
> Define a `ValidationError` type in go library, and implement `Unwrap` on `tApplicationException` so that when a `TApplicationException` is validation error it can also unwrap to that, that way when checking for `ValidationError` you can use its fields to determine what's the check (e.g. `vt.const`) and what's the field that failed validation.

sorry for the late reply, I have defined a `ValidationError` in application_exception.go and add a `NewValidationException` which is used to construct tApplicationExcetion with ValidationError. I also add a unwrap test in the validation test to make sure `Unwrap` works correctly.

lib/go/thrift/application_exception.go
```diff
@@ -59,9 +61,119 @@ type TApplicationException interface {
...
+		return
+	}
+	err = oprot.WriteStructEnd(ctx)
+	return
+}
+
+type tApplicationException struct {
+	message         string
+	type_           int32
+	validationError *ValidationError
```
fishy, line 176: this should not be specified to `validationError *ValidationError`, but just a generic `err error`. see `TTransportException` for an example: https://github.com/apache/thrift/blob/e0ee2c7514b812b5a18bfe9aab9594c5879dc34f/lib/go/thrift/transport_exception.go#L48

lib/go/thrift/application_exception.go
```diff
@@ -77,8 +189,23 @@ func (e tApplicationException) Error() string {
 	return defaultApplicationExceptionMessage[e.type_]
 }
 
+func (e tApplicationException) Unwrap() error {
+	if e.validationError != nil {
+		return e.validationError
+	}
+	return e
```
fishy, line 196: no you can't make `Unwrap` to return self. that will cause infinite loops. this function should simple be `return e.err`.

lib/go/thrift/application_exception.go
```diff
@@ -124,6 +252,17 @@ func (p *tApplicationException) Read(ctx context.Context, iprot TProtocol) error
 					return err
 				}
 			}
+		case 3:
```
fishy, line 255: It's likely a bad idea to write/read the additional `ValidationError` details over the wire (the implementation of `Read` and `Write` of `ValidationError`). the details should be purely local additional information for now. to be able to actually communicate that over the wire, we need to expand the spec to define the wire protocol for it.

lib/go/test/tests/validate_test.go
```diff
@@ -0,0 +1,304 @@
...
+func TestBasicValidator(t *testing.T) {
+	bt := validatetest.NewBasicTest()
+	if err := bt.Validate(); err != nil {
+		t.Error(err)
+	}
+	bt = validatetest.NewBasicTest()
+	bt.Bool1 = thrift.BoolPtr(false)
+	if err := bt.Validate(); err == nil {
+		t.Error("Expected vt.const error for Bool1")
+	} else if !strings.Contains(err.Error(), "vt.const") {
```
fishy, line 42: the whole point of adding `ValidationError` is to use something better than `strings.Contains` for unit tests. what we want here is something like this:

```go
} else {
  var ve *thrift.ValidationError
  if errors.As(err, &ve) {
    if ve.Check() != "vt.const" {
      t.Errorf(...)
    }
    if ve.Field() != "bool1" {
      t.Errorf(...)
    }
    ....
  } else {
    t.Errorf("Error cannot be unwrapped into *ValidationError: %v", err)
  }
}
```

fishy (changes requested):

simon0-o:
@fishy All fixes done, PTAL

compiler/cpp/src/thrift/generate/go_validator_generator.h
```diff
@@ -0,0 +1,72 @@
...
+#include <string>
+#include <vector>
+
+class go_validator_generator {
+public:
+  go_validator_generator(t_go_generator* gg) : go_generator(gg){};
+  void generate_struct_validator(std::ostream& out, t_struct* tstruct);
+
+  struct generator_context {
+    std::string field_symbol;
```
fishy, line 38: we probably should store the original field name instead of field symbol.

for example, for thrift struct of:
```thrift
struct foo {
  1: bool bool0 = true (vt.const = "true")
}
```

the field name itself would be `"bool0"` but the go symbol would be `"Bool0"`.

lib/go/thrift/application_exception.go
```diff
@@ -59,9 +62,35 @@ type TApplicationException interface {
...
+	check       string
+	fieldSymbol string
+}
+
+func (e *ValidationError) Check() string {
+	return e.check
+}
+
+func (e *ValidationError) Field() string {
+	if fs := strings.Split(e.fieldSymbol, "."); len(fs) > 1 {
```
fishy, line 76: why is this `Split` needed? in which case would we use `.` to construct field names/symbols?

lib/go/thrift/application_exception.go
```diff
@@ -77,8 +106,20 @@ func (e tApplicationException) Error() string {
...
 func NewTApplicationException(type_ int32, message string) TApplicationException {
-	return &tApplicationException{message, type_}
+	return &tApplicationException{message, type_, nil}
+}
+
+func NewValidationException(type_ int32, check string, field string, message string) TApplicationException {
+	return &tApplicationException{
+		type_:   type_,
+		message: message,
+		err:     &ValidationError{message, check, field},
```
fishy, line 121: nit: avoid using unkeyed struct constructors.

fishy (approved):
@dcelasun @Jens-G would you mind taking another look?

lib/go/thrift/application_exception.go
simon0-o: `fieldSymbol` indicates a string like 'BasicTest.Double1', which means that the field is the 'Double1' in type 'BasicTest'. And similarly 'BasicTest.Map0.key' indicates that the field is the key of field 'Map0' in type 'BasicTest'.
I also add a new method 'TypeName' to ValidationError to get the name of the type.

simon0-o:
@fishy I have run `cppcheck --force --quiet --inline-suppr --suppress="*:thrift/thriftl.cc" --error-exitcode=1 -j2 compiler/cpp/src` locally, but it didn't show that error. If there are still errors like that, please let me know.

lib/go/thrift/application_exception.go
fishy: why not just return the whole string, `BasicTest.Map0.key`? now we have 2 separated APIs to return `BasicTest` and `Map0`, what about the `key` part? this makes things unnecessarily complicated.

or just export the fields and remove the getters. we don't really need getters here anyways, if someone really want to mutate the error returned, they are on their own risks.

fishy (approved):

dcelasun (approved):

CHANGES.md
```diff
@@ -9,6 +9,7 @@
 - [THRIFT-5527](https://issues.apache.org/jira/browse/THRIFT-5527) - Compiler generated Process function will swallow exceptions defined in thrift IDL
 - [THRIFT-5605](https://issues.apache.org/jira/browse/THRIFT-5605) - Provide `ExtractIDLExceptionClientMiddleware` and `ExtractExceptionFromResult` to help client middlewares to gain access to exceptions defined in thrift IDL
 - [THRIFT-5609](https://issues.apache.org/jira/browse/THRIFT-5609) - `TJSONProtocol` is now safe to be used in `TDeserializePool`
+- [THRIFT-5423](https://issues.apache.org/jira/browse/THRIFT-5423) - Specify thrift parameter validation proposal and provide parameter validator for go language.
```
fishy, line 12: this is actually no longer going to make the 0.17.0 release and will be in 0.18.0 instead. @Jens-G is the recommendation to create the 0.18.0 section or just remove this line?

Jens-G: Usually not necessary, since the changes.md file will be [assembled during release process](https://github.com/apache/thrift/blob/master/doc/ReleaseManagement.md#release-candidate). 
But what you might want to include beforehand is [breaking changes](https://github.com/apache/thrift/blob/master/CHANGES.md#breaking-changes).

simon0-o: I have just removed this line since there are no breaking changes from the user side. Or should I create a 0.18.0 section and add this line to breaking changes

fishy:
Merged in as d5927a960.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -1384,6 +1121,14 @@ void t_go_generator::generate_xception(t_struct* txception) {
  */
 void t_go_generator::generate_go_struct(t_struct* tstruct, bool is_exception) {
   generate_go_struct_definition(f_types_, tstruct, is_exception);
+  // generate Validate function
+  std::string tstruct_name(publicize(tstruct->get_name(), false));
+  f_types_ << "func (p *" << tstruct_name << ") Validate() error {" << endl;
+  indent_up();
+  go_validator_generator(this).generate_struct_validator(f_types_, tstruct);
```
Jens-G, line 1128: I know I'm late to the party, but this line causes LNK2019 in Visual Studio 2022. Any idea why? I am staring at this for quite a while now and just can't find it. 

```
Code wird generiert...
Debug\go_validator_generator.obj : warning LNK4042: Objekt mehrmals angegeben; zusätzliche Objekte werden ignoriert.
t_go_generator.obj : error LNK2019: Verweis auf nicht aufgelöstes externes Symbol ""public: void __thiscall go_validator_generator::generate_struct_validator(class std::basic_ostream<char,struct std::char_traits<char> > &,class t_struct *)" (?generate_struct_validator@go_validator_generator@@QAEXAAV?$basic_ostream@DU?$char_traits@D@std@@@std@@PAVt_struct@@@Z)" in Funktion ""public: void __thiscall t_go_generator::generate_go_struct(class t_struct *,bool)" (?generate_go_struct@t_go_generator@@QAEXPAVt_struct@@_N@Z)".
thrift.exe : fatal error LNK1120: 1 nicht aufgelöste Externe

```

PR #2474, opened 2021-10-28
THRIFT-5474: TSimpleJsonProtocol with inner quotes does not produce valid JSON.

thealmightygrant:
There is an issue in the generation of JSON from the TSimpleJSONProtocol, where if there are quotes inside of a string, then those quotes will not be serialized correctly. These quotes need one more layer of escaping to produce valid JSON.

I have also added the Jackson library to these tests to ensure that any future changes to this library produce valid JSON. 


emmenlau:
Thanks for the effort, @thealmightygrant ! Its always nice to see people going to the effort of making a pull request.

PR #2471, opened 2021-10-21
Compile TWebSocketServer with CMake only when OpenSSL is found

kainjow:
When compiling the C++ library and OpenSSL isn't found, TWebSocketServer will fail as it includes OpenSSL headers.

PR #2476, opened 2021-11-02
moved websocket cpp server to openssl sections as it depends on it

techbech:
I've moved `TWebSocketServer` into the OpenSSL section of the CMake, as the web socket server depends on OpenSSL. This is currently breaking the optional building with OpenSSL. So right now even tho one has disabled OpenSSL, it still uses  `bio` and `evp` from OpenSSL.


kainjow:
Same as https://github.com/apache/thrift/pull/2471 ?

PR #2477, opened 2021-11-08
rust code performance

tsxiaofang:
Why is the performance of the generated rust code particularly low? When host a accesses host B, but when client and server are on the same host, there is no problem.


Jens-G:
> tsxiaofang wants to merge 70 commits into 0.15.0 from master

Most certainly not. This must be a mistake.

Jens-G:
> Why is the performance of the generated rust code particularly low? When host a accesses host B, but when client and server are on the same host, there is no problem.

Recommend to ask that question on the [mailing lists](https://thrift.apache.org/mailing). If you can provide a test case demonstrating the issue you experience, that would be even better.

PR #2478, opened 2021-11-13
THRIFT-5480

io-eft:
THRIFT-5480
netstd
TFramedTransport throws a transport exception when a message is bigger than the previous one, because the knownMessageSize is not reset after a read.
Added a call to update (reset) message size after a frame is read.


Jens-G:
+1 reasons outlined in the ticket

PR #2482, opened 2021-12-01
THRIFT-5482: Fix memory leak during SSL handshake in C++ library

anshulmgupta:
Client: C++
We observed continuous memory increase in the thrift server during SSL handshake and this memory increase was crashing the windows hosts as there was no memory left for other applications to work. During further investigation, we found that ERR_remove_state has been made stub in the later versions of OpenSSL, and recommended function to clear the thread state is OPENSSL_thread_stop. Once I made this change we did not observe any memory leak further.


emmenlau (approved):
Looks convincing. Should we merge, and further test this on master?

emmenlau:
Builds and tests fine for me (our internal tests, not standard thrift tests) against OpenSSL 1.1.1l.

anshulmgupta:
@emmenlau If the diff looks good to you, can you please merge the changes? Thanks!

PR #2486, opened 2021-12-11
Include readme in python packaged sources

alisaifee:
Since the README.md file is now being used for `long_description` (as per #2458) it needs to be included in package files.


alisaifee:
Sorry this isn't necessary - the problem only occurs when the `setup.py` uses the fallback to distutils

PR #2487, opened 2021-12-12
THRIFT-5488: Define PY_SSIZE_T_CLEAN to fix error with using PyObject_CallFunction in python 3.10

alisaifee:
[THRIFT-5488](https://issues.apache.org/jira/browse/THRIFT-5488)
## Description
When using the refill  python functions for the fallback scenario in `ProtocolBase<Impl>::readBytes` ([protocol.tcc](../blob/master/lib/py/src/ext/protocol.tcc#L279)), a `SystemError` is raised in python 3.10 (this was only a warning till python 3.9) due to the use of `#yi` for the output buffer argument. The removal of the warning and dropped support is described in [Python issue 40943](https://bugs.python.org/issue40943).

Extra test cases have been added to cover serialization/deserialization both with compact & binary protocol both with and without the c-extension

~~Additionally the environment variable `PYTHONWARNINGS=error` has been added to the configuration of the python cross tests so as to verify this fix, and also catch any deprecations early.~~ (This change was reverted - see #2490 for details)
 
## Testing
- Run the added unit tests and verify they pass for both python 3.9/3.10:
  ```
   cd lib/py
   make && python test/thrift_TSerializer.py
   ```
- Comment out the `PY_SSIZE_T_CLEAN` define in `lib/py/src/ext/binary.cpp` & `lib/py/src/ext/compact.cpp` 
- Re-build and re-run the tests with python 3.9 - verify that the tests pass, but a warning is displayed
- Re-build and re-rerun the tests with python 3.10 - verify that the tests: `test_TBinaryProtocolAccelerated` & `test_TCompactProtocolAccelerated` fail.

Additionally `make cross` can also be used to exercise the failure by settings warnings to errors by setting the environment variable `PYTHONWARNINGS=error`

- [x] Did you create an [Apache Thrift](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)

### DIFF
--- a/lib/py/test/test_thrift_file/TestServer.thrift
+++ b/lib/py/test/test_thrift_file/TestServer.thrift
@@ -16,7 +16,11 @@
 # specific language governing permissions and limitations
 # under the License.
 #
- 
+
+struct Message {
+   1: optional string body,
+   2: optional i64 num,
+}
 
 service TestServer{
 	string add_and_get_msg(1:string msg)
--- a/lib/py/Makefile.am
+++ b/lib/py/Makefile.am
@@ -30,6 +30,7 @@ py3-test: py3-build
 	$(PYTHON3) test/thrift_TZlibTransport.py
 	$(PYTHON3) test/thrift_TCompactProtocol.py
 	$(PYTHON3) test/thrift_TNonblockingServer.py
+	$(PYTHON3) test/thrift_TSerializer.py
 else
 py3-build:
 py3-test:
@@ -55,6 +56,7 @@ check-local: all py3-test
 	$(PYTHON) test/thrift_TZlibTransport.py
 	$(PYTHON) test/thrift_TCompactProtocol.py
 	$(PYTHON) test/thrift_TNonblockingServer.py
+	$(PYTHON) test/thrift_TSerializer.py
 
 
 clean-local:
--- a/lib/py/src/ext/binary.cpp
+++ b/lib/py/src/ext/binary.cpp
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+#define PY_SSIZE_T_CLEAN
 #include "ext/binary.h"
 namespace apache {
 namespace thrift {
--- a/lib/py/src/ext/compact.cpp
+++ b/lib/py/src/ext/compact.cpp
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+#define PY_SSIZE_T_CLEAN
 #include "ext/compact.h"
 
 namespace apache {
--- a/lib/py/test/thrift_TSerializer.py
+++ b/lib/py/test/thrift_TSerializer.py
@@ -1,0 +1,84 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+import unittest
+import os
+import sys
+
+gen_path = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "gen-py"
+)
+sys.path.append(gen_path)
+
+import _import_local_thrift  # noqa
+from thrift.protocol.TBinaryProtocol import TBinaryProtocolFactory
+from thrift.protocol.TBinaryProtocol import TBinaryProtocolAcceleratedFactory
+from thrift.protocol.TCompactProtocol import TCompactProtocolFactory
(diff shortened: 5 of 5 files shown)


lib/py/test/test_thrift_file/TestServer.thrift
```diff
@@ -16,8 +16,12 @@
...
- 
+
+struct Message {
+   1: optional string body,
+   2: optional i64 num,
+}
 
 service TestServer{
-	string add_and_get_msg(1:string msg)
+	Message add_and_get_msg(1:Message msg)
```
fishy, line 26: I don't think this change is explained in either the PR description of the ticket?

fishy (commented):
Thanks for the PR!

I wanted to suggest adding python 3.10 dockerfile(s) used by travis to make sure we test it in CI, but then looking at the existing dockerfiles we are only using python3 package from ubuntu versions, and I don't think any of the ubuntu versions has 3.10 yet :(

But in the linked python issue, I do see this:

>I prefer to raise a DeprecationWarning exception at runtime since this warning can be made an error when using -Werror: it eases detection of deprecated modules without preventing to build or use a C extension using the deprecated functions.

I wonder whether we can add `-Werror` in the CI scripts to force failing that in 3.8/3.9 to verify the fix, but I also don't know whether that's the only python warning we have (probably not 😅).

lib/py/test/test_thrift_file/TestServer.thrift
alisaifee: My mistake, this shouldn't have made it to the PR as it is not relevant to this issue.

alisaifee: Removed in b2672adfea84051c38b7d124f4599b544d11ebbd

alisaifee:
I realized I hadn't added any details about using the unit tests to verify the behavior before and after the patch - added now. 

Regarding the CI -  the build is failing at the moment due to a rust dependency issue (`clap`: `use of unstable library feature 'matches_macro'`) which I also get locally. However, I didn't see the where the python unit tests would be run. Adding `-Werror` does sound like a good direction though!

alisaifee:
@fishy I just verified that after getting the cross tests to run after patching the rust version issue in #2488 the issue can be reproduced by a failure in the cross tests for py3-threader_unframed_binary & compact. Using `PYTHONWARNINGS=error make cross` with python 3.9 also reproduces the same failures without this patch (albeit also for py2) - but is cleared when the patch is applied.

lib/py/test/test_thrift_file/TestServer.thrift
```diff
@@ -16,7 +16,11 @@
 # specific language governing permissions and limitations
 # under the License.
 #
- 
+
+struct Message {
+   1: optional string body,
+   2: optional i64 num,
+}
```
fishy, line 23: this change should also be reverted?

alisaifee: This struct is used in the test cases here: https://github.com/apache/thrift/pull/2487/files#diff-1364ef1b9a20b93088778203e7f99d70ecb498ba6736e11466979edc229f4600R36

alisaifee:
Are there any specific changes I should make to this patch?
Since the cross tests are now running on master (still one pending erlang related failure), I could also figure out how to add the `-Werror` or `PYTHONWARNINGS=error` in that CI step?

fishy (approved):
> Are there any specific changes I should make to this patch? Since the cross tests are now running on master (still one pending erlang related failure), I could also figure out how to add the `-Werror` or `PYTHONWARNINGS=error` in that CI step?

If we can add `-Werror` into CI that would be great.

If that's not possible (for example, this is not the only warning we are hitting), then the current state of the PR is good to be merged.

alisaifee:
Turns out adding environment variables for specific targets for the cross tests was already built in :tada:. I've created a PR against master with just the `PYTHONWARNINGS` set to `error` to compare against this branch at #2490

alisaifee:
Unfortunately the error couldn't be reproduced on CI since the python version (3.6.9) doesn't actually contain the warning. Regardless, I think having warnings as errors should be beneficial - especially once the CI pythons are updated.

fishy:
> Unfortunately the error couldn't be reproduced on CI since the python version (3.6.9) doesn't actually contain the warning. Regardless, I think having warnings as errors should be beneficial - especially once the CI pythons are updated.

In that case, I'd prefer to revert the CI changes. I don't want the CI to start fail suddenly when Ubuntu decided to upgrade python3 in their version from 3.6 to 3.7.

We _might_ want to consider the CI change in a separated PR so it can be reverted, but that still sounds quite risky to me, as when that upgrade in ubuntu actually happens no one would remember this is the commit to be reverted.

Sorry about the back-and-forth.

alisaifee:
> In that case, I'd prefer to revert the CI changes. I don't want the CI to start fail suddenly when Ubuntu decided to upgrade python3 in their version from 3.6 to 3.7.

That's a very good point that I hadn't considered at all. I've reverted that commit (force pushed) so that there's a clean single commit for you to merge.

> Sorry about the back-and-forth.

No worries at all! Thank you for the review.

fishy:
@alisaifee Can you rebase your branch on top of the latest master (with your CI fix for rust merged), so that the CI output of this PR would be more meaningful?

alisaifee:
> @alisaifee Can you rebase your branch on top of the latest master (with your CI fix for rust merged), so that the CI output of this PR would be more meaningful?

I already did that! Sadly there's still one erlang test case failing which I couldn't figure out. The difference between the CI failure now versus before the rust fix, is that the tests actually run now (previously they were failing in the test compilation stage)

fishy (approved):

PR #2490, opened 2021-12-16
Ensure python warnings are treated as errors

alisaifee:
## Description

Adds environment variable `PYTHONWARNINGS=error` for python cross tests to catch deprecation warnings that could lead to incompatibility in future versions.
  
:warning: This PR is created only to demonstrate the failure on master - but can be closed once #2487 is merged (which contains the change in the cross test config)

### DIFF
--- a/test/tests.json
+++ b/test/tests.json
@@ -248,7 +248,10 @@
         "TestServer.py",
         "--verbose",
         "--genpydir=gen-py"
-      ]
+      ],
+      "env": {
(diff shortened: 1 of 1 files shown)


alisaifee:
Unfortunately this doesn't demonstrate any warnings (specifically the one about `PY_SSIZE_T_CLEAN` since the warning was only introduced in python 3.8 in python/cpython#12473 - and the CI uses python 3.6.9

PR #2491, opened 2021-12-16
THRIFT-5467 Python: fix CannotSendHeader exception

csordasmarton:
Based on the python source for `http.client`, `HTTPConnection.putheader`
can only be called after a request has been started, and before it's
been sent. Otherwise it will throw a `http.client.CannotSendHeader`
exception.

If the server returns a `Set-Cookie` header, the client will always
fail with the `CannotSendHeader` exception because `HTTPConnection.putheader`
is called after reading the response.

With this patch we will call this method before the request has been sent.

### DIFF
--- a/lib/py/src/transport/THttpClient.py
+++ b/lib/py/src/transport/THttpClient.py
@@ -91,6 +91,7 @@ class THttpClient(TTransportBase):
         self.__http_response = None
         self.__timeout = None
         self.__custom_headers = None
+        self.headers = None
 
     @staticmethod
     def basic_proxy_auth_header(proxy):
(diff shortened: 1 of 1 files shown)


csordasmarton:
Is there anything I can do to get this merged?

csordasmarton:
@Jens-G Done. I also rebased my branch.

rajatm91:
I see in the Jira that this issue is fixed as part of v0.17.0. What are the timelines for to release of this version?

trams:
@jeking3 is another maintainer. @jeking3 , can you help us to release 0.17 to PyPi? thank you

PR #2492, opened 2021-12-16
THRIFT-5490: Use pooled buffer for THeaderTransport

fishy:
Client: go

Instead of binding 2 buffers (read/write) to each THeaderTransport, grab
one from the pool to be used for the whole read/write, and return it
back to the pool after the read/write is done. This would help reduce
the memory footprint from idle connections.

  

### DIFF
--- a/lib/go/thrift/header_transport.go
+++ b/lib/go/thrift/header_transport.go
@@ -28,7 +28,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 )
 
 // Size in bytes for 32-bit ints.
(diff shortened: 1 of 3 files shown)


fishy:
Note: similar things can be done in `TFrameTransport` (and maybe a few others), but I plan to do that in a follow up PR(s).

lib/go/thrift/header_transport.go
```diff
@@ -628,24 +633,25 @@ func (t *THeaderTransport) Flush(ctx context.Context) error {
...
-		writer, err := NewTransformWriter(&payload, t.writeTransforms)
+		writer, err := NewTransformWriter(payload, t.writeTransforms)
 		if err != nil {
 			return NewTTransportExceptionFromError(err)
 		}
-		if _, err := io.Copy(writer, &t.writeBuffer); err != nil {
+		if _, err := io.Copy(writer, t.writeBuffer); err != nil {
 			return NewTTransportExceptionFromError(err)
 		}
 		if err := writer.Close(); err != nil {
```
marcoferrer, line 657: Curious. If we return an error on line 655 would that mean the `writer` is never closed? Reason why I ask is because at first glance, `NewTransformWriter` could potentially allocate a new zlib writer (line 221) which would never get closed. But I could just be missing something in my understand of how this works

fishy: it's a local variable, so it will be garbage collected later after it's no longer reachable by any code.

dcelasun (approved):
LGTM

fishy:
Added a unit test to make sure that a `THeaderTransport` can be reused. Will merge this in after some staging test is completed.

fishy:
not seeing issues on staging. merging.

PR #2493, opened 2021-12-17
THRIFT-5491 : add benchmark for partial deserialization

bhalchandrap:
Adds a JMH based benchmark that shows cpu and memory efficiency of partial Thrift deserialization

### DIFF
--- a/lib/java/test/org/apache/thrift/test/DeserializationBenchmark.java
+++ b/lib/java/test/org/apache/thrift/test/DeserializationBenchmark.java
@@ -1,0 +1,165 @@
+package org.apache.thrift.test;
+
+import org.apache.thrift.Fixtures;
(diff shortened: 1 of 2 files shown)


lib/java/test/org/apache/thrift/test/DeserializationBenchmark.java
```diff
@@ -0,0 +1,165 @@
+package org.apache.thrift.test;
```
Jens-G, line 1: ASF license header missing

bhalchandrap: sorry for the oversight. added license header comment.

Jens-G:
rebase needed, no avctivity since 2022

PR #2496, opened 2021-12-31
THRIFT-5492: Add readEnd to TBufferedTransport

slicking:
client: cpp

Add a readEnd implementation to TBufferedTransport which resets the consumed
message size.  Without this it was noticed that calls to the transport's
consume method would slowly decrement the remainingMessageSize_ and the
remainingMessageSize_ was never reset.  A client could send many messages to
the server, eventually bringing remainingMessageSize_ below zero causing an
EOD_OF_FILE exception.

### DIFF
--- a/lib/cpp/src/thrift/transport/TBufferTransports.h
+++ b/lib/cpp/src/thrift/transport/TBufferTransports.h
@@ -270,6 +270,11 @@ public:
    */
   uint32_t readAll(uint8_t* buf, uint32_t len) { return TBufferBase::readAll(buf, len); }
 
+  uint32_t readEnd() override {
+    resetConsumedMessageSize();
+    return 0;
+  }
+
 protected:
   void initPointers() {
     setReadBuffer(rBuf_.get(), 0);
--- a/test/StressTest.thrift
+++ b/test/StressTest.thrift
@@ -31,5 +31,6 @@ service Service {
   list<i8>  echoList(1: list<i8> arg),
   set<i8>  echoSet(1: set<i8> arg),
   map<i8, i8>  echoMap(1: map<i8, i8> arg),
+  binary echoBinary(1: binary arg),
 }
 
--- a/test/cpp/src/StressTest.cpp
+++ b/test/cpp/src/StressTest.cpp
@@ -95,6 +95,9 @@ public:
   void echoList(vector<int8_t>& out, const vector<int8_t>& arg) override { out = arg; }
   void echoSet(set<int8_t>& out, const set<int8_t>& arg) override { out = arg; }
   void echoMap(map<int8_t, int8_t>& out, const map<int8_t, int8_t>& arg) override { out = arg; }
+  void echoBinary(string& out, const string& arg) override {
+    out = arg;
+  }
 
 private:
   count_map counts_;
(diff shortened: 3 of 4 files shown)


Jens-G:
<html>
<body>
<p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(36, 41, 47); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">This stage<span> </span><strong style="box-sizing: border-box; font-weight: 600;">failed</strong>.</p>

Job | Xcode | ENV | OS | State
-- | -- | -- | -- | --
7919.3 |   | SCRIPT="cmake.sh" | Linux | failed
7919.4 |   | SCRIPT="cmake.sh" | Linux | failed
7919.5 |   | SCRIPT="cmake.sh" | Linux | failed
7919.6 |   | SCRIPT="cmake.sh" | Linux | failed
7919.7 |   | SCRIPT="cmake.sh" | Linux | failed
7919.8 |   | SCRIPT="cmake.sh" | Linux | failed
7919.9 |   | SCRIPT="cmake.sh" | Linux | failed
7919.10 |   | SCRIPT="cmake.sh" | Linux | failed
7919.11 |   | SCRIPT="cmake.sh" | Linux | passed
7919.12 |   | SCRIPT="cmake.sh" | Linux | failed
7919.13 |   | SCRIPT="cmake.sh" | Linux | passed
7919.14 | xcode11.3 | SCRIPT="cmake.sh" | macOS | passed


</body>
</html>

slicking:
@Jens-G thanks for pointing out the failures.  Looking at several of them it appears to be because I added to test/StessTest.thrift but only updated test/cpp/src/StessTest.cpp and NOT lib/d/test/stress_test_server.d so there is an unimplemented function.
The fix I made was only under the c++ server; let me see if adding a test under lib/cpp/test/ would be better contained and cover only the c++ server.

goorjn:
Do you think Message Size check is implemented the right way ?
Is this not a protocol concern, rather than a distributed protocol / transport responsibility ?
The functionality is implemented through transport base class function definition. Don't you think it would be better to move these to a different object which responsibility would be to check message size validity ?
Several functions in the implementation are never called (in production code, I haven't checked the tests), code is hard to read and some paths are never reached (resetConsumedMessageSize(long newSize = -1) with actual newSize value).
I do not know enough about the whole Virtual protocol / transport implementation to propose relevant changes to the code, but the current one (I use version 0.14.3) broke my program !

jsiewkow:
I've moved from 0.11.0 to 0.20.0 and encountered the issue described here, fix in this PR seems to solve it, is there a plan for this change to be merged?

Jens-G:
I merged the fix part in 86b05bf, but left the rest, as there are unfortunately still open questions etc.

jsiewkow:
Thank you, I guess this will be in next thrift release, when do you plan it?

emmenlau:
Is the remainder of this PR still valid? Anyone able to comment on the remaining issues?

slicking:
Thanks @Jens-G for merging the fix; I'm glad it could be of help to a few others.
I've not had time to get back to this issue and update the test; with the fix merged though I think this PR can be closed.  The changes to update the test in this PR need to be reworked to be specific to the c++ server and, if anyone is able, can be taken up in a separate PR.

PR #2497, opened 2022-01-04
THRIFT-5495: close client when shutdown server in go lib

buptubuntu:
Client: [go]

If there is client connection and no data is send，we will encounter hang druing server stop:
1>If transport factory conf with socket timeout,we will hang until the deadline of the socket

2>If transport factory conf without socket timeout,we will hang forever

Stack As below:
goroutine 140800 [IO wait, 2706 minutes]:
internal/poll.runtime_pollWait(0x7fbf804fb100, 0x72)
  runtime/netpoll.go:234 +0x89
internal/poll.(*pollDesc).wait(0xc009087700, 0xc008196000, 0x0)
  internal/poll/fd_poll_runtime.go:84 +0x32
internal/poll.(*pollDesc).waitRead(...)
  internal/poll/fd_poll_runtime.go:89
internal/poll.(*FD).Read(0xc009087700, {0xc008196000, 0x10000, 0x10000})
  internal/poll/fd_unix.go:167 +0x25a
net.(*netFD).Read(0xc009087700, {0xc008196000, 0x0, 0xc0061089b8})
  net/fd_posix.go:56 +0x29
net.(*conn).Read(0xc007c98038, {0xc008196000, 0x0, 0xc006108978})
  net/net.go:183 +0x45
github.com/apache/thrift/lib/go/thrift.(*socketConn).Read(0x246aae0, {0xc008196000, 0xc0058b4ed0, 0x246aae0})
  github.com/apache/thrift@v0.15.0/lib/go/thrift/socket_conn.go:101 +0x44
github.com/apache/thrift/lib/go/thrift.(*TSocket).Read(0xc003555460, {0xc008196000, 0x10000, 0x10000})
  github.com/apache/thrift@v0.15.0/lib/go/thrift/socket.go:221 +0x67
bufio.(*Reader).Read(0xc005657320, {0xc001da1000, 0x1000, 0x203000})
  bufio/bufio.go:227 +0x1b4
github.com/apache/thrift/lib/go/thrift.(*TBufferedTransport).Read(0xc0035554a0, {0xc001da1000, 0x431e10, 0x64})
  github.com/apache/thrift@v0.15.0/lib/go/thrift/buffered_transport.go:67 +0x45
bufio.(*Reader).Read(0xc005657380, {0xc0090877f0, 0x4, 0x4b5bac0})
  bufio/bufio.go:227 +0x1b4
io.ReadAtLeast({0x30c0520, 0xc005657380}, {0xc0090877f0, 0x4, 0x4}, 0x4)
  io/io.go:328 +0x9a
io.ReadFull(...)
  io/io.go:347
github.com/apache/thrift/lib/go/thrift.(*TFramedTransport).readFrame(0xc009087780)
  github.com/apache/thrift@v0.15.0/lib/go/thrift/framed_transport.go:199 +0x3c
github.com/apache/thrift/lib/go/thrift.(*TFramedTransport).Read(0xc009087780, {0xc0090877f0, 0x1, 0x4})
  github.com/apache/thrift@v0.15.0/lib/go/thrift/framed_transport.go:148 +0x130
github.com/apache/thrift/lib/go/thrift.(*TFramedTransport).ReadByte(0xc009087780)
  github.com/apache/thrift@v0.15.0/lib/go/thrift/framed_transport.go:157 +0x2e
github.com/apache/thrift/lib/go/thrift.(*TCompactProtocol).readByteDirect(...)
  github.com/apache/thrift@v0.15.0/lib/go/thrift/compact_protocol.go:766
github.com/apache/thrift/lib/go/thrift.(*TCompactProtocol).ReadMessageBegin(0xc008765040, {0x311a118, 0xc00319ede0})
  github.com/apache/thrift@v0.15.0/lib/go/thrift/compact_protocol.go:367 +0x62

### DIFF
--- a/lib/go/README.md
+++ b/lib/go/README.md
@@ -132,3 +132,27 @@ if this interval is set to a value too low (for example, 1ms), it might cause
 excessive cpu overhead.
 
 This feature is also only enabled on non-oneway endpoints.
+
+A note about server stop implementations
+========================================
+
+[TSimpleServer.Stop](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TSimpleServer.Stop) will wait for all client connections to be closed after 
+the last received request to be handled, as the time spent by Stop
+ may sometimes be too long:
+* When socket timeout is not set, server might be hanged before all active
+  clients to finish handling the last received request.
+* When the socket timeout is too long (e.g one hour), server will
+  hang for that duration before all active clients to finish handling the
+  last received request.
+
+To prevent Stop from hanging for too long, you can set 
+thrift.ServerStopTimeout in your main or init function:
+
+    thrift.ServerStopTimeout = <max_duration_to_stop>
+
+If it's set to <=0, the feature will be disabled (by default), and server 
+will wait for all the client connections to be closed gracefully with 
+zero err time. Otherwise, the stop will wait for all the client 
+connections to be closed gracefully util thrift.ServerStopTimeout is 
+reached, and client connections that are not closed after thrift.ServerStopTimeout 
+will be closed abruptly which may cause some client errors.
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -20,6 +20,7 @@
 package thrift
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -48,15 +49,26 @@ var ErrAbandonRequest = errors.New("request abandoned")
 // If it's changed to <=0, the feature will be disabled.
 var ServerConnectivityCheckInterval = time.Millisecond * 5
 
+// ServerStopTimeout defines max stop wait duration used by
+// server stop to avoid hanging too long to wait for all client connections to be closed gracefully.
+//
+// It's defined as a variable instead of constant, so that thrift server
+// implementations can change its value to control the behavior.
+//
+// If it's set to <=0, the feature will be disabled(by default), and the server will wait for
+// for all the client connections to be closed gracefully.
+var ServerStopTimeout = time.Duration(0)
+
 /*
  * This is not a typical TSimpleServer as it is not blocked after accept a socket.
  * It is more like a TThreadedServer that can handle different connections in different goroutines.
  * This will work if golang user implements a conn-pool like thing in client side.
  */
 type TSimpleServer struct {
-	closed int32
-	wg     sync.WaitGroup
-	mu     sync.Mutex
+	closed   int32
+	wg       sync.WaitGroup
+	mu       sync.Mutex
+	stopChan chan struct{}
 
 	processorFactory       TProcessorFactory
 	serverTransport        TServerTransport
@@ -121,6 +133,7 @@ func NewTSimpleServerFactory6(processorFactory TProcessorFactory, serverTranspor
 		outputTransportFactory: outputTransportFactory,
 		inputProtocolFactory:   inputProtocolFactory,
 		outputProtocolFactory:  outputProtocolFactory,
+		stopChan:               make(chan struct{}),
 	}
 }
 
@@ -192,11 +205,25 @@ func (p *TSimpleServer) innerAccept() (int32, error) {
 		return 0, err
 	}
 	if client != nil {
-		p.wg.Add(1)
+		ctx, cancel := context.WithCancel(context.Background())
+		p.wg.Add(2)
+
 		go func() {
 			defer p.wg.Done()
+			defer cancel()
 			if err := p.processRequests(client); err != nil {
 				p.logger(fmt.Sprintf("error processing request: %v", err))
+			}
+		}()
+
+		go func() {
+			defer p.wg.Done()
+			select {
+			case <-ctx.Done():
+				// client exited, do nothing
+			case <-p.stopChan:
+				// TSimpleServer.Close called, close the client connection
+				client.Close()
 			}
 		}()
 	}
@@ -229,12 +256,31 @@ func (p *TSimpleServer) Serve() error {
 func (p *TSimpleServer) Stop() error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
+
 	if atomic.LoadInt32(&p.closed) != 0 {
 		return nil
 	}
 	atomic.StoreInt32(&p.closed, 1)
 	p.serverTransport.Interrupt()
-	p.wg.Wait()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		p.wg.Wait()
+	}()
+
+	if ServerStopTimeout > 0 {
+		timer := time.NewTimer(ServerStopTimeout)
+		select {
+		case <-timer.C:
+		case <-ctx.Done():
+		}
+		close(p.stopChan)
+		timer.Stop()
+	}
+
+	<-ctx.Done()
+	p.stopChan = make(chan struct{})
 	return nil
 }
 
--- a/lib/go/thrift/simple_server_test.go
+++ b/lib/go/thrift/simple_server_test.go
@@ -20,10 +20,16 @@
 package thrift
 
 import (
-	"testing"
+	"context"
 	"errors"
+	"net"
 	"runtime"
+	"sync"
+	"testing"
+	"time"
 )
+
+const networkWaitDuration = 10 * time.Millisecond
 
 type mockServerTransport struct {
 	ListenFunc    func() error
@@ -153,4 +159,131 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
 	go serv.Serve()
 	runtime.Gosched()
 	serv.Stop()
+}
+
+func TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop(t *testing.T) {
+	ln, err := net.Listen("tcp", "localhost:0")
+
+	if err != nil {
+		t.Fatalf("Failed to listen: %v", err)
+	}
+
+	proc := &mockProcessor{
+		ProcessFunc: func(in, out TProtocol) (bool, TException) {
+			in.ReadMessageBegin(context.Background())
+			return false, nil
+		},
+	}
+
+	trans := &mockServerTransport{
+		ListenFunc: func() error {
+			return nil
+		},
+		AcceptFunc: func() (TTransport, error) {
+			conn, err := ln.Accept()
+			if err != nil {
+				return nil, err
+			}
+
+			return NewTSocketFromConnConf(conn, nil), nil
+		},
+		CloseFunc: func() error {
+			return nil
+		},
+		InterruptFunc: func() error {
+			return ln.Close()
+		},
+	}
+
+	serv := NewTSimpleServer2(proc, trans)
+	go serv.Serve()
+	time.Sleep(networkWaitDuration)
+
+	netConn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil || netConn == nil {
+		t.Fatal("error when dial server")
+	}
+	time.Sleep(networkWaitDuration)
+
+	serverStopTimeout := 50 * time.Millisecond
+	backupServerStopTimeout := ServerStopTimeout
+	t.Cleanup(func() {
+		ServerStopTimeout = backupServerStopTimeout
+	})
+	ServerStopTimeout = serverStopTimeout
+
+	st := time.Now()
+	err = serv.Stop()
+	if err != nil {
+		t.Errorf("error when stop server:%v", err)
+	}
+
+	if elapsed := time.Since(st); elapsed < serverStopTimeout {
+		t.Errorf("stop cost less time than server stop timeout, server stop timeout:%v,cost time:%v", ServerStopTimeout, elapsed)
+	}
+}
+
+func TestStopTimeoutWithSocketTimeout(t *testing.T) {
+	ln, err := net.Listen("tcp", "localhost:0")
+
+	if err != nil {
+		t.Fatalf("Failed to listen: %v", err)
+	}
+
+	proc := &mockProcessor{
+		ProcessFunc: func(in, out TProtocol) (bool, TException) {
+			in.ReadMessageBegin(context.Background())
+			return false, nil
+		},
+	}
+
+	conf := &TConfiguration{SocketTimeout: 5 * time.Millisecond}
+	wg := &sync.WaitGroup{}
+	trans := &mockServerTransport{
+		ListenFunc: func() error {
+			return nil
+		},
+		AcceptFunc: func() (TTransport, error) {
+			conn, err := ln.Accept()
+			if err != nil {
+				return nil, err
+			}
+			defer wg.Done()
+			return NewTSocketFromConnConf(conn, conf), nil
+		},
+		CloseFunc: func() error {
+			return nil
+		},
+		InterruptFunc: func() error {
+			return ln.Close()
+		},
+	}
+
+	serv := NewTSimpleServer2(proc, trans)
+	go serv.Serve()
+	time.Sleep(networkWaitDuration)
+
+	wg.Add(1)
+	netConn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil || netConn == nil {
+		t.Fatal("error when dial server")
+	}
+	wg.Wait()
+
+	expectedStopTimeout := time.Second
+	backupServerStopTimeout := ServerStopTimeout
+	t.Cleanup(func() {
+		ServerStopTimeout = backupServerStopTimeout
+	})
+	ServerStopTimeout = expectedStopTimeout
+
+	st := time.Now()
+	err = serv.Stop()
+	if elapsed := time.Since(st); elapsed > expectedStopTimeout/2 {
+		t.Errorf("stop cost more time than socket timeout, socket timeout:%v,server stop timeout:%v,cost time:%v", conf.SocketTimeout, ServerStopTimeout, elapsed)
+	}
+
+	if err != nil {
+		t.Fatalf("error when stop server:%v", err)
+	}
 }
--- a/test/go/src/common/clientserver_test.go
+++ b/test/go/src/common/clientserver_test.go
@@ -75,7 +75,7 @@ func doUnit(t *testing.T, unit *test_unit) {
 		t.Errorf("Unable to start server: %v", err)
 		return
 	}
-	go server.AcceptLoop()
+	go server.Serve()
 	defer server.Stop()
 	client, trans, err := StartClient(unit.host, unit.port, unit.domain_socket, unit.transport, unit.protocol, unit.ssl)
 	if err != nil {


dcelasun:
Thanks for the PR!

This doesn't seem like a good use of `sync.Map`. From [godoc](https://pkg.go.dev/sync#Map):

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

In your case, the map is write-heavy, and reads only happen once during `Close`. It'd be better to replace it with a `map[TTransport]struct{}` and a `sync.Mutex`. It would also give us type safety.

buptubuntu:
> Thanks for the PR!
> 
> This doesn't seem like a good use of `sync.Map`. From [godoc](https://pkg.go.dev/sync#Map):
> 
> > The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
> 
> > The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.
> 
> In your case, the map is write-heavy, and reads only happen once during `Close`. It'd be better to replace it with a `map[TTransport]struct{}` and a `sync.Mutex`. It would also give us type safety.

Good suggestion，I will change the code later this week

dcelasun (approved):
LGTM. @fishy any thoughts?

lib/go/thrift/simple_server.go
```diff
@@ -193,7 +195,13 @@ func (p *TSimpleServer) innerAccept() (int32, error) {
 	}
 	if client != nil {
 		p.wg.Add(1)
+		p.clients[client] = struct{}{}
 		go func() {
+			defer func() {
+				p.mu.Lock()
+				defer p.mu.Unlock()
+				delete(p.clients, client)
```
fishy, line 203: Go has a runtime bug that a builtin map never actually shrinks when you delete from it: https://github.com/golang/go/issues/20135

Which means although the behavior is correct (you won't try to close those clients that's already disconnected during `Close`), for a long running server this map will keep growing indefinitely and eventually causing memory issues.

For that reason, a `sync.Map` will probably actually be better, as I understood it `sync.Map` actually occasionally replaces the whole map under the hood so it actually can be garbage collected.

Also I'm not sure that you can actually use a `TTransport` as the map key (for either builtin map or `sync.Map`), when you try to use a type that cannot be used as the map key it will cause a runtime panic instead of some compile time error. We might need to make some breaking change to expose the `TSocket`/`TSSLSocket` that's wrapped inside this `TTransport` and use its `.Addr().String()` as the map key (and ignore any `TTransport` that does not have a `TSocket`/`TSSLSocket` under the hood).

fishy (changes requested):

lib/go/thrift/simple_server.go
dcelasun: > `sync.Map` will probably actually be better, as I understood it `sync.Map` actually occasionally replaces the whole map under the hood so it actually can be garbage collected.

Looking at the implementation, I don't think `sync.Map` would help either. The underlying "dirty" map is only replaced after a number of `Load()` misses, which would never happen here as we only read once during `Close()`.

> Also I'm not sure that you can actually use a TTransport as the map key (for either builtin map or sync.Map), when you try to use a type that cannot be used as the map key it will cause a runtime panic instead of some compile time error.

I'm not seeing any Go test failures on Travis (though I didn't check thoroughly) so I assume all implementations of `TTransport` are hashable.

fishy: Looks like at least a `*TSocket` is comparable:

```
$ cat main.go 
package main

import (
        "fmt"

        "github.com/apache/thrift/lib/go/thrift"
)

func main() {
        sock := thrift.NewTSocketConf("localhost:6060", nil)
        m := map[thrift.TTransport]struct{}{
                sock: {},
        }
        fmt.Println(m)
}
$ go run main.go 
map[0xc000068120:{}]
```

But still, there's no guarantee that `TTransport` returned by `TServerSocket.Accept()` will always be comparable, and when it's not trying to use it as a key will cause a runtime panic and it will be hard to detect in tests. (For example, we have a pending PR to wrap `TServerSocket` to return a `TTransport` that counts the number of active clients: https://github.com/reddit/baseplate.go/pull/464)

So if this is really important to you, make it opt-in instead: This map is only used when people call an exported function from `TSimpleServer` to enable it (for example, `TSimpleServer.EnableClientCleanup()`, name tbd as I'm bad at names). And we should note in the function's doc about the implications:

1. The memory used by this map could be unbounded
2. If you enable this map, `TTransport` returned by your `TServerTransport` must be comparable or panics will happen

fishy: Even with the opt-in approach, consider how big a footgun using `TTransport` as the map key could be, I think it's still better to:

1. Make an breaking API change, add `Addr() net.Addr` to `TTransport` interface (`TMemoryBuffer.Addr()` should return `nil`, delegate transports like `TFramedTransport`, `TBufferedTransport`, `THeaderTransport`, etc. should just call their wrapped `TTransport.Addr()`)
2. We use `TTransport.Addr().String()` as the map key, and don't put the `TTransport` into the map if `Addr()` returns `nil` (this should only happen when we use `TMemoryBuffer` under the hood, I think, which should be fine)

@dcelasun what do you think?

Another alternative is to add something like `TTransport.String()` to the interface, but the behavior of that would be more undefined.

fishy: Another idea I had considered before for a different issue is to add an `Unwrap() TTransport` api to `TTransport`, similar to how `errors.Unwrap()` works. This way we can keep unwrapping until we find the `*TSocket` or `*TSSLSocket` wrapped under the hood and use its `Addr().String()` as the map key (and if there's no `*TSocket`/`*TSSLSocket`, then we don't put this client into the map).

dcelasun: Breaking `TTransport` will definitely be painful. When we added `context.Context` to function signatures, it broke so many downstream projects. This won't be as bad, but still.

Anyway, as long as we're adding new API to `TTransport` I like `Addr() net.Addr` better than `Unwrap() TTransport`. It feels more readable.

dcelasun: I'm also not a fan of the opt-in or documentation approaches. The former, because the current behaviour is wrong and should be fixed, and the latter because I don't think end users should worry about an OOM caused by an internal implementation detail.

fishy: I think depending on the real world situation "memory of a server grows unbounded and will eventually OOM" could be more severe than "the shutdown of the server will take forever and eventually timeout". But this is a fair point:

>I'm also not a fan of the opt-in or documentation approaches. The former, because the current behaviour is wrong and should be fixed, and the latter because I don't think end users should worry about an OOM caused by an internal implementation detail.

But in order to turn it default on, we must make the memory bounded. One way to do it is to add a background goroutine with `TSimpleServer` that use a ticker to copy the map over regularly, but when doing so it also need to hold the lock, and preventing `TSimpleServer` from processing any new requests (so it's kind of like a stop-the-world gc). The goroutine function should be something like this:

```go
func (p *TSimpleServer) mapCleanup() {
  p.mu.Lock()
  defer p.mu.Unlock()
  newMap := make(map[string]TTransport, len(p.clients))  // the actual type of the map pending on which type we decide in the end
  for k, v := range p.clients {
    newMap[k] = v
  }
  p.clients = newMap
}
```

It might not be too bad? This copy should be reasonably efficient in go, but I would like to see some real world impact first before merging it.

fishy: Or maybe instead of a ticker, make it run every N deletes. I'm not sure which would actually work better in real world scenarios.

dcelasun: > Or maybe instead of a ticker, make it run every N deletes. I'm not sure which would actually work better in real world scenarios.

I think every N deletes is better because with a ticker it can still OOM during traffic spikes.

fishy: with either a ticker or every N deletes, we should make the ticker time/N configurable.

so if we choose ticker approach, and some service know that it could handle a lot of connections, they can choose to set a shorter ticker time to cleanup more often.

I think a shorter ticker time might work better as with a shorter ticker time usually also comes with smaller map to copy over, so the whole "stop-the-world-gc-ish" thing can be quicker and less noticeable.

fishy: now I wonder how stdlib http server resolve this issue. maybe I should take a look at its implementation.

fishy: well, for http server (https://pkg.go.dev/net/http#Server.Shutdown):

>When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return.
>Shutdown does not attempt to close nor wait for hijacked connections such as WebSockets. The caller of Shutdown should separately notify such long-lived connections of shutdown and wait for them to close, if desired. See RegisterOnShutdown for a way to register shutdown notification functions.

fishy: instead of either of the stop-the-world-gc-ish approaches, a better approach might be to make the read not blocking forever. basically we could try to always add a timeout to the read, and when timeout happens check if the server is closed, and if not, we can go back to read again.

this might need some compiler change to accomplish, though. as it might need some changes to the processor. haven't really looked into the details yet.

fishy: Actually, I misinterpreted that go runtime bug. It's just that the capacity of the map only grows and never shrinks (so delete won't cause the map to shrink), but memory holding deleted items can be reused for future inserts. So although this is still unbounded, it will be more like a high watermark of how many concurrent connection the server handles, and unlikely to cause OOM. So using a map is fine. But we still need to handle the map key issue.

This is the code I used to verify the behavior (note the different allocations between the 2 benchmark tests):

```
$ go test -v -bench .
=== RUN   TestHelpers
    main_test.go:43: size: 1000
    main_test.go:46: size: 502
--- PASS: TestHelpers (0.00s)
goos: linux
goarch: amd64
pkg: test
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkMap
BenchmarkMap/fill-twice
    main_test.go:60: total size 2000, n 1, average size: 2000
    main_test.go:60: total size 200000, n 100, average size: 2000
    main_test.go:60: total size 5982000, n 2991, average size: 2000
    main_test.go:60: total size 14522000, n 7261, average size: 2000
BenchmarkMap/fill-twice-12                  7261            184987 ns/op          101042 B/op        113 allocs/op
BenchmarkMap/fill-delete-fill
    main_test.go:72: total size 1505, n 1, average size: 1505
    main_test.go:72: total size 150243, n 100, average size: 1502
    main_test.go:72: total size 8163265, n 5442, average size: 1500
    main_test.go:72: total size 12801894, n 8534, average size: 1500
BenchmarkMap/fill-delete-fill-12            8534            146318 ns/op           50348 B/op         91 allocs/op
PASS
ok      test    3.927s
$ cat main_test.go 
package main

import (
        "math/rand"
        "os"
        "testing"
)

const (
        n = 1000

        margin = 0.1
)

type mapType = map[int]struct{}

func fillMap(m mapType, n int) {
        for i := 0; i < n; i++ {
                m[rand.Int()] = struct{}{}
        }
}

func deleteHalf(m mapType) {
        for k := range m {
                if rand.Float64() >= 0.5 {
                        delete(m, k)
                }
        }
}

func compareMargin(tb testing.TB, want, got int, margin float64) {
        tb.Helper()

        diff := int(float64(want) * margin)
        if got < want-diff || got > want+diff {
                tb.Errorf("Want %d got %d margin %v", want, got, margin)
        }
}

func TestHelpers(t *testing.T) {
        m := make(mapType)
        fillMap(m, n)
        t.Logf("size: %d", len(m))
        compareMargin(t, n, len(m), margin)
        deleteHalf(m)
        t.Logf("size: %d", len(m))
        compareMargin(t, n/2, len(m), margin)
}

func BenchmarkMap(b *testing.B) {
        b.Run("fill-twice", func(b *testing.B) {
                b.ReportAllocs()
                var totalSize int
                for i := 0; i < b.N; i++ {
                        m := make(mapType)
                        fillMap(m, n)
                        fillMap(m, n)
                        totalSize += len(m)
                }
                b.Logf("total size %d, n %d, average size: %v", totalSize, b.N, totalSize/b.N)
        })
        b.Run("fill-delete-fill", func(b *testing.B) {
                b.ReportAllocs()
                var totalSize int
                for i := 0; i < b.N; i++ {
                        m := make(mapType)
                        fillMap(m, n)
                        deleteHalf(m)
                        fillMap(m, n)
                        totalSize += len(m)
                }
                b.Logf("total size %d, n %d, average size: %v", totalSize, b.N, totalSize/b.N)
        })
}

func TestMain(m *testing.M) {
        rand.Seed(1)
        os.Exit(m.Run())
}
```

fishy: another potential issue with the approach: closing the socket in one goroutine while another goroutine is reading from/writing to the socket might cause race conditions?

fishy: > another potential issue with the approach: closing the socket in one goroutine while another goroutine is reading from/writing to the socket might cause race conditions?

from the doc of `net.Conn`:
```
	// Close closes the connection.
	// Any blocked Read or Write operations will be unblocked and return errors.
	Close() error
```

so I guess this is fine.

But in that case we don't even need a map? we can just add one additional goroutine per client connection, basically:

1. In `TSimpleServer.Serve` we create a `context.WithCancel` and store both the ctx and cancel function `TSimpleServer`
2. cancel that ctx in `TSimpleServer.Close`
3. in every client accept goroutine, we start another goroutine to be unblocked by the server ctx to close the connection, basically changing https://github.com/apache/thrift/blob/999e6e3bce217acb35b44440fd656cf169d47ed8/lib/go/thrift/simple_server.go#L196-L201 into:

```go
		go func() {
			defer p.wg.Done()
			ctx, cancel := context.WithCancel(context.Background())
			defer cancel()
			go func() {
				select {
				case <-ctx.Done():
					// client exited, do nothing
				case <- p.ctx.Done():
					// TSimpleServer.Close called, close the client connection
					client.Close()
				}
			}()
			if err := p.processRequests(client); err != nil {
				p.logger(fmt.Sprintf("error processing request: %v", err))
			}
		}()
```

basically we trade a (kinda unbounded) map with one additional goroutine per client connection.

dcelasun: > we trade a (kinda unbounded) map with one additional goroutine per client connection. the big advantage is that we no longer need to make a breaking change to TTransport to make it usable in a map.

Sounds much more reasonable. @buptubuntu what do you think?

buptubuntu: I have also considered to add one additional goroutine per client connection，but not sure if that will cause performance problem. But if we want to fix the map key issue we will have to make big change to TTrancsport， that will be another issue that we will handle in the future.Besides the cost of the additinal go routine should be quite small comparing to the cost of user defined processor or the data read and write operations in transport.

So I think this is a better solution theoretically

I will try to get the cost of the additional goroutine for some extreme cases:
1>1000~10000 client connect to server but not send data for a long time
2>1000~10000 client connect to server and send 1k data per second

buptubuntu: I have tested the mem the goroutine cost using the code below:
```golang
package main

import (
	"context"
	"fmt"

	"runtime"
	"sync"
)

func main() {
	memConsumed := func() uint64 {
		runtime.GC()
		var s runtime.MemStats
		runtime.ReadMemStats(&s)
		return s.Sys
	}

	var channel <-chan interface{}
	var wg sync.WaitGroup

	const numGoroutines = 1000000 // 1M
	wg.Add(numGoroutines)

	before := memConsumed()
	for i := numGoroutines; i > 0; i-- {
		ctx, cancel := context.WithCancel(context.Background())
		go func() {
			wg.Done()
			cancel()
			<-ctx.Done()
			<-channel

		}()
	}
	wg.Wait()
	after := memConsumed()

	fmt.Printf("%.3fkb", float64(after-before)/numGoroutines/1024)
}
```

cost as below:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/buptubuntu/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/buptubuntu/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
</head>

<body link="#0563C1" vlink="#954F72">



goroutines | total mem | mem per goroutine
-- | -- | --
10000 | 35M | 3.5K
1000 | 5.3M | 5.3K



</body>

</html>

buptubuntu: I have also tested the cost per client using code below:

```golang
package main

import (
	"context"
	"fmt"
	"net"
	_ "net/http/pprof"
	"runtime"
	"strings"
	"time"

	"github.com/apache/thrift/lib/go/thrift"
)

func main() {
	memConsumed := func() uint64 {
		runtime.GC()
		var s runtime.MemStats
		runtime.ReadMemStats(&s)
		return s.Sys
	}

	ln, _ := net.Listen("tcp", "localhost:0")

	proc := &mockProcessor{
		ProcessFunc: func(in, out thrift.TProtocol) (bool, thrift.TException) {
			in.ReadMessageBegin(context.Background())
			time.Sleep(time.Hour)
			return false, nil
		},
	}

	trans := &mockServerTransport{
		ListenFunc: func() error {
			return nil
		},
		AcceptFunc: func() (thrift.TTransport, error) {
			conn, err := ln.Accept()
			if err != nil {
				// t.Errorf("error accept connection")
				return nil, err
			}
			return thrift.NewTSocketFromConnTimeout(conn, 0), nil
		},
		CloseFunc: func() error {
			return nil
		},
		InterruptFunc: func() error {
			return ln.Close()
		},
	}

	serv := thrift.NewTSimpleServer2(proc, trans)
	go serv.Serve()
	time.Sleep(time.Second)
	const numGoroutines = 1000
	conns := make([]net.Conn, 0, numGoroutines)
	before := memConsumed()

	for i := 0; i < numGoroutines; i++ {
		port := strings.Split(ln.Addr().String(), ":")[1]
		netConn, _ := net.Dial("tcp", "localhost:"+port)
		time.Sleep(time.Millisecond)
		conns = append(conns, netConn)
	}

	after := memConsumed()

	time.Sleep(time.Second)
	fmt.Printf("%.3fkb", float64(after-before)/numGoroutines/1024)
}

type mockProcessor struct {
	ProcessFunc func(in, out thrift.TProtocol) (bool, thrift.TException)
}

func (m *mockProcessor) Process(ctx context.Context, in, out thrift.TProtocol) (bool, thrift.TException) {
	return m.ProcessFunc(in, out)
}

func (m *mockProcessor) ProcessorMap() map[string]thrift.TProcessorFunction {
	return map[string]thrift.TProcessorFunction{
		"mock": thrift.WrappedTProcessorFunction{
			Wrapped: func(ctx context.Context, seqId int32, in, out thrift.TProtocol) (bool, thrift.TException) {
				return m.ProcessFunc(in, out)
			},
		},
	}
}

func (m *mockProcessor) AddToProcessorMap(name string, processorFunc thrift.TProcessorFunction) {}

type mockWrappedProcessorContextKey int

const (
	processorName mockWrappedProcessorContextKey = iota
)

// setMockWrappableProcessorName sets the "name" of the TProcessorFunction to
// call on a mockWrappableProcessor when calling Process.
//
// In a normal TProcessor, the request name is read from the request itself
// which happens in TProcessor.Process, so it is not passed into the call to
// Process itself, to get around this in testing, mockWrappableProcessor calls
// getMockWrappableProcessorName  to get the name to use from the context
// object.
func setMockWrappableProcessorName(ctx context.Context, name string) context.Context {
	return context.WithValue(ctx, processorName, name)
}

// getMockWrappableProcessorName gets the "name" of the TProcessorFunction to
// call on a mockWrappableProcessor when calling Process.
func getMockWrappableProcessorName(ctx context.Context) (string, bool) {
	val, ok := ctx.Value(processorName).(string)
	return val, ok
}

// mockWrappableProcessor can be used to create a mock object that fufills the
// TProcessor interface in testing.
type mockWrappableProcessor struct {
	ProcessorFuncs map[string]thrift.TProcessorFunction
}

// Process calls the TProcessorFunction assigned to the "name" set on the
// context object by setMockWrappableProcessorName.
//
// If no name is set on the context or there is no TProcessorFunction mapped to
// that name, the call will panic.
func (p *mockWrappableProcessor) Process(ctx context.Context, in, out thrift.TProtocol) (bool, thrift.TException) {
	name, ok := getMockWrappableProcessorName(ctx)
	if !ok {
		panic("MockWrappableProcessorName not set on context")
	}
	processor, ok := p.ProcessorMap()[name]
	if !ok {
		panic(fmt.Sprintf("No processor set for name %q", name))
	}
	return processor.Process(ctx, 0, in, out)
}

func (p *mockWrappableProcessor) ProcessorMap() map[string]thrift.TProcessorFunction {
	return p.ProcessorFuncs
}

func (p *mockWrappableProcessor) AddToProcessorMap(name string, processorFunc thrift.TProcessorFunction) {
	p.ProcessorFuncs[name] = processorFunc
}

var (
	_ thrift.TProcessor = (*mockProcessor)(nil)
	_ thrift.TProcessor = (*mockWrappableProcessor)(nil)
)

type mockServerTransport struct {
	ListenFunc    func() error
	AcceptFunc    func() (thrift.TTransport, error)
	CloseFunc     func() error
	InterruptFunc func() error
}

func (m *mockServerTransport) Listen() error {
	return m.ListenFunc()
}

func (m *mockServerTransport) Accept() (thrift.TTransport, error) {
	return m.AcceptFunc()
}

func (m *mockServerTransport) Close() error {
	return m.CloseFunc()
}

func (m *mockServerTransport) Interrupt() error {
	return m.InterruptFunc()
}

type mockTTransport struct {
	thrift.TTransport
}

func (m *mockTTransport) Close() error {
	return nil
}
```
cost as below:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/buptubuntu/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/buptubuntu/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
</head>

<body link="#0563C1" vlink="#954F72">



clients | total mem | mem per client
-- | -- | --
10000 | 157M | 15.7K
1000 | 24M | 24K



</body>

</html>

buptubuntu: For the cpu cost, as the operation of the additional goroutine is quite simple(close the client) and we just move it from the stop function,i think we can just ignore that

lib/go/thrift/socket.go
```diff
@@ -195,8 +195,9 @@ func (p *TSocket) IsOpen() bool {
 // Closes the socket.
 func (p *TSocket) Close() error {
 	// Close the socket
-	if p.conn != nil {
-		err := p.conn.Close()
+	conn := p.conn
```
fishy, line 198: I assume the intention of this change is to make it thread safe, but this does not achieve that (if you run `go test -race` with the change in this PR it still complains race condition, because we have 2 goroutines calling `TSocket.Close` concurrently).

https://github.com/apache/thrift/pull/2500 will address this issue.

lib/go/thrift/simple_server.go
```diff
@@ -234,7 +250,9 @@ func (p *TSimpleServer) Stop() error {
 	}
 	atomic.StoreInt32(&p.closed, 1)
 	p.serverTransport.Interrupt()
+	close(p.stopChan)
 	p.wg.Wait()
+	p.stopChan = make(chan struct{})
```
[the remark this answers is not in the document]
buptubuntu: Because is stopped server can serve later, and the closed channel can not be closed again, so we have to make a new channel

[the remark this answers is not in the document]
fishy: in that case this should be done in `Serve` instead.

lib/go/thrift/socket.go
buptubuntu: go test --race ./lib/go/thrift/...
ok      github.com/apache/thrift/lib/go/thrift  2.343s

seems ok

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,50 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
 	runtime.Gosched()
 	serv.Stop()
 }
+
+func TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop(t *testing.T) {
+	ln, err := net.Listen("tcp", "localhost:0")
+	if err != nil {
+		t.Errorf("error when listen")
```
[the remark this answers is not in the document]
buptubuntu: done

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,50 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+	}
+
+	trans := &mockServerTransport{
+		ListenFunc: func() error {
+			return nil
+		},
+		AcceptFunc: func() (TTransport, error) {
+			conn, err := ln.Accept()
+			if err != nil {
+				// t.Errorf("error accept connection")
```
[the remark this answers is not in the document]
buptubuntu: done

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,50 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+		InterruptFunc: func() error {
+			return ln.Close()
+		},
+	}
+
+	serv := NewTSimpleServer2(proc, trans)
+	go serv.Serve()
+
+	port := strings.Split(ln.Addr().String(), ":")[1]
+	netConn, err := net.Dial("tcp", "localhost:"+port)
```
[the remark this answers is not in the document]
buptubuntu: done

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,50 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+		ListenFunc: func() error {
+			return nil
+		},
+		AcceptFunc: func() (TTransport, error) {
+			conn, err := ln.Accept()
+			if err != nil {
+				// t.Errorf("error accept connection")
+				return nil, err
+			}
+			return NewTSocketFromConnTimeout(conn, 0), nil
```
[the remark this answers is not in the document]
buptubuntu: done

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,50 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+
+	serv := NewTSimpleServer2(proc, trans)
+	go serv.Serve()
+
+	port := strings.Split(ln.Addr().String(), ":")[1]
+	netConn, err := net.Dial("tcp", "localhost:"+port)
+	if err != nil || netConn == nil {
+		t.Errorf("error when dial server")
+	}
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
buptubuntu: done

lib/go/thrift/socket.go
buptubuntu: OK,i have undo the change

lib/go/thrift/simple_server.go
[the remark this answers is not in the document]
buptubuntu: I think it is better to init the stopChan in construction, if not TestMultipleStop will panic because of nil chan

What's more, current HEAD allows multi **Serve**, if we make it in **Serve**，the stopChan may change and will never be closed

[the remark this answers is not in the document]
buptubuntu: i don't think so
Firstly, the stopChan should be created when the server is created
Secondly, it is **stop** that close the stopChan, so **stop** should also be responsible to recreate the stopChan
After all if we call serve several times before stop, it will create stopChan several times without close it, that is quite strange

lib/go/thrift/socket.go
fishy: Race detector detected issue in this change and the issue went away after I cherry-picked #2500 in:

```
~/work/thrift/lib/go/thrift$ go test -race
==================
WARNING: DATA RACE
Write at 0x00c0002c2260 by goroutine 74:
  github.com/apache/thrift/lib/go/thrift.(*TSocket).Close()
      /home/fishy/work/thrift/lib/go/thrift/socket.go:203 +0x79
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept.func2()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:215 +0x1a1

Previous read at 0x00c0002c2260 by goroutine 107:
  github.com/apache/thrift/lib/go/thrift.(*TSocket).Close()
      /home/fishy/work/thrift/lib/go/thrift/socket.go:198 +0x32
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).processRequests·dwrap·14()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:307 +0x48
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).processRequests()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:349 +0x8ce
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept.func1()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:203 +0x108

Goroutine 74 (running) created at:
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:208 +0x396
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).AcceptLoop()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:224 +0xee
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).Serve()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:241 +0xbd
  github.com/apache/thrift/lib/go/thrift.TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop·dwrap·30()
      /home/fishy/work/thrift/lib/go/thrift/simple_server_test.go:191 +0x39

Goroutine 107 (finished) created at:
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:200 +0x296
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).AcceptLoop()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:224 +0xee
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).Serve()
      /home/fishy/work/thrift/lib/go/thrift/simple_server.go:241 +0xbd
  github.com/apache/thrift/lib/go/thrift.TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop·dwrap·30()
      /home/fishy/work/thrift/lib/go/thrift/simple_server_test.go:191 +0x39
==================
--- FAIL: TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop (0.02s)
    testing.go:1152: race detected during execution of test
FAIL
exit status 1
FAIL    github.com/apache/thrift/lib/go/thrift  2.597s
~/work/thrift/lib/go/thrift$ git cherry-pick 1387ff557
[pr/2497 c356299ba] go: Make socketConn.Close thread-safe
 Date: Sat Jan 8 01:03:57 2022 -0800
 3 files changed, 14 insertions(+), 19 deletions(-)
~/work/thrift/lib/go/thrift$ go test -race
PASS
ok      github.com/apache/thrift/lib/go/thrift  2.468s
```

lib/go/thrift/simple_server.go
```diff
@@ -54,10 +55,10 @@ var ServerConnectivityCheckInterval = time.Millisecond * 5
...
  */
 type TSimpleServer struct {
-	closed int32
-	wg     sync.WaitGroup
-	mu     sync.Mutex
-
+	closed                 int32
+	wg                     sync.WaitGroup
+	mu                     sync.Mutex
+	stopChan               chan struct{}
```
fishy, line 61: nit: I would still prefer to have a blank line after this line so we have the logical "groups" of the fields of a `TSimpleServer`.

fishy (approved):
this one should only be merged after #2500. we probably should enable race detector in travis.

fishy:
actually we already have race detector enabled on travis and it's falling (https://app.travis-ci.com/github/apache/thrift/jobs/554765808):

```
make[4]: Entering directory '/thrift/src/lib/go'
/usr/local/bin/go test -mod=mod -race ./thrift
==================
WARNING: DATA RACE
Read at 0x00c000509580 by goroutine 92:
  github.com/apache/thrift/lib/go/thrift.(*TSocket).Close()
      /thrift/src/lib/go/thrift/socket.go:198 +0x32
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).processRequests·dwrap·15()
      /thrift/src/lib/go/thrift/simple_server.go:307 +0x48
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).processRequests()
      /thrift/src/lib/go/thrift/simple_server.go:349 +0x8ce
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept.func1()
      /thrift/src/lib/go/thrift/simple_server.go:203 +0x108
Previous write at 0x00c000509580 by goroutine 130:
  github.com/apache/thrift/lib/go/thrift.(*TSocket).Close()
      /thrift/src/lib/go/thrift/socket.go:203 +0x79
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept.func2()
      /thrift/src/lib/go/thrift/simple_server.go:215 +0x1a1
Goroutine 92 (running) created at:
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept()
      /thrift/src/lib/go/thrift/simple_server.go:200 +0x296
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).AcceptLoop()
      /thrift/src/lib/go/thrift/simple_server.go:224 +0xee
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).Serve()
      /thrift/src/lib/go/thrift/simple_server.go:241 +0xbd
  github.com/apache/thrift/lib/go/thrift.TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop·dwrap·31()
      /thrift/src/lib/go/thrift/simple_server_test.go:191 +0x39
Goroutine 130 (finished) created at:
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept()
      /thrift/src/lib/go/thrift/simple_server.go:208 +0x396
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).AcceptLoop()
      /thrift/src/lib/go/thrift/simple_server.go:224 +0xee
  github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).Serve()
      /thrift/src/lib/go/thrift/simple_server.go:241 +0xbd
  github.com/apache/thrift/lib/go/thrift.TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop·dwrap·31()
      /thrift/src/lib/go/thrift/simple_server_test.go:191 +0x39
==================
--- FAIL: TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop (0.02s)
    testing.go:1152: race detected during execution of test
FAIL
FAIL	github.com/apache/thrift/lib/go/thrift	2.081s
FAIL
```

fishy:
also cc @Jens-G we should also wait for this one for 0.16.0, but this one is already pretty close :)

fishy:
@buptubuntu now that #2500 is merged, please rebase this branch on top of the latest master so that it can pass the test with race detector on travis.

buptubuntu:
> @buptubuntu now that #2500 is merged, please rebase this branch on top of the latest master so that it can pass the test with race detector on travis.

OK

fishy:
hmm, this test failed on travis:

```
/usr/local/bin/go test -mod=mod -v ./src/common/...
=== RUN   TestAllConnection
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6cc5da]
goroutine 9 [running]:
github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept.func1()
	/thrift/src/lib/go/thrift/simple_server.go:205 +0xba
created by github.com/apache/thrift/lib/go/thrift.(*TSimpleServer).innerAccept
	/thrift/src/lib/go/thrift/simple_server.go:201 +0x1c8
FAIL	github.com/apache/thrift/test/go/src/common	0.015s
FAIL
```

but I cannot reproduce it locally and it doesn't make sense to me

fishy:
Oh I think I know why (but I wonder why did this not fail before)

@buptubuntu please apply this diff to the PR:

```diff
diff --git a/test/go/src/common/clientserver_test.go b/test/go/src/common/clientserver_test.go
index 609086bad..64b326a81 100644
--- a/test/go/src/common/clientserver_test.go
+++ b/test/go/src/common/clientserver_test.go
@@ -75,7 +75,7 @@ func doUnit(t *testing.T, unit *test_unit) {
                t.Errorf("Unable to start server: %v", err)
                return
        }
-       go server.AcceptLoop()
+       go server.Serve()
        defer server.Stop()
        client, trans, err := StartClient(unit.host, unit.port, unit.domain_socket, unit.transport, unit.protocol, unit.ssl)
        if err != nil {
```

buptubuntu:
> Oh I think I know why (~but I wonder why did this not fail before~ it didn't fail before because we never called `TSimpleServer.logger` in that test before)
> 
> @buptubuntu please apply this diff to the PR:
> 
> ```diff
> diff --git a/test/go/src/common/clientserver_test.go b/test/go/src/common/clientserver_test.go
> index 609086bad..64b326a81 100644
> --- a/test/go/src/common/clientserver_test.go
> +++ b/test/go/src/common/clientserver_test.go
> @@ -75,7 +75,7 @@ func doUnit(t *testing.T, unit *test_unit) {
>                 t.Errorf("Unable to start server: %v", err)
>                 return
>         }
> -       go server.AcceptLoop()
> +       go server.Serve()
>         defer server.Stop()
>         client, trans, err := StartClient(unit.host, unit.port, unit.domain_socket, unit.transport, unit.protocol, unit.ssl)
>         if err != nil {
> ```

ok，later today

buptubuntu:
> ```diff
> Serve
> ```

done

lib/go/thrift/simple_server.go
```diff
@@ -192,13 +195,27 @@ func (p *TSimpleServer) innerAccept() (int32, error) {
 		return 0, err
 	}
 	if client != nil {
-		p.wg.Add(1)
+		ctx, cancel := context.WithCancel(context.Background())
+		p.wg.Add(2)
```
fishy, line 199: style nit: it's probably less error prone to use `wg.Add(1)` right before every `go`, e.g. write it this way:

```go
ctx, cancel := context.WithCancel(context.Background())

p.wg.Add(1)
go func() {
  defer p.wg.Done()
  // actual work
}()

p.wg.Add(1)
go func() {
  defer p.wg.Done()
  // actual work
}()
```

this avoids the potential bug that someone adds another goroutine in the future but forgot to change the `2` to `3`.

also style-wise, it's really better to always immediately do `defer cancel()` right after `ctx, cancel := ...` (which is a style guide to avoid resource leaks), which is not possible in the current 2 separated goroutines way but possible if you wrap the second goroutine as a goroutine inside the first one (which was what I suggested in https://github.com/apache/thrift/pull/2497#discussion_r779887382 :) )

I don't feel too strongly for either of the nits here, though.

fishy (approved):

lib/go/thrift/simple_server.go
buptubuntu: ctx cancel is done by close a chan
```golang
func (c *cancelCtx) cancel(removeFromParent bool, err error) {
	if err == nil {
		panic("context: internal error: missing cancel error")
	}
	c.mu.Lock()
	if c.err != nil {
		c.mu.Unlock()
		return // already canceled
	}
	c.err = err
	d, _ := c.done.Load().(chan struct{})
	if d == nil {
		c.done.Store(closedchan)
	} else {
		close(d)
	}
	for child := range c.children {
		// NOTE: acquiring the child's lock while holding parent's lock.
		child.cancel(false, err)
	}
	c.children = nil
	c.mu.Unlock()

	if removeFromParent {
		removeChild(c.Context, c)
	}
}
```
And yes, if the ctx is caneled, it will not close the inner channel again, and will do nothing and return. So we still have to create a new ctx after stop the server, because the ctx.Done() channel is already stopped.

divide p.wg.Add(2) into two p.wg.Add(1), it seems ok for now

fishy:
@dcelasun what do you think? I'm ok to merge this in as-is in its current state.

dcelasun:
Yeah, LGTM as well.

fishy (changes requested):
Wait. Sorry I just realized that just closing all clients during `TSimpleServer.Stop` would be a behavior change and make things worth.

In the current code, when `TSimpleServer.Stop` is called, it's a graceful shutdown: it makes `TSimpleServer` to stop accepting new connections (via the atomic int and `Interrupt` call), but it still waits for any request that's currently handling to finish. And letting any request that's currently handling to finish is an important part of graceful shutdown.

Now we just close all the connections (this is true with either the goroutine approach or original map approach), which will cause any currently handling connection to be closed abruptly. from client's pov, they will not receive the full response for this request (it could be either they haven't received any response at all, or they received half of the response then the unexpected EOF happens), and this will cause those requests to fail (because we didn't let them to finish).

So when any server upgrade to this version, this will increase their error rate whenever autoscaling is scaling down. This is a worse scenario than graceful shutdown just timeouts (kubernnetes has a timeout to wait for graceful shutdown, so in current version as long as the server can finish all the pending requests in that time they won't fail any requests just because of a shutdown happens).

I don't know what's the best way to fix that. Maybe add a setting to `TSimpleServer` to sleep for a period of time before trying to abruptly close all connections is a ok short term solution. for long term we might need to make some breaking changes to `TProtocol`/`TTransport` to add the graceful shutdown semantic to them.

@dcelasun your thoughts?

buptubuntu:
> Wait. Sorry I just realized that just closing all clients during `TSimpleServer.Stop` would be a behavior change and make things worse.
> 
> In the current code, when `TSimpleServer.Stop` is called, it's a graceful shutdown: it makes `TSimpleServer` to stop accepting new connections (via the atomic int and `Interrupt` call), but it still waits for any request that's currently handling to finish. And letting any request that's currently handling to finish is an important part of graceful shutdown.
> 
> Now we just close all the connections (this is true with either the goroutine approach or original map approach), which will cause any currently handling connection to be closed abruptly. from client's pov, they will not receive the full response for this request (it could be either they haven't received any response at all, or they received half of the response then the unexpected EOF happens), and this will cause those requests to fail (because we didn't let them to finish).
> 
> So when any server upgrade to this version, this will increase their error rate whenever autoscaling is scaling down. This is a worse scenario than graceful shutdown just timeouts (kubernnetes has a timeout to wait for graceful shutdown, so in current version as long as the server can finish all the pending requests in that time they won't fail any requests just because of a shutdown happens).
> 
> I don't know what's the best way to fix that. Maybe add a setting to `TSimpleServer` to sleep for a period of time before trying to abruptly close all connections is a ok short term solution. for long term we might need to make some breaking changes to `TProtocol`/`TTransport` to add the graceful shutdown semantic to them.
> 
> @dcelasun your thoughts?

buptubuntu:
sorry, wrong operation

buptubuntu:
> I don't know what's the best way to fix that. Maybe add a setting to `TSimpleServer` to sleep for a period of time before trying to abruptly close all connections is a ok short term solution. for long term we might need to make some breaking changes to `TProtocol`/`TTransport` to add the graceful shutdown semantic to them.

the key is to prevent the server to hang for too long or hang for ever，because that will block other some pipeline in a process

i think the former is a better choice

buptubuntu:
> > I don't know what's the best way to fix that. Maybe add a setting to `TSimpleServer` to sleep for a period of time before trying to abruptly close all connections is a ok short term solution. for long term we might need to make some breaking changes to `TProtocol`/`TTransport` to add the graceful shutdown semantic to them.
> 
> the key is to prevent the server to hang for too long or hang for ever，because that will block in some pipeline inside a process.As the listener is closed, new client can not connect to ther server(imaging a standal

> 

for the error rate, from the server's pov，the client may also close abruptly to cause server error.

Maybe we can add func called **CloseImmediatly** for user to choice

fishy:
>for the error rate, from the server's pov，the client may also close abruptly to cause server error.

That's much less important for the majority of servers and that's already supported by connectivity check: https://github.com/apache/thrift/blob/master/lib/go/README.md#a-note-about-server-handler-implementations

I don't know what's your production setup. For a typical stateless thrift server running on kubernetes, each individual servers (pods in kubernetes teminology) are running behind a shared "service" that's acting as load balancer. clients talks to this service instead of individual pods, and the service will choose a pod to connect the client to. When a pod is shutting down, the service will no longer assign any new clients to it, but will use other pods to handle those requests. The pods are also usually on horizontal auto scaler (hpa) that auto scales up and down based on the traffic.

In majority of the cases, there are only two scenarios a pod would be shutdown (this is where `TSimpleServer.Stop` is called):

1. the hpa decides to scale down (for example, you are moving away from peak business hours), so it reduce the number of pods (choose some pods to shut them down).
2. you are doing a rolling release, so the pods are being replaced.

In either case, graceful shutdown makes sure that no requests gets interrupted by the graceful shutdown, which is commonly referred to as "zero downtime". When kubernetes decides to kill a pod (in either case), it first sends a shutdown signal to your running process, then wait for shutdown timeout. Your code handles the signal, and call `TSimpleServer.Stop` as a result of that, and wait for the shutdown/cleanup to complete. The only downtime (failed requests) are caused by it takes longer than the shutdown timeout to finish (if the process does not stop after the shutdown timeout, kubernetes controller will just kill the process forcefully).

buptubuntu:
> > for the error rate, from the server's pov，the client may also close abruptly to cause server error.
> 
> That's much less important for the majority of servers and that's already supported by connectivity check: https://github.com/apache/thrift/blob/master/lib/go/README.md#a-note-about-server-handler-implementations
> 
> I don't know what's your production setup. For a typical stateless thrift server running on kubernetes, each individual servers (pods in kubernetes teminology) are running behind a shared "service" that's acting as load balancer. clients talks to this service instead of individual pods, and the service will choose a pod to connect the client to. When a pod is shutting down, the service will no longer assign any new clients to it, but will use other pods to handle those requests. The pods are also usually on horizontal auto scaler (hpa) that auto scales up and down based on the traffic.
> 
> In majority of the cases, there are only two scenarios a pod would be shutdown (this is where `TSimpleServer.Stop` is called):
> 
> 1. the hpa decides to scale down (for example, you are moving away from peak business hours), so it reduce the number of pods (choose some pods to shut them down).
> 2. you are doing a rolling release, so the pods are being replaced.
> 
> In either case, graceful shutdown makes sure that no requests gets interrupted by the graceful shutdown, which is commonly referred to as "zero downtime". When kubernetes decides to kill a pod (in either case), it first sends a shutdown signal to your running process, then wait for shutdown timeout. Your code handles the signal, and call `TSimpleServer.Stop` as a result of that, and wait for the shutdown/cleanup to complete. The only downtime (failed requests) are caused by it takes longer than the shutdown timeout to finish (if the process does not stop after the shutdown timeout, kubernetes controller will just kill the process forcefully).

Our application will receive config from a config server to stop the thrift server and setup    some other components instead without kill the process

buptubuntu:
> > for the error rate, from the server's pov，the client may also close abruptly to cause server error.
> 
> That's much less important for the majority of servers and that's already supported by connectivity check: https://github.com/apache/thrift/blob/master/lib/go/README.md#a-note-about-server-handler-implementations
> 
> I don't know what's your production setup. For a typical stateless thrift server running on kubernetes, each individual servers (pods in kubernetes teminology) are running behind a shared "service" that's acting as load balancer. clients talks to this service instead of individual pods, and the service will choose a pod to connect the client to. When a pod is shutting down, the service will no longer assign any new clients to it, but will use other pods to handle those requests. The pods are also usually on horizontal auto scaler (hpa) that auto scales up and down based on the traffic.
> 
> In majority of the cases, there are only two scenarios a pod would be shutdown (this is where `TSimpleServer.Stop` is called):
> 
> 1. the hpa decides to scale down (for example, you are moving away from peak business hours), so it reduce the number of pods (choose some pods to shut them down).
> 2. you are doing a rolling release, so the pods are being replaced.
> 
> In either case, graceful shutdown makes sure that no requests gets interrupted by the graceful shutdown, which is commonly referred to as "zero downtime". When kubernetes decides to kill a pod (in either case), it first sends a shutdown signal to your running process, then wait for shutdown timeout. Your code handles the signal, and call `TSimpleServer.Stop` as a result of that, and wait for the shutdown/cleanup to complete. The only downtime (failed requests) are caused by it takes longer than the shutdown timeout to finish (if the process does not stop after the shutdown timeout, kubernetes controller will just kill the process forcefully).

In fact，that just let the k8s handle the hang problem，but not dealed in thrift.

fishy:
>In fact，that just let the k8s handle the hang problem，but not dealed in thrift.

Yes, that's what we are trying to fix here. I'm just trying to explain to you why abruptly closing all connections from the server could be worse than the hang problem.

buptubuntu:
> explain

Yeah，I think there are generally two solutions:
1> add a func called: **CloseImmediatly**，it will close all clients, that will cause client err， but it is a seperate func, and user can choose.
2> add a var called **ServerCloseTimeout** like the ServerConnectivityCheckInterval，and its default value is 0，means that the server will wait all client closed even the underlying net.Conn is hanged; or if user set ServerCloseTimeout positive, the server will wait for a while before close all the clients

Which one do you think is a better choice?

buptubuntu:
> Yes, that's what we are trying to fix here. I'm just trying to explain to you why abruptly closing all connections from the server could be worse than the hang problem.

If the hang problem is worse than abruptly closing is not absolute，it depend on user， maybe user want a quick restart more than zero error even in k8s.

As for our situation, hang is absolutely worse than abruptly closing.

fishy:
@buptubuntu could you please clarify whether you are working on the new approach(-es)?

buptubuntu:
> @buptubuntu could you please clarify whether you are working on the new approach(-es)?

As I mentioned above:

Yeah，I think there are generally two solutions:
*  add a func called: CloseImmediatly，it will close all clients, that will cause client err， but it is a seperate func, and user can choose.
* add a var called ServerCloseTimeout like the ServerConnectivityCheckInterval，and its default value is 0，means that the server will wait all client closed even the underlying net.Conn is hanged; or if user set ServerCloseTimeout positive, the server will wait for a while before close all the clients

@fishy @dcelasun

fishy:
between the 2 I would certainly prefer the 2nd one.

dcelasun:
Likewise, the second one.

buptubuntu:
The CI seems fail on cpp test, does that matter? @fishy @dcelasun

dcelasun:
There are some unstable tests. As long as Go tests pass, it's fine.

fishy (changes requested):

lib/go/thrift/simple_server.go
```diff
@@ -234,7 +252,14 @@ func (p *TSimpleServer) Stop() error {
 	}
 	atomic.StoreInt32(&p.closed, 1)
 	p.serverTransport.Interrupt()
+
+	if ServerCloseTimeout > 0 {
+		<-time.After(ServerCloseTimeout)
+		close(p.stopChan)
+	}
```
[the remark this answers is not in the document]
buptubuntu: Reasonable

lib/go/thrift/simple_server.go
```diff
@@ -229,12 +247,29 @@ func (p *TSimpleServer) Serve() error {
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		p.wg.Wait()
+	}()
+
+	if ServerCloseTimeout > 0 {
+		select {
+		case <-time.After(ServerCloseTimeout):
```
fishy, line 265: Say `ServerCloseTimeout` is 1min and all clients closed after 5s, so we go with the other case, this timer will still occupy the resources for the whole 1min and only gets recycled after the timer has fired, even though it's no longer needed after 5s. This is why we should defer `timer.Stop()` so we can release the resources sooner.

lib/go/thrift/simple_server.go
```diff
@@ -47,16 +48,18 @@ var ErrAbandonRequest = errors.New("request abandoned")
 //
 // If it's changed to <=0, the feature will be disabled.
 var ServerConnectivityCheckInterval = time.Millisecond * 5
+var ServerCloseTimeout = time.Duration(0)
```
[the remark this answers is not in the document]
fishy: and please also add some explanation to the README file similar to https://github.com/apache/thrift/blob/master/lib/go/README.md#a-note-about-server-handler-implementations.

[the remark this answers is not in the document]
buptubuntu: done

lib/go/README.md
```diff
@@ -132,3 +132,17 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+The server stop will wait for all client connections to be closed after the last received or to be received request to be hanled, 
+so the time spent by stop may be too long in some situation:
+* when socket timeout is not set,server will be hanged before all active clients to finish handling the last received or to be received request.
+* when the socket timeout is too long saying one hour, server will be hanged for one hour before all active clients to finish handling the last received or to be received request.
+
+To avoid the server stop to be hanged for too long, you can take advantage of thrift.ServerCloseTimeout in your main or init function:
+
+    thrift.ServerCloseTimeout = <max_duration_to_stop>
+
+If it's changed to <=0, the feature will be disabled, and server will wait for all the client connections to be closed gracefully with zero err time.Otherwise, the stop will wait for all the client connections to be closed gracefully util thrift.ServerCloseTimeout is reached, and the client connections that are not closed after thrift.ServerCloseTimeout will be closed abruptly which may cause some client err rate.
```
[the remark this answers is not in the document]
fishy: also in markdown only a blank line is actual line break (paragraph separate), so you don't have to write the whole paragraph in a single line. I usually prefer to add line breaks after comma or period to keep each lines reasonably short.

lib/go/thrift/simple_server.go
```diff
@@ -48,15 +49,26 @@ var ErrAbandonRequest = errors.New("request abandoned")
...
 
+// ServerCloseTimeout defines max close wait duration used by
+// server stop to avoid hanging too long to wait for all client connections to be closed gracefully.
+//
+// It's defined as a variable instead of constant, so that thrift server
+// implementations can change its value to control the behavior.
+//
+// If it's changed to <=0, the feature will be disabled, and the server will wait for
+// for all the client connections to be closed gracefully.
+var ServerCloseTimeout = time.Duration(0)
```
fishy, line 60: On second thought a better name is probably `ServerStopTimeout`, since the function using it is `Stop`, not `Close`.

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,112 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
 	runtime.Gosched()
 	serv.Stop()
 }
+
+func TestNoHangDuringStopFromClientNoDataSendDuringAcceptLoop(t *testing.T) {
+	ln, _ := net.Listen("tcp", "localhost:0")
```
fishy, line 163: don't discard errors, even in a test, do something like this instead:
```suggestion
	ln, err := net.Listen("tcp", "localhost:0")
	if err != nil {
		t.Fatalf("Failed to listen: %v", err)
	}
```

same for the other test function.

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,112 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+	time.Sleep(10 * time.Millisecond)
+
+	netConn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil || netConn == nil {
+		t.Fatal("error when dial server")
+	}
+	time.Sleep(10 * time.Millisecond)
+
+	st := time.Now()
+	ServerCloseTimeout = 50 * time.Millisecond
```
fishy, line 203: nit: when a test involves changing the value of a global variable, I'd prefer to change its value back so we don't have tests relying on the default value becomes flaky because of the order of test executions, so this should be something like:
```suggestion
	backupServerCloseTimeout = ServerCloseTimeout
	t.Cleanup(func() {
		ServerCloseTimeout = backupServerCloseTimeout
	})
	ServerCloseTimeout = 50 * time.Millisecond
```

same for the other test.

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,112 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+	err = serv.Stop()
+	if err != nil {
+		t.Logf("error when stop server:%v", err)
+		t.FailNow()
+	}
+	if time.Since(st) < 50*time.Millisecond {
+		t.Logf("stop cost less time than socket timeout, server close timeout:%v,cost time:%v", ServerCloseTimeout, time.Since(st))
```
fishy, line 210: >less time than socket timeout

this is actually "stop/close timeout", not "socket timeout"

fishy: also calling `time.Since` again will give you a different value and it might be more than 50ms now and that will make the error message confusing. so do something like this instead:
```
if elapsed := time.Since(st); elapsed < 50*time.Millisecond {
  t.Errorf("...", ServerCloseTimeout, elapsed)
}
```

(also there's no reason to use `t.Logf` then `t.FailNow()`, just use `t.Errorf`)

fishy: also we should define `50*time.Millisecond` as a function scoped constant as we are using it several times inside this function.

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +158,112 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+	netConn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil || netConn == nil {
+		t.Fatal("error when dial server")
+	}
+	wg.Wait()
+
+	st := time.Now()
+	ServerCloseTimeout = 50 * time.Millisecond
+	err = serv.Stop()
+	if time.Since(st) >= 10*time.Millisecond {
```
fishy, line 261: same comments as the previous function, plus:

this check of `>=10ms` can make this test really flaky, because when the test is running on CI, the system can be resource constrained and it can actually take more than 10ms for `serv.Stop()` to return.

I would suggest to set `ServerCloseTimeout` to a much longer duration (say, 1 second), and fail the test if it takes more than half of that (so that would be 500ms, but just use `ServerCloseTimeout/2` here).

buptubuntu: Yes, I actually use time.Since(st) >= 6*time.Millisecond and the test result is not stable

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
 excessive cpu overhead.
 
 This feature is also only enabled on non-oneway endpoints.
+
+A note about server stop implementations
+========================================
+
+[TSimpleServer.Stop](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TSimpleServer.Stop) will wait for all client connections to be closed after 
+the last received or to be received request to be hanled, so the time spent by stop
```
fishy, line 140: 1. typo: `s/hanled/handled/`
2. "after the last received" sounds a bit weird to me.

I would word it like this:

```
When TSimpleServer.Stop is called,
it stops accepting any new client connections,
but will wait for all existing client connections to be closed.
The client connections will only be closed if:

1. The client closes the connection; or
2. The SocketTimeout set on the server is hit.

You can set global variable thrift.ServerStopTimeout early
in your main function to control how long `TSimpleServer.Stop`
would wait:

    // code example here

By default (when it's <=0), it waits indefinitely.
When you set it to a positive value (for example, 5 seconds),
TSimpleServer.Stop will close all the remaining client connections
after the timeout.
Please note that this will cause any ongoing requests to fail,
so in general you should set it to a value that's long enough for
your handlers to handle the requests.
```

lib/go/thrift/simple_server.go
```diff
@@ -48,15 +49,26 @@ var ErrAbandonRequest = errors.New("request abandoned")
 // If it's changed to <=0, the feature will be disabled.
 var ServerConnectivityCheckInterval = time.Millisecond * 5
 
+// ServerStopTimeout defines max close wait duration used by
```
fishy, line 52: nit: "max close wait" -> "max stop wait"

lib/go/thrift/simple_server_test.go
```diff
@@ -154,3 +160,130 @@ func TestNoHangDuringStopFromDanglingLockAcquireDuringAcceptLoop(t *testing.T) {
...
+	time.Sleep(networkWaitDuration)
+
+	wg.Add(1)
+	netConn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil || netConn == nil {
+		t.Fatal("error when dial server")
+	}
+	wg.Wait()
+
+	expectedStopTimeout := time.Second
```
fishy, line 273: semantically this is not the "expected" stop timeout (the actual expectation is around the socket timeout), so I would rename it to just `serverStopTimeout`.

fishy (approved):
looks good besides some documentation/naming nitpicks.

lib/go/README.md
buptubuntu: Emmm, current head will check if `closed` after calling `processor.Process` finish receving and handling a request in `TSimpleServer.processRequests`, if `closed` is set to 1, `inputTransport.Close()` and `outputTransport.Close()` will close the client, so i use the term `last received or to be received request`

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
 
 This feature is also only enabled on non-oneway endpoints.
+
+A note about server stop implementations
+========================================
+
+[TSimpleServer.Stop](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TSimpleServer.Stop) will wait for all client connections to be closed after 
+the last received or to be received request to be handled, so the time spent by stop
+ may be too long in some situation:
+* when socket timeout is not set,server will be hanged before all active
```
fishy, line 142: nit: add a space after `,` (and any other punctuation): `s/set,server/set, server/`

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+  clients to finish handling the last received or to be received request.
+* when the socket timeout is too long saying one hour, server will be
```
fishy, line 144: 1. `s/saying/say/`
2. `s/will/might/`, also it should be "be hanging" instead of "be hanged", I think?

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+* when the socket timeout is too long saying one hour, server will be 
+  hanged for one hour before all active clients to finish handling the
+  last received or to be received request.
+
+To avoid the server stop to be hanged for too long, you can take 
+advantage of thrift.ServerStopTimeout in your main or init function:
+
+    thrift.ServerStopTimeout = <max_duration_to_stop>
+
+If it's set to <=0, the feature will be disabled(by default), and server
```
fishy, line 153: nit: add space before `(`, also I _think_ from grammar perspective it should be just "(default)" instead of "(by default)".

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+  last received or to be received request.
+
+To avoid the server stop to be hanged for too long, you can take 
+advantage of thrift.ServerStopTimeout in your main or init function:
+
+    thrift.ServerStopTimeout = <max_duration_to_stop>
+
+If it's set to <=0, the feature will be disabled(by default), and server 
+will wait for all the client connections to be closed gracefully with 
+zero err time.Otherwise, the stop will wait for all the client
```
fishy, line 155: `s/time.Otherwise/time. Otherwise/`

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+To avoid the server stop to be hanged for too long, you can take 
+advantage of thrift.ServerStopTimeout in your main or init function:
+
+    thrift.ServerStopTimeout = <max_duration_to_stop>
+
+If it's set to <=0, the feature will be disabled(by default), and server 
+will wait for all the client connections to be closed gracefully with 
+zero err time.Otherwise, the stop will wait for all the client 
+connections to be closed gracefully util thrift.ServerStopTimeout is 
+reached, and the client connections that are not closed after thrift.
```
fishy, line 157: don't put line break after `thrift.` as that will cause a space being added (so it's rendered as `thrift. ServerStopTimeout`).

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+ServerStopTimeout will be closed abruptly which may cause some client err 
+rate.
```
fishy, line 159: "cause client errors", you can't cause a "err rate" :)

fishy (approved):
@dcelasun what do you think?

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
 excessive cpu overhead.
 
 This feature is also only enabled on non-oneway endpoints.
+
+A note about server stop implementations
+========================================
+
+[TSimpleServer.Stop](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TSimpleServer.Stop) will wait for all client connections to be closed after 
+the last received or to be received request to be handled, so the time spent by stop
+ may be too long in some situation:
```
dcelasun, line 141:
```suggestion
the last received request to be handled, as the time spent by Stop
 may sometimes be too long:
```

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+* when socket timeout is not set,server will be hanged before all active
+  clients to finish handling the last received or to be received request.
+* when the socket timeout is too long saying one hour, server will be 
+  hanged for one hour before all active clients to finish handling the
+  last received or to be received request.
```
dcelasun, line 146:
```suggestion
* When the socket timeout is too long (e.g one hour), server will 
  hang for that duration before all active clients finish handling the
  last received received request.
```

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
 excessive cpu overhead.
 
 This feature is also only enabled on non-oneway endpoints.
+
+A note about server stop implementations
+========================================
+
+[TSimpleServer.Stop](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TSimpleServer.Stop) will wait for all client connections to be closed after 
+the last received or to be received request to be handled, so the time spent by stop
```
dcelasun, line 140:
```suggestion
the last received request to be handled, so the time spent by stop
```

lib/go/README.md
```diff
@@ -132,3 +132,28 @@ if this interval is set to a value too low (for example, 1ms), it might cause
...
+ may be too long in some situation:
+* when socket timeout is not set,server will be hanged before all active
+  clients to finish handling the last received or to be received request.
+* when the socket timeout is too long saying one hour, server will be 
+  hanged for one hour before all active clients to finish handling the
+  last received or to be received request.
+
+To avoid the server stop to be hanged for too long, you can take 
+advantage of thrift.ServerStopTimeout in your main or init function:
```
dcelasun, line 149:
```suggestion
To prevent Stop from hanging for too long, you can set 
thrift.ServerStopTimeout in your main or init function:
```

lib/go/README.md
dcelasun: What does `with zero err time` mean here?

lib/go/README.md
dcelasun, line 157:
```suggestion
reached, and client connections that are not closed after thrift.
```

dcelasun (changes requested):

lib/go/README.md
buptubuntu: It means all connections is closed gracefully

buptubuntu:
@dcelasun any more suggestions?

dcelasun (approved):
LGTM

lib/go/README.md
```diff
@@ -132,3 +132,27 @@ if this interval is set to a value too low (for example, 1ms), it might cause
 excessive cpu overhead.
 
 This feature is also only enabled on non-oneway endpoints.
+
+A note about server stop implementations
+====                                            ====================================
```
fishy, line 137: what's with the extra spaces?

buptubuntu: fixed

buptubuntu:
@fishy any other suggestions?

PR #2498, opened 2022-01-06
Use iotest.OneByteReader instead of self implemented one

fishy:
Client: go

This is a trivial unit test improvement from the last commit. Of course
I only discovered the existence of testing/iotest package after I
re-invented the wheel.


dcelasun (approved):
> I only discovered the existence of testing/iotest package

First I've heard of it as well. I checked upstream to see if it was added recently but no, it's been there since 2009 :)

PR #2500, opened 2022-01-08
go: Make socketConn.Close thread-safe

fishy:
Client: go

We used to rely on setting the connection inside TSocket/TSSLSocket as
nil after Close is called to mark the connection as closed, but that is
not thread safe and causing TSocket.Close/TSSLSocket.Close cannot be
called concurrently. Use an atomic int to mark closure instead.

### DIFF
--- a/lib/go/thrift/socket_conn.go
+++ b/lib/go/thrift/socket_conn.go
@@ -21,6 +21,7 @@ package thrift
 
 import (
 	"net"
+	"sync/atomic"
 )
 
 // socketConn is a wrapped net.Conn that tries to do connectivity check.
@@ -28,6 +29,7 @@ type socketConn struct {
 	net.Conn
 
 	buffer [1]byte
+	closed int32
 }
 
 var _ net.Conn = (*socketConn)(nil)
@@ -64,7 +66,7 @@ func wrapSocketConn(conn net.Conn) *socketConn {
 // It's the same as the previous implementation of TSocket.IsOpen and
 // TSSLSocket.IsOpen before we added connectivity check.
 func (sc *socketConn) isValid() bool {
-	return sc != nil && sc.Conn != nil
+	return sc != nil && sc.Conn != nil && atomic.LoadInt32(&sc.closed) == 0
 }
 
 // IsOpen checks whether the connection is open.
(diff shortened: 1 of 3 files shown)


dcelasun (approved):

lib/go/thrift/socket_conn.go
```diff
@@ -100,3 +102,12 @@ func (sc *socketConn) Read(p []byte) (n int, err error) {
 
 	return sc.Conn.Read(p)
 }
+
+func (sc *socketConn) Close() error {
+	if !sc.isValid() {
+		// Already closed
+		return net.ErrClosed
+	}
+	atomic.StoreInt32(&sc.closed, 1)
```
buptubuntu, line 111: Maybe atomic.CompareAndSwapInt32 is better

buptubuntu (changes requested):
only atomic.CompareAndSwapInt32 is atomic, first load and later store is not

Jens-G:
Quick question regarding 0.16.0: Wait for this to complete?

buptubuntu (changes requested):
only atomic.CompareAndSwapInt32 is atomic, first load and later store is not

fishy:
> only atomic.CompareAndSwapInt32 is atomic, first load and later store is not

while technically that's correct, that doesn't matter here.

the problem we are solving is that inside `TSocket.Close` and `TSSLSocket.Close`, we call underlying socket's close, and set the underlying socket to nil. The race condition is about two goroutines setting the socket to nil concurrently, not about the socket's Close being called concurrently, as that's already supported by the stdlib.

so the fix here is that we use atomic int being set to non-zero to replace the old socket is nil check, and eliminate the race condition.

so we can still get into the situation that two goroutines call `Close` concurrently, they both pass the `isValid` check, and get to the point of setting the atomic int to 1. this is still fine because two goroutines calling store is fine (the winner change it from 0 to 1 and the loser doesn't really make any changes), and then they will both call the underlying Close and that's still fine (the winner will close the socket and return nil error, the loser will get an already closed error).

we can change it to compare and swap here, and return the already closed error here directly (like in the failed `isValid` check case), but the already closed error returned by the underlying Close function contains extra info about the connection itself, which is slightly better than the "naked" error we are returning here.

PR #2502, opened 2022-01-10
THRIFT-5496: Generate sufficient digits after the decimal point to n…

rgumpertz:
…ot truncate numbers like 123.45e-22 to 0.0000000000000000

See Jira THRIFT-5496  

- [X ] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?


compiler/cpp/src/thrift/generate/t_erl_generator.cc
```diff
@@ -389,27 +390,36 @@ void t_erl_generator::close_generator() {
...
+    double_output_stream << std::fixed;
+    #if 0 //maintain compatibility with previous versions of emit_double_as_string that could generate extra trailing digits
+      const int min_precision = std::numeric_limits<double>::digits10 + 1;
+    #else //don't add unneeded 0's
+      const int min_precision = 1;
+    #endif
+    if (std::isfinite(value) && value != 0.0) {
+      double_output_stream <<
+          std::setprecision(std::max(min_precision,
+                                     std::numeric_limits<double>::digits10 + 1
```
emmenlau, line 416: Better have `std::numeric_limits<double>::digits10 + 1` in a const int variable at the top of the code?

compiler/cpp/src/thrift/generate/t_erl_generator.cc
```diff
@@ -389,27 +390,36 @@ void t_erl_generator::close_generator() {
...
   #else
-      double_output_stream << std::fixed;
+    //question from Thrift@Rick.Gumpertz.com: Does anybody know why we avoid scientific notation here?
+    double_output_stream << std::fixed;
+    #if 0 //maintain compatibility with previous versions of emit_double_as_string that could generate extra trailing digits
+      const int min_precision = std::numeric_limits<double>::digits10 + 1;
+    #else //don't add unneeded 0's
+      const int min_precision = 1;
+    #endif
+    if (std::isfinite(value) && value != 0.0) {
```
emmenlau, line 413: Rather than this check of `value != 0.0`, would it be better to check the value of `static_cast<int>(floor(log10(abs(value)))`? Its numerically not very stable to perform math with values very close to zero, so evaluating the *result* of the math may be better than the source?

emmenlau (changes requested):
Looks basically reasonable, with minor changes.
Note that I did *not* evaluate the code path for MSVC version < 2015. Is thrift still supporting such old compilers or should this code path be removed?

compiler/cpp/src/thrift/generate/t_erl_generator.cc
```diff
@@ -389,27 +390,36 @@ void t_erl_generator::close_generator() {
...
+      double_output_stream <<
+          std::setprecision(std::max(min_precision,
+                                     std::numeric_limits<double>::digits10 + 1
+                                     - static_cast<int>(floor(log10(abs(value))))
+                                    )
+                           );
+    } else {
+      double_output_stream << std::setprecision(min_precision);
+  }
```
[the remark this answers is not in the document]
rgumpertz: That's fine with me

compiler/cpp/src/thrift/generate/t_erl_generator.cc
rgumpertz: That would make the constant have wider scope when it is intrinsically limited to this code.  I would leave it where it is.  Nevertheless, I'll defer to you if you consider it important.

compiler/cpp/src/thrift/generate/t_erl_generator.cc
rgumpertz: I agree that, in general, equality checking for doubles is a bad idea.  In this case, however, it is very specifically testing for 0.0, not something near 0.0: a tiny number on either side of 0 should display with a whole lot of digits after the decimal point; 0.0 should not.  Note also that log10(0.0) is undefined!  If there were a std::iszero function in <cmath>, I would have used that, but there is not.  I see no reasonable alternative to what I wrote.

emmenlau: Ok I see your point here, and I did not consider that `log10(0.0)` is undefined. Lets leave this as it is.

compiler/cpp/src/thrift/generate/t_erl_generator.cc
emmenlau: Yes, I can see that there is a tradeoff with the good scoping. The reason for my suggestion is that `std::numeric_limits<double>::digits10 + 1` is a rather "complex" term that is used in multiple places. It could add the readability and ease future maintenance to assign it to a more expressive variable, like `default_floating_precision` (is that what it is?). That also makes the long bracketed term a bit shorter :-)
Its not strictly required, but it would help me read the code.

emmenlau:
Overall it looks good, thanks for the patch!

Jens-G:
```
[...]\include\stdint.h(48): warning C4005: 'INT8_MIN': macro redefinition 
[...]\include\stdint.h(49): warning C4005: 'INT16_MIN': macro redefinition
[...]\include\stdint.h(50): warning C4005: 'INT32_MIN': macro redefinition
[...]\include\stdint.h(52): warning C4005: 'INT8_MAX': macro redefinition
[...]\include\stdint.h(53): warning C4005: 'INT16_MAX': macro redefinition
[...]\include\stdint.h(54): warning C4005: 'INT32_MAX': macro redefinition
[...]\include\stdint.h(56): warning C4005: 'UINT8_MAX': macro redefinition
[...]\include\stdint.h(57): warning C4005: 'UINT16_MAX': macro redefinition
[...]\include\stdint.h(58): warning C4005: 'UINT32_MAX': macro redefinition
[...]\thrift/generate/t_generator.h(334): error C2589: '(': illegal token on right side of '::' 
[...]\thrift/generate/t_generator.h(334): error C2059: syntax error: '::' 
```

emmenlau:
Any update on the issue that @Jens-G found?

Jens-G:
@rgumpertz  Ping

Jens-G:
Consider sending another PR, closing this one unmerged due to inactivity.

PR #2507, opened 2022-01-20
THRIFT-5500 uncompilable code when a .thrift struct named 'System' is present

Jonas-Marty:
https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-5500

Fixes by adding `global::` in front of the call to `System.Object.Equals(...)`


emmenlau (approved):
Seems reasonable.

PR #2509, opened 2022-01-26
THRIFT-5504 CA2254 Message template should be compile time constant

Jens-G:
Client netstd
Patch: Jens Geyer


Jens-G:
also merged into 0.16.0

PR #2511, opened 2022-01-29
THRIFT-5506 + THRIFT-5505

Jens-G:
* THRIFT-5506 C26495 variable "t_field::req_" not initialized, t_field.h:40
Compiler General
Patch: Jens Geyer

* THRIFT-5505 error: 'close_generator' overrides a member function but is not marked 'override'
Client: netstd
Patch: Jens Geyer

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_netstd_generator.h
+++ b/compiler/cpp/src/thrift/generate/t_netstd_generator.h
@@ -69,16 +69,16 @@ public:
   map<string, int> get_keywords_list() const;
 
   // overrides
-  void init_generator();
-  void close_generator();
-  void generate_consts(vector<t_const*> consts);
+  void init_generator() override;
+  void close_generator() override;
+  void generate_consts(vector<t_const*> consts) override;
   void generate_consts(ostream& out, vector<t_const*> consts);
-  void generate_typedef(t_typedef* ttypedef);
-  void generate_enum(t_enum* tenum);
+  void generate_typedef(t_typedef* ttypedef) override;
+  void generate_enum(t_enum* tenum) override;
   void generate_enum(ostream& out, t_enum* tenum);
-  void generate_struct(t_struct* tstruct);
-  void generate_xception(t_struct* txception);
-  void generate_service(t_service* tservice);
+  void generate_struct(t_struct* tstruct) override;
+  void generate_xception(t_struct* txception) override;
+  void generate_service(t_service* tservice) override;
 
   // additional files
   void generate_extensions_file();
@@ -148,7 +148,7 @@ public:
   string func_name(t_function* tfunc, bool suppress_mapping = false);
   string func_name(std::string fname, bool suppress_mapping = false);
   string convert_to_pascal_case(const string& str);
-  string get_enum_class_name(t_type* type);
+  string get_enum_class_name(t_type* type) override;
 
 protected:
   std::string autogen_comment() override {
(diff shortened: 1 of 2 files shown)


jeffalder:
@Jens-G When I download the 0.16.0 source and build locally on my Mac, I get the THRIFT-5505 errors. Do you have any plans to release a 0.16.1 that includes this fix, by any chance?

We don't use the `netstd` generator so I stubbed it out by removing the files. That's how I was able to get it to build.

Jens-G:
> Do you have any plans to release a 0.16.1 that includes this fix, by any chance?

Not yet, no. 

You probably could cherry pick the patch without much problems yourself, or what am I overlooking?

jeffalder:
Yeah, it might be a better idea for us to maintain a fork.

Is there any way we can help prevent these kinds of problems in the future, before the release? Do you need additional integration tests, or specific git workflows or something?

Jens-G:
> Is there any way we can help prevent these kinds of problems in the future, before the release? Do you need additional integration tests, or specific git workflows or something?

Great question. It would help to see the builds fail in the existing CI runs. Not sure what needs to be changed (assuming its not a compiler specific thing) but that would certainly help.

jeffalder:
I was building on an M1 Mac. 

I see there are homebrew bottles built; how do those happen? Can you point me to that?

I also see in `.travis.yml` that the osx builds were causing timeouts, but that looks like a relatively old comment. How was that working?

Jens-G:
Don't know either. I sent a request to the dev mailing list.

jeffalder:
I see how homebrew is building bottles. The [homebrew recipe](https://github.com/Homebrew/homebrew-core/blob/master/Formula/thrift.rb#L69) includes [ENV.deparallelize](https://rubydoc.brew.sh/Superenv.html#deparallelize-instance_method), which removes the MAKEFLAGS environment variable. In my build log, I see this:

```
clang++ called with: -std=c++11 -DHAVE_CONFIG_H -I./src -Wall -Wextra -pedantic -Werror -g -O2 -MT src/thrift/generate/thrift-t_netstd_generator.o -MD -MP -MF src/thrift/generate/.deps/thrift-t_netstd_generator.Tpo -c -o src/thrift/generate/thrift-t_netstd_generator.o src/thrift/generate/t_netstd_generator.cc
superenv removed:  -Wall -Wextra -pedantic -Werror -g -O2
superenv added:    -pipe -w -Os -stdlib=libc++ -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk --sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -isystem/opt/homebrew/include -isystem/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -I/opt/homebrew/opt/openssl@1.1/include
superenv executed: clang++ -pipe -w -Os -std=c++11 -stdlib=libc++ -std=c++11 -DHAVE_CONFIG_H -I./src -MT src/thrift/generate/thrift-t_netstd_generator.o -MD -MP -MF src/thrift/generate/.deps/thrift-t_netstd_generator.Tpo -c -o src/thrift/generate/thrift-t_netstd_generator.o src/thrift/generate/t_netstd_generator.cc -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk --sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -isystem/opt/homebrew/include -isystem/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -I/opt/homebrew/opt/openssl@1.1/include
```

So `ENV.deparallelize` is removing the `-Wall -Wextra -pedantic -Werror` settings, which means that the command-line flag that triggered that error has been removed for homebrew builds.

PR #2513, opened 2022-02-03
THRIFT-5509: Document about IsOpen returns false

fishy:
Client: go

Document that IsOpen returns false does not necessarily mean the
connection is already explicitly closed.

This is Path 1 of THRIFT-5509.

[skip ci]


fishy:
NOTE: Only one of https://github.com/apache/thrift/pull/2513 and https://github.com/apache/thrift/pull/2514 should be merged, not both.

fishy:
Closing this one in favor of #2514

PR #2514, opened 2022-02-03
THRIFT-5509: Close connection in IsOpen

fishy:
Client: go

When the connectivity check failed in IsOpen, close the connection
explicitly to avoid connection leaks.

This is Path 2 of THRIFT-5509.

### DIFF
--- a/lib/go/thrift/socket_conn.go
+++ b/lib/go/thrift/socket_conn.go
@@ -20,6 +20,7 @@
 package thrift
 
 import (
+	"errors"
 	"net"
 	"sync/atomic"
 )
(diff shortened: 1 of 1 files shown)


fishy:
NOTE: Only one of https://github.com/apache/thrift/pull/2513 and https://github.com/apache/thrift/pull/2514 should be merged, not both.

lib/go/thrift/socket_conn.go
```diff
@@ -83,7 +84,17 @@ func (sc *socketConn) IsOpen() bool {
 	if !sc.isValid() {
 		return false
 	}
-	return sc.checkConn() == nil
+	if err := sc.checkConn(); err != nil {
+		if !errors.Is(err, net.ErrClosed) {
+			// The connectivity check failed and the error is not
+			// that the conncetion is already closed, we need to
```
dcelasun, line 90:
```suggestion
			// that the connection is already closed, we need to
```

dcelasun (approved):
LGTM

fishy:
We are testing an [equivalent fix](https://github.com/reddit/baseplate.go/pull/477) internally (on our client pool management logic). So far the test shows that it indeed fixes the leak, but we'll keep it running for the day and merge both fixes tomorrow morning.

fishy:
Merging this in now. Will also cherry-pick the fix into 0.16.0 branch.

PR #2519, opened 2022-02-08
THRIFT-5512: Update java dependencies

fishy:
Update java dependencies according to the suggestions from:
https://mvnrepository.com/artifact/org.apache.thrift/libthrift/0.15.0


fishy:
OK this version does not introduce new breakages in travis, we are still seeing the same old erl-tls failures.

Jens-G:
LGTM but I don't use Java so I can't really tell if it is correct

Jens-G (approved):

fishy:
cherry-picked into 0.16.0 branch as a8e3b758a.

Jens-G:
Thanks!

PR #2520, opened 2022-02-08
THRIFT-5509: Suppress noisy log from go's TSimpleServer

fishy:
Client: go

This is a follow up of 6f33b0470. After we proactively closed the client
connection, processor.Process could return NOT_OPEN as a result, and
those errors being logged will cause the log to be very noisy.

This will also be cherry-picked into 0.16.0 branch.


dcelasun (approved):

nanassito (approved):

fishy:
cherry-picked into `0.16.0` branch as 303139410.

PR #2522, opened 2022-02-11
THRIFT-5515: Oneway requests can stuck in TNonblockingServer with TSSLSocket.

tkovacs-2:
TConnection::workSocket reads all pending oneway requests.
In this way it is possible to avoid that situation when the openssl reads all data, potentially contain more requests, from the socket receive buffer and the workSocket() won't be triggered again after the first request is processed.


RandyAbernethy (commented):
Functionally looks good (switch from recursion to a loop and add additional check in the SOCKET_RECV case). The comment at line 481 should be updated to explain the continue rather than the call to workSocket(). Would be nice to respect the indent style and move the entire switch under the while indent(looks a little hacky with the double closing curly braces at line 576). Otherwise LGTM.

emmenlau:
Agree with @RandyAbernethy .

An additional comment: With the loop, the variables `int got, left, sent` and `uint32_t fetch` are not initialized again. I did not double-check that this is fine. @tkovacs-2 , did you check that they are always initialized to useful values in the loop? Otherwise, to be sure, move the declaration and initialization inside the loop?

tkovacs-2:
Thank you for the comments.
The variables are my mistake, I put back them into the loop. Anyway they initialized later with valid values, so they would be ok even outside.

tkovacs-2:
@RandyAbernethy @emmenlau
I have modified the PR according to your comments. Could you please review again?

RandyAbernethy:
LGTM

PR #2489, opened 2021-12-12
THRIFT-2059: Support for Enums in Python >= 3.4

kainjow:
Currently in Python, the generated enum classes for enum thrift objects, are not being used in deserialization (or serialization). This means that the string representations of enums are never used. Moving one step further than the patch contained in THRIFT-2059, this creates python IntEnum classes and retrieves the string representation on deserialization while storing the integer representation on serialization.

[Recreated original PR from #1788 by @constd]
  

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [ ] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [x] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #2523, opened 2022-02-16
THRIFT-5139: THRIFT-4181: Python3 type hints

cspwizard:
Add type hinting and native enums (IntEnum) for python generated code
  


kainjow:
Cool! There's a PR here for enum changes: #2489

Jens-G:
Stupid question (probably): Can we make it the default or what reason does the switch have? Are there any potential incompatibilities? Just asking.

cspwizard:
Just in case if somebody use lower versions of Python. However it make sense to  make it default, and for older versions users can use the previous version. For me both ways are good, what is your opinion? @Jens-G 

I'll rebase

Jens-G:
If there is a reason, leave it. I just want to make sure we don't introduce unnecessary switches. Seems not the case. As I said, stupid question :-)

cspwizard:
rebased and fixed a bug, for some reason on de/serialization it used enum variable as string

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2839,4 +2927,5 @@ THRIFT_REGISTER_GENERATOR(
     "                     Package prefix for generated files.\n"
     "    old_style:       Deprecated. Generate old-style classes.\n"
     "    enum:            Generates Python's IntEnum, connects thrift to python enums. Python 3.4 and higher.\n"
+    "    type_hints:        Generate type hints and type checks in write method.\n"
```
kainjow, line 2930: nit: can you make the spacing line up?
```suggestion
    "    type_hints:      Generate type hints and type checks in write method.\n"
```

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -127,6 +128,9 @@ class t_py_generator : public t_generator {
         gen_tornado_ = true;
       } else if( iter->first.compare("coding") == 0) {
         coding_ = iter->second;
+      } else if( iter->first.compare("type_hints") == 0) {
+        gen_type_hints_ = true;
+        gen_enum_ = true;
```
kainjow, line 133: should the help text below mention that the enum option is also enabled if this option is enabled?

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2742,8 +2765,73 @@ string t_py_generator::type_name(t_type* ttype) {
...
+      return "bool";
+    case t_base_type::TYPE_I8:
+    case t_base_type::TYPE_I16:
+    case t_base_type::TYPE_I32:
+    case t_base_type::TYPE_I64:
+      return "int";
+    case t_base_type::TYPE_DOUBLE:
+      return "float";
+    }
+  } else if (type->is_enum()) {
```
kainjow, line 2817: could this be combined with the check below since they both return the same value?

compiler/cpp/src/thrift/generate/t_py_generator.cc
cspwizard: added it, but it should be obvious as without enum type type hinting won't work.

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2742,8 +2765,71 @@ string t_py_generator::type_name(t_type* ttype) {
...
+    case t_base_type::TYPE_I32:
+    case t_base_type::TYPE_I64:
+      return "int";
+    case t_base_type::TYPE_DOUBLE:
+      return "float";
+    }
+  } else if (type->is_enum() || type->is_struct() || type->is_xception()) {
+    return type_name(type);
+  } else if (type->is_map()) {
+    return "dict[" + type_to_py_type(((t_map*)type)->get_key_type()) + ", " + type_to_py_type(((t_map*)type)->get_val_type()) + "]";
```
jeking3, line 2820: This should be using `Dict[...]` and `from typing import Dict` ?

Same below for `Set` and `List` ?

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2497,6 +2513,11 @@ void t_py_generator::generate_serialize_field(ostream& out, t_field* tfield, str
  * @param prefix  String prefix to attach to all fields
  */
 void t_py_generator::generate_serialize_struct(ostream& out, t_struct* tstruct, string prefix) {
+  if(gen_type_hints_) {
+    indent(out) << "if not type(" << prefix << ") is " << type_to_py_type(tstruct)
```
jeking3, line 2517: This is not a type hint, it is a run-time check.  I don't think it belongs in this set of changes.  This is likely to break code relying on duck typing such as mocking.

jeking3 (changes requested):
There's more than just type hint changes in here, I found a case that I think will break many things including mock unit tests.

compiler/cpp/src/thrift/generate/t_py_generator.cc
cspwizard: With Python 3.7 or newer, you can use dict as a generic type if you start your module with [from __future__ import annotations](https://www.python.org/dev/peps/pep-0563/), and as of Python 3.9, dict (as well as other standard containers) [supports being used as generic type even without that directive](https://www.python.org/dev/peps/pep-0585/).

compiler/cpp/src/thrift/generate/t_py_generator.cc
cspwizard: you are right duck typing wouldn't work, but I think it worse it. As this keeps the contract consistent. For example we have structs like this: 
```
struct Date { 
   1: i32 days_since_AD1
}

struct License {
    1: string code
    2: string name
    3: Date date
}
```

in case if we don't have this check we could set anything that have serialize method to the `date` field and probably it would be able to serialize, but when this message would be deserialized deserizer would fail (in the best case scenario) in the worst and most common there would be some unusable garbage in the data.

ShaneEverittM:
What's blocking this from going forward?

Jens-G:
> What's blocking this from going forward?

Pending change requests?

cspwizard:
> Pending change requests?

what requests? I missed them probably

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2742,8 +2765,71 @@ string t_py_generator::type_name(t_type* ttype) {
...
+  }
+
+  return "";
+}
+
+/**
+ * Converts the parse type to a Python type hint
+ */
+string t_py_generator::type_to_py_hint(t_type* type) {
+  return "typing.Optional[" + type_to_py_type(type) + "]";
```
ShaneEverittM, line 2788: Seems wrong that this is unconditionally `Optional`. For example, as a service argument type hint. This will cause the type hint to show None as  valid, but a server will throw an exception due to the missing argument.

cspwizard: server wouldn't throw as the argument will be there, optional means that it could be None, and it will be default value of an argument type on receive. 
However if you have a better solution you are welcome.

ShaneEverittM: When testing my server was returning an exception message if I provided None.

cspwizard: Can you give more details? I've tested it with c# server on both reference/value types, everything worked fine

ShaneEverittM: Sure, I'll run it when I get home. Ping me if I forget. It was a rust server and a python client with a single optional i32 argument on a service that just added 1 and returned it.

ShaneEverittM: Here's my setup.
Thrift file:
```thrift
service Increment {
    i32 add(1: i32 value);
}
```
Python Client:
```python
from serial import Serial
from thrift.protocol.TCompactProtocol import TCompactProtocol
from thrift.transport.TTransport import TFramedTransport, TFileObjectTransport

from commands.Increment import Client


def main():
    client = Client(TCompactProtocol(TFramedTransport(TFileObjectTransport(Serial("COM2", 115200)))))
    assert client.add(None) == 1, "Expect default value on server to be 0, returning 1"


if __name__ == '__main__':
    main()
```
Generated from [this](https://github.com/cspwizard/thrift/tree/feature/python3) branch, running:
```shell
./thrift --gen py:type_hints -out pyclient/ idl/commands.thrift
```

Rust Server:
```rust
use anyhow::Result;
use thrift::{
    protocol::{TCompactInputProtocol, TCompactOutputProtocol},
    server::TProcessor,
    transport::{TFramedReadTransport, TFramedWriteTransport}
};

use commands::{IncrementSyncHandler, IncrementSyncProcessor};
use rpc::VirtualCOMPort;
use thrift_import::import;

import!(commands);

fn main() -> Result<()> {
    let port = serialport::new("COM4", 115200).open_native()?;

    let read_port = port.try_clone_native().expect("Failed to duplicate handle");
    let write_port = port;

    let mut i_prot = TCompactInputProtocol::new(TFramedReadTransport::new(VirtualCOMPort(read_port)));
    let mut o_prot = TCompactOutputProtocol::new(TFramedWriteTransport::new(VirtualCOMPort(write_port)));

    let processor = IncrementSyncProcessor::new(Increment);

    loop {
        processor.process(&mut i_prot, &mut o_prot)?;
    }
}

struct Increment;

impl IncrementSyncHandler for Increment {
    fn handle_add(&self, value: i32) -> thrift::Result<i32> {
        Ok(value + 1)
    }
}
```

Where `VirtualCOMPort` is a wrapper that stubs `flush` with noop because my virtual COM port software doesn't support that operation for some reason. The wrapper looks like this:
```rust
use std::io::{self, Read, Write};

use serialport::COMPort;

pub struct VirtualCOMPort(pub COMPort);

impl Write for VirtualCOMPort {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        self.0.write(buf)
    }

    fn flush(&mut self) -> io::Result<()> {
        // Flushing does not work with virtual COM ports.
        Ok(())
    }
}

impl Read for VirtualCOMPort {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        self.0.read(buf)
    }
}
```
The rust is generated using a build script and an importing macro that I wrote, but it boils down to this call to the thrift compiler:
```shell
./thrift --gen rs -out $OUT_DIR ./idl/commands.thrift
```

When I run the client as shown above, the server returns this exception:
```shell
thrift.Thrift.TApplicationException: ProtocolError { kind: Unknown, message: "missing required field IncrementAddArgs.value" }
```
This is in conflict with the generated signature of `Client.add`:
```python
def add(self, value: typing.Optional[int]) -> typing.Optional[int]:
```
Let me know if you need any more information.

cspwizard: Here are some results of my investigation: rust implementation for some reason apply T_OPT_IN_REQ_OUT to input parameter and this cause this particular issue. Probably bug in rust implementation?

Jens-G: In short, service method arguments are not "optional". They are mandatory, at least that's the current de-facto standard with Thrift. Note that the syntax would allow both optional and default values but hardly any language (if any at all) has any support for it. Hence, given the current state I would vote for the optional here being incorrect.

Note I'm not saying this is the ultimative decision and law until end of times. It's just what is consistent with the rest of Thrift as it exists today. Thios certainly could be changed, but that will be some pile of work for sure.

cspwizard: OK, should we implement this check in "write" method too then?

cspwizard: updated PR, however left the TODO for making args really required with appropriate validation

cspwizard:
@Jens-G , @ShaneEverittM are there any other comments or it could be completed?

compiler/cpp/src/thrift/generate/t_py_generator.cc
arkuhn: Is this change request still pending @jeking3 ?

pnovotnak:
What remains to get this merged? It would be wonderful to see this land

roshanjrajan-zip:
@jeking3 Did you have any more feedback on this PR? Would love to have this! :D

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2497,6 +2524,11 @@ void t_py_generator::generate_serialize_field(ostream& out, t_field* tfield, str
  * @param prefix  String prefix to attach to all fields
  */
 void t_py_generator::generate_serialize_struct(ostream& out, t_struct* tstruct, string prefix) {
+  if(gen_type_hints_) {
+    indent(out) << "if not type(" << prefix << ") is " << type_to_py_type(tstruct)
```
arkuhn, line 2528: I think I agree with @jeking3  that this goes beyond type hinting and is a behavioral change that might deserve its own PR and discussion. Any opposition to backing this part of the diff out so the non-controversial type hints can land?

Jens-G:
This branch has conflicts that must be resolved
Conflicting files
compiler/cpp/src/thrift/generate/t_py_generator.cc

Jens-G:
This is obsolete now, right?

cspwizard:
obsolete

PR #2524, opened 2022-02-21
THRIFT-5520 [java lib] add support for converting to java optional

Jimexist:

### DIFF
--- a/lib/java/src/org/apache/thrift/Option.java
+++ b/lib/java/src/org/apache/thrift/Option.java
@@ -19,6 +19,8 @@
 
 package org.apache.thrift;
 
+import java.util.Optional;
+
 /**
  * Implementation of the Option type pattern
  */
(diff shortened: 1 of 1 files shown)


Jimexist:
as a follow up, see:
- https://github.com/apache/thrift/pull/2525

lib/java/src/org/apache/thrift/Option.java
```diff
@@ -54,6 +56,19 @@ public T or(T other) {
...
+    /**
+     * Turn this Option into Java 8 Optional type
+     * @return Java 8+ Optional Type
+     */
+    public Optional<T> toOptional() {
+        if (isDefined()) {
+            return Optional.of(get());
+        } else {
+            return Optional.empty();
+        }
```
ctubbsii, line 69: Could also do:

```suggestion
        return Optional.ofNullable(isDefined() ? get() : null);
```

ctubbsii (approved):
This looks good to me, with or without my proposed alternative

lib/java/src/org/apache/thrift/Option.java
Jimexist: I'd keep this as is because it is more explicit while the proposed change results in two if-comparison, one in here and another in the optional class itself

PR #2525, opened 2022-02-21
THRIFT-5521: [java gen] use jdk8 option type in java generator code

Jimexist:
use jdk8 option type in java generator code


ttzztztz:
I really like this change, it could be passed between functions which receives `java.util.Optional` without explicitly do type conversion 👍

Jimexist:
not sure if the build can be fixed by merging
- https://github.com/apache/thrift/pull/2522
first

Jimexist:
hi @Jens-G - any feedback on this? or maybe do you suggest someone else to cover this type of changes. Thanks!

ctubbsii (commented):
I like this, but would probably prefer it be on by default. I haven't used the `optional` keyword in my `.thrift` files yet, but would probably always want the JDK8 version. I'm guessing it's off by default because of backwards-compatibility... but Thrift has broken stuff like that before, so I wouldn't have a problem with it breaking it here.

Also, you could use `Optional.ofNullable` to simplify the generated code. However, the generator code might look more complicated, because there's two cases for the Thrift Option type and only one case for the JDK8 Optional type (unless the Thrift Option type has an equivalent).

Jimexist:
> I'm guessing it's off by default because of backwards-compatibility... but Thrift has broken stuff like that before, so I wouldn't have a problem with it breaking it here.

thanks for the feedback, regarding

> I'm guessing it's off by default because of backwards-compatibility... but Thrift has broken stuff like that before, so I wouldn't have a problem with it breaking it here.

currently thrift requires JDK8 so this won't introduce new breakage

Jimexist:
> but would probably prefer it be on by default.

@ctubbsii do you suggest that i [flip the default](https://github.com/apache/thrift/pull/2525/files#diff-c59dffe78e4250a1f0450c496300e137aa1631b60488108e237a292a1c2e6789R105) here in this PR?

ctubbsii:
> > but would probably prefer it be on by default.
> 
> @ctubbsii do you suggest that i [flip the default](https://github.com/apache/thrift/pull/2525/files#diff-c59dffe78e4250a1f0450c496300e137aa1631b60488108e237a292a1c2e6789R105) here in this PR?

I meant backwards-compatibility, in the sense that the generated code wouldn't have a different API, which could cause problems for user code that interfaced with the generated code.

I would wait for consensus before flipping. I am just one voice, and others may prefer to keep the original behavior they've become accustomed to. I have the luxury of not caring about the previous behavior, because I've never used optional fields in my IDL files before. I expect that my preference is in the minority.

Jimexist:
> > > but would probably prefer it be on by default.
> > 
> > 
> > @ctubbsii do you suggest that i [flip the default](https://github.com/apache/thrift/pull/2525/files#diff-c59dffe78e4250a1f0450c496300e137aa1631b60488108e237a292a1c2e6789R105) here in this PR?
> 
> I meant backwards-compatibility, in the sense that the generated code wouldn't have a different API, which could cause problems for user code that interfaced with the generated code.
> 
> I would wait for consensus before flipping. I am just one voice, and others may prefer to keep the original behavior they've become accustomed to. I have the luxury of not caring about the previous behavior, because I've never used optional fields in my IDL files before. I expect that my preference is in the minority.

Understood. I'd intend to _not_ flipping the default, so this and https://github.com/apache/thrift/pull/2524 are only _adding_ features, not altering any existing ones or changing default behaviors. Deprecation of default behaviors or legacy options can happen much later down the road.

Jimexist:
i'm not sure if the build failure is due to this change

Jens-G:
> i'm not sure if the build failure is due to this change

Strange. Could you force push to trigger CI again?

Jimexist:
> > i'm not sure if the build failure is due to this change
> 
> 
> 
> Strange. Could you force push to trigger CI again?

I've done that twice last week by rebasing onto master. Strange that I got no err or log details in the Travis CI.

Jimexist:
after rebasing it seems fixed

Jimexist:
> FWIW, I would strongly advise against rebasing as a mechanism for triggering CI builds. Rebasing makes it difficult for reviewers to see diffs with new changes to pull requests they've already reviewed, as it resets the changeset. It's very frustrating, especially for large pull requests.
> 
> As an alternative, committers can re-trigger the Travis and Mergeable checks using the "Checks" tab in the Pull Request. Appveyor builds can be re-triggered using Appveyor's dashboard (though, I think Appveyor should probably be replaced with GitHub Actions Windows builds).
> 
> Contributors can re-trigger builds by adding a new empty commit to their branch instead of rebasing. These empty commits do not need to be merged... they can be squashed out when the PR is accepted. GitHub's interface has a "Squash and merge" option that would do this, as well as let the committer doing the merge polish the commit message style. Of course, this could also be done on the command-line.

@ctubbsii that's a fair point - and I wasn't aware of such issues before so thanks for pointing out.

My reasoning for using rebase based approach for managing pull requests - (and I realize that such arguments can easily slide into a bikeshedding so just to be clear i'll be willing to accommodate either way of working as long as it's clearly mandated or preferred, just sharing here for knowledge):
1. it's clearer than say merging master back to the pull request branch, as there's no merge commit, and you always know that you are diffing against latest master
2. GitHub actually helps reviewers manage their review history - the "viewed" toggle button on the top right is something (afaik) that can persist across rebases - so no longer do you need to keep track of what's reviewed and what's changed - with that said, i acknowledge that GitHub isn't something that everyone uses for reviewing the code and i guess in this case i'll refrain from rebasing later

Jimexist:
Hi @Jens-G , any chance that this can be merged soon? Thanks!

Jens-G:
I'm not the only committer around.

Jens-G:
Could you place language-specific tests somewhere under /lib/java/... ? The /test folder is already a bit too crowded, and the idea behind was always that it contains cross-language stuff only. All lang-specific tests should really be in their respective lib folders.

ctubbsii:
> I'm not the only committer around.

I might be able to help some more. However, I'm not really sure what the normal procedures are in this community, so I've been uncomfortable doing more.

Jimexist:
@Jens-G and @ctubbsii thank you _both_ for all your time and efforts invested into reviewing my code last few weeks, I truly appreciate that.

As PMC (or any other PMC can chime in as well), do you think you can shed some light on how to better contribute here (both as a committer like @ctubbsii or non-committers like me). I can't find it neither in https://thrift.apache.org/about or some other links, but it maybe was just me.

Some examples where I don't know if I should go ahead and make changes:
1. there are JUnit 3 and 4 style unit tests in Java repo, but the mainstream version is now JUnit 5
2. some cross tests drivers in Java lib is not out of date and can't run with the python driver now - not sure if they should be removed

I know that it's non-trivial for projects like Thrift given its multi-language and not all members are familiar with all languages - but given I don't see a lot of reviews on Java side of the code lately I wonder if there's any mechanism to make sure enough coverage / attention / review there is for major languages.

Jens-G:
> there are JUnit 3 and 4 style unit tests in Java repo, but the mainstream version is now JUnit 5

Usually if something is going to get outdated it's a good decision to stop using the old stuff, simply to reduce maintenance burden. It's however always also a case-by-case decision. It depends. Whatever makes most sense. Especially as it is "only" about tests I probably would drop 3 and 4 altogether and only use 5, but please take that personal opinion with a grain of salt. 

> some cross tests drivers in Java lib is not out of date [...]

Any way to get that fixed?

> given I don't see a lot of reviews on Java side of the code lately I wonder if there's any 
> mechanism to make sure enough coverage / attention / review there is for major languages.

Yes, that's a constant fight. Some areas are covered by reviewers more than others. I also wonder a bit ... Java is a mainstream language, but we have less activity as say Go and Rust lately. On the other hand, you hardly can't people force to do voluntary work. 

>  you can shed some light on how to better contribute here 

Well. pretty simple actually. 

First, we have http://thrift.apache.org/docs/HowToContribute which covers the contribution part.

Patches and PRs are reviewed by committers either because someone picks it himself or gets asked and then does it. If all sides are happy, it gets merged and the JIRA ticket updated. Only breaking changes should be documented in CHANGES.md, all else in that document will be filled up when we do a release. Of course anyone else is also welcome to review stuff. Given enough eyes all bugs are shallow, as the saying goes. For each committer, it is hard to say if a change beyond a certain level of complexity is good or bad if it is in some subsystem he/she does not understand. One usually leaves these tickets to people more skilled in that area.

You might have some luck if you do a post to the dev list asking about a reviewer.

PR #2528, opened 2022-02-22
THRIFT-5574: add ubuntu focal build

Jimexist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@
 
 sudo: required
 # https://docs.travis-ci.com/user/reference/linux
-dist: xenial
+dist: focal
 language: cpp
 
 services:
@@ -47,7 +47,7 @@ env:
     - SCRIPT="cmake.sh"
     - BUILD_ARG=""
     - BUILD_ENV="-e CC=gcc -e CXX=g++ -e THRIFT_CROSSTEST_CONCURRENCY=4"
-    - DISTRO=ubuntu-bionic
+    - DISTRO=ubuntu-focal
     - BUILD_LIBS="CPP C_GLIB JAVA PYTHON TESTING TUTORIALS"  # only meaningful for CMake builds
     - TRAVIS_BUILD_STAGE=test
     # DOCKER_REPO (this works for all builds as a source for docker images - you can override for fork builds in your Travis settings)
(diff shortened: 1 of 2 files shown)


Jens-G:
```
The following packages have unmet dependencies:
 erlang : Depends: erlang-odbc
E: Unable to correct problems, you have held broken packages.
The command '/bin/sh -c apt-get install -y --no-install-recommends   `# Erlang dependencies`   erlang &&   wget https://s3.amazonaws.com/rebar3/rebar3 -O /usr/bin/rebar3 &&   chmod 755 /usr/bin/rebar3 &&   rebar3 --version' returned a non-zero code: 100
```

Jens-G (changes requested):
See above

Jens-G:
Still broken. Also there is a possible conflict with #2541 ahead.

Jimexist:
> Still broken. Also there is a possible conflict with #2541 ahead.

yes so let me wait for that to merge first and then rebase & fix

Jimexist:
i think the latest CI fails because ubuntu:focal ships with openSSL 1.1.1f instead of 1.1.1 in e.g. xenial. This requires some test changes.

Jimexist:
cc @jeking3 the original author of `SecurityTest`. The failing tests is https://app.travis-ci.com/github/apache/thrift/jobs/567449611 - i am not sure if we need to drop some support or need to update the way parameters are constructed

Jimexist:
probably need to be done after:
- #2588 
is merged first

Jimexist:
@Jens-G do you think this can be merged? the refresh script did run correctly. the main errors are the same with `master`, i.e. Java errors, for which I'm trying to find time to fix.

Jens-G:
+1

Jimexist:
FYI @Jens-G @fishy or anyone who's interested, i'm merging this, given the current state of CI, i'd rather be iterating towards this (plus bionic) and move away from xenial.

Jens-G:
I would really appreciate if we could squash commits that belong together again before merging.

Jimexist:
> I would really appreciate if we could squash commits that belong together again before merging. And label it with the ticket number, btw.

thanks for the suggestion, i'd make sure to follow it next time.

Jimexist:
> I would really appreciate if we could squash commits that belong together again before merging. And label it with the ticket number, btw.

I just double checked that I indeed used GitHub's squash commit button, and the hash is f027deeaa8bf21769d83a35ac37ede21f035c0ad which is a single commit onto master. it only shows here in the pull request as a multi-commit pull request while in the master branch it's one. Did i miss anything?

PR #2529, opened 2022-02-22
THRIFT-5524: [java lib] add jdk8-with-postfix for option_type

Jimexist:
java lib: add jdk8-with-postfix for option_type so that instead of just using jdk8 `Optional<T>` type, it'll generate methods with `Optional()` postfix, in parallel with the original getter methods.

this pull request is based on:
- #2525 

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -46,6 +46,9 @@ using std::vector;
 
 static const string endl = "\n"; // avoid ostream << std::endl flushes
 
+static const string thrift_option_class = "org.apache.thrift.Option";
+static const string jdk_option_class = "java.util.Optional";
+
 /**
  * Java code generator.
  *
@@ -68,7 +71,7 @@ public:
     sorted_containers_ = false;
     java5_ = false;
     reuse_objects_ = false;
-    use_option_type_ = false;
+    use_option_type_ = NO_OPTION_TYPE;
     undated_generated_annotations_  = false;
     suppress_generated_annotations_ = false;
     rethrow_unhandled_exceptions_ = false;
@@ -95,7 +98,15 @@ public:
         // keep both reuse_objects and reuse-objects (legacy) for backwards compatibility
         reuse_objects_ = true;
       } else if( iter->first.compare("option_type") == 0) {
-        use_option_type_ = true;
+        if( iter->second.compare("jdk8") == 0) {
+          use_option_type_ = JDK8;
+        } else if( iter->second.compare("jdk8withpostfix") == 0) {
+          use_option_type_ = JDK8_WITH_POSTFIX;
+        } else if( iter->second.compare("thrift") == 0 || iter->second.compare("") == 0) {
+          use_option_type_ = THRIFT;
+        } else {
+          throw "option_type must be 'jdk8', 'jdk8withpostfix', or 'thrift'";
+        }
       } else if( iter->first.compare("rethrow_unhandled_exceptions") == 0) {
         rethrow_unhandled_exceptions_ = true;
       } else if( iter->first.compare("generated_annotations") == 0) {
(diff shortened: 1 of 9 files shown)


ctubbsii (commented):
I'm not a fan of this. I think I'd like to have one or the other only. With the proposed changes in #2524, the postfix option seems unnecessary.

Jens-G:
Build is green, yet I'm unsure how to move forward with this.

Jimexist:
> Build is green, yet I'm unsure how to move forward with this.

thanks for the review @Jens-G 

both:
- this #2529 and
- #2525 

are additive, back-compatible changes, and #2529 is a subset of this (or this change is based on #2529). So either we can merge that first or merge this as is and close that.

Jens-G:
Ok thanks but I also mean w/regard to @ctubbsii 's comments.

Jimexist:
> Ok thanks but I also mean w/regard to @ctubbsii 's comments.

I think our use cases might differ.

Currently the setup (in this pull request) is to have four possible values for the option argument:
1. Nullable getters if argument is absent
1. Thrift option returning getters  if argument is present and the default case is this (back compatible)
1. JDK option returning getters if argument is present with jdk8 value
1. Both 1 and 3 except option returning getters will be defined with Optional postfix


I would argument 4 is needed because in my use case both Java code and kotlin code will be consumers and Java needs option returning getters to do proper option chaining while having null returning getters makes kotlin recognize the property as is and it does option chaining from there.

In all of the cases they are additive and users can choose from the behavior so I guess it'll be needed and useful.

ctubbsii:
> I would argument 4 is needed because in my use case both Java code and kotlin code will be consumers and Java needs option returning getters to do proper option chaining while having null returning getters makes kotlin recognize the property as is and it does option chaining from there.

I'm not so sure it's needed, because if you didn't have option 4, then:

1. If you select the original nullable types, the option chaining works fine in Kotlin, but in Java, you'd just do: `var x = Optional.ofNullable(MyThriftType.getMyField())`

2. If you select the JDK8 option, then everything works nicely in Java, but in Kotlin, I think you could just do something like: `val x: SomeType? = MyThriftType.getMyField().orElse(null)` (or one of the [other examples here](https://stackoverflow.com/q/38767376/196405))

So, even if you didn't have option 4, you'd still have a choice between two relatively convenient options.

So, I don't think having option 4 is particularly compelling, especially since it creates getters that do the same thing as option 3, but with a completely different name. Having both option 3 and option 4 is likely to be confusing, and over time, it's likely to incur technical debt, with very little benefit, other than to support users who are generating a library intended for use with both Kotlin and Java, which is a *very* niche use case.

If I were using Kotlin (and I'm not), I would probably use the JDK8 option (option 3), and convert them to Kotlin nullables using [this suggestion](https://stackoverflow.com/a/38767769/196405).

Jimexist:
> > I would argument 4 is needed because in my use case both Java code and kotlin code will be consumers and Java needs option returning getters to do proper option chaining while having null returning getters makes kotlin recognize the property as is and it does option chaining from there.
> 
> 
> 
> I'm not so sure it's needed, because if you didn't have option 4, then:
> 
> 
> 
> 1. If you select the original nullable types, the option chaining works fine in Kotlin, but in Java, you'd just do: `var x = Optional.ofNullable(MyThriftType.getMyField())`
> 
> 
> 
> 2. If you select the JDK8 option, then everything works nicely in Java, but in Kotlin, I think you could just do something like: `val x: SomeType? = MyThriftType.getMyField().orElse(null)` (or one of the [other examples here](https://stackoverflow.com/q/38767376/196405))
> 
> 
> 
> So, even if you didn't have option 4, you'd still have a choice between two relatively convenient options.
> 
> 
> 
> So, I don't think having option 4 is particularly compelling, especially since it creates getters that do the same thing as option 3, but with a completely different name. Having both option 3 and option 4 is likely to be confusing, and over time, it's likely to incur technical debt, with very little benefit, other than to support users who are generating a library intended for use with both Kotlin and Java, which is a *very* niche use case.
> 
> 
> 
> If I were using Kotlin (and I'm not), I would probably use the JDK8 option (option 3), and convert them to Kotlin nullables using [this suggestion](https://stackoverflow.com/a/38767769/196405).

It's a matter of doing more in the generated code or less and in this case I actually agree with you that keeping tech debt low in the generated code is preferred.

I hold no strong position in merging this so if 3 is to be merged I'm okay to abandon this approach. Let me close this pull request.

PR #2530, opened 2022-02-22
THRIFT-5525: java gen to use reuse_objects instead of reuse-objects as a consistent param casing

Jimexist:
java gen to use reuse_objects instead of reuse-objects as a consistent param casing, but keep supporting original naming; motiviation is that almost all current params are using `_` not `-`

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -90,7 +90,8 @@ public:
         sorted_containers_ = true;
       } else if( iter->first.compare("java5") == 0) {
         java5_ = true;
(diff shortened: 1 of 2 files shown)


Jens-G (changes requested):
Proposal: Accept both forms to prevent breaking changes

Jimexist:
> Proposal: Accept both forms to prevent breaking changes

> Proposal: Accept both forms.

Hi @Jens-G thanks for the suggestion. in fact it's not a breaking change, as in https://github.com/apache/thrift/pull/2530/files#diff-c59dffe78e4250a1f0450c496300e137aa1631b60488108e237a292a1c2e6789R93-R94 both forms are indeed accepted. however do you think i should include this information in the document (`-help` print out) instead?

PR #2531, opened 2022-02-22
THRIFT-5526: java gen to use private_members instead of private-members as a consistent param casing

Jimexist:
java gen to use private_members instead of private-members as a consistent param casing  

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -78,7 +78,8 @@ public:
         bean_style_ = true;
       } else if( iter->first.compare("android") == 0) {
         android_style_ = true;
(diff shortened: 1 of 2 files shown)


Jens-G:
You are being aware that you are about to introduce a breaking change? Proposal: Accept both forms.

Jens-G (changes requested):
Proposal: Accept both forms.

Jimexist:
> Proposal: Accept both forms.

Hi @Jens-G thanks for the suggestion. in fact it's not a breaking change, as in https://github.com/apache/thrift/pull/2531/files#diff-c59dffe78e4250a1f0450c496300e137aa1631b60488108e237a292a1c2e6789R81-R82 both forms are indeed accepted. however do you think i should include this information in the document (`-help` print out) instead?

PR #2534, opened 2022-02-24
THRIFT-5519 Java async client loses exceptions in void methods

alex-kormukhin:
Fix [THRIFT-5519 Java async client loses exceptions in void methods](https://issues.apache.org/jira/browse/THRIFT-5519).

### DIFF
--- a/test/VoidMethExceptionsTest.thrift
+++ b/test/VoidMethExceptionsTest.thrift
@@ -1,0 +1,13 @@
+namespace java thrift.test.voidmethexceptions
+
+exception TExampleException {
+  1: required string message;
+}
+
+service TAppService01 {
+  string returnString(1: string msg, 2: bool throwException) throws (1:TExampleException error);
+  void returnVoidThrows(1: string msg, 2: bool throwException) throws (1:TExampleException error);
+  void returnVoidNoThrowsRuntimeException(1: string msg, 2: bool throwException);
+  void returnVoidNoThrowsTApplicationException(1: string msg, 2: bool throwException);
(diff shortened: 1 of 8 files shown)


test/VoidMethExceptionsTest.thrift
```diff
@@ -0,0 +1,13 @@
...
+exception TExampleException {
+  1: required string message;
+}
+
+service TAppService01 {
+  string returnString(1: string msg, 2: bool throwException) throws (1:TExampleException error);
+  void returnVoidThrows(1: string msg, 2: bool throwException) throws (1:TExampleException error);
+  void returnVoidNoThrowsRuntimeException(1: string msg, 2: bool throwException);
+  void returnVoidNoThrowsTApplicationException(1: string msg, 2: bool throwException);
+  oneway void onewayVoidNoThrows(1: string msg, 2: bool throwException);
```
Jens-G, line 12: Just to be clear: A oneway does not return anything. Including exceptions. There is no return. None.

alex-kormukhin: Yes, test checks exactly that (no client-side exceptions when server throws, no client-side wait when server hangs). Method name reflects correct signature for such methods.

alex-kormukhin:
Try to fix this with https://github.com/apache/thrift/pull/2536, but may be infrastructure problem (https://github.com/johnrengelman/shadow/issues/221).

Jens-G:
Could you force-push? That will trigger another CI run.

Jens-G:
Tons of "no @param for ..." warnings

PR #2536, opened 2022-02-28
THRIFT-5530 could not resolve plugin artifact 'com.github.johnrengelm…

alex-kormukhin:
Try to fix [THRIFT-5530 could not resolve plugin artifact 'com.github.johnrengelman.shadow:com.github.johnrengelman.shadow.gradle.plugin:4.0.4'](https://issues.apache.org/jira/browse/THRIFT-5530).

### DIFF
--- a/lib/java/build.gradle
+++ b/lib/java/build.gradle
@@ -39,7 +39,7 @@ plugins {
     id 'java'
     id 'maven'
     id 'signing'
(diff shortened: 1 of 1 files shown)


Jens-G:
That patch only introduces new problems, so I would say we leave it as it is for now.

PR #2537, opened 2022-03-01
THRIFT-5532: Fix headers bug in Thrift/HttpClient.pm

finncolman:
This fixes a bug in the headers being passed into HTTP::Request. Previously these were being set as a hash reference which is incorrect as it should be either an HTTP::Headers object or an array reference. The details are in the Jira issue.


finncolman:
The Travis CI build failure doesn't seem to have anything to do with my PR?

PR #2539, opened 2022-03-05
THRIFT-5283 add support for Unix Domain Sockets in lib/rs

tokcum:
Client: rs

The Thrift crate does not support Unix Domain Sockets (UDS). After reviewing and expressing my thoughts on THRIFT-3877, I added support for UDS.

This is not a breaking change. Inspired by how actix_web::HttpServer supports UDS, I added an additional fn to listen to UDS. I reorganized some other existing code to be able to reuse it for UDS. I also had to implement the trait TIoChannel for UnixStream. This went to transport::socket.

As UnixStream is only available on Unix, I tagged the code with #[cfg(unix)]. However, I was not able to test the build on Windows. I hope for the CI to detect any build issues on Windows.

### DIFF
--- a/lib/rs/src/server/threaded.rs
+++ b/lib/rs/src/server/threaded.rs
@@ -17,9 +17,13 @@
 
 use log::warn;
 
-use std::net::{TcpListener, TcpStream, ToSocketAddrs};
+use std::net::{TcpListener, ToSocketAddrs};
+use std::path::Path;
 use std::sync::Arc;
 use threadpool::ThreadPool;
+
+#[cfg(unix)]
+use std::os::unix::net::UnixListener;
 
 use crate::protocol::{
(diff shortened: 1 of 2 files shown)


tokcum:
I reviewed the CI results. The build failures are not related to this PR, I think. Meanwhile I also cross compiled lib/rs to Windows with the following cargo targets:

- i686-pc-windows-gnu
- i686-pc-windows-msvc
- x86_64-pc-windows-gnu
- x86_64-pc-windows-msvc

This releaved that I should have also tagged the use statement for Path.

```
warning: unused import: `std::path::Path`
  --> src/server/threaded.rs:21:5
   |
21 | use std::path::Path;
   |     ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

Jens-G:
You picked the wrong ticket number.

PR #2543, opened 2022-03-10
workaround MSVC compiler bug

srnwk:
On recent MSVC versions, the thrift compiler fails to build under C++17 or C++20 -- as far as I can tell this is a bug in MSVC.
Splitting that clever "read the whole file into a string"-line into several seems to avoid it, with no functional change.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_generator.h
+++ b/compiler/cpp/src/thrift/generate/t_generator.h
@@ -419,7 +419,9 @@ public:
     old_file.open(output_file_path.c_str(), std::ios::in);
 
     if (old_file) {
-      std::string const old_file_contents(static_cast<std::ostringstream const&>(std::ostringstream() << old_file.rdbuf()).str());
+      std::ostringstream oss;
+      oss << old_file.rdbuf();
+      std::string const old_file_contents(oss.str());
       old_file.close();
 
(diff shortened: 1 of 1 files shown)


Jens-G:
I use it with VS2022 and had not experienced any issues. 

Could you enlighten us what the problem is exactly? "fails fails to build" is a bit too vague (and grammatically questionable :-))

srnwk:
> I use it with VS2022 and had not experienced any issues.
> 
> Could you enlighten us what the problem is exactly? "fails fails to build" is a bit too vague (and grammatically questionable :-))

Here's a minimal repro https://godbolt.org/z/b76n9oaKv
```
<source>(9,102): error C2064: term does not evaluate to a function taking 0 arguments
        std::string const old_file_contents(static_cast<std::ostringstream const&>(std::ostringstream() << old_file.rdbuf()).str());
                                                                                                     ^
```

When compiling with /std:c++20 (or C++17 with /permissive-), this blows up with the above error.
(There's no reason I can tell why it should, clang and GCC agree).

emmenlau:
Its an unlikely problem, but at the same time the fix seems relatively harmless (and makes the code even a bit more readable). I'm not opposed to merging. @Jens-G what do you think?

PR #2545, opened 2022-03-11
THRIFT-5283: add support for Unix Domain Sockets in lib/rs

tokcum:
Client: rs

The Thrift crate does not support Unix Domain Sockets (UDS). After reviewing and expressing my thoughts on THRIFT-5283, I added support for UDS.

This is not a breaking change. Inspired by how actix_web::HttpServer supports UDS, I added an additional fn to listen to UDS. I reorganized some other existing code to be able to reuse it for UDS. I also had to implement the trait TIoChannel for UnixStream. This went to transport::socket.

As UnixStream is only available on Unix, I tagged the code with #[cfg(unix)]. I also cross compiled lib/rs to Windows with the following cargo targets:

    i686-pc-windows-gnu
    i686-pc-windows-msvc
    x86_64-pc-windows-gnu
    x86_64-pc-windows-msvc

Last but not least I ran the rust <-> rust test harnesses as well as the cross test harnesses. This looks good. At least, my contribution didn't introduce new fails.

-> already existed as THRIFT-5283
-> not a breaking change.

### DIFF
--- a/lib/rs/src/server/threaded.rs
+++ b/lib/rs/src/server/threaded.rs
@@ -17,9 +17,14 @@
 
 use log::warn;
 
-use std::net::{TcpListener, TcpStream, ToSocketAddrs};
+use std::net::{TcpListener, ToSocketAddrs};
 use std::sync::Arc;
 use threadpool::ThreadPool;
+
+#[cfg(unix)]
+use std::os::unix::net::UnixListener;
+#[cfg(unix)]
+use std::path::Path;
 
 use crate::protocol::{
     TInputProtocol, TInputProtocolFactory, TOutputProtocol, TOutputProtocolFactory,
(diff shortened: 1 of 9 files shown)


tokcum:
I tried to figure out why one of the Travis CI jobs failed. To be honest, I've no clue. I do not see a connection to the PR.

tokcum:
Actually this PR is a blocker for my osquery-rust crate. I appreciate any feedback on how to bring this PR forward.

allengeorge (commented):
Looks good! Suggest replacing use of comparison with "" with `Option` and `if let` for the UDS usages.

Thank you!

tokcum:
> Looks good! Suggest replacing use of comparison with "" with `Option` and `if let` for the UDS usages.
> 
> Thank you!

Thanks Allen, will do and update PR.

tokcum:
> Looks good! Suggest replacing use of comparison with "" with `Option` and `if let` for the UDS usages.
> 
> Thank you!

Surprisingly the linter discouraged me to use if let and pointed me to .is_none() just to complain again and proposing a match statement. This didn't happen when I compiled with cargo build but when I used make to run the tests. It looks like additional rules apply depending on which tools are used.

Just for reference, the two messages I received where:

```
error: redundant pattern matching, consider using `is_none()`

error: You checked before that `unwrap()` cannot fail. Instead of checking and unwrapping, it's better to use `if let` or `match`.
```

So I used the match statement.

allengeorge:
@tokcum Sorry for the delay, and thank you so much for your contribution!

PR #2548, opened 2022-03-15
THRIFT-5539: Decouple TDuplicateToProtocol out of TDebugProtocol

fishy:
Client: go

Currently go's TDebugProtocol does two things: log the read/writes, and
duplicate all the reads/writes to another TProtocol. For people who only
need the second feature, even if they use NopLogger for the logging
part, they still need to pay the price of all the fmt.Sprintf calls used
by logging, and on some cases those fmt.Sprintf calls alone can cause
significant CPU to be wasted (in one of our services fmt.Sprintf called
by TDebugProtocol used ~10% of CPU).

Create a dedicated TDuplicateToProtocol to reduce cpu waste, and mark
TDebugProtocol.DuplicateTo as deprecated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Apache Thrift Changelog
 
+## 0.17.0
+
+### Go
+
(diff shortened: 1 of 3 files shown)


dcelasun (approved):
Seems straightforward enough, though you should mention it in `CHANGES.md`

PR #2552, opened 2022-03-31
THRIFT-5543: add type parameter to java lib field metadata

Jimexist:
add type parameter to java lib field metadata


lib/java/src/org/apache/thrift/meta_data/FieldMetaData.java
```diff
@@ -68,21 +64,24 @@ public static void addStructMetaDataMap(Class<? extends TBase> sClass, Map<? ext
...
     // Similarly, do not trigger sClass.newInstance() while holding a lock on structMap,
     // it will lead to the same deadlock.
     // See: https://issues.apache.org/jira/browse/THRIFT-5430 for details.
-    if (!structMap.containsKey(sClass)){ // Load class if it hasn't been loaded
-      try{
-        sClass.newInstance();
-      } catch (InstantiationException e){
+    if (!structMap.containsKey(sClass)) { // Load class if it hasn't been loaded
+      try {
+        sClass.getDeclaredConstructor().newInstance();
```
Jimexist, line 75: `T.newInstance()` is deprecated since Java 9.

ctubbsii: Not the one from the constructor, only the one dangling off the class itself. The above should be fine.

lib/java/src/org/apache/thrift/meta_data/FieldMetaData.java
```diff
@@ -68,21 +64,24 @@ public static void addStructMetaDataMap(Class<? extends TBase> sClass, Map<? ext
...
+      } catch (InstantiationException e) {
         throw new RuntimeException("InstantiationException for TBase class: " + sClass.getName() + ", message: " + e.getMessage());
-      } catch (IllegalAccessException e){
+      } catch (IllegalAccessException e) {
         throw new RuntimeException("IllegalAccessException for TBase class: " + sClass.getName() + ", message: " + e.getMessage());
+      } catch (ReflectiveOperationException e) {
+        throw new RuntimeException("ReflectiveOperationException for TBase class: " + sClass.getName() + ", message: " + e.getMessage());
```
ctubbsii, line 81:
```suggestion
      } catch (ReflectiveOperationException e) {
        throw new RuntimeException(e.getClass().getSimpleName() + " for TBase class: " + sClass.getName()", e);
```

This can set the cause, so you don't need to serialize the message of the cause inside this exception's message. Also, both of the above exceptions are instances of ReflectiveOperationException, so they can be deleted in favor of just this one.

ctubbsii:
@Jimexist I did my best to merge master into this to resolve merge conflicts. I'm not sure if you have more work to do to this one.

Jimexist:
> @Jimexist I did my best to merge master into this to resolve merge conflicts. I'm not sure if you have more work to do to this one.

@ctubbsii thanks for the effort, i've fixed one merge error but now it's ready.

lib/java/src/main/java/org/apache/thrift/meta_data/FieldMetaData.java
```diff
@@ -103,8 +105,12 @@ public static Map<? extends TFieldIdEnum, FieldMetaData> getStructMetaDataMap(
                 + sClass.getName()
                 + ", message: "
                 + e.getMessage());
+      } catch (ReflectiveOperationException e) {
```
ctubbsii, line 108: I think this could be cleaned up further. This ReflectiveOperationException also replaces the above two individual catch clauses, which are both ReflectiveOperationExceptions.

ctubbsii (changes requested):

lib/java/src/main/java/org/apache/thrift/meta_data/FieldMetaData.java
Jimexist: @ctubbsii  thanks for the catch - fixed now. must've been due to merge conflict.

ctubbsii (approved):

PR #2553, opened 2022-04-01
THRIFT-5544: add java code gen param to support including field annotation as metadata

Jimexist:
add java code gen param to support including field annotation as metadata

Currently the code generator has a map of string to string field that denotes the "annotations" for a field: https://github.com/apache/thrift/blob/90ea2e8398eda32da8be0b3514516e7ad932a869/compiler/cpp/src/thrift/parse/t_field.h#L109

It is made available to generators such as JSON (in https://github.com/apache/thrift/blob/90ea2e8398eda32da8be0b3514516e7ad932a869/compiler/cpp/src/thrift/generate/t_json_generator.cc#L267-L274) to allow (runtime) interpretation of such metadata.

In Java, similar mechanism exists to register field metadata (via https://github.com/apache/thrift/blob/90ea2e8398eda32da8be0b3514516e7ad932a869/compiler/cpp/src/thrift/generate/t_java_generator.cc#L2862-L2865 which calls `org.apache.thrift.meta_data.FieldMetaData#addStructMetaDataMap` method to register into a centralized places. Such metadata is useful, e.g. the microservice framework Ameria [uses it][1] to dynamically proxy requests.

However the existing field metadata only contain information about field id, type, and name, but it does not contain annotations. Adding annotations will be useful for situations e.g. (these are useful enablement at user side, make possible if annotation is included)

- some static metadata is useful at runtime for dynamic operations, e.g. like what `deprecated` means in the context of a field, adding `obfuscated` allows the user site to decide to obfuscate detailed information
- allows for more richer and more type specific validation and transformation logic, e.g. since we can only have string as field type, adding `date_format="YYYY-MM-DD"` allows the thrift struct to be interpreted as a local date and thus can automatically be converted to java.time.LocalDate

This should be a back-compatible change as the default behavior is not to include the annotation. A second constructor can be added to allow included annotation to exist and register.

[1]: https://github.com/line/armeria/blob/5c2efebc7ba2350540d0554f530c173a49eb4c19/thrift0.13/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java#L553-L554

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -78,6 +78,7 @@ public:
     suppress_generated_annotations_ = false;
     rethrow_unhandled_exceptions_ = false;
     unsafe_binaries_ = false;
+    annotations_as_metadata_ = false;
     for (iter = parsed_options.begin(); iter != parsed_options.end(); ++iter) {
       if (iter->first.compare("beans") == 0) {
         bean_style_ = true;
@@ -124,6 +125,8 @@ public:
         }
       } else if (iter->first.compare("unsafe_binaries") == 0) {
         unsafe_binaries_ = true;
+      } else if (iter->first.compare("annotations_as_metadata") == 0) {
+        annotations_as_metadata_ = true;
       } else {
         throw "unknown option java:" + iter->first;
       }
@@ -188,6 +191,7 @@ public:
   void generate_java_struct_read_object(std::ostream& out, t_struct* tstruct);
   void generate_java_meta_data_map(std::ostream& out, t_struct* tstruct);
   void generate_field_value_meta_data(std::ostream& out, t_type* type);
+  void generate_metadata_for_field_annotations(std::ostream& out, t_field* field);
   std::string get_java_type_string(t_type* type);
   void generate_java_struct_field_by_id(ostream& out, t_struct* tstruct);
   void generate_reflection_setters(std::ostringstream& out,
@@ -439,6 +443,7 @@ private:
   bool suppress_generated_annotations_;
   bool rethrow_unhandled_exceptions_;
   bool unsafe_binaries_;
+  bool annotations_as_metadata_;
 };
 
 /**
@@ -2856,6 +2861,11 @@ void t_java_generator::generate_java_meta_data_map(ostream& out, t_struct* tstru
 
     // Create value meta data
     generate_field_value_meta_data(out, field->get_type());
+
+    // Include the annotation into metadata when asked
+    if (annotations_as_metadata_) {
+      generate_metadata_for_field_annotations(out, field);
+    }
     out << "));" << endl;
   }
 
@@ -2922,6 +2932,33 @@ std::string t_java_generator::get_java_type_string(t_type* type) {
                              + "\" passed to t_java_generator::get_java_type_string!");
     // This should never happen!
   }
+}
+
+void t_java_generator::generate_metadata_for_field_annotations(std::ostream& out, t_field* field) {
+  if (field->annotations_.size() == 0) {
+    return;
+  }
+  out << ", " << endl;
+  indent_up();
+  indent_up();
+  indent(out) << "java.util.stream.Stream.<java.util.Map.Entry<java.lang.String, "
+                 "java.lang.String>>builder()"
+              << endl;
+
+  indent_up();
+  indent_up();
+  for (auto& annotation : field->annotations_) {
+    indent(out) << ".add(new java.util.AbstractMap.SimpleImmutableEntry<>(\"" + annotation.first
+                       + "\", \"" + annotation.second + "\"))"
+                << endl;
+  }
+  indent(out) << ".build().collect(java.util.stream.Collectors.toMap(java.util.Map.Entry::getKey, "
+                 "java.util.Map.Entry::getValue))";
+  indent_down();
+  indent_down();
+
+  indent_down();
+  indent_down();
 }
 
 void t_java_generator::generate_field_value_meta_data(std::ostream& out, t_type* type) {
@@ -5740,4 +5777,6 @@ THRIFT_REGISTER_GENERATOR(
     "    generated_annotations=[undated|suppress]:\n"
     "                     undated: suppress the date at @Generated annotations\n"
     "                     suppress: suppress @Generated annotations entirely\n"
-    "    unsafe_binaries: Do not copy ByteBuffers in constructors, getters, and setters.\n")
+    "    unsafe_binaries: Do not copy ByteBuffers in constructors, getters, and setters.\n"
+    "    annotations_as_metadata:\n"
+    "                     Include Thrift field annotations as metadata in the generated code.\n")
--- a/lib/java/gradle/generateTestThrift.gradle
+++ b/lib/java/gradle/generateTestThrift.gradle
@@ -86,6 +86,7 @@ task generateJava(group: 'Build') {
     thriftCompile(it, 'EnumContainersTest.thrift')
     thriftCompile(it, 'JavaBinaryDefault.thrift')
     thriftCompile(it, 'VoidMethExceptionsTest.thrift')
+    thriftCompile(it, 'AnnotationTest.thrift')
     thriftCompile(it, 'partial/thrift_test_schema.thrift')
 }
 
@@ -132,4 +133,13 @@ task generateUnsafeBinariesJava(group: 'Build') {
     ext.outputBuffer = new ByteArrayOutputStream()
 
     thriftCompile(it, 'UnsafeTypes.thrift', 'java:unsafe_binaries', genUnsafeSrc)
+}
+
+task generateWithAnnotationMetadata(group: 'Build') {
+    description = 'Generate with annotation enabled and add to the default source'
+    generate.dependsOn it
+
+    ext.outputBuffer = new ByteArrayOutputStream()
+
+    thriftCompile(it, 'JavaAnnotationTest.thrift', 'java:annotations_as_metadata', genSrc)
 }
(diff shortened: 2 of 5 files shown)


Jens-G:
>  but a quick check shows it's defaults would change a lot of thrift code, and not always for the better. 
> I'm not sure if the Thrift community would endorse its use,

There were similar approaches a while ago and my (personal) feelings about automated C++ code formatting tools abilities are, well, let's call it "mixed feelings". But it's not only a C++ specific thing, I had the same experience with other tools and other languages as well. Sometimes the problems start with the style guide supplied by the manufacturer itself that ist a) the default and b) horribly broken. (Yes, I'm looking at you, Emba). 

The best tool I've seen in that regard is surely `gofmt`. The style actually makes sense to begin with, and it also offers not a single formatting-related option that could be configured. None. Zero. Kind of "get some sane style without the bikeshedding" approach.

Jimexist:
> > but a quick check shows it's defaults would change a lot of thrift code, and not always for the better.
> > I'm not sure if the Thrift community would endorse its use,
> 
> There were similar approaches a while ago and my (personal) feelings about automated C++ code formatting tools abilities are, well, let's call it "mixed feelings". But it's not only a C++ specific thing, I had the same experience with other tools and other languages as well. Sometimes the problems start with the style guide supplied by the manufacturer itself that ist a) the default and b) horribly broken. (Yes, I'm looking at you, Emba).
> 
> The best tool I've seen in that regard is surely `gofmt`. The style actually makes sense to begin with, and it also offers not a single formatting-related option that could be configured. None. Zero. Kind of "get some sane style without the bikeshedding" approach.

@Jens-G did you mean to comment on https://github.com/apache/thrift/pull/2559#pullrequestreview-938818662 instead?

Jimexist:
> 

I'd say that both [rustfmt](https://github.com/rust-lang/rustfmt) and gofmt are in this category and it contribute to the fact that Rust's one of the most love languages (per stackoverflow)

Jens-G:
> did you mean to comment on [...]

Probably. Must have confused open tabs.

ctubbsii (commented):
I think this could use a larger explanation. From these changes, including the added help documentation in the generator options, it's not clear what 'annotation metadata' is. What are common examples of annotation metadata, and how would these be used? The JIRA doesn't have any details in it, either: https://issues.apache.org/jira/browse/THRIFT-5544

Jimexist:
> I think this could use a larger explanation. From these changes, including the added help documentation in the generator options, it's not clear what 'annotation metadata' is. What are common examples of annotation metadata, and how would these be used? The JIRA doesn't have any details in it, either: https://issues.apache.org/jira/browse/THRIFT-5544

@ctubbsii thanks for the comment. detailed explanation added to jira ticket and inline

Jimexist:
> > > I think this could use a larger explanation. From these changes, including the added help documentation in the generator options, it's not clear what 'annotation metadata' is. What are common examples of annotation metadata, and how would these be used? The JIRA doesn't have any details in it, either: https://issues.apache.org/jira/browse/THRIFT-5544
> > 
> > 
> > @ctubbsii thanks for the comment. detailed explanation added to jira ticket and inline
> 
> Thanks for the additional information. However, I'm still confused. Is this intended to serialize information about Java annotations `@Override`, `@Deprecated`, `@Nullable`, etc.? Are these annotations in the Java code that is generated from an `*.thrift` IDL file? Is the annotation defined in the `*.thrift` file or expected to be manually added after code generation?
> 
> I'm just not clear on the end-to-end use of this. Do you have a simple example?

@ctubbsii i created a simple example at https://gist.github.com/Jimexist/7acffaa0c4967193937a05f56098853b with the newly added flag in this pull request.

ctubbsii:
> @ctubbsii i created a simple example at https://gist.github.com/Jimexist/7acffaa0c4967193937a05f56098853b with the newly added flag in this pull request.

Thanks. That example helps clear up some things. I was confused by the use of the terms "java" and "field annotation", because I was thinking [Java's Field `@Annotations`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Field.html#getAnnotation-java.lang.Class-).

I would call these "tags" rather than "annotations", in order to reduce confusion. However, if this is already an existing IDL language construct, then the name may already be established as "field annotations". If that's the case, I can propose some slight wording improvements in the gen-java parameter help doc.

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -78,6 +78,7 @@ class t_java_generator : public t_oop_generator {
     suppress_generated_annotations_ = false;
     rethrow_unhandled_exceptions_ = false;
     unsafe_binaries_ = false;
+    annotation_metadata_ = false;
```
ctubbsii, line 81: Maybe:
```suggestion
    annotations_as_metadata_ = false;
```

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -2922,6 +2932,33 @@ std::string t_java_generator::get_java_type_string(t_type* type) {
   }
 }
 
+void t_java_generator::generate_field_annotations(std::ostream& out, t_field* field) {
```
ctubbsii, line 2935:
```suggestion
void t_java_generator::generate_metadata_for_field_annotations(std::ostream& out, t_field* field) {
```

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -5737,4 +5774,6 @@ THRIFT_REGISTER_GENERATOR(
     "    generated_annotations=[undated|suppress]:\n"
     "                     undated: suppress the date at @Generated annotations\n"
     "                     suppress: suppress @Generated annotations entirely\n"
-    "    unsafe_binaries: Do not copy ByteBuffers in constructors, getters, and setters.\n")
+    "    unsafe_binaries: Do not copy ByteBuffers in constructors, getters, and setters.\n"
+    "    annotation_metadata:\n"
+    "                     Include annotation metadata in the generated code.\n")
```
ctubbsii, line 5779:
```suggestion
    "    annotations_as_metadata:\n"
    "                     Include annotations as metadata in the generated code.\n")
```

lib/java/gradle/generateTestThrift.gradle
```diff
@@ -133,3 +134,12 @@ task generateUnsafeBinariesJava(group: 'Build') {
 
     thriftCompile(it, 'UnsafeTypes.thrift', 'java:unsafe_binaries', genUnsafeSrc)
 }
+
+task generateWithAnnotation(group: 'Build') {
```
ctubbsii, line 138:
```suggestion
task generateWithAnnotationMetadata(group: 'Build') {
```

ctubbsii (commented):
Here's some proposed naming / phrasing changes to clear up potential confusion.

Jimexist:
> However, if this is already an existing IDL language construct, then the name may already be established as "field annotations".



> > @ctubbsii i created a simple example at https://gist.github.com/Jimexist/7acffaa0c4967193937a05f56098853b with the newly added flag in this pull request.
> 
> Thanks. That example helps clear up some things. I was confused by the use of the terms "java" and "field annotation", because I was thinking [Java's Field `@Annotations`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Field.html#getAnnotation-java.lang.Class-).
> 
> I would call these "tags" rather than "annotations", in order to reduce confusion. However, if this is already an existing IDL language construct, then the name may already be established as "field annotations". If that's the case, I can propose some slight wording improvements in the gen-java parameter help doc.

I agree it's confusing but i guess that ship is sailed: the name in C++ and also in all of the file names it's named as "annotation". I guess the best method is to try best to make it less ambiguous in Java.

Jimexist:
> However, if this is already an existing IDL language construct, then the name may already be established as "field annotations".



> > @ctubbsii i created a simple example at https://gist.github.com/Jimexist/7acffaa0c4967193937a05f56098853b with the newly added flag in this pull request.
> 
> Thanks. That example helps clear up some things. I was confused by the use of the terms "java" and "field annotation", because I was thinking [Java's Field `@Annotations`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Field.html#getAnnotation-java.lang.Class-).
> 
> I would call these "tags" rather than "annotations", in order to reduce confusion. However, if this is already an existing IDL language construct, then the name may already be established as "field annotations". If that's the case, I can propose some slight wording improvements in the gen-java parameter help doc.

I agree it's confusing but i guess that ship is sailed: the name in C++ and also in all of the file names it's named as "annotation". I guess the best method is to try best to make it less ambiguous in Java.

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -5737,4 +5774,6 @@ THRIFT_REGISTER_GENERATOR(
     "    generated_annotations=[undated|suppress]:\n"
     "                     undated: suppress the date at @Generated annotations\n"
     "                     suppress: suppress @Generated annotations entirely\n"
-    "    unsafe_binaries: Do not copy ByteBuffers in constructors, getters, and setters.\n")
+    "    unsafe_binaries: Do not copy ByteBuffers in constructors, getters, and setters.\n"
+    "    annotations_as_metadata:\n"
+    "                     Include annotations as metadata in the generated code.\n")
```
ctubbsii, line 5779: Just a tiny bit more clarification here would help I think.

```suggestion
    "                     Include Thrift field annotations as metadata in the generated code.\n")
```

ctubbsii (commented):
I have only this last final suggestion to help further clarify things in the documentation. But otherwise, this looks great to me.

ctubbsii (approved):

Jimexist:
had to do a rebase onto master due to conflict.

Jimexist:
seems like the build failure was due to `kerl` failing to compile erlang in time...

Jimexist:
now this is fixed, ready for merging

PR #2554, opened 2022-04-01
fix binary protocol's first byte value description in thrift-rpc.md

Blacker1230:

based on TBinaryProtocol.java, the strictWrite would write VERSION_1 | message.type  first. and the first byte would be 0x80. but in thrift-rpc.md, it's described as 0x81.

### DIFF
--- a/doc/specs/thrift-rpc.md
+++ b/doc/specs/thrift-rpc.md
@@ -151,7 +151,7 @@ always slightly faster.
 ## Compatibility
 
 A server could automatically determine whether a client talks the binary protocol or the compact protocol by
-investigating the first byte. If the value is `1000 0001` or `0000 0000` (assuming a name shorter than ±16 MB) it is the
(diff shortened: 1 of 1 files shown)


Blacker1230:
VERSION_1 = 0x80010000
int version = VERSION_1 | message.type;
writeI32(version);

Jens-G:
> VERSION_1 = 0x80010000
> int version = VERSION_1 | message.type;
> writeI32(version);

[   ] You did understand the difference between 2 bytes and an `int`

See also: https://github.com/apache/thrift/blob/master/doc/specs/thrift-binary-protocol.md#message

Jens-G:
Well, good catch. 👍 I would even extend it to the whole section being misleading. For example, inspecting the first byte can indicate compact, but it cannot indicate binary with 100% certainty, because the binary version in strict mode is two bytes.

PR #2555, opened 2022-04-04
THRIFT-5546: automate pypi publishing [ci skip]

jeking3:
When a pre-release is created on GitHub for the repo it will publish to the TestPyPI service, and if that pre-release is promoted to a release, it will publish to the PyPI (real) service.

Using the technique described at https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status to use third party github actions in the build.

Apache Infra team [added](https://issues.apache.org/jira/browse/INFRA-23070) the CLOUDTRUTH_API_KEY secret to the apache/thrift repo.  Apache didn't have a way to help us manage our own deployment keys so I used a third party service to make it easier for us.

Here's a pre-release test build:

https://github.com/apache/thrift/actions/runs/2091974166

For context here's the project and parameters (without secret values shown) in CloudTruth so you can see what they are defined as for names to correspond with the workflow:

image

---

- [X] If your change does not involve any code, include `[ci skip]` anywhere in the commit message to free up build resources.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,52 @@
+#
+# Apache Thrift release publishing workflow
+#
+# When a release is created in GitHub, it should be created in pre-release
+# mode first.  This will trigger a "prereleased" event which will cause this
+# workflow to run and publish packages to various package managers.  You
+# must check the Actions tab in GitHub to see the result of the workflow.
+#
+# github.event.action is either "prereleased" or "released" and corresponds
+# with a named environment in CloudTruth
+#
+#           |         Destinations
+# Language  | Prereleased   | Released
+# ----------+---------------+---------------
+# py        | test.pypi.org | pypi.org
+#
+---
+name: Publish
+
+on:
+  release:
+    types:
+      - prereleased
+      - released
+
+permissions:
+  contents: read
+
+jobs:
+  pypi:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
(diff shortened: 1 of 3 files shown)


jeking3:
This probably needs to change and put the pypi secrets into Apache's secret store directly so we can access them, instead of going through an intermediary.

SvenRoederer:
WOuld love to see this implemented, as currently pypi only offers Thrift 0.16.0 (which has been released some months ago)

Jens-G:
@jeking3 we're working on 0.18.1 but 0.18.0 is ready. No py changes in 0.18.1  so the choice is yours :-)

trams:
How can I help to make this possible? We would like to unblock thrift upgrades on our end

trams:
I think now there is a way to declare a Github Action trusted and avoid adding tokens anywhere. See this documentation https://docs.pypi.org/trusted-publishers/using-a-publisher/

Their example is using pypa/gh-action-pypi-publish@release/v1 which I am not sure would work correctly with our thrift repo because we have a package inside lib/py not the root directory
But I think we are not limited to use only this GitHub action

bwangelme:
@jeking3 

Hello, is there any progress on this issue?

We need a newer python thrift client which is great 0.16.

Jens-G:
@jeking3 Not sure how we should proceed here. Is this ready to be merged? Anything that we have to do?

trams:
I believe now we have thrift==0.21 in PyPi so I assumed someone completed this task

PR #2558, opened 2022-04-10
THRIFT-4100: Extract Rust read/write struct methods into a trait

zachary-kuhn:
Client: rs

In order for code to be generic over any generated Thrift object, extract the read and write methods into a `trait`.

[THRIFT-4100](https://issues.apache.org/jira/browse/THRIFT-4100)


martinvonz:
This seems to have made it into version 0.17.0. Any idea when that will be on crates.io?

PR #2559, opened 2022-04-12
THRIFT-5551: java generator to add @Override where possible

Jimexist:
java generator to add @Override where possible

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -1280,6 +1280,7 @@ void t_java_generator::generate_tuple_scheme_read_value(ostream& out, t_struct* 
   indent_up();
   indent(out) << "switch (setField) {" << endl;
   indent_up();
+  
   const vector<t_field*>& members = tstruct->get_members();
   vector<t_field*>::const_iterator m_iter;
 
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -318,6 +318,7 @@ public:
   std::string java_package();
   std::string java_suppressions();
   std::string java_nullable_annotation();
+  std::string java_override_annotation();
   std::string type_name(t_type* ttype,
                         bool in_container = false,
                         bool in_init = false,
@@ -468,6 +469,10 @@ string t_java_generator::java_nullable_annotation() {
   return "@org.apache.thrift.annotation.Nullable";
 }
 
+string t_java_generator::java_override_annotation() {
+  return "@java.lang.Override";
+}
+
 /**
  * Nothing in Java
  */
@@ -542,6 +547,7 @@ void t_java_generator::generate_enum(t_enum* tenum) {
   indent(f_enum) << " * Get the integer value of this enum value, as defined in the Thrift IDL."
                  << endl;
   indent(f_enum) << " */" << endl;
+  indent(f_enum) << java_override_annotation() << endl;
   indent(f_enum) << "public int getValue() {" << endl;
   indent(f_enum) << "  return value;" << endl;
   indent(f_enum) << "}" << endl << endl;
@@ -934,6 +940,7 @@ void t_java_generator::generate_union_constructor(ostream& out, t_struct* tstruc
   indent(out) << "  super(other);" << endl;
   indent(out) << "}" << endl;
 
+  indent(out) << java_override_annotation() << endl;
   indent(out) << "public " << tstruct->get_name() << " deepCopy() {" << endl;
   indent(out) << "  return new " << tstruct->get_name() << "(this);" << endl;
   indent(out) << "}" << endl << endl;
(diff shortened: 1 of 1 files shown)


Jimexist:
> These changes are harder to review than necessary, because they contain a lot of style changes. I would recommend minimizing the changes in this PR by retaining the existing style, so it's easier to review. For style standardization/improvements, I'd suggest creating a separate PR that does only that and no substantive changes.

Thanks for the suggestion - I wonder it too. The formatting changes were made by my editor, and I don't have a good way to separate it from the actual changes now.

If it's okay do you think I can create a separate, clean, style only change and merge that first. And then this change can be much cleaner? @ctubbsii

ctubbsii:
> If it's okay do you think I can create a separate, clean, style only change and merge that first. And then this change can be much cleaner?

Yeah, I think that would be a good option. If a standard style can be enforced by people with any editor, something like what the formatter-maven-plugin provides for Java code, that would be even better, I would think.

Jimexist:
> 

yes @ctubbsii the tool is `clang-format` and it's also centrally configured [here](https://github.com/apache/thrift/blob/master/.clang-format) - which is also what my editor formatting is relying on.

- I'm creating https://github.com/apache/thrift/pull/2560 as style only format to precede this change

ctubbsii:
@Jimexist I think you need to rebase and force-push to get this up-to-date.

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -468,6 +469,10 @@ string t_java_generator::java_nullable_annotation() {
   return "@org.apache.thrift.annotation.Nullable";
 }
 
+string t_java_generator::java_override_annotation() {
+  return "@java.lang.Override";
```
ctubbsii, line 473: Everything in `java.lang` is imported by default. You don't need the fully-qualified class names for anything in there. But, I see it's done elsewhere (for String). I suppose it doesn't hurt.

```suggestion
  return "@Override";
```

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -1273,7 +1280,6 @@ void t_java_generator::generate_tuple_scheme_read_value(ostream& out, t_struct*
   indent_up();
   indent(out) << "switch (setField) {" << endl;
   indent_up();
-
```
ctubbsii, line 1276: I think you lost a line in your change. I'm not sure if clang-format would re-add it. It'd be good to re-run it after your changes, just to be sure this PR is still well-formatted.

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -2108,8 +2117,8 @@ void t_java_generator::generate_java_struct_compare_to(ostream& out, t_struct* t
  *
  * @param tstruct The struct definition
  */
-void t_java_generator::generate_java_struct_reader(ostream& out, t_struct* tstruct) {
-  (void)tstruct;
+void t_java_generator::generate_java_struct_reader(ostream& out, t_struct* /*tstruct*/) {
```
ctubbsii, line 2120: You commented out the variable name, but left the variable's type. I'm mostly a Java dev. Is that a C++ quirk that's allowed for unused variables?

ctubbsii (commented):
Overall, this looks pretty good to me. I haven't verified that all the new uses are actually overrides, but it will be pretty obvious if they aren't... it will generate code that won't compile if they aren't.

compiler/cpp/src/thrift/generate/t_java_generator.cc
Jimexist: both are valid ways to suppress warnings about unused variables (in this case an error because `-Werror` flag is present).

https://stackoverflow.com/questions/308277/what-are-the-consequences-of-ignoring-warning-unused-parameter/308286#308286

i'd prefer the commenting out way because an extra line with `(void)` is a bit arcane.

compiler/cpp/src/thrift/generate/t_java_generator.cc
Jimexist: > it's done elsewhere (for String)

using fully qualified names (even for `java.lang.*`) allows for structs to be named `String` (although it's not a good practice) and used in the same package - otherwise it won't compile (some of the tests in `test/` folder does this).

For annotations i think it's fine to not to use fully qualified name.

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -5080,9 +5110,7 @@ t_java_generator::isset_type t_java_generator::needs_isset(t_struct* tstruct,
 }
 
 void t_java_generator::generate_java_struct_clear(std::ostream& out, t_struct* tstruct) {
-  if (!java5_) {
-    indent(out) << "@Override" << endl;
```
Jimexist, line 5084: `@Override` is `_since_ 5.0`

compiler/cpp/src/thrift/generate/t_java_generator.cc
ctubbsii: That's fine... I just didn't know you could comment out the variable name, but leave the type in place. You definitely can't do that in Java, but it does make sense... you're keeping the same signature just not using the variable. :smiley_cat:

ctubbsii (approved):

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -3172,6 +3189,7 @@ void t_java_generator::generate_service_async_client(t_service* tservice) {
   indent(f_service_) << "    this.clientManager = clientManager;" << endl;
   indent(f_service_) << "    this.protocolFactory = protocolFactory;" << endl;
   indent(f_service_) << "  }" << endl;
+  indent(f_service_) << "  @Override" << endl;
```
Jens-G, line 3192: That's a test, right? Seen and fixed.

PR #2565, opened 2022-04-13
Fix bug in THttpClient proxy handling

pspeter:
Previously, creating an instance of this class would throw whenever the http(s)_proxy environment variables were set, since it tried concatenating a string with a bytes array.

### DIFF
--- a/lib/py/src/transport/THttpClient.py
+++ b/lib/py/src/transport/THttpClient.py
@@ -100,7 +100,7 @@ class THttpClient(TTransportBase):
         ap = "%s:%s" % (urllib.parse.unquote(proxy.username),
                         urllib.parse.unquote(proxy.password))
         cr = base64.b64encode(ap.encode()).strip()
-        return "Basic " + cr
+        return "Basic " + cr.decode()
 
     def using_proxy(self):
         return self.realhost is not None


lib/py/src/transport/THttpClient.py
```diff
@@ -100,7 +100,7 @@ def basic_proxy_auth_header(proxy):
         ap = "%s:%s" % (urllib.parse.unquote(proxy.username),
                         urllib.parse.unquote(proxy.password))
         cr = base64.b64encode(ap.encode()).strip()
-        return "Basic " + cr
+        return "Basic " + cr.decode()
```
fishy, line 103: this is a python 2 vs. 3 issue. the old code works in python 2 and the new code works in python 3. currently we still officially support both so you probably need to find a way to make it work in both 2 and 3 (maybe `six` can help in this case?)

we do have a plan to drop python 2 support, so that won't be needed after we do that. but we just don't have the time to actually work on that yet (you are welcomed to take that on :) https://issues.apache.org/jira/browse/THRIFT-5537)

pspeter: Ah, that makes sense. I'll see if I can make this work using `six`.

pspeter:
@fishy the return type of the function should now always be `str` in Python 2 and 3, so it is consistent to the current behavior in Python2 while also fixing the function in Python 3.

fishy:
This is the correct _type_, yes, but it's not the correct string:

```
$ python3
Python 3.10.4 (main, Mar 24 2022, 13:07:27) [GCC 11.2.0] on linux
>>> import base64
>>> str(base64.b64encode(b'foo'))
"b'Zm9v'"
```

pspeter:
You're right, `ensure_str` it is.

Python 3:
```
>>> cr = b"abc"
>>> "Basic " + six.ensure_str(cr)
'Basic abc'
```


Python 2:
```
>>> cr = "abc"
>>> "Basic " + six.ensure_str(cr)
'Basic abc'
```

fishy (approved):
Thanks for the fix!

Can you also add a unit test for the `basic_proxy_auth_header` function?

Jimexist (approved):

PR #2575, opened 2022-04-18
THRIFT-5563: fix deprecation and enable xlint for java library

Jimexist:
fix deprecation and enable xlint for java library

### DIFF
--- a/lib/java/gradle/sourceConfiguration.gradle
+++ b/lib/java/gradle/sourceConfiguration.gradle
@@ -60,7 +60,16 @@ tasks.withType(JavaCompile) {
     if (JavaVersion.current() > JavaVersion.VERSION_1_8) {
         options.compilerArgs.addAll(['--release', '8'])
     }
-    // options.compilerArgs.addAll('-Xlint:unchecked')
+    options.compilerArgs.addAll([
+            '-Werror',
+            '-Xlint:deprecation',
+            '-Xlint:cast',
+            '-Xlint:empty',
+            '-Xlint:fallthrough',
+            '-Xlint:finally',
+            '-Xlint:overrides',
+            // we can't enable -Xlint:unchecked just yet
+    ])
 }
 
 // ----------------------------------------------------------------------------
--- a/lib/java/src/org/apache/thrift/partial/ThriftMetadata.java
+++ b/lib/java/src/org/apache/thrift/partial/ThriftMetadata.java
@@ -25,23 +25,20 @@ import org.apache.thrift.TFieldIdEnum;
 import org.apache.thrift.TFieldRequirementType;
 import org.apache.thrift.TUnion;
 import org.apache.thrift.meta_data.FieldMetaData;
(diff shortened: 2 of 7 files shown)


lib/java/gradle/sourceConfiguration.gradle
```diff
@@ -60,7 +60,16 @@ tasks.withType(JavaCompile) {
...
-    // options.compilerArgs.addAll('-Xlint:unchecked')
+    options.compilerArgs.addAll([
+            '-Werror',
+            '-Xlint:deprecation',
+            '-Xlint:cast',
+            '-Xlint:empty',
+            '-Xlint:fallthrough',
+            '-Xlint:finally',
+            '-Xlint:overrides',
+            // we can't enable -Xlint:unchecked just yet
```
ctubbsii, line 71: There might be a shorthand for this... something like: `-Xlint:all,-unchecked`
However, I haven't experimented with this enough to know for sure.

lib/java/src/org/apache/thrift/partial/ThriftMetadata.java
```diff
@@ -463,14 +463,12 @@ protected void toPrettyString(StringBuilder sb, int level) {
...
-      T instance = null;
+      T instance;
 
       try {
         Class<T> structClass = getStructClass(this.data);
-        instance = (T) structClass.newInstance();
-      } catch (InstantiationException e) {
-        throw new RuntimeException(e);
-      } catch (IllegalAccessException e) {
+        instance = structClass.newInstance();
```
ctubbsii, line 470: Should this use `.getDeclaredConstructor().newInstance()`? Or is that too new for Thrift's minimum Java version?

lib/java/src/org/apache/thrift/partial/ThriftMetadata.java
```diff
@@ -463,14 +463,12 @@ protected void toPrettyString(StringBuilder sb, int level) {
...
+      } catch (InstantiationException | IllegalAccessException e) {
```
ctubbsii, line 471: These could just catch all `ReflectiveOperationException`

ctubbsii (commented):
Seems like good improvements

lib/java/gradle/sourceConfiguration.gradle
Jimexist: thanks but i've tried locally, `'-Xlint:all,-unchecked',` does not work as it will still enforce `unchecked`.

in the longer term i'd plan to remove all unchecked code styles so the current explicit enumeration seems fine to me.

lib/java/src/org/apache/thrift/partial/ThriftMetadata.java
Jimexist: fixed @ctubbsii

lib/java/src/org/apache/thrift/partial/ThriftMetadata.java
Jimexist: fixed

PR #2579, opened 2022-04-19
Add missing underscore in call to ReadField_N

cptaffe:
Fix typo which caused undefined method errors in certain cases.

Generated code would encounter the following compilation error:
```
go/some_package/some_package-remote/some_package-remote.go:12:34: containerStruct0.ReadField1 undefined (type *some_package.SomeArgs has no field or method ReadField1)
```

The type of `containerStruct0` would instead have the following method:
```
func (p *SomeArgs)  ReadField_1(ctx context.Context, iprot thrift.TProtocol) error {
```


fishy:
@cptaffe can you please provide a minimal example of thrift file that causes this? In most cases we don't add underscore after `ReadField` so blindly adding an underscore would most likely be wrong.

fishy (changes requested):
@cptaffe I would suggest you to create a [jira ticket](https://issues.apache.org/jira/projects/THRIFT/issues) with an example thrift file to reproduce the issue. this is not a simple typo.

PR #2583, opened 2022-04-21
THRIFT-5570: update java lib document about gradle usage

Jimexist:
update java lib document about gradle usage


doc/install/debian.md
```diff
@@ -18,7 +18,7 @@ Debian 7/Ubuntu 12 users need to manually install a more recent version of autom
 If you would like to build Apache Thrift libraries for other programming languages you may need to install additional packages. The following languages require the specified additional packages:
 
  * Java
-	* packages: gradle 
+	* packages: gradle (version 6.9.2)
```
ctubbsii, line 21: I don't know that the Gradle version is a fixed requirement. Can clarify that this was the version tested:

```suggestion
	* packages: gradle (version 6.9.2 is confirmed to work)
```

lib/java/.gitignore
```diff
@@ -0,0 +1,53 @@
...
+gradle-app.setting
+
+# Cache of project
+.gradletasknamecache
+
+# Eclipse Gradle plugin generated files
+# Eclipse Core
+.project
+# JDT-specific (Eclipse Java Development Tools)
+.classpath
```
ctubbsii, line 45: Eclipse has a .settings/ directory also:

```suggestion
.classpath
.settings/
```

lib/java/README.md
```diff
@@ -42,11 +42,41 @@ The Thrift Java source is not build using the GNU tools, but rather uses
 the Gradle build system, which tends to be predominant amongst Java
 developers.
 
+Currently we use gradle 6.9.2 to build the Thrift Java source. The usual way to setup gradle
+project is to include the gradle-wrapper.jar in the project and then run the gradle wrapper to
+bootstrap setting up gradle binaries. However to avoid putting binary files into the source tree we
+have ignored the gradle wrapper files. You are expected to install it manually, as described in
+the [gradle documentation](https://docs.gradle.org/current/userguide/installation.html), or
+following this step (which is also done in the travis CI docker images):
```
ctubbsii, line 50: Minor punctuation, capitalization, and phrasing improvements:

```suggestion
Currently, we use Gradle 6.9.2 to build the Thrift Java source. The usual way to setup Gradle
project is to include the gradle-wrapper.jar in the project and then run the Gradle Wrapper to
bootstrap setting up Gradle binaries. However, to avoid putting binary files into the source tree, we
have excluded the Gradle Wrapper files. You are expected to install Gradle manually, as described in
the [Gradle documentation](https://docs.gradle.org/current/userguide/installation.html), or
following the following steps (which is also done in the Travis CI Docker images):
```

lib/java/README.md
```diff
@@ -42,11 +42,41 @@ The Thrift Java source is not build using the GNU tools, but rather uses
...
+# check binary integrity
+echo "8b356fd8702d5ffa2e066ed0be45a023a779bba4dd1a68fd11bc2a6bdc981e8f  /tmp/gradle-$GRADLE_VERSION-bin.zip" | sha256sum -c -
+# unzip and install
+unzip -d /tmp /tmp/gradle-$GRADLE_VERSION-bin.zip
+mv /tmp/gradle-$GRADLE_VERSION /usr/local/gradle
+ln -s /usr/local/gradle/bin/gradle /usr/local/bin
+```
+
+After the above step, `gradle` binary will be available in `/usr/local/bin/`. You can further choose
+to locally create the gradle wrapper (even if they are ignored) using:
```
ctubbsii, line 67:
```suggestion
After the above step, the `gradle` binary will be available in `/usr/local/bin/`. You may then choose
to locally create the Gradle Wrapper, if you wish to run commands with `gradlew` instead of `gradle`, using:
```

lib/java/README.md
```diff
@@ -42,11 +42,41 @@ The Thrift Java source is not build using the GNU tools, but rather uses
...
+
 To compile the Java Thrift libraries, simply do the following:
 
-    gradle
+```bash
+gradle
+```
 
-Yep, that's easy. Look for libthrift-<version>.jar in the build/libs directory.
+Yep, that's easy. Look for `libthrift-<version>.jar` in the build/libs directory.
```
ctubbsii, line 79: We probably shouldn't say it's easy... that's subjective. Users may not agree, and might interpret this as being arrogant.
```suggestion
Look for `libthrift-<version>.jar` in the build/libs directory.
```

ctubbsii (commented):
Doc changes look good to me. I made only minor suggestions.

Jimexist:
had to rebase to pickup fix from #2585

lib/java/.gitignore
```diff
@@ -0,0 +1,53 @@
+# Created by https://www.toptal.com/developers/gitignore/api/java,gradle
+# Edit at https://www.toptal.com/developers/gitignore?templates=java,gradle
+
```
ctubbsii, line 3: This change adds a new .gitignore file in the lib/java directory. This is consistent with the recently added one for lib/kotlin. However, they do differ in some ways. And, a lot of this is redundant with the top-level .gitignore file.

I think it'd be better to remove both of these .gitignore files and add the relevant patterns into the .gitignore file at the root of the repository, rather than add this here.

ctubbsii (changes requested):
The doc changes are fine, but I made a comment about the .gitignore files. I don't want to include the new .gitignore file here. I'd prefer to reconcile the differences with the one in the lib/kotlin directory and the root directory, and just merge it into the one at the root for consistency. Can be done as part of this PR, or a separate one. But, I don't think the new file should be added here.

lib/java/.gitignore
Jimexist: fixed in https://github.com/apache/thrift/pull/2583/commits/85c18bd6b15d45d34098e902c4337cdb7401ce2b

ctubbsii (approved):

PR #2541, opened 2022-03-07
THRIFT-5536 Fix cross language tests

nsuke:
* ~Update Erlang lib to be compatible for OTP-24~
  * ~`This` makes it incompatible to OTP-20 and less unfortunately~
* Pin OTP and rebar3 versions in Docker images
* Fix Go zlib cross tests

There were additional issues so I'm focusing on fixing CI in this PR.

### Failure1: dpkg deps

Dockerfile changes introduced this error in dpkg.sh
```
dpkg-checkbuilddeps: error: Unmet build dependencies: **erlang-base**
```
https://app.travis-ci.com/github/apache/thrift/jobs/562353250

Since we're not building any erlang dpkgs, I'm going to simply remove this from dpkg dependencies.

### Failure2: Go zlib cross test

```
Executing: /thrift/src/test/go/bin/testclient --protocol=binary --transport=zlib --zlib --port=35362
...
flag provided but not defined: -zlib
...
Return code: 2 (negative values indicate kill by signal)
```
https://app.travis-ci.com/github/apache/thrift/jobs/562353240#L21363

(Why it was working before in the first place?)
By explicitly parsing `--zlib` redundant flag, it works.

----

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [x] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #2585, opened 2022-04-22
THRIFT-5572: add travis_wait for rebuilding docker image

Jimexist:
because building image requires kerl to build erlang, which requires >10
min to build and meanwhile not printing to stdout, failing the travis
build

Currently the travis CI is broken because the step when kerl is building erlang OTP from source but the build process produces no output for >10 mins and travis will then fail the build if no output is detected.

The max number of build errors has been reached so now the travis CI build is suspended.

The easy fix is to allow travis to wait a bit longer (the erlang build itself takes around 1100 seconds).

this is a follow up on:
- https://github.com/apache/thrift/pull/2541

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,10 @@ services:
   - docker
 
 install:
-  - if [[ `uname` == "Linux" ]]; then build/docker/refresh.sh; fi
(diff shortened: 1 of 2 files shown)


Jimexist:
i tried to build erlang locally using kerl and it's still taking significant time - so i assume this fix isn't temporary and instead should be checked in for good...

PR #2592, opened 2022-04-28
THRIFT-5564: setup compiler-building github action [skip ci]

Jimexist:
setup a basic github action. this is part of THRIFT-5564 but does not completely close it yet.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [x] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [x] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #2593, opened 2022-04-28
THRIFT-5564: setup java and kotlin lib building

Jimexist:
setup java and kotlin lib building

this is a second step after:
- #2592 


ctubbsii:
I'm noticing the job name is quite long, which makes it hard to see the matrix params in the PR display. You can probably shorten it from "Build Thrift Compiler and Libraries" to just "Build" or similar.

You can also reduce the number of items in the matrix, and the output, by not making the Java distribution a matrix. It's probably sufficient to just use a single stable Java distribution. If temurin works, then just use that, but don't need to put it in the matrix.

You can also probably remove the Java 8 builds, since Java 11 should be using `--release 8` when it builds anyway, if the intent is to support Java 8. It'd be more useful to include Java 17 (which is the current LTS), and Java 11 (former LTS, still popular).

Jimexist:
> It'd be more useful to include Java 17 (which is the current LTS)

for now we can't because we are on gradle 6.9; we'll have to upgrade to gradle 7 first but that in turn replies on migrating from `maven` to `maven-publish` plugin

Jens-G:
Just want to drop the usual reminder of maintaining the ReleaseManagement.md because we will need this.

Jimexist:
> Just want to drop the usual reminder of maintaining the ReleaseManagement.md because we will need this.

thanks for the reminder @Jens-G, for now this and the currently open pull requests only adds CI part, not the CD part, also the travis CI is intact - so I have no changes to the `ReleaseManagement.md` just yet. When I do later make the final switch, I'll post to dev@thrift and update accordingly. For now, there's no change needed.

.github/workflows/build.yml
```diff
@@ -1,35 +1,28 @@
...
+name: 'Build'
+
 on:
   push:
     branches: [ '*' ]
   pull_request:
     branches: [ '*' ]
+
+env:
+  BUILD_DEPS: automake bison flex g++ git libboost-all-dev libevent-dev libssl-dev libtool make pkg-config
```
ctubbsii, line 10: Clever! I didn't know about `env`

ctubbsii (approved):

.github/workflows/build.yml
Jimexist: Thanks! Do you mind merging this some time later because I don't have write access

PR #2588, opened 2022-04-26
Remove python2

Jimexist:
Remove python2

- [ ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  (not required for trivial changes)
- [ ] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [ ] Did you squash your changes to a single commit?  (not required, but preferred)
- [ ] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #2596, opened 2022-04-29
THRIFT-5564: add GitHub action for python 2.x and 3.x

Jimexist:
- based on and closes https://github.com/apache/thrift/pull/2594
  


Jimexist:
before merging
- #2588 
@fishy do you want to have a look at this? this _might_ help my work on #2588 as well

.github/workflows/build.yml
```diff
@@ -128,3 +128,171 @@ jobs:
 
       - name: Run make kotlin
         run: make -C lib/kotlin
+
+  lib-rust:
```
fishy, line 132: this seems unrelated to what the pr title/description says it does? but since the jira ticket itself covers all languages I don't really see this as a problem :)

.github/workflows/build.yml
```diff
@@ -128,3 +128,171 @@ jobs:
...
+        run: |
+          chmod a+x compiler/cpp/thrift
+          compiler/cpp/thrift -version
+
+      - name: Run make rust
+        run: make -C lib/rs
+
+  lib-python:
+    needs: compiler
+    runs-on: ubuntu-20.04
```
fishy, line 202: should we also make this part of the matrix? my understanding is that this is not the only version of ubuntu we want to support.

.github/workflows/build.yml
```diff
@@ -128,3 +128,171 @@ jobs:
...
+          name: thrift-compiler
+          path: compiler/cpp
+
+      - name: Run thrift-compiler
+        run: |
+          chmod a+x compiler/cpp/thrift
+          compiler/cpp/thrift -version
+
+      - name: Run make python
+        run: make -C lib/py
```
fishy, line 298: does this run the unit tests or does it only make sure that the code builds?

.github/workflows/build.yml
Jimexist: yes it's actually based on:
- #2594 

but i think we can merge in one pass as well

.github/workflows/build.yml
Jimexist: i'm adding matrix to the compiler part but for python i think it's okay to have just one - given how libraries and compilers are in many cases compiled, installed, and used separately. the matrix should be focusing on multiple python versions for this part.

PR #2598, opened 2022-05-03
Update ordered-float dependency to latest version

eranrund:
Version 2.0.0 had RUSTSEC advisory, see https://github.com/rustsec/advisory-db/blob/main/crates/ordered-float/RUSTSEC-2020-0082.md

2.0.1 resolved the issue but I figured this is a good opportunity to move to the latest version.

I did not create a JIRA issue since this is a trivial change, but I am happy to do that if needed.

### DIFF
--- a/lib/rs/Cargo.toml
+++ b/lib/rs/Cargo.toml
@@ -16,7 +16,7 @@ keywords = ["thrift"]
 byteorder = "1.3"
 integer-encoding = "3.0"
 log = {version = "0.4", optional = true}
-ordered-float = "2.0"
+ordered-float = "3.0"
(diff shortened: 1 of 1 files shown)


Jimexist:
one thing we can do before merge is to rebase this onto or merge first:
- #2594

allengeorge:
@Jimexist Does your last comment still stand? Should the rust Github Actions PR be merged first, and then this one?

Jimexist:
> @Jimexist Does your last comment still stand? Should the rust Github Actions PR be merged first, and then this one?

my plan is to get:
- #2628 

merged first then i can fix and merge
- #2594 

which will then catch all future rust related code changes

PR #2602, opened 2022-05-08
THRIFT-5564: setup cross test in GitHub action

Jimexist:


.github/workflows/build.yml
```diff
@@ -131,11 +131,111 @@ jobs:
...
+          path: lib/kotlin
+
+      - name: Set back executable flags
+        run: |
+          chmod a+x \
+            lib/java/build/run* \
+            lib/kotlin/cross-test-client/build/install/TestClient/bin/* \
+            lib/kotlin/cross-test-server/build/install/TestServer/bin/*
+
+      - name: Run precross
```
Jimexist, line 223:
```suggestion
      - name: Run cross test
```

.github/workflows/build.yml
```diff
@@ -131,11 +131,111 @@ jobs:
...
+
+  cross-test:
+    needs:
+      - lib-java-kotlin
+    strategy:
+      matrix:
+        test_regex:
+          - "(binary)"
+          - "(header|json)"
+          - "(compact|multiplexed)"
```
Jimexist, line 190:
```suggestion
          - "-(binary)"
          - "-(header|json)"
          - "-(compact|multiplexed)"
```

.github/workflows/build.yml
```diff
@@ -131,11 +131,111 @@ jobs:
...
+
+  cross-test:
+    needs:
+      - lib-java-kotlin
+    strategy:
+      matrix:
+        test_regex:
+          - "-(binary)"
+          - "-(header|json)"
+          - "-(compact|multiplexed)"
```
ctubbsii, line 190: I wonder if the failure is because the leading dash is being treated as a CLI arg instead of part of the regex pattern?

```suggestion
          - "[-](binary)"
          - "[-](header|json)"
          - "[-](compact|multiplexed)"
```

.github/workflows/build.yml
```diff
@@ -131,11 +131,111 @@ jobs:
...
+
+  cross-test:
+    needs:
+      - lib-java-kotlin
+    strategy:
+      matrix:
+        test_regex:
+          - "[-](binary)"
+          - "[-](header|json)"
+          - "[-](compact|multiplexed)"
```
ctubbsii, line 190: These patterns seem to be working now, but I'm not quite clear on why have these run as a matrix build at all? Why not just run all of the tests?

Also, "multiplexed" and "header" patterns didn't seem to match on any tests. I'm not sure if these patterns cover all the tests, or if there's some tests that failed to match and therefore didn't run.

Jimexist: i'm currently just following up the travis setup, the split is also for the more anticipated cross tests in future.

i can revert back to having just one big test runs though.

ctubbsii: It's fine. I was just curious what the reasoning was.

ctubbsii (approved):

PR #2605, opened 2022-05-10
THRIFT-5569: Fix size check in TCompactProtocol.ReadListBegin

fishy:
Client: go

This was a copy-paste error in the original implementation in
6583f4e52345c3b05a76f0b188836599628356e8.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -3375,9 +3375,6 @@ void t_go_generator::generate_deserialize_container(ostream& out,
   // Declare variables, read header
   if (ttype->is_map()) {
     out << indent() << "_, _, size, err := iprot.ReadMapBegin(ctx)" << endl;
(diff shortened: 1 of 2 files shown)


Jens-G:
By checking error we are sure size is not negative now? Not sure if I getr the idea right.

fishy:
> By checking error we are sure size is not negative now? Not sure if I getr the idea right.

`checkSizeForProtocol` checks both that size is negative and size is too large (> max size allowed), and returns the proper TProtocolException if either case is true: https://github.com/apache/thrift/blob/dae1437ca03834527b654cf5bc8d7f41c9412a20/lib/go/thrift/configuration.go#L305-L319

it's used in all of `T{Binary|Compact|JSON}Protocol.Read{Map|List|Set}Begin`, just that for `TCompactProtocol.ReadListBegin` we have a bug that we checked the wrong size (`size32` here is actually a const, not the size we read from the container).

dcelasun (approved):

Jens-G (approved):

PR #2606, opened 2022-05-13
THRIFT-5584: use gradle toolchain to specify Java 11 with --release 8

Jimexist:
use gradle toolchain to specify Java 11 with --release 8

### DIFF
--- a/lib/java/gradle/codeQualityChecks.gradle
+++ b/lib/java/gradle/codeQualityChecks.gradle
@@ -26,8 +26,6 @@ spotbugsMain {
     }
 }
 
-apply plugin: 'pmd'
-
 pmd {
     ignoreFailures = true
     toolVersion = '6.0.0'
--- a/lib/java/gradle/codeQualityChecks.gradle
+++ b/lib/java/gradle/codeQualityChecks.gradle
@@ -5,10 +5,8 @@
 
 dependencies {
     spotbugs configurations.spotbugsPlugins.dependencies
-    spotbugsPlugins 'com.h3xstream.findsecbugs:findsecbugs-plugin:1.11.0'
+    spotbugsPlugins 'com.h3xstream.findsecbugs:findsecbugs-plugin:1.12.0'
 }
-
-apply plugin: 'com.github.spotbugs'
 
 // see https://spotbugs-gradle-plugin.netlify.app/com/github/spotbugs/snom/spotbugsextension
 spotbugs {
--- a/lib/java/gradle/codeQualityChecks.gradle
+++ b/lib/java/gradle/codeQualityChecks.gradle
@@ -42,9 +42,7 @@ tasks.withType(Pmd) {
 
 spotless {
     java {
-        target project.fileTree(project.rootDir) {
-          include 'src/**/*.java'
-        }
+        target 'src/*/java/**/*.java'
         googleJavaFormat()
     }
 }
--- a/lib/java/gradle/codeQualityChecks.gradle
+++ b/lib/java/gradle/codeQualityChecks.gradle
@@ -42,7 +42,7 @@ tasks.withType(Pmd) {
 
 spotless {
     java {
-        target 'src/*/java/**/*.java'
+        target 'src/**/*.java'
         googleJavaFormat()
     }
 }
--- a/lib/java/gradle/sourceConfiguration.gradle
+++ b/lib/java/gradle/sourceConfiguration.gradle
@@ -39,7 +39,7 @@ tasks.withType(JavaCompile).configureEach {
     options.encoding = 'UTF-8'
     options.debug = true
     options.deprecation = true
-    assert JavaVersion.current().isJava11Compatible(), "we are using Java 11 to compile, this is to ensure that"
+    assert JavaVersion.current().isJava11Compatible(), "JDK 11 or higher is required to build"
     // the following is to build with Java 8 specifications, even when building with JDK9 or later
     options.release = 8
     options.compilerArgs += [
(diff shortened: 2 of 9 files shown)


lib/java/gradle/sourceConfiguration.gradle
```diff
@@ -21,21 +21,28 @@
...
+        languageVersion = JavaLanguageVersion.of(11)
+    }
+}
 
-tasks.withType(JavaCompile) {
+tasks.withType(JavaCompile).configureEach {
     options.encoding = 'UTF-8'
     options.debug = true
     options.deprecation = true
+    assert JavaVersion.current().isJava11Compatible(), "we are using Java 11 to compile, this is to ensure that"
```
ctubbsii, line 42:
```suggestion
    assert JavaVersion.current().isJava11Compatible(), "JDK 11 or higher is required to build"
```

lib/java/src/crossTest/java/org/apache/thrift/test/TestClient.java
```diff
@@ -823,4 +808,10 @@ public static void main(String[] args) {
 
     System.exit(returnCode);
   }
+
+  private static byte[] getBytesData() {
+    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream(256);
+    IntStream.range(-128, 128).forEach(byteArrayOutputStream::write);
+    return byteArrayOutputStream.toByteArray();
+  }
```
ctubbsii, line 816: These seem unrelated, but it's a nice improvement. None of this requires Java 11, right?

ctubbsii (changes requested):
Looks good to me, for the most part. I would suggest a slight change to the error message.

Also, there are some conditionals in the Gradle build files that check for Java 11. Those can be removed now.

We'd also need to remove the CI checks that run using JDK 8, like that the Ubuntu Xenial builds, unless you want to restore your previous workaround to force JDK 11 to be installed.

lib/java/src/crossTest/java/org/apache/thrift/test/TestClient.java
Jimexist: no, otherwise the cross-test CI in GitHub workflow will fail

Jimexist:
> Looks good to me, for the most part. I would suggest a slight change to the error message.
> 
> Also, there are some conditionals in the Gradle build files that check for Java 11. Those can be removed now.
> 
> We'd also need to remove the CI checks that run using JDK 8, like that the Ubuntu Xenial builds, unless you want to restore your previous workaround to force JDK 11 to be installed.

we don't need to update the xenial build, since using java toolchain means we are decoupling the JDK used to run gradle from the one used to compile Java. in the case of xenial, JDK8 is used to run gradle, which will download JDK 11 on its own before compiling Java.

ctubbsii:
> we don't need to update the xenial build, since using java toolchain means we are decoupling the JDK used to run gradle from the one used to compile Java. in the case of xenial, JDK8 is used to run gradle, which will download JDK 11 on its own before compiling Java.

Ah, ok. That's convenient. I assumed it worked like Maven toolchains: detected but not installed.

Jimexist:
> > we don't need to update the xenial build, since using java toolchain means we are decoupling the JDK used to run gradle from the one used to compile Java. in the case of xenial, JDK8 is used to run gradle, which will download JDK 11 on its own before compiling Java.
> 
> Ah, ok. That's convenient. I assumed it worked like Maven toolchains: detected but not installed.

yes there's a section in [here][1] describing the auto provisioning feature in case you are interested

[1]: https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning

lib/java/gradle/sourceConfiguration.gradle
```diff
@@ -21,21 +21,28 @@
...
+        languageVersion = JavaLanguageVersion.of(11)
+    }
+}
 
-tasks.withType(JavaCompile) {
+tasks.withType(JavaCompile).configureEach {
     options.encoding = 'UTF-8'
     options.debug = true
     options.deprecation = true
+    assert JavaVersion.current().isJava11Compatible(), "JDK 11 or higher is required to build"
```
Jimexist, line 42: i guess do we have to drop this assertion as the `current` refers to the one used by gradle not by toolchain.

```suggestion
```

lib/java/gradle/codeQualityChecks.gradle
```diff
@@ -46,9 +42,7 @@ tasks.withType(Pmd) {
 
 spotless {
     java {
-        target project.fileTree(project.rootDir) {
-          include 'src/**/*.java'
-        }
+        target 'src/*/java/**/*.java'
```
Jimexist, line 45:
```suggestion
        target 'src/**/*.java'
```

ctubbsii:
> why not split dependency updates or other code changes - not related to Gradle/Java update from real Gradle Java 11 upgrade?

If they are small and trivial, like these, I don't think it matters much. All of these are related to updating the build.

ctubbsii (approved):

nicolasb29:
Hi,

It seems that the version 0.18 is compiled in Java 11 even if --release 8 is present :
- libthrift 0.18 :
javap -verbose org/apache/thrift/transport/THttpClient.class | grep -i major
  major version: **55**
- libthrift 0.17 :
javap -verbose org/apache/thrift/transport/THttpClient.class | grep -i major
  major version: **52**

https://javaalmanac.io/bytecode/versions/

ctubbsii:
@nicolasb29 That's not a bug. Thrift moved to Java 11 in #2686 (THRIFT-5644)

PR #2607, opened 2022-05-15
remove unused java generator suppressions

Jimexist:
remove unused java generator suppressions

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -336,7 +336,6 @@ public:
    */
 
   std::string java_package();
-  std::string java_suppressions();
   std::string java_nullable_annotation();
   std::string java_override_annotation();
   std::string type_name(t_type* ttype,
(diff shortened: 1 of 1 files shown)


compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -485,11 +484,6 @@ string t_java_generator::java_package() {
   return "";
 }
 
-string t_java_generator::java_suppressions() {
-  return "@SuppressWarnings({\"cast\", \"rawtypes\", \"serial\", \"unchecked\", \"unused\"})"
-         + endl;
-}
```
ctubbsii, line 491: How certain are you that these aren't used? I'm pretty sure some of them are still applicable in at least some situations.

Jimexist: > How certain are you that these aren't used? I'm pretty sure some of them are still applicable in at least some situations.

yeah you are right - i tried to localize the warning suppression only to find it's not worth it.

ctubbsii: Some of these warnings should be fixed, though with better generated code. Like... why are we generating a logger that is unused? Or why are we using raw generics instead of the specific generic types when we know them? Or, why would we need to suppress serial? Why not just use the generated serialUid or a 1-up counter for custom Exceptions?

All this would take work, but eventually, we could probably get rid of these warnings suppressions by improving the generator in other parts.

PR #2608, opened 2022-05-18
THRIFT-5502: Log SocketException at WARN level only

slachiewicz:
and proposition to move logging to one place

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/server/TThreadPoolServer.java
+++ b/lib/java/src/main/java/org/apache/thrift/server/TThreadPoolServer.java
@@ -271,7 +271,7 @@ public class TThreadPoolServer extends TServer {
       }
     }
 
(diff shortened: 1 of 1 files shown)


ctubbsii:
> and proposition to move logging to one place

Why?

slachiewicz:
instead of the private method isIgnorableException we can just have a method to log proper message and add mentioned in THRIFT-5502 ignore for specific Socket exceptions. 
if not this way - please suggest how this ticket should be solved

ctubbsii:
> and proposition to move logging to one place

I misunderstood this. I thought you were removing duplicate code and consolidating into a single method, but it didn't make sense, since this code is only called once.

What you're actually doing is combining the exception handling logic, so it's all handled in one place. That makes sense to me now.

PR #2619, opened 2022-06-08
THRIFT-4086: Add missing calls to get_true_type when generating valid…

nanreh:
…ator + metadata code

Client: java

The change is trivial: just added a couple of missing calls to `get_true_type` in the Java generator so that we get exactly the same code regardless of what order structs are defined (specifically structs that refer to each other and create forward references). Without this fix, it's possible to get different Java from the generator. Specifically, the generator would produce incorrect metadata and would exclude validation calls on sub-structs. More detail can be found at this test repo I put together to demonstrate the issue:

https://github.com/nanreh/thrift_struct_refs

### DIFF
--- a/lib/java/gradle/generateTestThrift.gradle
+++ b/lib/java/gradle/generateTestThrift.gradle
@@ -26,6 +26,8 @@ ext.genReuseSrc = file("$buildDir/gen-javareuse")
 ext.genFullCamelSrc = file("$buildDir/gen-fullcamel")
 ext.genOptionTypeJdk8Src = file("$buildDir/gen-option-type-jdk8")
 ext.genUnsafeSrc = file("$buildDir/gen-unsafe")
+ext.genStructOrderTestASrc = file("$buildDir/resources/test/struct-order-test/a")
+ext.genStructOrderTestBSrc = file("$buildDir/resources/test/struct-order-test/b")
 
 // Add the generated code directories to the test source set
 sourceSets {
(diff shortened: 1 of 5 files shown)


lib/java/gradle/generateTestThrift.gradle
```diff
@@ -26,6 +26,8 @@ ext.genReuseSrc = file("$buildDir/gen-javareuse")
 ext.genFullCamelSrc = file("$buildDir/gen-fullcamel")
 ext.genOptionTypeJdk8Src = file("$buildDir/gen-option-type-jdk8")
 ext.genUnsafeSrc = file("$buildDir/gen-unsafe")
+ext.genStructOrderTestASrc = file("$buildDir/resources/test/struct-order-test/a")
+ext.genStructOrderTestBSrc = file("$buildDir/resources/test/struct-order-test/b")
```
Jimexist, line 30: maybe try to use a less nested subdir?

nanreh: Thanks for the feedback. Are you suggesting that instead of:
```
$buildDir/resources/test/struct-order-test/a/
$buildDir/resources/test/struct-order-test/b/
```
it could be improved as:
```
$buildDir/resources/test/struct-order-a/
$buildDir/resources/test/struct-order-b/
```
... or is there another change that would be preferable? thx

Jens-G: @Jimexist I guess @nanreh is waiting on any further comments/requests from your side?

PR #2620, opened 2022-06-09
THRIFT-5595: Avoid peek request with SSLSocket

MikaelSmith:
The update to TSocket in https://github.com/MikaelSmith/thrift/commit/01d53f483a7531ad4899b522060e8913dca309fb errors for TSSLSocket with
```
ValueError: non-zero flags not allowed in calls to recv() on <class 'ssl.SSLSocket'>
```
Handles ValueError from ssl.SSLSocket to fix isOpen when using TSSLSocket.

### DIFF
--- a/lib/py/test/test_socket.py
+++ b/lib/py/test/test_socket.py
@@ -25,6 +25,7 @@ class TSocketTest(unittest.TestCase):
             acc.start()
 
             sock = TSocket(host="localhost", port=acc.port)
+            self.assertFalse(sock.isOpen())
             sock.open()
             sock.setTimeout(timeout)
 
--- a/lib/py/src/transport/TSocket.py
+++ b/lib/py/src/transport/TSocket.py
@@ -94,6 +94,9 @@ class TSocket(TSocketBase):
                 if exc.errno in (errno.EWOULDBLOCK, errno.EAGAIN):
                     return True
                 return False
+            except ValueError:
+                # SSLSocket fails on recv with non-zero flags; fallback to the old behavior
+                return True
         finally:
             self.handle.settimeout(original_timeout)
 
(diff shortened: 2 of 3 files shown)


lib/py/src/transport/TSSLSocket.py
```diff
@@ -266,6 +266,10 @@ def __init__(self, host='localhost', port=9090, *args, **kwargs):
         TSocket.TSocket.__init__(self, host, port, unix_socket,
                                  socket_keepalive=socket_keepalive)
 
+    # Override TSocket because ssl.SSLSocket does not allow non-zero flags in calls to recv()
+    def isOpen(self):
+        return self.handle is not None
```
fishy, line 271: I would slightly prefer to just catch the exception in `TSocket.isOpen`, but I don't feel strongly either way.

lib/py/test/test_socket.py
```diff
@@ -25,6 +25,7 @@ def test_isOpen_checks_for_readability(self):
             acc.start()
 
             sock = TSocket(host="localhost", port=acc.port)
+            self.assertFalse(sock.IsOpen())
```
fishy, line 28: this is a typo. it should be `isOpen` instead of `IsOpen`. it's currently causing CI failures: https://app.travis-ci.com/github/apache/thrift/jobs/572946806 & https://app.travis-ci.com/github/apache/thrift/jobs/572946807

lib/py/src/transport/TSSLSocket.py
MikaelSmith: I guess that makes some sense in that we'd largely expect the implementations to be in-sync.

fishy (approved):
lgtm. just waiting for travis to finish running.

MikaelSmith:
I verified this fixes the issue in my scenario that was failing. I didn't dig into how to run these python tests directly, hoping CI would cover it.

fishy:
Thanks for the comments and added unit tests!

MikaelSmith:
Thanks for being so helpful with code review and getting this merged!

PR #2633, opened 2022-07-06
THRIFT-5603: add operator == for TEnumIterator

africamonkey:
We want to enable the compile option `-D_GLIBCXX_DEBUG` to find potential bugs, but we meet compile errors: [error_log.txt](https://issues.apache.org/jira/secure/attachment/13046360/13046360_error_log.txt)

Let's see the code in apache arrow project: https://github.com/apache/arrow/blob/984b59a2b4edb03479947c46007fc6142447f759/cpp/src/generated/parquet_types.cpp#L36

It constructs a std::map from a pair of iterators. However, in the debug mode of std::map (say, std::__debug::map), it will add additional check (such as boundary check) and need an operator == for these iterators.

So, I want to add an operator ==.

Apache Jira ticket: https://issues.apache.org/jira/browse/THRIFT-5603

### DIFF
--- a/lib/cpp/src/thrift/Thrift.h
+++ b/lib/cpp/src/thrift/Thrift.h
@@ -62,10 +62,8 @@ public:
     return ii_ == rhs.ii_ && n_ == rhs.n_;
   }
 
-  bool operator!=(const TEnumIterator& end) const {
-    THRIFT_UNUSED_VARIABLE(end);
-    assert(end.n_ == -1);
-    return (ii_ != n_);
+  bool operator!=(const TEnumIterator& rhs) const {
+    return ii_ != rhs.ii_ || n_ != rhs.n_;
   }
 
   std::pair<int, const char*> operator*() const { return std::make_pair(enums_[ii_], names_[ii_]); }
--- a/lib/cpp/src/thrift/Thrift.h
+++ b/lib/cpp/src/thrift/Thrift.h
@@ -59,11 +59,15 @@ public:
   int operator++() { return ++ii_; }
 
   bool operator==(const TEnumIterator& rhs) const {
-    return ii_ == rhs.ii_ && n_ == rhs.n_;
+    bool is_end = ii_ == n_ || n_ == -1;
+    bool is_rhs_end = rhs.ii_ == rhs.n_ || rhs.n_ == -1;
+    return (ii_ == rhs.ii_ && n_ == rhs.n_) || (is_end && is_rhs_end);
   }
 
   bool operator!=(const TEnumIterator& rhs) const {
-    return ii_ != rhs.ii_ || n_ != rhs.n_;
+    bool is_end = ii_ == n_ || n_ == -1;
+    bool is_rhs_end = rhs.ii_ == rhs.n_ || rhs.n_ == -1;
+    return (ii_ != rhs.ii_ || n_ != rhs.n_) && (!is_end || !is_rhs_end);
   }
 
   std::pair<int, const char*> operator*() const { return std::make_pair(enums_[ii_], names_[ii_]); }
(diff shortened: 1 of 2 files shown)


africamonkey:
Can anyone review this PR and merge? Thank you!

Jens-G:
Not sure ... the != and the == are actually testing different things. That looks suspicious to me ..

emmenlau:
@africamonkey , can you comment on the concern from @Jens-G . Generally your suggestion looks reasonable but why are operators `==` and `!=` testing different conditions?

africamonkey:
> @africamonkey , can you comment on the concern from @Jens-G . Generally your suggestion looks reasonable but why are operators `==` and `!=` testing different conditions?

I changed the implementation. Please take a look, thank you!

africamonkey:
@emmenlau Do you want to merge this pull request? If I can't get any response or comment for another week, I will close this pull request.

emmenlau:
Dear @africamonkey , I'm really sorry but I'm not all too deep in that code and find it slightly difficult to wrap my head around the implementation. Can you explain the logic, either in a comment in the code, or here in the chat?

Basically my questions are:
 - Why was the old implementation incorrect and/or is your new implementation better?
 - Why is the `!=` operator not just calling the `==` operator and returning the negated result?

africamonkey:
Hi @emmenlau 

> Why was the old implementation incorrect and/or is your new implementation better?

The old implementation assumes that the right hand side is the end of the iterator. So it's obviously incorrect. Basically, a `!=` operator should have commutative law, which is: `lhs != rhs <=> rhs != lhs`. 

The old one:
```C++
  bool operator!=(const TEnumIterator& end) {
    THRIFT_UNUSED_VARIABLE(end);
    assert(end.n_ == -1);  // assume this is the end iterator
    return (ii_ != n_);  // only check the left hand side because we assumed the right hand side is end iterator
  }
```

I'm writing a new operator which can handle general cases, which can also satisfy the commutative law. 

The new logic is that:
```C++
  bool operator==(const TEnumIterator& rhs) const {
    bool is_end = ii_ == n_ || n_ == -1;  // check if left hand side is an end iterator. 
    bool is_rhs_end = rhs.ii_ == rhs.n_ || rhs.n_ == -1;  // check if right hand side is an end iterator. 
    return (ii_ == rhs.ii_ && n_ == rhs.n_) || (is_end && is_rhs_end);
  }
```
Note that https://github.com/apache/arrow/blob/984b59a2b4edb03479947c46007fc6142447f759/cpp/src/generated/parquet_types.cpp#L36 uses `n_ = -1` to represent the end iterator, so we need to regard either `ii_ == n_` or `n_ == -1` as an end iterator. They should be the same. 
Obviously, if `ii_` and `n_` are the same, they should be the same iterators. Otherwise, if they are both end iterator, they should be regarded as the same iterators. 

> Why is the `!=` operator not just calling the `==` operator and returning the negated result?

I pushed a new commit. Now the `!=` operator is just calling the `==` operator.

ctubbsii:
> Friendly ping @ctubbsii @Jens-G for review. Thanks!

I'm not familiar with this C code.

praetorian20:
@Jens-G  could you please merge this PR? v0.20 fails to compile when _GLIBCXX_DEBUG is defined and this patch fixes the failure. The code appears to be correct.

emmenlau:
I've looked through the code for as much as I can, without stepping with a debugger. As far as I can tell, it looks correct and clean now. I agree that the old code seems rather limited, for example in that it `asserts` that the given argument is the end iterator. The new code does look more clean and more versatile.

All with a grain of salt, because I did not step through the code. But if it furthermore was tested by @africamonkey , I'd vote for merging.

PR #2637, opened 2022-08-01
THRIFT-5605: Client middleware to extract exceptions

fishy:
Client: go

Provide ExtractIDLExceptionClientMiddleware client middleware
implementation to extract exceptions defined in thrift IDL into err
return so they are accessible from other client middlewares.

### DIFF
--- a/lib/go/thrift/middleware.go
+++ b/lib/go/thrift/middleware.go
@@ -19,7 +19,9 @@
 
 package thrift
 
-import "context"
+import (
+	"context"
+)
 
 // ProcessorMiddleware is a function that can be passed to WrapProcessor to wrap the
 // TProcessorFunctions for that TProcessor.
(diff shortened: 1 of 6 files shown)


lib/go/thrift/middleware.go
```diff
@@ -107,3 +110,76 @@ func WrapClient(client TClient, middlewares ...ClientMiddleware) TClient {
...
+//     }
+//
+// The thrift compiler generated go code for the result TStruct would be like:
+//
+//     type MyServiceFooResult struct {
+//       Success *FooResponse `thrift:"success,0" db:"success" json:"success,omitempty"`
+//       Error1 *Exception1 `thrift:"error1,1" db:"error1" json:"error1,omitempty"`
+//       Error2 *Exception2 `thrift:"error2,2" db:"error2" json:"error2,omitempty"`
+//     }
+func getClientError(result TStruct, err error) error {
```
fishy, line 130: @dcelasun I'm also debating whether to export this as a public API, and if yes, what's a good name for it.

dcelasun: I think it's fine to make it public, but maybe move it to `exception.go` and call it `GetClientException`?

fishy: What do you think of `ExtractExceptionFromResult`?

dcelasun: Sounds good!

dcelasun (approved):

PR #2638, opened 2022-08-02
Update supported go version to 1.18.* and 1.19.*

fishy:
Client: go

Also provide generic version of Pointer helper function.


dcelasun:
Hmm, I'm not so sure about this. It's been less than 5 months since 1.18 was released. We are due for a new Thrift version and that might be too soon to introduce code that won't compile with <=1.17.

fishy:
> Hmm, I'm not so sure about this. It's been less than 5 months since 1.18 was released. We are due for a new Thrift version and that might be too soon to introduce code that won't compile with <=1.17.

This is according to our supported go releases policy at https://github.com/apache/thrift/blob/master/lib/go/README.md#suppored-go-releases

With the release of go 1.19 today, go 1.17 is no longer supported officially.

dcelasun (approved):
Well alright then :)

PR #2644, opened 2022-08-18
Use openssl and zlib targets

lorteddie:
If available the targets of OpenSSL and ZLIB should be used instead of library paths

### DIFF
--- a/lib/c_glib/CMakeLists.txt
+++ b/lib/c_glib/CMakeLists.txt
@@ -97,10 +97,16 @@ target_link_libraries(thrift_c_glib PUBLIC ${SYSLIBS})
 
 # If Zlib is not found just ignore the Zlib stuff
 if(WITH_ZLIB)
-    include_directories(SYSTEM ${ZLIB_INCLUDE_DIRS})
     ADD_LIBRARY_THRIFT(thrift_c_glib_zlib ${thrift_c_glib_zlib_SOURCES})
-    target_link_libraries(thrift_c_glib_zlib ${SYSLIBS} ${ZLIB_LIBRARIES})
-    target_link_libraries(thrift_c_glib_zlib thrift_c_glib)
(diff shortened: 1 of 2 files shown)


emmenlau:
Yep, this is a very reasonable change, and seems fully backwards compatible.

@lorteddie , the build errors are mostly unrelated. But could you kindly rebase on latest master and force push once more, to see if they maybe go away?

emmenlau:
The AppVeyor test is successful except for an unrelated issue.

emmenlau:
Thanks @lorteddie ! The Travis build shows a problem in the current implementation, see https://app.travis-ci.com/github/apache/thrift/jobs/582329079:
```
CMake Error at lib/c_glib/CMakeLists.txt:105 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "thrift_c_glib_zlib".  All uses of target_link_libraries with a
  target must be either all-keyword or all-plain.
  The uses of the plain signature are here:
   * lib/c_glib/CMakeLists.txt:101 (target_link_libraries)
   * lib/c_glib/CMakeLists.txt:102 (target_link_libraries)
```
I think this error did not exist before, is that possible?

emmenlau:
From a brief look, I think an inconsistency may have been there even before your change. After your change, the line  `target_link_libraries(thrift_c_glib_zlib ${SYSLIBS})` is inconsistent with the other statements of `target_link_libraries`, because some use the `PUBLIC` keyword and others do not. Is that possibly the problem?

I think most of cmake uses the `PUBLIC` keyword nowadays in thrift, so maybe add that where it's missing?

lorteddie:
thanks for the feedback, I added the PUBLIC keyword to make the calls more consistent

emmenlau:
The build errors seem all unrelated, but I'll pull this first into my local CI and test there, to be sure.

PR #2652, opened 2022-09-07
THRIFT-5208: capitalize exception ctor

yawaramin:
Fix https://issues.apache.org/jira/browse/THRIFT-5208

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_ocaml_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_ocaml_generator.cc
@@ -1665,7 +1665,7 @@ string t_ocaml_generator::type_name(t_type* ttype) {
   }
 
   string name = ttype->get_name();
-  if (ttype->is_service()) {
+  if (ttype->is_service() || ttype->is_xception()) {
     name = capitalize(name);
   } else {
     name = decapitalize(name);


Jens-G:
Testcase would be nice. Possible?

yawaramin:
@Jens-G any ideas where to start? I'm not a C++ guy so pretty unfamiliar with the tooling. Case in point, https://github.com/apache/thrift/blob/master/compiler/cpp/tests/README.md?plain=1#L52

I'm not sure how to do that! E.g. how do I run specifically the netstd tests to see how it works?

yawaramin:
OK for the record existing tests are building and running with the following setup:

Configure test build:

```
cd compiler/cpp/tests
cmake -DCMAKE_PREFIX_PATH=/usr/local/opt/bison -DCMAKE_CXX_STANDARD=11 .
```

Build and run tests:

```
cmake --build .
ctest
```

_But,_ the existing netstd/netcore tests seem to have a naming mixup which is preventing any tests from _actually_ running, which I checked by deliberately introducing an incorrect input into the `t_netcore_generator_functional_tests_helpers.cc` file and seeing that no tests failed.

I will try to write a test for the OCaml and see if I can actually get it to fail first before trying to add test coverage.

Jens-G:
> But, the existing netstd/netcore tests seem to have a naming mixup which is preventing any tests from actually running, which I checked by deliberately introducing an incorrect input into the t_netcore_generator_functional_tests_helpers.cc file and seeing that no tests failed.

Was not really aware about this piece of code. Never looked into it. Since it was written for netcore it might be broken anyway, there are a number of differences between the two implementations. If anybody wants to fix any issues here, please do. I most certainly won't do it myself.

yawaramin:
I wasn't really asking for a fix nor volunteering, just keeping a record of my long and winding path to figuring out how to run the codegen unit tests :-)

PR #2661, opened 2022-09-13
fix staleness check in cmake/ctest

Jimexist:
fix staleness check in cmake/ctest, the line was mistakenly removed in https://github.com/apache/thrift//commit/944b8e68a099392d80153ebcf26f32ff7f1d893a#diff-f5cd0615ff0ea8e5fb4cc8aae1b13f897a6386fe99238b93e1ea09a71d8c44c3L1

### DIFF
--- a/compiler/cpp/test/compiler/Included.thrift
+++ b/compiler/cpp/test/compiler/Included.thrift
@@ -19,6 +19,8 @@
  * under the License.
  */
 
+const string foo = "bar"
+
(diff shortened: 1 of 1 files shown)


Jimexist:
cmake flows failures will be caught by GitHub actions as well so we should rely on that to do fast sanity check

Jens-G (approved):
thanks!

PR #2662, opened 2022-09-13
[build infra] focal to use openjdk 11 instead of 17 to get rid of false errors

Jimexist:
focal to use openjdk 11 instead of 17 to get rid of false errors

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,7 +6,7 @@ Build-Depends: dotnet-runtime-6.0, dotnet-sdk-6.0, debhelper (>= 9), build-essen
     pkg-config, libtool, bison, flex, libboost-dev | libboost1.56-dev | libboost1.63-all-dev,
     python-all, python-setuptools, python-all-dev, python-all-dbg,
     python3-all, python3-setuptools, python3-all-dev, python3-all-dbg,
-    openjdk-8-jdk | openjdk-8-jdk-headless | openjdk-11-jdk | openjdk-11-jdk-headless | default-jdk,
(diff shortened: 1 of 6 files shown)


debian/control
```diff
@@ -6,7 +6,7 @@ Build-Depends: dotnet-runtime-6.0, dotnet-sdk-6.0, debhelper (>= 9), build-essen
     pkg-config, libtool, bison, flex, libboost-dev | libboost1.56-dev | libboost1.63-all-dev,
     python-all, python-setuptools, python-all-dev, python-all-dbg,
     python3-all, python3-setuptools, python3-all-dev, python3-all-dbg,
-    openjdk-8-jdk | openjdk-8-jdk-headless | openjdk-11-jdk | openjdk-11-jdk-headless | default-jdk,
+    openjdk-8-jdk | openjdk-8-jdk-headless | openjdk-11-jdk | openjdk-11-jdk-headless | openjdk-17-jdk | openjdk-17-jdk-headless | default-jdk,
```
emmenlau, line 9: Do we even still need `openjdk-8-jdk` and `openjdk-8-jdk-headless`, or could they be dropped at the same time? Just asking, I'm not in the details!

Jimexist: i think it can be dropped once we move away from xenial and welcome focal and have both focal and bionic available

emmenlau: Ok, thanks for checking!

Jimexist:
I'm merging this as this has now fixed the jdk issue. the next thing up is the swift error.

PR #2670, opened 2022-09-20
Thrift 4547: Add Swift cross tests

kinoroy:
- Adds a full Swift cross test suite, implementing nearly all of the tests (except for one-way functions which the Swift generator doesn't currently support [THRIFT-5468](https://issues.apache.org/jira/browse/THRIFT-5468)).
- Adds Swift to the list of cross test languages running on Github Actions.
- Adds UUID support to the Swift generator and client lib ([THRIFT-5629](https://issues.apache.org/jira/browse/THRIFT-5629)). I thought about separating this into a separate pull request but these changes are needed for the cross tests to pass against the kotlin/java libraries which currently run on CI.
- Fixes [THRIFT-5630](https://issues.apache.org/jira/browse/THRIFT-5630) where the TSocketServer does not work on Linux. Again this is included here because without it the cross tests won't pass on CI
- Fixes a few small issues in the Swift library cross compatibility that weren't detected before tests
- Makes a small change to insanity test in the kotlin/java cross test servers because they don't provide the value of `userMap` or `xtructs` fields in the Insanity object. These fields are required in the Swift generated code (because the field is not marked as optional in the TestThrift.thrift specification and the Swift generator doesn't support default values yet) 

Possible breaking change: 
- In the `TType` enum, the integer value of 17 was already taken by `TType.utf16` but this was incompatible with java/kotlin which use the value 17 for the `uuid` type. I'm not sure if this type is in use or considered part of the public API because it's not listed in the spec: https://thrift.apache.org/docs/idl.html 

I understand this is a fairly large PR, I'm open to suggestions on splitting it up if necessary, either merging in the tests first (which wouldn't pass), or merging in fixes individually first (but then there won't be any tests run against those PRs)

Thanks to @janosvitok for the WIP of the cross test suite, of which the PR is based off of. 


.gitignore
```diff
@@ -49,6 +49,7 @@ test-driver
 erl_crash.dump
 project.lock.json
 
+.Dockerfile.sha512
```
Jimexist, line 52: wondering what will generate this?

lib/java/src/test/java/org/apache/thrift/server/ServerTestBase.java
```diff
@@ -246,7 +246,7 @@ public Map<Long, Map<Numberz, Insanity>> testInsanity(Insanity argument) {
       first_map.put(Numberz.TWO, argument);
       first_map.put(Numberz.THREE, argument);
 
-      Insanity looney = new Insanity();
+      Insanity looney = new Insanity(new HashMap<>(), Arrays.asList());
```
Jimexist, line 249: would this break if not added?

lib/swift/Sources/TBinaryProtocol.swift
```diff
@@ -246,6 +256,20 @@ public class TBinaryProtocol: TProtocol {
     return data
   }
   
+  public func read() throws -> UUID {
+    let data = try self.transport.readAll(size: 16)
+    let lsb = data[0..<data.count/2]
+    let msb = data[(data.count/2)..<data.count]
```
Jimexist, line 262:
```suggestion
    let lsb = data[0..<8]
    let msb = data[8..<16]
```

lib/swift/Sources/TProtocol.swift
```diff
@@ -41,7 +41,8 @@ public enum TType: Int32 {
   case set      = 14
   case list     = 15
   case utf8     = 16
-  case utf16    = 17
+  //case utf16    = 17
```
Jimexist, line 44: is this a breaking change?

lib/swift/Sources/TProtocol.swift
```diff
@@ -132,12 +137,13 @@ public extension TProtocol {
   func skip(type: TType) throws {
     switch type {
     case .bool:   _ = try read() as Bool
-    case .i8:   _ = try read() as UInt8
+    case .i8:   _ = try read() as Int8
```
Jimexist, line 140: was this a bug?

test/swift/CrossTests/Sources/TestServer/main.swift
```diff
@@ -0,0 +1,59 @@
...
+    if parameters.showHelp {
+      parameters.printHelp()
+      return
+    }
+
+    let service = ThriftTestImpl()
+    let processor = ThriftTestProcessor(service: service)
+  
+      
+      switch (parameters.proto, parameters.transport) {
```
Jimexist, line 37: wrong indentation?

Jimexist (approved):
overall looks good to me

lib/java/src/test/java/org/apache/thrift/server/ServerTestBase.java
kinoroy: Yes, since this field is marked as non-optional in the TestThrift.thrift spec, the Swift generator makes them required fields in Swift for deserialization.  I'm not sure if this is the intended behaviour, maybe "non-optional" fields should be mapped to optionals in Swift since they can apparently be omitted by servers according to the spec for "default requiredness".  We could just comment out the insanity test for now until changes are made to the Swift generator to be more compliant with the spec

lib/swift/Sources/TProtocol.swift
kinoroy: yeah I'm not sure if the spec changed at some point, but the docs now define `i8` as an "An 8-bit signed integer" https://thrift.apache.org/docs/types so I think it should be Int8

.gitignore
kinoroy: The `build/docker/refresh.sh` command from here https://github.com/apache/thrift/tree/master/build/docker Although it seems like we're moving away from Travis CI / Docker anyways so doesn't matter too much

lib/swift/Sources/TProtocol.swift
Jens-G: > utf16

Strings in Thrift ALWAYS go as UTF8 over the wire. Utf16 is not a valid Thrift type, so adding it to TType seems odd.

> Because right now neither utf16 nor uuid are documented types here: https://thrift.apache.org/docs/idl.html

Wrong link, this is release 0.17.0. The one at https://github.com/apache/thrift/blob/master/doc/specs/idl.md is for master.

lib/swift/Sources/TSocketServer.swift
```diff
@@ -96,34 +96,29 @@ open class TSocketServer<InProtocol: TProtocol, OutProtocol: TProtocol, Processo
...
+    
+    // tell socket to listen
+    acceptConnectionInBackgroundAndNotify(handle: socketFileHandle)
   }
-
-  deinit {
-    NotificationCenter.default.removeObserver(self)
+  
+  open func acceptConnectionInBackgroundAndNotify(handle: FileHandle) {
+    DispatchQueue(label: "ConnectionAcceptQueue").async {
```
kainjow, line 107: should this label match the form of the label used above? also is `Queue` in the name redundant? `TSocketServer.processing`
maybe `TSocketServer.connectionAccept`

lib/swift/Sources/TSocketServer.swift
```diff
@@ -96,34 +96,29 @@ open class TSocketServer<InProtocol: TProtocol, OutProtocol: TProtocol, Processo
...
+      let acceptedFD = accept(handle.fileDescriptor, nil, nil)
+      DispatchQueue.main.async {
```
kainjow, line 109: maybe instead of hard coding DispatchQueue this could be a public property? maybe a client would want to use a custom queue?

kinoroy: I don't think this should be exposed. We have to do it this way rn because NotificationCenter doesn't seem to be delivering notifications on Linux targets. We'd probably prefer to use the system implementation of `FileHandle. acceptConnectionInBackgroundAndNotify` and so in the future if this is fixed on Linux in the CoreFoundation implementation then we might want to move back to that and having this queue exposed as public API would be a breaking change.

lib/swift/Sources/TSocketServer.swift
kinoroy: Yeah good point can update that

lib/swift/Sources/TProtocol.swift
kinoroy: @Jens-G Is the justification for this change satisfactory for you? To be clear, this PR *removes* UTF16 support, and replaces it with UUID to bring Swift in compliance with the spec

lib/swift/Sources/TSocketServer.swift
Jens-G: The `byte` type was **always** signed, like all Thrift integer types. So we renamed the type specifier a while ago from byte to `i8` mostly to remove that constant source of confusion - a byte is considered unsigned in many languages. Nevertheless, for compatibility reasons you still can use `byte` in the IDL.

lib/swift/Sources/TProtocol.swift
Jens-G: The `byte` type was **always** signed, like all Thrift integer types. So we renamed the type specifier a while ago from byte to `i8` mostly to remove that constant source of confusion - a byte is considered unsigned in many languages. Nevertheless, [for compatibility reasons](https://www.mail-archive.com/dev@thrift.apache.org/msg50144.html) you still can use `byte` in the IDL.

kinoroy:
@Jimexist @Jens-G is there anything else you need from me to be able to merge this? 
I'd like to contribute more to the Swift lib and it'd be great to have the cross tests running for future PRs :)

Jimexist (approved):

.github/workflows/build.yml
```diff
@@ -183,6 +183,19 @@ jobs:
...
+
+      - name: Upload swift precross artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: swift-precross
+          if-no-files-found: error
+          path: |
+            test/swift/CrossTests/.build/x86_64-unknown-linux-gnu/debug/TestServer
+            test/swift/CrossTests/.build/x86_64-unknown-linux-gnu/debug/TestClient
+
```
Jimexist, line 197: actually do you mind pull this out to a separate step other than lib-java-and-kotlin?

kinoroy: Done

Jimexist:
i'll wait 1 more day for @Jens-G 's feedback before merging

PR #2673, opened 2022-09-21
THRIFT-5634: thrift docs picture was broken

iceprosurface:
  - https://issues.apache.org/jira/browse/THRIFT-5634

change :

https://thrift.apache.org/docs/doc/images/cgrn.png -> https://thrift.apache.org/doc/images/cgrn.png
https://thrift.apache.org/docs/doc/images/cred.png -> https://thrift.apache.org/doc/images/cred.png

### DIFF
--- a/LANGUAGES.md
+++ b/LANGUAGES.md
@@ -66,289 +66,289 @@ Thrift's core protocol is TBinary, supported by all languages except for JavaScr
 <tr align=center>
 <td align=left><a href="https://github.com/apache/thrift/blob/master/lib/c_glib/README.md">C (glib)</a></td>
 <!-- Since -----------------><td>0.6.0</td>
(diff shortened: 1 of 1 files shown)


Jimexist:
i wonder what is a better fix, to change all relative paths to absolute one or to just update somewhere once to fix the `docs` prefix?

iceprosurface:
> i wonder what is a better fix, to change all relative paths to absolute one or to just update somewhere once to fix the `docs` prefix?

This is determined by the location of the static resource (image).

For example:

`https://thrift.apache.org/docs/Languages.html` was under `/docs/`, so that relative assets will load with prefix  `/docs/`

`doc/images/cgrn.png`-> `/docs/doc/images/cgrn.png` ❌

However if we change correct relative path -> `../doc/images/cgrn.png`, this will break github preview. 


Move `doc/images` to `docs/images`, then add the `/docs` prefix to every static resource(image), maybe this is a better fix.

Change all doc will break search result of search engine.

PR #2675, opened 2022-09-23
[java][lib] use check not test in gradle

Jimexist:
use check not test in gradle


Jimexist:
> this test fails because `ktfmt-gradle` actually will check for `generated-sources` dir but we'd like to exclude that for checking but not for compilation

i put up the fix: to put the generated sources to test source set which resolves the issue

ctubbsii (commented):
I don't know enough about Gradle to understand the differences between test and check, but don't see anything to object to here.

Jimexist:
> I don't know enough about Gradle to understand the differences between test and check, but don't see anything to object to here.

note to myself: currently this is still failing because the gradle check will include spotless check on generated java files. there's no easy way to filter them out unless changes are to be made in the gradle plugin.

Jens-G:
Rebase needed, no actvity since 2022

PR #2676, opened 2022-09-23
THRIFT-4623: Create new instance of default value for mutable types

grzegorzlyczba:
Client: py

Default values have a single instance what's fine for non-mutable types, but any change made for list, class, etc are propagated to all instances
```
>>> from gen.test import ttypes
>>> b1 = ttypes.B()
>>> b1
B(itm=A(value=None))
>>> b1.itm.value = 1
>>> b2 = ttypes.B()
>>> b2
B(itm=A(value=None))
>>> b1
B(itm=A(value=1))
>>>
```

With this change, generator will set `None` as default value for such params and add a check in `__init__` method to create new instance when the value is `None`.

```py
def __init__(self, itm=None,):
    if itm is None:
        itm = A(**{
        })
    self.itm = itm
```
It works for types defined in the same file, the only limitation is to create recurrence types with a default value of the same type because it generates inf recursion.

Apache Jira ticket: https://issues.apache.org/jira/browse/THRIFT-4623

  

grzegorzlyczba:
I removed the last element in `thrift_spec` tuples with default instance (it was used only in `__init__` to check the param value, but it was always failing). I don't see any other place where it's still in use.

Jens-G:
Needs rebase

Jens-G:
This still breaks CI. 

```
src/thrift/generate/t_py_generator.cc:2749:9: error: ‘type’ was not declared in this scope; did you mean ‘wctype’?
 2749 |     if (type->is_struct() || type->is_list() || type->is_xception() || type->is_set()) {
 ```

Jens-G:
Feel free to update the PR and reopen.

PR #2680, opened 2022-09-26
THRIFT-5661: lib: cpp: TOutput: add zephyr-specific strerror_s implementation

cfriedt:
In Zephyr, optimize `TOutput::strerror_s` to minimize (stack) space. The Zephyr string error table is in ROM in any case, so string values will not be overwritten (if they happen to be compiled-in).
 

cfriedt:
rebased

PR #2686, opened 2022-09-28
THRIFT-5644: java to use jdk 17 and upgrade minimal language level to 11

Jimexist:
java to use jdk 17, this is a second try after #2662 

### DIFF
--- a/build/docker/README.md
+++ b/build/docker/README.md
@@ -180,7 +180,7 @@ Last updated: October 1, 2017
 | erlang    | OTP-18        | OTP-23        |       |
 | go        | 1.15.10       | 1.16.2        |       |
 | haxe      | 3.2.1         | 3.4.4         | THRIFT-4352: avoid 3.4.2 |
-| java      | 1.8.0\_191     | 11.0.3        |       |
+| java      | 1.8.0\_191     | 17           |       |
 | js        | Node.js 6.17.1, V8 5.1.281.111, npm 3.10.10 | Node.js 10.18.0, V8 6.8.275.32, npm 6.13.4 |     |
 | lua       |               | 5.2.4         | Lua 5.3: see THRIFT-4386 |
 | netstd    | 6.0           | 6.0           |       |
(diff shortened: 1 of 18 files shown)


lib/java/gradle/sourceConfiguration.gradle
```diff
@@ -53,13 +53,6 @@ tasks.withType(JavaCompile).configureEach {
     ]
 }
 
-tasks.withType(Javadoc) {
-    failOnError false
-    options.addStringOption('Xdoclint:none', '-quiet')
-    options.addStringOption('encoding', 'UTF-8')
-    options.addStringOption('charSet', 'UTF-8')
-}
-
```
ctubbsii, line 62: Why not keep this?

ctubbsii (commented):
I think it's reasonable to use the latest LTS Java for development on the next version of Thrift. I also think it would be reasonable to make Java 11 the minimum instead of 8 at this point.

Jimexist:
turns out that upgrading to java 17 won't solve the TSL runtime error during cross test

Jimexist:
because we want to push for raising minimal Java version to 11 we'd need to base on #2690

Jimexist:
@fishy @ctubbsii i have updated the minimal JDK requirement to be 11, please take a look

ctubbsii (commented):
I'm in favor of bumping the min java to 11 for runtime, and 17 for build time. I'm not sure I understand the issue with the cross test failure, but it seems like that's a separate issue from this PR.

build/docker/README.md
```diff
@@ -180,7 +180,7 @@ Last updated: October 1, 2017
 | erlang    | OTP-18        | OTP-23        |       |
 | go        | 1.15.10       | 1.16.2        |       |
 | haxe      | 3.2.1         | 3.4.4         | THRIFT-4352: avoid 3.4.2 |
-| java      | 1.8.0\_191     | 11.0.3        |       |
+| java      | 1.8.0\_191     | 17           |       |
```
fishy, line 183: we should also update the min version here? I think we dropped support to jdk8 with this change.

fishy (approved):

build/docker/README.md
Jimexist: this column is for xenial but we have deprecated it recently...

PR #2695, opened 2022-10-08
THRIFT-5648: add uuid support in rust

Jimexist:

### DIFF
--- a/lib/rs/Cargo.toml
+++ b/lib/rs/Cargo.toml
@@ -15,6 +15,7 @@ keywords = ["thrift"]
 [dependencies]
 byteorder = "1.3"
 integer-encoding = "3.0.3"
+uuid = "1"
 log = {version = "0.4", optional = true}
 ordered-float = "3.0"
 threadpool = {version = "1.7", optional = true}
@@ -22,3 +23,6 @@ threadpool = {version = "1.7", optional = true}
 [features]
 default = ["server"]
 server = ["threadpool", "log"]
+
+[dev-dependencies]
+uuid = { version = "*", features = ["v4"] }
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -17,11 +17,11 @@
  * under the License.
  */
 
-#include <string>
 #include <iostream>
+#include <string>
 
-#include "thrift/platform.h"
 #include "thrift/generate/t_generator.h"
+#include "thrift/platform.h"
 
 using std::map;
 using std::ofstream;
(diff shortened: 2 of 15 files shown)


lib/rs/Cargo.toml
```diff
@@ -15,10 +15,14 @@ keywords = ["thrift"]
 [dependencies]
 byteorder = "1.3"
 integer-encoding = "3.0.3"
+uuid = { version = "1.2.0" }
```
Jimexist, line 18:
```suggestion
uuid = "1"
```

Jimexist:
@allengeorge per discussion in #2700, @fishy suggested that we do not bring any 3rd dependencies (in go library). here we are facing the same issue, do you believe we shall remove the uuid dependency in rust?

fishy:
> @allengeorge per discussion in #2700, @fishy suggested that we do not bring any 3rd dependencies (in go library). here we are facing the same issue, do you believe we shall remove the uuid dependency in rust?

go library is slightly different in that it already has zero third party dependency right now so there's a big incentive to keep it that way, which does not apply to rust library (which I _believe_ already have third party dependencies).

allengeorge:
I'm ok with adding a dependency (we shouldn't add them without thought, of course). The Rust UUID crate is dual-licensed ASL 2.0/MIT, so that's not an issue.

Jimexist:
> I'm ok with adding a dependency (we shouldn't add them without thought, of course). The Rust UUID crate is dual-licensed ASL 2.0/MIT, so that's not an issue.
> 
> We can always reimplement the UUID portions necessary and remove the dependency over time.

thanks for the context. @allengeorge in this case do you think this pr can be merged as is?

Jimexist:
hi @allengeorge and @Jens-G please take a look at this, thanks1

my-vegetable-has-exploded:
I'd like to know why this pr was closed. thanks

Jimexist:
> I'd like to know why this pr was closed. thanks

hi, simple answer - i was not able to get approvers for this.

my-vegetable-has-exploded:
> > I'd like to know why this pr was closed. thanks
> 
> hi, simple answer - i was not able to get approvers for this.

what a pity！

Jens-G:
> hi, simple answer - i was not able to get approvers for this.

For starters, I am not a Rust guy, and @allengeorge already chimed in. That's why I removed the review request. 

Next, we have plenty of pending PRs, which is a whole topic on itself. Getting no approvers applies to most of them, but the real question is why? IMHO this is not a indication of a quality issue, but a quantity issue. If we had enough active committers taking care, we would be able to give better responses than "there just was nobody around to take care".

Jens-G:
I just skimmed over the changes and what could be better is the fact, that the compiler part also has a lot of refactorings in it. It's a bit hard to see what the real changes are for uuid and what is just shuffling comments around etc. I'd say we give it another chance. Shall we?

PR #2697, opened 2022-10-08
THRIFT-5649: add github action for go

Jimexist:
add github action for go

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,75 @@ jobs:
           path: compiler/cpp/thrift
           retention-days: 3
 
+  lib-go:
+    needs: compiler
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '>=1.17.0'
+
+      - name: Run bootstrap
+        run: ./bootstrap.sh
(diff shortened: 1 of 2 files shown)


Jimexist:
the cross test failure is irrelevant to the changes in this pr

.github/workflows/build.yml
```diff
@@ -51,6 +51,80 @@ jobs:
...
 
+  lib-go:
+    needs: compiler
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '>=1.17.0'
```
fishy, line 62: the current supported go versions are 1.18.x and 1.19.x: https://github.com/apache/thrift/blob/master/LANGUAGES.md, not 1.17.x.

but looks like it actually picks the latest version (1.19.1) from the logs, so I think we probably actually need two actions, one for `lib-go-1.18` with `<1.19.0` and one for `lib-go-1.19` with `>=1.19.0`, or something like that.

Jimexist: thanks for the suggestion - will setup a new jira to follow up

Jimexist: see https://issues.apache.org/jira/browse/THRIFT-5651

PR #2700, opened 2022-10-09
THRIFT-5650: add uuid support in go (part I - library)

Jimexist:
add uuid support in go, this is part I where only library code is updated.

will update compiler generation code later.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,7 +1,5 @@
 github.com/golang/mock v1.5.0 h1:jlYHihg//f7RRwuPfptm04yp4s7O6Kw8EZiVYIGcH0g=
 github.com/golang/mock v1.5.0/go.mod h1:CWnOUgYIOo4TcNZ0wHX3YZCqsaM1I1Jvs6v3mP3KVu8=
-github.com/google/uuid v1.3.0 h1:t6JiXgmwXMjEs8VusXIJk2BXHsn+wx8BZdTaoZ5fu7I=
-github.com/google/uuid v1.3.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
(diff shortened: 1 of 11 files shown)


lib/go/thrift/binary_protocol.go
```diff
@@ -26,6 +26,8 @@ import (
 	"fmt"
 	"io"
 	"math"
+
+	"github.com/google/uuid"
```
fishy, line 30: I don't think we want to use a third party library for this:

1. so far we don't have any third party dependency (the only one is test only, not used by the library itself)
2. there are several "competing" uuid libraries and we don't really want to force our user to a particular one.

 just `type UUID [16]byte` and it would be easy enough to be casted to the majority of the third party uuid libraries.

the only thing we need to implement is converting from and to string.

go.sum
```diff
@@ -1,5 +1,7 @@
 github.com/golang/mock v1.5.0 h1:jlYHihg//f7RRwuPfptm04yp4s7O6Kw8EZiVYIGcH0g=
 github.com/golang/mock v1.5.0/go.mod h1:CWnOUgYIOo4TcNZ0wHX3YZCqsaM1I1Jvs6v3mP3KVu8=
+github.com/google/uuid v1.3.0 h1:t6JiXgmwXMjEs8VusXIJk2BXHsn+wx8BZdTaoZ5fu7I=
+github.com/google/uuid v1.3.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
```
Jimexist, line 4:
```suggestion
```

lib/go/thrift/uuid.go
```diff
@@ -0,0 +1,102 @@
...
+
+package thrift
+
+import (
+	"encoding/hex"
+	"errors"
+)
+
+// most of the code is copied from https://github.com/google/uuid in order to reduce dependency on
+// external libraries
```
fishy, line 28: if we are copying code then the license notification would need to be different to reflect the code we copied.

I really did not intend to copy any code. the majority of the complexity in uuid third party libraries are in the generating part, which we don't need in thrift, and that's why I prefer to **reimplement** the parts we need in thrift over depending on a third party library. do you need this merged in urgently? do you mind me taking this ticket later this week instead?

Jimexist: @fishy no i'm not in a hurry so please feel free to take your time and grab this pr and make changes to it. thanks!

PR #2701, opened 2022-10-09
Bump go version to 1.19.2 and update hashes

HamzaAnis:
Updated the max Golang version to latest 1.19.2

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
 module github.com/apache/thrift
 
-go 1.18
+go 1.19.2
 
 require github.com/golang/mock v1.5.0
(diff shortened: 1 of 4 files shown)


Jimexist (approved):

go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/apache/thrift
 
-go 1.18
+go 1.19.2
```
fishy, line 3: this is wrong:
1. this cannot include point releases
2. this is the minimal version of go we support, and we don't want to drop support for go1.18 yet, per our support policy at https://github.com/apache/thrift/blob/master/lib/go/README.md#suppored-go-releases

fishy: @Jimexist this is also reported by the github action workflow for go: https://github.com/apache/thrift/actions/runs/3214860554/jobs/5257904020

HamzaAnis: @fishy I believe we can revert this. I created this PR considering two minor go releases but you guys are considering two major go releases.

fishy: 1. in semver `1` is major and `18`/`19` is minor
2. https://github.com/apache/thrift/pull/2702

PR #2708, opened 2022-10-12
THRIFT-5650: Implement UUID in Go library

fishy:
Client: go

This is only the library part of THRIFT-5650. I still have some open questions for the compiler part so that will be done later.

While I'm here, also made some changes to go CI process:

* Update ubuntu-bionic to use go 1.18 for travis
* Update ubuntu-jammy to use the latest go 1.19.x for travis
* Run both go 1.18 and 1.19 for github actions
* Also run test/go tests for github actions

---
- Closes #2700


lib/go/test/tests/protocol_mock.go
```diff
@@ -1,513 +1,703 @@
...
- * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied. See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-
-// Automatically generated by MockGen. DO NOT EDIT!
-// Source: thrift (interfaces: TProtocol)
+// Code generated by MockGen. DO NOT EDIT.
```
fishy, line 1: Since we updated `TProtocol` and have to rerun mockgen, I keep this file to be the exact output of mockgen, and thus removed the license header.

@Jens-G I'm not 100% sure if that's in compliance to Apache license policy (this file is purely machine generated). if that's not allowed, I can add the license header back.

lib/go/thrift/uuid.go
```diff
@@ -0,0 +1,115 @@
...
+// For example, assuming you are using
+// https://pkg.go.dev/github.com/google/uuid to generate a v4 UUID for an
+// optional thrift field:
+//
+//	id, err := uuid.NewRandom()
+//	if err != nil {
+//	  // TODO: handle errors
+//	}
+//	myRequest.Uuid = thrift.Pointer(thrift.Tuuid(id))
+type Tuuid [16]byte
```
fishy, line 44: I'm not super happy with the name here, but the reality is that we have a type (here) and a constant (the value of TType enum) both should be named `UUID`, and only one of them can be named `UUID` and the other need to find a different name.

I think it's better to change the type name to `Tuuid`, as a way to emphasize that this is not really a full UUID implementation. @dcelasun what do you think?

fishy (commented):
cc @Jimexist

lib/go/thrift/type.go
```diff
@@ -38,9 +38,7 @@ const (
 	MAP    = 13
 	SET    = 14
 	LIST   = 15
-	UTF8   = 16
-	UTF16  = 17
-	//BINARY = 18   wrong and unusued
+	UUID   = 16
```
Jimexist, line 41: for Java it's 17, FYI

Jimexist (approved):

Jimexist:
exciting! looking forward to implementing it in precross tests so we can bring down the number of failing tests

lib/go/thrift/type.go
fishy: It's [16 in the spec](https://github.com/apache/thrift/blob/0223d6346675c5089c2a2de5fb6f3858e599c9a9/doc/specs/thrift-binary-protocol.md?plain=1#L204). does java pass cross test?

Jimexist: https://github.com/apache/thrift/blob/master/lib/java/src/main/java/org/apache/thrift/protocol/TType.java#L37-L38

.github/workflows/build.yml
```diff
@@ -54,12 +54,17 @@ jobs:
   lib-go:
     needs: compiler
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        go:
+          - '1.18'
+          - '1.19'
```
Jimexist, line 61: do you want to only do upload for precross for 1.19?

fishy: probably, but I'm not sure how to do that with github actions 😅

dcelasun (approved):

lib/go/thrift/type.go
kinoroy: @fishy Crosstests until recently were only running Java/Kotlin so obviously that was successful running crosstests against themselves even with incorrect spec. Swift was recently added but was implemented to match Java so also has UUID=17 so all those tests were passing

dcelasun: UUID=17 has already been included in a stable release. We should either correct that with a breaking change, or update the spec and the Go implementation to be consistent.

Neither option is attractive.

fishy: >UUID=17 has already been included in a stable release

is that true? I thought all UUID related changes are merged **after** 0.18.0.

dcelasun: > is that true? I thought all UUID related changes are merged after 0.18.0.

I only did a cursory check. `TType.java` was last modified on 9 June with `UUID=17` and 0.17.0 was released on 18 September.

dcelasun: NVM, the commit date is not the merge date. We didn't include it in a stable release. Sorry :)

fishy: looking at https://github.com/apache/thrift/commits/v0.17.0 I _think_ the only thing uuid related included in the stable release was the spec, so we definitely should fix java/kotlin/swift to match the spec instead :)

Jimexist: tracked in https://issues.apache.org/jira/browse/THRIFT-5653

Jimexist:
@fishy we can merge this and then start to work on compiler code and then move on to fixing crosstests?

PR #2710, opened 2022-10-13
fix various java codestyle issues

Jimexist:
fix various java codestyle issues
- remove unreachable code or deadcode
- remove unnecessary `final` key words on static method
- remove unnecessary `public` key words on interface
- make fields final if possible
- add `@Overrides` when applicable
- collapse multiple catch arms

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/TEnumHelper.java
+++ b/lib/java/src/main/java/org/apache/thrift/TEnumHelper.java
@@ -19,11 +19,13 @@
 
 package org.apache.thrift;
 
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
 /** Utility class with static methods for interacting with TEnum */
 public class TEnumHelper {
+
+  /* no instantiation */
+  private TEnumHelper() {}
 
   /**
    * Given a TEnum class and integer value, this method will return the associated constant from the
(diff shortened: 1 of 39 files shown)


lib/java/src/main/java/org/apache/thrift/TEnumHelper.java
```diff
@@ -38,11 +41,7 @@ public static TEnum getByValue(Class<? extends TEnum> enumClass, int value) {
     try {
       Method method = enumClass.getMethod("findByValue", int.class);
       return (TEnum) method.invoke(null, value);
-    } catch (NoSuchMethodException nsme) {
-      return null;
-    } catch (IllegalAccessException iae) {
-      return null;
-    } catch (InvocationTargetException ite) {
+    } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException nsme) {
```
ctubbsii, line 44: I think these can be replaced with:

```suggestion
    } catch (ReflectiveOperationException nsme) {
```

lib/java/src/main/java/org/apache/thrift/server/TSaslNonblockingServer.java
```diff
@@ -171,6 +171,7 @@ private void acceptNewConnection() {
             LOGGER.warn("Failed to accept incoming connection.", e);
           }
         } else {
+          // TODO fix with try-with-resource for the return value of .channel
```
ctubbsii, line 174: TODOs in code are easily forgotten. Better to create an issue in the issue tracker for the project than to embed a TODO.

Jimexist: let me revert this line - I don't think I can fully test whether the selector resource _will_ leak. let me re-check later.
```suggestion
```

ctubbsii (approved):

PR #2714, opened 2022-10-14
THRIFT-5652 IDL uuid literals can be improved

Jens-G:
Patch: Jens Geyer

### DIFF
--- a/compiler/cpp/src/thrift/thriftl.ll
+++ b/compiler/cpp/src/thrift/thriftl.ll
@@ -117,8 +117,6 @@ intconstant   ([+-]?[0-9]+)
 hexconstant   ([+-]?"0x"[0-9A-Fa-f]+)
 dubconstant   ([+-]?[0-9]*(\.[0-9]+)?([eE][+-]?[0-9]+)?)
 identifier    ([a-zA-Z_](\.[a-zA-Z_0-9]|[a-zA-Z_0-9])*)
-uuidconstant  (([0-9A-Fa-f]{8})(([\-]([0-9A-Fa-f]{4})){3})[\-]([0-9A-Fa-f]{12}))
-guidconstant  ("{"([0-9A-Fa-f]{8})(([\-]([0-9A-Fa-f]{4})){3})[\-]([0-9A-Fa-f]{12})"}")
 whitespace    ([ \t\r\n]*)
 sillycomm     ("/*""*"*"*/")
 multicm_begin ("/*")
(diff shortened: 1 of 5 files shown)


test/ConstantsDemo.thrift
```diff
@@ -65,7 +65,9 @@ const map<i32,thing> GEN_WHAT = { 35 : { 'hello' : 325, 'goodbye' : 325352 } }
 
 const set<i32> GEN_SET = [ 235, 235, 53235 ]
 
-const uuid GEN_UUID = "00000000-4444-CCCC-ffff-0123456789ab"
+// uuids are accepted with or without curly braces
+const uuid GEN_UUID =  00000000-4444-CCCC-ffff-0123456789ab
+const uuid GEN_GUID = {00112233-4455-6677-8899-aaBBccDDeeFF}
```
fishy, line 70: 1. can we force it still to be in string format? (e.g. adding leading and trailing `"` to the ll file)
2. if we decided that we want to support the form with curly braces, then I'll need to add support to that to the go library.
3. will the compiler still take this as a string token?

compiler/cpp/src/thrift/thriftl.ll
```diff
@@ -248,6 +250,18 @@ literal_begin (['\"])
...
+  /* Placed before intconstant, because uuids start with either "{" or directly with some hex digit */
+  yylval.uconst = strdup(yytext);
+  return tok_uuid_constant;
+}
+
+{guidconstant} {
+  yylval.uconst = strdup(&yytext[1]);
+  yylval.uconst[36] = '\000';
+  return tok_uuid_constant;
+}
```
fishy, line 263: oh I see you stripped the curly braces from the token so language libraries do not need to support the parsing of this form.

test/ConstantsDemo.thrift
Jens-G: > an we force it still to be in string format? (e.g. adding leading and trailing " to the ll file)
Why?

fishy: several reasons:
1. This would likely become a very common mistake, that developers would think that they are supposed to write it in this way: `const uuid GEN_UUID = "00000000-4444-CCCC-ffff-0123456789ab"`
2. The editor supports for the new syntax will take time to catch up (see how github is highlighting the syntax right now for an example), which will further make 1 more likely to happen.

Jens-G: We can't fly to Mars because we don't have syntax highlighting for it .... :-D

Jens-G: Done. That would have saved me some time if we had made this clear from the beginning.

fishy (approved):

test/ConstantsDemo.thrift
fishy: Mars is overrated anyways :-D

PR #2717, opened 2022-10-20
THRIFT-5653: UUID should use value 16 in Swift binary protocol

kinoroy:
Removes non-standard utf8 type with raw value 16 and replaces uuid raw value with 16. 
Brings Swift in compliance with spec: https://github.com/apache/thrift/blob/0223d6346675c5089c2a2de5fb6f3858e599c9a9/doc/specs/thrift-binary-protocol.md?plain=1#L204

This is technically a breaking change by removing the utf8 type, however this type was non-standard and not documented in the spec.

This causes new failing crosstests with Swift-Java with the binary protocol due to Java not yet being updated to match the spec. 

### DIFF
--- a/lib/swift/Sources/TCompactProtocol.swift
+++ b/lib/swift/Sources/TCompactProtocol.swift
@@ -208,8 +208,6 @@ public class TCompactProtocol: TProtocol {
     case .map:    return .map
     case .set:    return .set
     case .list:   return .list
-    case .utf8:   return .binary
-      //case .utf16:  return .binary
     case .uuid:   return .uuid
     }
   }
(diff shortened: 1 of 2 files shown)


fishy (approved):
Thank you for doing this!

(the following can be fixed in new PR(s) instead) travis has been failing with this swift error:
```
/thrift/src/lib/swift/Sources/TSocketServer.swift:83:60: error: type 'CFSocketError' (aka 'Int') has no member 'success'
      if CFSocketSetAddress(sock, cfaddr) != CFSocketError.success { //kCFSocketSuccess {
                                             ~~~~~~~~~~~~~ ^~~~~~~
```
since it looks like you are familiar with swift (😅), do you mind taking a look at that?

kinoroy:
@fishy yes I can look at that, the reason is that Swift on Linux is not as stable as it is on macOS yet, they sometimes make  breaking updates to the foundation library on Linux. I will look to see if we can standardize the version of Swift we use across Travis and Github actions

Jimexist (approved):

PR #2718, opened 2022-10-21
THRIFT-5650: Implement UUID in Go compiler

fishy:
Client: go

### DIFF
--- a/lib/go/thrift/json_protocol.go
+++ b/lib/go/thrift/json_protocol.go
@@ -524,6 +524,8 @@ func (p *TJSONProtocol) TypeIdToString(fieldType TType) (string, error) {
 		return "set", nil
 	case LIST:
 		return "lst", nil
+	case UUID:
+		return "uid", nil
 	}
 
 	e := fmt.Errorf("Unknown fieldType: %d", int(fieldType))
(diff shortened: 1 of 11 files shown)


fishy:
hmm go-swift cross-test failed.

@Jimexist do you know how to access the detailed log from github actions?

fishy:
I found the url to the logs uploaded by failed cross-test by enabling debug logging, but I don't have access to it.

kinoroy:
go-swift is failing because Go cross test server does not implement testUuid() method. This causes an error when the Swift client calls it as part of cross:
```
TApplicationError (1): Unknown Method
Message: Unknown function testUuid
```

fishy:
@kinoroy thanks that's easy to fix.

just out of curiosity, did you run the crosstests locally or did you get that from the logs from github actions?

kinoroy:
@fishy Sorry I forgot to mentioned, you can get the logs from GitHub on the job page once its complete just click "cross-test-log" to download :) 
Screen Shot 2022-10-22 at 5 00 26 PM

lib/go/thrift/json_protocol.go
```diff
@@ -524,6 +524,8 @@ func (p *TJSONProtocol) TypeIdToString(fieldType TType) (string, error) {
 		return "set", nil
 	case LIST:
 		return "lst", nil
+	case UUID:
+		return "uid", nil
```
fishy, line 528: apparently we don't have a written spec for TJSONProtocol, so I followed this part from the implementation of java library.

fishy:
It looks like rust is also using the wrong TType value for uuid so the rs-go tests are failing.

Jimexist (approved):

PR #2719, opened 2022-10-22
THRIFT-5657: Use Swift 5.7 in all CI (github actions + travis)

kinoroy:
Standardizes the version of Swift used across all CI (actions + travis) to Swift 5.7 (the latest version)

This should fix the current Travis Swift build failure on Linux due to an older version being used:
```
/thrift/src/lib/swift/Sources/TSocketServer.swift:83:60: error: type 'CFSocketError' (aka 'Int') has no member 'success'
      if CFSocketSetAddress(sock, cfaddr) != CFSocketError.success { //kCFSocketSuccess {
                                             ~~~~~~~~~~~~~ ^~~~~~~
```

### DIFF
--- a/build/docker/ubuntu-jammy/Dockerfile
+++ b/build/docker/ubuntu-jammy/Dockerfile
@@ -252,9 +252,12 @@ RUN apt-get install -yq \
   libxml2-dev && \
   cd / && \
   wget --quiet https://download.swift.org/swift-5.7-release/ubuntu2204/swift-5.7-RELEASE/swift-5.7-RELEASE-ubuntu22.04.tar.gz && \
-  tar xf swift-5.7-RELEASE-ubuntu22.04.tar.gz --strip-components=1 && \
-  rm swift-5.7-RELEASE-ubuntu22.04.tar.gz && \
-  swift --version
+  tar xf swift-5.7-RELEASE-ubuntu22.04.tar.gz && \
+  mv swift-5.7-RELEASE-ubuntu22.04 /usr/share/swift && \
+  rm swift-5.7-RELEASE-ubuntu22.04.tar.gz
+
+ENV PATH /usr/share/swift/usr/bin:$PATH
+RUN swift --version
 
 # Locale(s) for cpp unit tests
 RUN apt-get install -y --no-install-recommends \
--- a/LANGUAGES.md
+++ b/LANGUAGES.md
@@ -343,7 +343,7 @@ Thrift's core protocol is TBinary, supported by all languages except for JavaScr
 <td align=left><a href="https://github.com/apache/thrift/blob/master/lib/swift/README.md">Swift</a></td>
 <!-- Since -----------------><td>0.12.0</td>
 <!-- Build Systems ---------><td><img src="/doc/images/cgrn.png" alt="Yes"/></td><td><img src="/doc/images/cred.png" alt=""/></td>
-<!-- Language Levels -------><td colspan=2>4.2.1</td>
+<!-- Language Levels -------><td colspan=2>5.7</td>
 <!-- Field types -----------><td><img src="/doc/images/cred.png" alt=""/></td>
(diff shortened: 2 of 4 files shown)


kinoroy:
Hm seems like the repo only allows specific actions to be run, and doesn't allow the `swift-actions/setup-swift` action. 

This action is recommended by Github: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-swift so it seems reputable. Is it possible to change the repo settings to allow it?

Jimexist:
@kinoroy thank you for your contribution. at the moment i have no control over what GitHub action is allowed but i guess the fastest way to get this pr merged is to just manually configure swift - which i also did for `gradle`: download the binary, shasum check, and set execution flag

.github/workflows/build.yml
```diff
@@ -275,6 +275,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Setup Swift
+        uses: swift-actions/setup-swift@v1.18.0
+        with:
+          swift-version: "5.7"
```
jeking3, line 281: I would advise using the build containers over building an environment with GHA for every CI job (as it was, in the past).  Builds will take longer if you have to pull in every language like this, as opposed to downloading a one-time generated build container with everything in it.  With a build container anyone can use it to build the entire package locally and the local build is the same as the CI build.  With GHA scripted build environments like this, only GHA can run the build.

build/docker/ubuntu-jammy/Dockerfile
```diff
@@ -251,10 +251,13 @@ RUN apt-get install -yq \
   libpython2-dev \
   libxml2-dev && \
   cd / && \
-  wget --quiet https://download.swift.org/swift-5.7-release/ubuntu2204/swift-5.7-RELEASE/swift-5.7-RELEASE-ubuntu22.04.tar.gz && \
-  tar xf swift-5.7-RELEASE-ubuntu22.04.tar.gz --strip-components=1 && \
-  rm swift-5.7-RELEASE-ubuntu22.04.tar.gz && \
-  swift --version
+  wget --quiet https://download.swift.org/swift-5.7-release/ubuntu2004/swift-5.7-RELEASE/swift-5.7-RELEASE-ubuntu20.04.tar.gz && \
```
jeking3, line 254: Is 20.04 jammy?

.github/workflows/build.yml
kinoroy: I agree I like the old Travis way of having a docker container with everything setup in it. I'm not sure if you can do that on GitHub actions but I'd support that

build/docker/ubuntu-jammy/Dockerfile
kinoroy: oh good catch, 22.04 is jammy, 20.04 is focal

kinoroy:
I could remove the setup from GitHub actions from this PR because the current GitHub runner already includes Swift 5.7 but that might change in the future. Then we can resolve the issues with Github actions separately. 

The issue is mainly with Travis which includes a much older version that won't compile the current code

.github/workflows/build.yml
Jimexist: GitHub supports [hosted runners][1] but that requires infra work from apache i guess

[1]: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

Jimexist: if we only run on GitHub hosted runners, then the only way is to follow the current approach: [adding steps to customize the environment][1]

[1]: https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners

Jimexist:
> I could remove the setup from GitHub actions from this PR because the current GitHub runner already includes Swift 5.7 (but that might change in the future which is why I wanted to explicitly standardize it). Then we can resolve the issues with Github actions separately.
> 
> The issue is mainly with Travis which includes a much older version that won't compile the current code

i'm okay with this direction.

kinoroy:
I updated the PR, removing changes to GitHub actions, leaving only the docker container changes for Travis

Jimexist:
i'm merging this so that future changes on swift can iterate. thanks @kinoroy

PR #2723, opened 2022-10-28
THRIFT-5666: lib: cpp: prefer poll.h over sys/poll.h

cfriedt:
The `<sys/poll.h>` header is not actually where `poll(2)` must be defined according to the spec. However, it may be the case that some random UNIX-like OS uses `<sys/poll.h>` instead.

Both use cases can be supported simply by including the proper header, if it exists.

Tested with both the `autotools.sh` and `cmake.sh` build strategies, and is fine.

  

### DIFF
--- a/lib/cpp/src/thrift/transport/TNonblockingServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TNonblockingServerSocket.cpp
@@ -29,6 +29,9 @@
 #ifdef HAVE_SYS_UN_H
 #include <sys/un.h>
 #endif
+#ifdef HAVE_POLL_H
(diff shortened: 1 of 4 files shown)


cfriedt:
@Jens-G - can you please review? I can rebase if necessary, but otherwise, this is a relatively trivial change (`poll()` is declared in `<poll.h>` not `<sys/poll.h>`

cfriedt:
Actually, I'll respin this to just add `#ifdef HAVE_POLL_H` in a few places.

cfriedt:
@emmenlau - are you able to review? It's pretty trivial.

PR #2724, opened 2022-10-28
Use multi-module to cleanup top level go.mod

fishy:
Client: go

The go library itself does not have any third-party dependencies. We
have one third party dependency from the test code, which kind of
polluted from top-level go.mod file to the users of thrift go library.

We previous tried to clean that up by creating go.mod file at
lib/go/thrift, which caused issues to the release process and thus
reverted.

Use multi-module to separate tests requiring mock to their own modules
so that we can keep the top-level go.mod file clean.

Also some minor fixes on the github actions go workflow.


fishy:
cc @catenacyber this change is likely to break oss-fuzz integration slightly, you might need to tweak the args a bit after this is merged. Please let me know if the breakage is more than some args tweak.

catenacyber:
oss-fuzz build of thrift looks fine today :-)

fishy:
> oss-fuzz build of thrift looks fine today :-)

well that's because it's not merged yet, I believe oss-fuzz only run against the master version?

let me know if it causes problems after merged.

catenacyber:
Still looks good as of today

PR #2725, opened 2022-11-08
THRIFT-5667: Make ThriftConfig.cmake relocatable

kou:
Client: cpp
  
The current ThriftConfig.cmake isn't relocatable. It means that we can't move ThriftConfig.cmake installed by "make install".

In general, users don't need to move ThriftConfig.cmake installed by "make install" but some packaging systems such as conda and MSYS2 needs to move. Because users of these packaging systems may use different prefix. For example, conda may be used with ~alice/my-conda and ~bob/conda as base directory.

CMake has a feature to make Config.cmake relocatable: https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html

We can use it to make ThriftConfig.cmake relocatable.


kou:
Could someone review this?

kou:
Could someone take a look at this?

PR #2727, opened 2022-11-16
THRIFT-3956 Java keywords that are legal in IDL can lead to generated code that will not compile

Jens-G:
I might have overlooked a few cases. I used the test case from that ticket.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -39,6 +39,7 @@ using std::map;
 using std::ostream;
 using std::ostringstream;
 using std::setfill;
+using std::set;
 using std::setw;
 using std::string;
 using std::stringstream;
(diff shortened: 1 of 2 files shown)


compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -503,6 +510,89 @@ string t_java_generator::java_override_annotation() {
...
+{
+    return java_keywords;
+}
+
+void t_java_generator::init_keywords()
+{
+    java_keywords.clear();
+
+    // according to Oracle docs
+    java_keywords["abstract"] = 1;
```
Jimexist, line 523: can we just create an array and populate the map from that?

Jens-G: sure. in netstd there is a simikliar place that you could handle the same way

Jimexist (approved):

PR #2732, opened 2022-11-22
THRIFT-5670: lib: cpp: Fix wrong timeout error message

stiga-huang:
We've seen a strange timeout error message like this:
```
TSocket::read() THRIFT_EAGAIN (timed out) after %f ms: Unknown error 30000
```
The log comes from this line:
```cpp
GlobalOutput.perror("TSocket::read() THRIFT_EAGAIN (timed out) after %f ms", recvTimeout_);
```
The second argument of GlobalOutput.perror() is an errno. We should not use `recvTimeout_` on it:
https://github.com/apache/thrift/blob/a2a1f535813511aa274ed328b2318cb2beb7bb08/lib/cpp/src/thrift/TOutput.h#L40
```cpp
class TOutput {
  ...
  void perror(const char* message, int errno_copy);  // <-- The second arg is errno
  ...
  void printf(const char* message, ...);
};
THRIFT_EXPORT extern TOutput GlobalOutput;
```
The implementation of perror():
```cpp
void TOutput::perror(const char* message, int errno_copy) {
  std::string out = message + std::string(": ") + strerror_s(errno_copy);
  f_(out.c_str());
}
```
It converts the errno to the error string, and 30000 is an unknown errno.

What we want is to print the timeout threshold. We should use `GlobalOutput.printf()` instead.

This PR fixes the wrong usage of `GlobalOutput.perror()`. The error message will be
```
TSocket::read() THRIFT_EAGAIN (timed out) after 30000 ms
```

### DIFF
--- a/lib/cpp/src/thrift/transport/TSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSocket.cpp
@@ -569,7 +569,7 @@ try_again:
         throw TTransportException(TTransportException::INTERRUPTED, "Interrupted");
       }
     } else /* ret == 0 */ {
-      GlobalOutput.perror("TSocket::read() THRIFT_EAGAIN (timed out) after %f ms", recvTimeout_);
+      GlobalOutput.printf("TSocket::read() THRIFT_EAGAIN (timed out) after %d ms", recvTimeout_);
       throw TTransportException(TTransportException::TIMED_OUT, "THRIFT_EAGAIN (timed out)");
     }
 


lib/cpp/src/thrift/transport/TSocket.cpp
```diff
@@ -569,7 +569,7 @@ uint32_t TSocket::read(uint8_t* buf, uint32_t len) {
         throw TTransportException(TTransportException::INTERRUPTED, "Interrupted");
       }
     } else /* ret == 0 */ {
-      GlobalOutput.perror("TSocket::read() THRIFT_EAGAIN (timed out) after %f ms", recvTimeout_);
+      GlobalOutput.printf("TSocket::read() THRIFT_EAGAIN (timed out) after %d ms", recvTimeout_);
```
emmenlau, line 572: Dear @stiga-huang can you kindly help me understand why `printf` is better here than `perror`?

emmenlau (changes requested):

lib/cpp/src/thrift/transport/TSocket.cpp
stiga-huang: Sorry that I just mentioned it in the JIRA (https://issues.apache.org/jira/browse/THRIFT-5670). We saw a strange log when using cpp lib of thrift-0.16.0:
```
I1114 15:50:53.952184 16548 thrift-util.cc:196] TSocket::read() THRIFT_EAGAIN (timed out) after %f ms: Unknown error 30000
```
`recvTimeout_` (30000) is incorrectly used as `errno_copy` in `GlobalOutput.perror()`. `%f` is also not substituted as expected.
https://github.com/apache/thrift/blob/0.16.0/lib/cpp/src/thrift/TOutput.h#L40-L45
`printf` is what exactly we need. BTW, `recvTimeout_` is a `int` so replace `%f` to `%d`.

stiga-huang:
Ping @emmenlau @Jens-G

stiga-huang:
Updated the PR description. Hopefully it's more clear about the issue.

lib/cpp/src/thrift/transport/TSocket.cpp
emmenlau: This looks good, should be merged.

emmenlau (approved):
LGTM.

PR #2733, opened 2022-11-22
Fix cpp header issues

cfriedt:
2 commits to address 2 issues:
* lib: cpp: transport: include thrift/config.h in TFDTransport
  - Fixes https://app.travis-ci.com/github/apache/thrift/jobs/589120685
* lib: cpp: automake: ship thrift/numeric_cast.h
  - Fixes https://app.travis-ci.com/github/apache/thrift/jobs/589120686

cc @Jens-G

### DIFF
--- a/lib/cpp/Makefile.am
+++ b/lib/cpp/Makefile.am
@@ -133,6 +133,7 @@ libthriftqt5_la_LDFLAGS   = -release $(VERSION) $(BOOST_LDFLAGS) $(QT5_LIBS)
 include_thriftdir = $(includedir)/thrift
 include_thrift_HEADERS = \
                          $(top_builddir)/config.h \
+                         src/thrift/numeric_cast.h \
                          src/thrift/thrift-config.h \
                          src/thrift/thrift_export.h \
                          src/thrift/TDispatchProcessor.h \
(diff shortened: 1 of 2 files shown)


Jens-G:
I merged it because it changes the situation to the better. However I would say there is sth else:
https://app.travis-ci.com/github/apache/thrift/jobs/589165720
https://app.travis-ci.com/github/apache/thrift/jobs/589165723

The c_glib tests use a C++ server to test against. Seems there are some adjustments needed too, because now it complains about missing sleep(), fork() and throw()

Jens-G:
@cfriedt : Thoughts?

cfriedt:
> I merged it because it changes the situation to the better. However I would say there is sth else:
> https://app.travis-ci.com/github/apache/thrift/jobs/589165720
> https://app.travis-ci.com/github/apache/thrift/jobs/589165723
> 
> The c_glib tests use a C++ server to test against. Seems there are some adjustments needed too, because now it complains about missing sleep(), fork() and throw()
> 

@Jens-G - I noticed that as well. Shall I put together a fix for c_glib? I'm not sure yet what the root cause is yet.

(Sorry for the late response, i had left my comment in draft by accident)

Jens-G:
That would be super awesome.

PR #2740, opened 2022-12-20
netstd Server: Add IPv6

dejank-isystem:
Unlike servers in other languages (tested: cpp & python), netstd (Csharp) listens only on IPv4. This can cause errors and delays on clients if they use "localhost" for hostname.

  PS: Could you update PyPI when this gets released (I guess with version 0.18.0)?

### DIFF
--- a/lib/netstd/Thrift/Transport/Server/TServerSocketTransport.cs
+++ b/lib/netstd/Thrift/Transport/Server/TServerSocketTransport.cs
@@ -44,7 +44,8 @@ namespace Thrift.Transport.Server
             try
             {
                 // Make server socket
-                _server = new TcpListener(IPAddress.Any, port);
+		_server = new TcpListener(IPAddress.IPv6Any, port);
+                _server.Server.SetSocketOption(SocketOptionLevel.IPv6, SocketOptionName.IPv6Only, false);
                 _server.Server.NoDelay = true;
             }
             catch (Exception)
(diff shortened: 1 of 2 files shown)


dejank-isystem:
One mistake, when testing python server, I forgot to close netstd server. Error was in this that python does not listen both on IPv4 and IPv6, but only on IPv6. This can also be fixed with the `IPV6_V6ONLY` option set to `False`.

Also note that listening on both protocols is only valid when no hostname is given.

Jens-G:
Thanks!

Jens-G:
https://stackoverflow.com/a/28281485/499466

tisdall:
ok.. it's worse then I thought...  IPv4 addresses also fail:
```python
from thrift.transport.TSocket import TServerSocket
server_socket = TServerSocket(host='0.0.0.0')
server_socket.listen()
```

Output:
```
  File "/Users/ttisdall/repos/bespin/mu.py", line 3, in <module>
    server_socket.listen()
  File "/Users/ttisdall/.local/share/virtualenvs/bespin-uIM29DXQ/lib/python3.9/site-packages/thrift/transport/TSocket.py", line 232, in listen
    self.handle.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
OSError: [Errno 22] Invalid argument
```

tisdall:
Ah...  it's fixed in #2770 but not in a current release yet.

PR #2741, opened 2022-12-23
Improve enum support for json generator

cwe-dixa:
The json generator currently outputs the type `i32` for all enum fields. This commit changes it to output `enum` instead and output a `class` property with a reference to the actual enum type, similar to how it works for structs.

This is a breaking change for the json generator because it introduces a new type id (`enum`) which also includes a `class` reference.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_json_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_json_generator.cc
@@ -245,7 +245,7 @@ void t_json_generator::end_array() {
 
 void t_json_generator::write_type_spec_object(const char* name, t_type* ttype) {
   ttype = ttype->get_true_type();
-  if (ttype->is_struct() || ttype->is_xception() || ttype->is_container()) {
+  if (ttype->is_struct() || ttype->is_xception() || ttype->is_container() || ttype->is_enum()) {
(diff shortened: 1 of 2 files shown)


Jens-G:
LGTM

cwe-dixa:
There seems to be a bunch of errors in the cross test for a bunch of unrelated languages. The GA job log (https://github.com/apache/thrift/actions/runs/3764382167/jobs/6492240975) mentions that additional log files were uploaded for the failed cases somewhere. Is there someplace where I can view them?

cwe-dixa:
Is there anything I can do to help this along? :-)

anthonyalayo:
@cwe-dixa is there a way I can use this in the javascript/typescript world? I currently have some serialized thrift strings, and I would like to deserialize them with their enums readable. 

I tried the following:
```
export function read(ThriftClass: any, buffer: Buffer): string {
  const client = new ThriftClass();
  const transport = new thrift.TFramedTransport(buffer);
  const protocol = new thrift.TCompactProtocol(transport);

  client.read(protocol);

  return client;
```

But this causes all enums to be I32. Even though the types in the .d.ts files are there, the javascript doesn't do anything with them.

PR #2745, opened 2023-01-17
lib: cpp: transport: file descriptor leak in TServerSocket::close()

cfriedt:
Close a file descriptor leak in TServerSocket::close(). This was discovered in the final stages of porting Thrift to the Zephyr RTOS, which has a finite-sized file descriptor table.

Details at the URL below.
https://issues.apache.org/jira/browse/THRIFT-5677

cc: @Jens-G 

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -745,6 +745,9 @@ void TServerSocket::close() {
   if (childInterruptSockWriter_ != THRIFT_INVALID_SOCKET) {
     ::THRIFT_CLOSESOCKET(childInterruptSockWriter_);
   }
+  if (pChildInterruptSockReader_ != nullptr) {
+    ::THRIFT_CLOSESOCKET(*pChildInterruptSockReader_);
+  }
   serverSocket_ = THRIFT_INVALID_SOCKET;
   interruptSockWriter_ = THRIFT_INVALID_SOCKET;
   interruptSockReader_ = THRIFT_INVALID_SOCKET;


emmenlau (approved):
Looks good to me.

emmenlau:
@cfriedt , I guess this is beyond your current scope, but for the sake of clean code, would it not be better to bind the `std::shared_ptr` to a custom destructor that calls `::THRIFT_CLOSESOCKET` on the socket upon destruction? And then again, why is the operating system not doing that in the first place?

Please let me know if I'm completely on the wrong track here...

cfriedt:
> @cfriedt , I guess this is beyond your current scope, but for the sake of clean code, would it not be better to bind the `std::shared_ptr` to a custom destructor that calls `::THRIFT_CLOSESOCKET` on the socket upon destruction?

Maybe - that sounds a bit more complicated.

> And then again, why is the operating system not doing that in the first place?

I suspect that is exactly the issue - devs might expect the OS to close dangling FDs for them. However, neither Linux nor macOS do this either (not before the process ends). I have not checked what the status is for Windows.
 
> Please let me know if I'm completely on the wrong track here...

I would guess yea.

cfriedt:
At first glance, it seems like the build failures are unrelated, but I could take a deeper dive if I'm mistaken.

cfriedt:
@emmenlau - interesting, it seems that there is already a custom deleter, so perhaps it's the C++ runtime that is the problem.

```
    pChildInterruptSockReader_
        = std::shared_ptr<THRIFT_SOCKET>(new THRIFT_SOCKET(sv[0]), destroyer_of_fine_sockets);
```

I think I may close this PR, as it seems that could be the problem. Sorry for the trouble.

PR #2747, opened 2023-01-20
THRIFT-5678: lib: cpp: TConnectedClient: warning due to non-virtual dtor

cfriedt:
Commit 042580f53441efe1bc5c80c89351fcb30740659e removed the `virtual` keyword from the declaration of `~TConnectedClient()`.
    
While mostly benign, it does cause a warning in some versions  of GCC (non-virtual destructors can result in undefined behaviour), which can throw off CI sometimes when building with `-Werror`.

### DIFF
--- a/lib/cpp/src/thrift/TNonCopyable.h
+++ b/lib/cpp/src/thrift/TNonCopyable.h
@@ -30,7 +30,7 @@ namespace thrift {
 class TNonCopyable {
 protected:
   TNonCopyable() = default;
-  ~TNonCopyable() = default;
+  virtual ~TNonCopyable() = default;
 
   TNonCopyable(const TNonCopyable&) = delete;
   TNonCopyable& operator=(const TNonCopyable&) = delete;
--- a/lib/cpp/src/thrift/async/TConcurrentClientSyncInfo.h
+++ b/lib/cpp/src/thrift/async/TConcurrentClientSyncInfo.h
@@ -36,7 +36,7 @@ class TConcurrentClientSyncInfo;
 class TConcurrentSendSentry {
 public:
   explicit TConcurrentSendSentry(TConcurrentClientSyncInfo* sync);
-  ~TConcurrentSendSentry();
+  virtual ~TConcurrentSendSentry();
 
   void commit();
 
@@ -48,7 +48,7 @@ private:
 class TConcurrentRecvSentry {
 public:
   TConcurrentRecvSentry(TConcurrentClientSyncInfo* sync, int32_t seqid);
-  ~TConcurrentRecvSentry();
+  virtual ~TConcurrentRecvSentry();
 
   void commit();
 
--- a/lib/cpp/src/thrift/async/TEvhttpServer.h
+++ b/lib/cpp/src/thrift/async/TEvhttpServer.h
@@ -50,7 +50,7 @@ public:
    */
   TEvhttpServer(std::shared_ptr<TAsyncBufferProcessor> processor, int port);
 
-  ~TEvhttpServer();
+  virtual ~TEvhttpServer();
 
   static void request(struct evhttp_request* req, void* self);
   int serve();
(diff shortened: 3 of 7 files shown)


cfriedt:
@Jens-G @emmenlau - let me know your preferences. E.g. ensure all destructors are virtual, inline in `.h` vs in `.cpp`, etc.

Jens-G:
Not sure what the status is here?

cfriedt:
Zephyr v3.3 rc1 was just vmcut so I was a bit busy.

Additional work on this PR is required:
* Declare all dtors virtual in .h
* Implementations in .cpp (no .h inlines)

Jens-G:
Any idea how to continue with this? I would rather not decide about this one alone.

CJCombrink:
My 2c: `override` implies `virtual`. It is both a duplication and not conventional pattern to have both specified for the same function (I think SonarQube will warn on this and say the `virtual` should be removed).
See [override specifier](https://en.cppreference.com/w/cpp/language/override)
> In a member function declaration or definition, override specifier ensures that the function is virtual and is overriding a virtual function from a base class.

If this update is made at least the PR will touch a bit fewer files since this change is massive

Also refer to this code guideline: [C.128: Virtual functions should specify exactly one of virtual, override, or final](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final)

CJCombrink:
Looking at https://github.com/apache/thrift/commit/042580f53441efe1bc5c80c89351fcb30740659e yes `virtual` was removed from [lib/cpp/src/thrift/server/TConnectedClient.h](https://github.com/apache/thrift/commit/042580f53441efe1bc5c80c89351fcb30740659e#diff-049d157963556d333b246dfbe1a97ffce6c4f79dade8dea957b9abce6dab5795) but `override` was added.
This enforces that the destructors are virtual for at least that type.

> it does cause a warning in some versions of GCC

I think one would need to check specifically which classes gives this warning, since I expect it would not be TConnectedClient (since it is correct, virtual destructor)

Looking through this PR, I see there are actually a few classes that does not have virtual destructors, like [TConcurrentClientSyncInfo.h](https://github.com/apache/thrift/pull/2747/files#diff-410c896a5005757dab7dbc1e158977394e96b3f6d53d52c1714276c522d2b9f1). The question would probably be if they are base classes anywhere. Thus a change like this can be valuable.

@cfriedt I would suggest revert any change that adds virtual to a destructor already marked as `override` and then see where that leaves the PR.

CJCombrink (changes requested):
I would suggest revert any change that adds `virtual` to a destructor already marked as `override` and then see where that leaves the PR.

cfriedt:
Sure - thanks for the feedback.

There is one other change that I would need to do to upgrade the thrift version used in Zephyr after this. That is, namely, to redo the change where I added an alternative to some boost function (it's been so long I'm having a hard time remembering which).

And after that, I was thinking it would be interesting to try and coalesce fbthrift here too, because the newer compiler apparently generates really efficient c++ code with coroutines.

cfriedt:
@CJCombrink - my PR needs approval to run tests. Can you please approve? This change definitely looks cleaner without the `virutal ... override` destructors. I'm not 100% sure if the virtual keyword is required on the windows class variants, but I think it could probably be removed as well if there are no inheritors. What do you think?

cfriedt:
@Jens-G @emmenlau - any chance someone could approve and run tests?

emmenlau:
Looks indeed good for me, so thumbs up from my side.

cfriedt:
Any idea whether those CI errors exist already in the main branch?

Please let me know if there are any additional changes needed

CJCombrink (approved):

emmenlau:
> Please let me know if there are any additional changes needed

I would be so bold to say that your change can not create build errors. Still, I briefly checked the CI, and at least some of the build errors are unrelated. For example the MSVC build was fine except for a python-test failling.

Looks good to me, can be merged.

PR #2760, opened 2023-02-18
Fix Security Tests for OpenSSLv3

tobim:
OpenSSLv3 does stricter protocol matching in the handshake.

This change adapts the test matrix to work with either both widespread versions of OpenSSL.

Partially fixes [THRIFT-5672](https://issues.apache.org/jira/browse/THRIFT-5672).
  

### DIFF
--- a/lib/cpp/test/SecurityFromBufferTest.cpp
+++ b/lib/cpp/test/SecurityFromBufferTest.cpp
@@ -198,15 +198,16 @@ BOOST_AUTO_TEST_CASE(ssl_security_matrix) {
   try {
     // matrix of connection success between client and server with different SSLProtocol selections
         static_assert(apache::thrift::transport::LATEST == 5, "Mismatch in assumed number of ssl protocols");
+        bool ossl1 = OPENSSL_VERSION_MAJOR == 1;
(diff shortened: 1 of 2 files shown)


Jens-G:
Thanks! 

While we're at it ... why do we need that same matrix twice? Any possibility to consolidate that into just one easily?

Jens-G:
@tobim 

> C:\projects\thrift\lib\cpp\test\SecurityTest.cpp(223): error C2065: 'OPENSSL_VERSION_MAJOR': undeclared identifier [C:\projects\build\MSVC2017\x64\lib\cpp\test\SecurityTest.vcxproj]

tobim:
I guess this define is included in the prepocessed source by default on linux/unix but not on msvc. `#include <openssl/opensslv.h>` in both files should fix it, but I don't have access to that platform right now so I can't test it.

Docs are at: https://www.openssl.org/docs/man3.0/man3/OPENSSL_version_major.html

Jens-G:
> Docs are at: https://www.openssl.org/docs/man3.0/man3/OPENSSL_version_major.html

Tests are at: https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/46410992/job/mhmu1h5fghb3snpk

PR #2764, opened 2023-03-02
Add NewTServerSocketUnix() and NewTServerSocketTimeoutUnix()

sobomax:
The change adds ability to create servers that are listening on local unix domain sockets. We cannot implement it via existing API since necessary members of the TServerSocket() struct are not public (addr & clientTimeout).

### DIFF
--- a/lib/go/thrift/server_socket.go
+++ b/lib/go/thrift/server_socket.go
@@ -52,6 +52,18 @@ func NewTServerSocketFromAddrTimeout(addr net.Addr, clientTimeout time.Duration)
 	return &TServerSocket{addr: addr, clientTimeout: clientTimeout}
 }
 
+func NewTServerSocketUnix(listenAddr string) (*TServerSocket, error) {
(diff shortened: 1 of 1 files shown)


fishy:
Doesn't [`thrift.NewTServerSocketFromAddrTimeout`](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#NewTServerSocketFromAddrTimeout) work for unix domain sockets? you just create a `net.Addr` from [`net.ResolveUnixAddr`](https://pkg.go.dev/net#ResolveUnixAddr) to be passed in?

sobomax:
> Doesn't [`thrift.NewTServerSocketFromAddrTimeout`](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#NewTServerSocketFromAddrTimeout) work for unix domain sockets? you just create a `net.Addr` from [`net.ResolveUnixAddr`](https://pkg.go.dev/net#ResolveUnixAddr) to be passed in?

@fishy you are absolutely right! somehow I've overlooked it, thank you.

PR #2766, opened 2023-03-06
THRIFT-4838: Add unix socket support for Swift

kinoroy:

## What

Adds unix domain sockets as a low level transport option for the Swift library

## Why 

To improve Swift language support in Thrift. 

This change will, for example, enable the use of the popular [OSQuery library](https://github.com/osquery/osquery) natively in Swift (currently only supports Obj-C++)

## How

Extend the existing `TSocketServer` with a new initializer, taking a `path` input to a Unix domain socket path, and connecting to it. A new `UnixSocket` wrapper class was created for wrapping the different socket implementations in Linux and macOS.

## Compatibility

All changes are additive or equivalent; there are no breaking changes. 

## Testing

Enables `domain` socket tests in the cross-test suite for Swift.

Some cross-tests are failing due to other languages not implementing the `UUID` type or their implementation doesn't match the spec.
After rebasing on master it seems the cross-tests also seem to not be running right now in CI due to a failure to build the go library


Jens-G:
merged in 29d8773

PR #2772, opened 2023-04-05
THRIFT-5696: Allow custom TConfiguration for TByteBuffer.java

rizaon:
Client: java

TByteBuffer.java does not allow passing custom TConfiguration into constructor
https://github.com/apache/thrift/blob/66aac4465926404c2bb0c450e80fac2c2824c04c/lib/java/src/main/java/org/apache/thrift/transport/TByteBuffer.java#L18-L22 

Default TConfiguration limit message size is 100MB maximum. TByteBuffer initialization will fail with "MaxMessageSize reached" error for ByteBuffer longer than 100MB.

```
org.apache.thrift.transport.TTransportException: MaxMessageSize reached
	at org.apache.thrift.transport.TEndpointTransport.resetConsumedMessageSize(TEndpointTransport.java:58)
	at org.apache.thrift.transport.TEndpointTransport.updateKnownMessageSize(TEndpointTransport.java:71)
	at org.apache.thrift.transport.TByteBuffer.<init>(TByteBuffer.java:24) 
```

This PR adds one more constructor in TByteBuffer.java that accept custom TConfiguration.

### DIFF
--- a/lib/java/src/test/java/org/apache/thrift/transport/TestTByteBuffer.java
+++ b/lib/java/src/test/java/org/apache/thrift/transport/TestTByteBuffer.java
@@ -1,10 +1,12 @@
 package org.apache.thrift.transport;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import org.apache.thrift.TConfiguration;
 import org.junit.jupiter.api.Test;
 
 public class TestTByteBuffer {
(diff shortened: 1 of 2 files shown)


lib/java/src/test/java/org/apache/thrift/transport/TestTByteBuffer.java
```diff
@@ -39,4 +41,29 @@ public void testOverflow() throws Exception {
             () -> byteBuffer.write("Hello World".getBytes(StandardCharsets.UTF_8)));
     assertEquals("Not enough room in output buffer", e.getMessage());
   }
+
+  @Test
+  public void testSmallTConfiguration() throws Exception {
+    // Test that TByteBuffer init fail with small max message size.
+    final TConfiguration configSmall =
+        new TConfiguration(
+            4, TConfiguration.DEFAULT_MAX_FRAME_SIZE, TConfiguration.DEFAULT_RECURSION_DEPTH);
```
fishy, line 50: I'm not sure if we have a java style guide or something but this style looks weird to me, but that might be just me (as I don't write java nowadays).

fishy (approved):
Thanks for the patch! Looks good to me, I'll wait for the tests to pass and also to see if others have objections on it.

lib/java/src/test/java/org/apache/thrift/transport/TestTByteBuffer.java
rizaon: I think there is checkstyle built in the gradle script. This pass `gradle :spotlessApply`.

rizaon:
@fishy The check failure in Build / lib-go (1.19) does not seem to be related with this patch.
Can we have another try running the checks?

fishy:
> @fishy The check failure in Build / lib-go (1.19) does not seem to be related with this patch. Can we have another try running the checks?

that's unrelated and I already fixed it, you can either rebase your branch on top of latest master to retry, or just ignore it.

rizaon:
Thank you for accepting this PR!

PR #2777, opened 2023-04-07
Migration to JakartaEE 8

bogdandrutu:
Migrate to JakartaEE 8 dependencies which use the same namespace as the JavaEE 8.

### DIFF
--- a/lib/java/gradle.properties
+++ b/lib/java/gradle.properties
@@ -28,11 +28,11 @@ maven-repository-id=apache.releases.https
 httpclient.version=4.5.13
 httpcore.version=4.4.15
 slf4j.version=1.7.36
-servlet.version=4.0.1
+servlet.version=4.0.3
 tomcat.embed.version=9.0.43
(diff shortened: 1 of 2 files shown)


ctubbsii (commented):
I don't know enough about Gradle, but it seems like it'd be a good idea to try to use [the BOM that JakartaEE provides](https://central.sonatype.com/artifact/jakarta.platform/jakarta.jakartaee-bom/8.0.0/overview), for managing the dependency versions.

Maybe something like:

```gradle
implementation group: 'jakarta.platform', name: 'jakarta.jakartaee-bom', version: '8.0.0'
```

bogdandrutu:
Unfortunately the 8.0.0 bom has few problems and would like to not spread it in the world. See https://github.com/jakartaee/jakartaee-api/issues/136 and https://github.com/jakartaee/jakartaee-api/pull/49 (not included in the bom). Because of that I would avoid it until an 8.1.0 (or 8.0.1) is released.

PR #2779, opened 2023-04-09
THRIFT-5699: java lib and build tool chain: gradle 8.0.2

Jimexist:
- upgrade gradle to version 8
- ~upgrade toolchain java version to 19~
- upgrade shadow plugin due to incompatible gradle dependency

note this does not change built java library version which is pinned to 11

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -144,7 +144,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 19
+          java-version: 17
           cache: "gradle"
 
       - name: Install dependencies
@@ -431,7 +431,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          # here we intentionally use an older version so that we also verify java 19 compiles to it
+          # here we intentionally use an older version so that we also verify Java 17 compiles to it
           java-version: 11
           cache: "gradle"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -144,7 +144,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 17
+          java-version: 19
           cache: "gradle"
 
       - name: Install dependencies
@@ -431,7 +431,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          # here we intentionally use an older version so that we also verify java 17 compiles to it
+          # here we intentionally use an older version so that we also verify java 19 compiles to it
           java-version: 11
           cache: "gradle"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,7 +156,7 @@ jobs:
       - name: Setup gradle
         run: |
           wget https://services.gradle.org/distributions/gradle-$GRADLE_VERSION-bin.zip -q -O /tmp/gradle-$GRADLE_VERSION-bin.zip
-          (echo "7ba68c54029790ab444b39d7e293d3236b2632631fb5f2e012bb28b4ff669e4b  /tmp/gradle-$GRADLE_VERSION-bin.zip" | sha256sum -c -)
+          (echo "ff7bf6a86f09b9b2c40bb8f48b25fc19cf2b2664fd1d220cd7ab833ec758d0d7  /tmp/gradle-$GRADLE_VERSION-bin.zip" | sha256sum -c -)
           unzip -d /tmp /tmp/gradle-$GRADLE_VERSION-bin.zip
           sudo mv /tmp/gradle-$GRADLE_VERSION /usr/local/gradle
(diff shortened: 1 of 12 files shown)


.github/workflows/build.yml
```diff
@@ -137,14 +137,14 @@ jobs:
...
-      GRADLE_VERSION: "7.6"
+      GRADLE_VERSION: "8.0.2"
     steps:
       - uses: actions/checkout@v3
 
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 17
+          java-version: 19
```
fishy, line 147: 19 is not an LTS version, we should stick with LTS versions?

Jimexist: for maintainability i think we should stick with 11 (which is guaranteed using `-release` flag). i trust JDK's capability of down-compiling with newer Java versions.

for build tool chain i think it is okay to move along with newest stable release.

Jimexist: the `-release` flag was introduced by @ctubbsii

fishy: my main concern is that someone accidentally introduced some code that requires feature not available in LTS versions, but if that's already guaranteed by `-release` flag then I think this is fine.

the other concern is that the non-LTS versions all have a relatively much shorter support lifespan so we need to update those to supported versions much more frequently. For example, according to https://www.oracle.com/java/technologies/java-se-support-roadmap.html 19 is no longer supported?

Jimexist: we can move the java 21 in September but atm i think it is fine given there's no much of a support needed, given the compiled `.class` files are in Java 11. The only support needed is in the build system in this repo.

fishy: out of support would also mean that it won't get security fixes, so if a security bug is discovered in jdk 19 now the fix would only land on 20 and LTS versions, not 19, I assume? But maybe Oracle's "sustaining support" means it would still get security fixes?

if we are not using any new features introduced since 17 (the latest LTS version right now), I'm not sure why do we _need to_ use java 19 here?

slachiewicz: this is to check if we are prepared to run build/tests with upcoming releases.
I would recommend to use 20 here

doc/install/README.md
```diff
@@ -28,8 +28,8 @@ These are only required if you choose to build the libraries for the given langu
     * zlib (optional)
     * Qt (optional)
 * Java
-    * Java 17
-    * Gradle 7.6
+    * Java 19
```
slachiewicz, line 31: please leave 11/17 here

lib/java/gradle/unitTests.gradle
```diff
@@ -65,7 +65,7 @@ test {
         outputs.upToDateWhen { false }
     }
 
-    // This is required for Mockito to run under Java 17
+    // This is required for Mockito to run under Java 19
```
slachiewicz, line 68: leave as "17+"

lib/java/gradle/sourceConfiguration.gradle
```diff
@@ -31,7 +31,7 @@
 // also a runtime CI that's based on Java 11 to ensure that.
 java {
     toolchain {
-        languageVersion = JavaLanguageVersion.of(17)
+        languageVersion = JavaLanguageVersion.of(19)
```
slachiewicz, line 34: so with this definition - java defined with github actions setup is onoy used for Gradle process and not to build/run tests?

.github/workflows/build.yml
ctubbsii: I don't see any reason to test with the newer JDK that's available. But, I would suggest prioritizing testing against a JDK that developers are more likely to be using... like the current LTS. If we want to build against both, it's pretty easy to set up a Matrix build in GitHub Actions.

.github/workflows/build.yml
```diff
@@ -431,7 +431,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
-          # here we intentionally use an older version so that we also verify java 17 compiles to it
+          # here we intentionally use an older version so that we also verify java 19 compiles to it
```
ctubbsii, line 434: Could omit the version here, so it doesn't need to be updated so often. Just say:

```suggestion
          # intentionally use Java 11 here to ensure that the newer JDK builds for Java 11
```

lib/java/gradle/sourceConfiguration.gradle
ctubbsii: I think this forces developers to build using JDK 19. I would prefer not to do that. I think many developers are perfectly content to build using the latest LTS, which I think is still 17. The build should still work with later versions, but I don't think the build should require anything later than 17. I'm not very experienced with Gradle, but I think that's what this does.

Can we bump to the newest Gradle without forcing devs to develop using newer than JDK 17?

Jimexist:
@fishy @ctubbsii java version rolled back to 17

ctubbsii (approved):

Fokko:
Nice, thanks for picking this up @Jimexist

PR #2783, opened 2023-04-13
[THRIFT-5693] - fix bug in serialization of enum default values

andrewcunnin:
This change fixes a bug where the default values set in enum fields were unable to be serialized. Previously default enum values were being stored as enum objects upon initialization, where Thrift generally expects I32 values. With the changes in this pull request, serialization and deserialization are fixed and the python 3.4 style enums continue to work as expected.


test/ThriftTest.thrift
```diff
@@ -112,7 +112,7 @@ struct CrazyNesting {
   // Do not insert line break as test/go/Makefile.am is removing this line with pattern match
   3: required list<map<set<i32> (python.immutable = ""), map<i32,set<list<map<Insanity,string>(python.immutable = "")> (python.immutable = "")>>>> list_field,
   4: binary binary_field
-  5: uuid uuid_field
+  // 5: uuid uuid_field
```
andrewcunnin, line 115:
```suggestion
  5: uuid uuid_field
```

andrewcunnin, line 115:
```suggestion
  5: uuid uuid_field
```

fishy:
we actually don't have python set up in github actions yet so this is not really tested (we only tested that it did not break other languages that's setup in github actions)

can you show the diff between old and new compiler generated python code?

andrewcunnin:
> 

Gotcha, here's a diff of the ThriftTest.thrift generated Python:
https://gist.github.com/andrewcunnin/802f4d8fa779c181e2272b5219b52a07

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -2662,6 +2662,11 @@ string t_py_generator::declare_argument(t_field* tfield) {
 string t_py_generator::render_field_default_value(t_field* tfield) {
   t_type* type = get_true_type(tfield->get_type());
   if (tfield->get_value() != nullptr) {
+    if (type->is_enum() && gen_enum_) {
+        std::ostringstream out;
+        out << tfield->get_value()->get_integer();
+        return out.str();
+        }
```
fishy, line 2669: I see this diff from `ttypes.py`:
```diff
@@ -2931,7 +2931,7 @@ class OptionalEnum(object):
     )
 
 
-    def __init__(self, e=    Numberz.FIVE,):
+    def __init__(self, e=5,):
         self.e = e
 
     def read(self, iprot):
```
and also this diff:
```diff
@@ -7198,7 +7198,7 @@ OptionalBinary.thrift_spec = (
 all_structs.append(OptionalEnum)
 OptionalEnum.thrift_spec = (
     None,  # 0
-    (1, TType.I32, 'e', None,     Numberz.FIVE, ),  # 1
+    (1, TType.I32, 'e', None, 5, ),  # 1
 )
 fix_spec(all_structs)
 del all_structs
```
which I think is caused by this part of the change. from the diff I don't think this change is necessary? (it's probably still correct, but the old code generated seems to be better?)

test/py/SerializationTest.py
```diff
@@ -451,6 +460,8 @@ def _enumerate_enum(enum_class):
             self.assertEquals(obj, objcopy)
 
 
+
+
```
fishy, line 464: same here, you are making it 4 blank lines between the class above and `suite`.

fishy (commented):
the code looks good to me from the diff generated, but I'm not familiar with the python code enough so I'd still prefer to have some tests to verify it, and I failed to run python3 tests locally similar to how we didn't get the CI setup work in https://github.com/apache/thrift/pull/2787#issuecomment-1511840231.

andrewcunnin:
For the testing, is running the cross test and running `make -k check` in `test/py` sufficient? Or were you referring to a different set of tests that I should be running separately?

fishy:
> For the testing, is running the cross test and running `make -k check` in `test/py` sufficient? Or were you referring to a different set of tests that I should be running separately?

we have tests under both `test/py` and `lib/py/test`, and then we also have cross tests between different language libraries.

Jens-G:
Needs rebase

Jens-G:
This collides with the 284e6b30b8d6fe4522daf0c6982aa7eac8b8d01e commit which seems to address the same issue.

PR #2785, opened 2023-04-14
THRIFT-5702: Support Java 8

Fokko:
I think it makes sense to support Java 8 as well. It is still being supported until 2030, and it looks like we're not using any Java 11+ features.


Jimexist:
we kind of removed this not long ago. is there a possibility that downstream projects migrate to java 11?

Fokko:
@Jimexist Eventually all projects will migrate, but this isn't the case today. Parquet, Iceberg, and Hadoop are still very much on Java 8, and probably many more.

wgtmac:
> we kind of removed this not long ago. is there a possibility that downstream projects migrate to java 11?

That probably means `parquet-mr` cannot upgrade until java 8 is supported. Sigh.

Fokko:
@wgtmac I'm also all in for progressive upgrading of packages, but unfortunately, I think it is too early to drop Java 8.

Jimexist:
@ctubbsii and @fishy any concerns blocking this?

fishy:
I _thought_ the plan was that we are dropping support to java 8 then start using java 11 features.

If we don't have any plan for java 11 features, then sure.

ctubbsii:
> I _thought_ the plan was that we are dropping support to java 8 then start using java 11 features.
> 
> If we don't have any plan for java 11 features, then sure.

I thought that was the plan as well. I don't think there's a reason to stay at Java 8 for future releases.

ctubbsii:
> we kind of removed this not long ago. is there a possibility that downstream projects migrate to java 11?

That's my thought as well. If projects aren't ready to migrate to Java 11, they can just continue using the older versions of Thrift. That shouldn't stop newer versions of Thrift from moving forward on Java 11. If all projects wait on all other projects, nobody ever moves forward.

Fokko:
So Java 8 has been removed since [May 2022 in Thrift](https://github.com/apache/thrift/commit/c4e96c75758d94a79dae925b1daae0c950823057), with the argument to start using Java 11 features. From today and onwards this hasn't been used (it still compiles with Java 8). Java 8 is still supported (for security patches) until 2030 and many projects are still using it.

According to Maven central many people are still on `1.16.0` or `1.14.1`:
image

I took the time and effort to come over here to create a PR, but probably a lot of people don't care.

Why add another barrier to make it even harder for people to adopt newer versions? I think for `parquet-mr` it is very interesting to add support for UUIDs. However, we can't drop Java8 yet, and the price for not upgrading is not that high because the new features in the later Java versions are nice to have, but not required.

If Thrift cares about interoperability, and having support across many languages (including versions), then we should support Java8.

fishy:
> So Java 8 has been removed since [May 2022 in Thrift](https://github.com/apache/thrift/commit/c4e96c75758d94a79dae925b1daae0c950823057), with the argument to start using Java 11 features. From today and onwards this hasn't been used (it still compiles with Java 8). Java 8 is still supported (for security patches) until 2030 and many projects are still using it.
> 
> According to Maven central many people are still on `1.16.0` or `1.14.1`: image
> 
> I took the time and effort to come over here to create a PR, but probably a lot of people don't care.
> 
> Why add another barrier to make it even harder for people to adopt newer versions? I think for `parquet-mr` it is very interesting to add support for UUIDs. However, we can't drop Java8 yet, and the price for not upgrading is not that high because the new features in the later Java versions are nice to have, but not required.
> 
> If Thrift cares about interoperability, and having support across many languages (including versions), then we should support Java8.

The price is on both sides. Yes on one side dropping the support to java 8 means that projects still on java 8 can no longer update to newer version of thrift. But on the other side, not dropping the support to java 8 means that thrift project can not use any new feature that's introduced after java 8.

This is a matter we discussed in [thrift's maillist](https://lists.apache.org/thread/qlrtxc1yvrzdjkf1gqd2zcnvr49lz5y7), @Jimexist listed some of the features we _plan_ to use that requires the dropping of java 8.

wgtmac:
CMIW, at least it does not take too much work to add Java 8 support back to thrift 0.18.x version as Java 11 features are not yet used. Then downstream projects can have more time to migrate smoothly by upgrading to 0.18.x first and then prepare Java 11 upgrade. It would be harder for them to migrate from 0.16.x in the future.

Jimexist:
@Fokko could you fix the CI first?

Jimexist:
i am leaning towards merging this pull request. i don't have a solid plan of introducing java 11 only code soon, if anyone has that plan he/she can restart the upgrade process later.

Fokko:
> could you fix the CI first?

Done!

Jimexist (approved):
looks good to me but i suggest inviting one or two more to approve before merge

ctubbsii:
I still think it's a mistake to move backwards. We're just going to hit up against this again and need to have the conversation all over again.

There is value in moving the baseline, even if we haven't yet leveraged the newer Java features. If a contributor comes along and wants to make an improvement that uses Java 11, now they need to convince the community to move to Java 11 **and** adopt their changes. If we've already moved the baseline modestly (and 8 to 11 is pretty modest), then all they need to do is convince Thrift to adopt their changes on that newer baseline. If we don't move the baseline ahead of time, then we're asking contributors to justify moving to 11 whenever they want to make a change that uses 11. Those contributors are likely to face questions like: "can you rewrite this in a Java 8 way?" or "is this really important enough of a feature to justify moving to 11?". It's extra work for every contributor to justify the baseline move for *their* changes.

Projects with requirements on older Java versions can just use older Thrift releases. They *don't have to upgrade* Thrift so aggressively. If they need Java 8, they can use a version of Thrift that is appropriate for Java 8. They don't have to use the newer version. What kind of project is updating its dependencies so aggressively, but insists on its dependencies update themselves conservatively? Either those projects can update aggressively, and expect their dependencies to do the same, or they can update conservatively, and stay on the older version. Either way, it should not be a problem for Thrift to update to a newer Java version for its next release. This is no different than with other libraries. Consider Jetty. Jetty 11 uses JakartaEE (jakarta) instead of JavaEE (javax) libraries. We would obviously recognize the argument as flawed if somebody insisted that Jetty 11 continue using JavaEE libraries because JavaEE is still ubiquitous. No, those projects needing to stay using JavaEE would just use Jetty 10, and not upgrade to Jetty 11. Again, I will say it: you do not have to update. But that's *your* choice because it's better for *your* project. Thrift should still be able to update in its next releases, because that's better for the Thrift project.

slachiewicz (approved):

Fokko:
@ctubbsii Don't think we're going to agree on this one.

> Those contributors are likely to face questions like: "can you rewrite this in a Java 8 way?" or "is this really important enough of a feature to justify moving to 11?".

This didn't happen since May 2022 since it still compiles perfectly fine against Java 8...

> Again, I will say it: you do not have to update. But that's your choice because it's better for your project. Thrift should still be able to update in its next releases, because that's better for the Thrift project.

I also disagree there. I think it is best for Thrift as well to cater to the widest possible audience. Interoperability?

>  They don't have to upgrade Thrift so aggressively. 

Of course, this is true, but as you can see in the maven central screenshot above, there are security vulnerabilities. Also, it potentially blocks upgrading the transitive dependencies of Thrift. Also, I like to use UUIDs.

Anyway. I've updated the CI. Against Kotlin it is still Java11+ because the formatter doesn't work with Java8.

Jimexist:
> @ctubbsii Don't think we're going to agree on this one.
> 
> > Those contributors are likely to face questions like: "can you rewrite this in a Java 8 way?" or "is this really important enough of a feature to justify moving to 11?".
> 
> This didn't happen since May 2022 since it still compiles perfectly fine against Java 8...
> 
> > Again, I will say it: you do not have to update. But that's your choice because it's better for your project. Thrift should still be able to update in its next releases, because that's better for the Thrift project.
> 
> I also disagree there. I think it is best for Thrift as well to cater to the widest possible audience. Interoperability?
> 
> > They don't have to upgrade Thrift so aggressively.
> 
> Of course, this is true, but as you can see in the maven central screenshot above, there are security vulnerabilities. Also, it potentially blocks upgrading the transitive dependencies of Thrift. Also, I like to use UUIDs.
> 
> Anyway. I've updated the CI. Against Kotlin it is still Java11+ because the formatter doesn't work with Java8.

I think @ctubbsii has a point in that there's no perfect timing for thrift to upgrade - any time any one adds some new functionality it'll be hard to anyone still on Java 8 to pick it up if the upgrade is done. And sadly also there's no way to distribute with source instead of `.class` via maven central.

Can you find a counter argument that say maybe by 2024 it is a good time for thrift to upgrade to Java 11+? How about if later when I want to introduce https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html into the package?

pan3793 (approved):
+1, Thrift is one the most important infrastructure components in Hadoop ecosystem, and the whole Hadoop ecosystem stays on Java8 today.

ctubbsii:
> +1, Thrift is one the most important infrastructure components in Hadoop ecosystem, and the whole Hadoop ecosystem stays on Java8 today.

Hadoop [specifically says it supports running on Java 11 for 3.3 and up]( https://cwiki.apache.org/confluence/display/HADOOP/Hadoop+Java+Versions). It is not necessary to stay on Java 8 to work within the Hadoop ecosystem.

pan3793:
@ctubbsii it says support running on Java11, still supports Java8. And HADOOP-16795 is open, the latest Hadoop does not pass compiling in JDK11 now.

The latest AWS EMR uses Java8 by default.
https://docs.aws.amazon.com/emr/latest/ReleaseGuide/configuring-java8.html

ctubbsii:
> @ctubbsii it says support running on Java11, still supports Java8. And HADOOP-16795 is open, the latest Hadoop does not pass compiling on JDK11 now.

Yeah. That info was also at the link I provided. But it's not relevant. You don't need to build Thrift and Hadoop in the same environment for them to be compatible. You just need them to run in the same environment, and Hadoop runs just fine in Java 11.

> The latest AWS EMR uses Java8 by default. https://docs.aws.amazon.com/emr/latest/ReleaseGuide/configuring-java8.html

I'm not sure how this is relevant at all. EMR's default configuration has no bearing on whether Hadoop requires Thrift to stay on 8 or not. EMR can be configured to run on 11, and you can run Hadoop without EMR.

pan3793:
> I'm not sure how this is relevant at all.

@ctubbsii All things I listed prove that Java8 is adopted widely and has good support in the Hadoop ecosystem today, to be clear, I don't oppose thrift support new Java versions, I just think we should not drop Java8 support.

ctubbsii:
> @ctubbsii All things I listed prove that Java8 is adopted widely and has good support in the Hadoop ecosystem today

But nobody is disputing that Java 8 is widely adopted *today*. Development on new versions of Thrift aren't for *today*. New releases of Thrift would be released to be adopted in the *future*.

The fact that Hadoop has been made to run on Java 11 means that related projects no longer need to be held back on Java 8 and risk being incompatible. It is now possible for projects like Thrift to move to 11, and still work with Hadoop and other projects. Hadoop does not require Thrift to stay on Java 8. So... if the Thrift project has other reasons for moving (like simplifying release test matrices, or lowering the burden on contributors who want to use Java 11, which I think are good reasons to move to 11), that should be possible and not be a problem for the Hadoop ecosystem (or other similarly situated projects and their respective ecosystems).

pan3793:
> Development on new versions of Thrift aren't for today. New releases of Thrift would be released to be adopted in the future.

@ctubbsii That's a good point, but in the meanwhile, as a widely adopted component, *today*'s versions should get actively maintained too.

What are the versions for *today*? And what's the support policy of *today*'s versions?

Based on the statistics of [Maven Central](https://mvnrepository.com/artifact/org.apache.thrift/libthrift), the most adopted versions are 0.9.x and 0.12.x, can we treat them as versions for *today*? Can I request a security-patched/bug-fix version for them?

image

Thrift 0.13.0 made lots of [breaking changes](https://github.com/apache/thrift/blob/v0.13.0/CHANGES.md#breaking-changes) including THRIFT-4725 in Java, that's one of the reasons why the lower versions are adopted widely today, even they have known CVEs. Drops Java8 support is another significant breaking change.

Another fact is the indirect consumer of Thrift may take a quite long time to adopt the new Thrift when breaking changes are introduced.

Let's say that Apache Hive is a direct consumer of Apache Thrift, the latest stable release of Hive is [3.1.3](https://mvnrepository.com/artifact/org.apache.hive/hive-metastore/3.1.3), which adopts Thrift 0.9.3.

Apache Spark depends on Hive 2.3.9(which also uses Thrift 0.9.3) now, Spark adopts Thrift 0.12.0 because it is compatible w/ 0.9.3, but can not use 0.13 or any above versions, because of the breaking change THRIFT-4725. (See detail in SPARK-37090)

Assume the next Hive stable version 4.0.0 available in 2023, which likely ships [Thrift 0.16](https://mvnrepository.com/artifact/org.apache.hive/hive-metastore/4.0.0-alpha-2), because it still supports Java8. It may take several years for the downstream project(e.g. Spark) of Hive to adopt Hive 4.0 and therefore adopts Thrift 0.16. But what about the downstream of Spark? Will 0.16 get maintained on those days?

ctubbsii:
> What are the versions for _today_? And what's the support policy of _today_'s versions?

This is a relatively small project. I am not aware of any formal support policy, but I would happily embrace maintenance releases for recent versions with critical bugs to fix. That requires more active committers, and voting PMC members (not just regular contributors).

> Based on the statistics of [Maven Central](https://mvnrepository.com/artifact/org.apache.thrift/libthrift), the most adopted versions are 0.9.x and 0.12.x, can we treat them as versions for _today_? Can I request a security-patched/bug-fix version for them?

To make that happen, there needs to be sufficient demand for them, and more support to prepare releases. I've been following the mailing lists for some time now, and I have not seen a high demand for maintenance releases. Preparing release candidates seems to be done by relatively few people. That could change. I think the PMC should decide how they want to address that.

> Thrift 0.13.0 made lots of [breaking changes](https://github.com/apache/thrift/blob/v0.13.0/CHANGES.md#breaking-changes) including THRIFT-4725 in Java, that's one of the reasons why the lower versions are adopted widely today, even they have known CVEs. Drops Java8 support is another significant breaking change.

In my experience, *every* version of Thrift has been accompanied by a breaking change. This is one of the reasons why I don't understand why people seem to want to upgrade it so aggressively, such that they need to impose constraints on the anticipated future versions of Thrift. I would upgrade Thrift very conservatively, due to these breaking changes. By the time you upgrade to version of Thrift that requires Java 11, I would expect consuming projects to already be ready to move to 11, so it shouldn't be a problem.

I think a better approach, rather than hold back future versions of Thrift, is to encourage more maintenance releases on previous versions. That would solve the Java dependency problem *and* solve the breaking change issues that appear in each Thrift release.

Jens-G:
>  the most adopted versions are 0.9.x and 0.12.x, can we treat them as versions for today? Can I request a security-patched/bug-fix version for them?

Releases are currently done about twice a year, patches only when its really, really necessary. Although we have released patch versions in the past, the usual policy is to maintain only the last one. 

Since we also got a nice email from INFRA some time ago in that regard, we now also consequently remove any old versions from the ASF download servers, i.e. you can't even download them anymore that way. Only whats out there and available through 3rd party package managers and CDNs like maven, nuget or the like, and of course [right here from GitHub](https://github.com/apache/thrift/releases).

ctubbsii:
> Releases are currently done about twice a year, patches only when its really, really necessary. Although we have released patch versions in the past, the usual policy is to maintain only the last one.
> 
> Since we also got a nice email from INFRA some time ago in that regard, we now also consequently remove any old versions from the ASF download servers, i.e. you can't even download them anymore that way. Only whats out there and available through 3rd party package managers and CDNs like maven, nuget or the like, and of course [right here from GitHub](https://github.com/apache/thrift/releases).

You can still keep up the latest version in any release series that you consider "currently active". Accumulo has the latest 1.10.x and 2.1.x right now. Hadoop has the latest 3.3.x, 3.2.x, and  2.10.x. ZooKeeper has 3.5.x, 3.6.x, 3.7.x, and 3.8.x (they might have a few extra that they need to remove, as per that INFRA policy, though). The Thrift PMC could make a similar decision to consider certain versions as "currently active", even if they are only active for critical fixes.

pan3793:
> In my experience, every version of Thrift has been accompanied by a breaking change. This is one of the reasons why I don't understand why people seem to want to upgrade it so aggressively

@ctubbsii How do you handle the CVEs then? The security team in some companies forcibly bans the jars which were reported CVEs, even if the project which uses Thrift does not really get affected. Similar questions on bugs.

> ... such that they need to impose constraints on the anticipated future versions of Thrift.
> I think a better approach, rather than hold back future versions of Thrift, is to encourage more maintenance releases on previous versions.

The fact is Thrift lacks support for *today*'s widely adopted version, in the meanwhile, it aggressively makes breaking changes, I suppose the option you suggested needs more community resources than "avoid breaking changes as much as possible".

ctubbsii:
> > In my experience, every version of Thrift has been accompanied by a breaking change. This is one of the reasons why I don't understand why people seem to want to upgrade it so aggressively
> 
> @ctubbsii How do you handle the CVEs then? The security team in some companies forcibly bans the jars which were reported CVEs, even if the project which uses Thrift does not really get affected. Similar questions on bugs.

Case-by-case mitigation, if affected. Can patch/local fork or mitigate in calling code, or mitigate with firewall, or something else. Just depends on the impact. Ideally, we advocate upstream do maintenance releases to make this easier on everybody downstream.

Jens-G:
> You can still keep up the latest version in any release series that you consider "currently active"

But that's what we do.- The most recent one.

ctubbsii:
> > You can still keep up the latest version in any release series that you consider "currently active"
> 
> But that's what we do.- The most recent one.

Maybe I wasn't clear enough. What I was trying to say is that you could have more than one release series considered "currently active" and leave up the most recent of *each* series. That would still be compatible with INFRA release distribution policy.

For example, if the PMC decided that 0.9.x and 0.12.x are considered currently active in addition to the latest, 0.18.x, then you could have 0.9.3.1, 0.12.0, and 0.18.1 up there. INFRA distribution policy would only dictate that 0.18.0, 0.9.0, 0.9.1, 0.9.2, and 0.9.3 are archived (plus all 0.13.x-0.17.x). It doesn't require you to archive 0.9.3.1 and 0.12.0 if the PMC still considers those to be active.

What seems to be happening today is that Thrift only considers 0.18.x to be currently active, so 0.18.1 is available, but nothing earlier. But that can change with a simple PMC decision.

Jens-G:
Propose to move that discussion to mailing list.

Jimexist:
can you also update https://github.com/apache/thrift/blob/4f63573f5a49fb564e7b65b9573769963511dbea/.github/workflows/build.yml#L438 to 8 to make sure it runs with JVM 8

Jimexist:
For this specific change, i wonder if we can leverage [maven artifact classifier](https://docs.gradle.org/current/userguide/feature_variants.html) to support both versions?

ctubbsii:
> For this specific change, i wonder if we can leverage [maven artifact classifier](https://docs.gradle.org/current/userguide/feature_variants.html) to support both versions?

That's not necessary, and a bit overkill for libthrift.jar to produce separate artifacts. The artifacts will basically be equivalent, as Thrift doesn't use any complicated language-specific features/assumptions. The only difference would be byte code version, and Java 8 byte code runs just fine in Java 11, so it's just not necessary.

Fokko:
@Jimexist Ah, missed that. Done!

Jimexist:
At least I don't think Java and kotlin cross tests should fail - can you check that as well?

Fokko:
@Jimexist Looks like Kotlin requires still Java 8+:

```
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: org/apache/thrift/test/TestClientKt has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:601)
```

Fokko:
Hey all, we would still love to get this in. With the last release of Parquet, a lot of people were struggling to get Thrift installed, so maintaining Java 8 support until we start using Java 11 features, would be awesome!

vinooganesh:
> Propose to move that discussion to mailing list.

Did this ever make it to a mailing list conversation? Was there a decision that you'd be able to message out?

Jens-G:
> Did this ever make it to a mailing list conversation?

Sure, It's in the archives. 

BTW I wonder why one would need 3 tickets with the exact same content?

Fokko:
Thanks @Jens-G for pushing this forward, appreciate it 🙏🏻

PR #2787, opened 2023-04-14
THRIFT-5564: add GitHub action for python 2.x and 3.x

Jimexist:
- closes #2596

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -428,3 +428,61 @@ jobs:
           name: cross-test-log
           path: test/log/
           retention-days: 3
+
+  lib-python:
+    needs: compiler
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ["2.x", "3.x"]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -yq
+          sudo apt-get install -y --no-install-recommends $BUILD_DEPS
+          sudo apt-get install -y --no-install-recommends curl
+
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Python setup
+        run: |
+          python -m pip install --upgrade pip setuptools wheel flake8
+          python --version
+          pip --version
+
+      - name: Run bootstrap
+        run: ./bootstrap.sh
+
+      - name: Run configure 2.x
+        if: matrix.python-version == '2.x'
+        run: |
+          ./configure $(echo $CONFIG_ARGS_FOR_LIBS | sed 's/without-python/with-python/')
+
+      - name: Run configure 3.x
+        if: matrix.python-version == '3.x'
+        run: |
+          ./configure $(echo $CONFIG_ARGS_FOR_LIBS | sed 's/without-python3/with-python3/')
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: thrift-compiler
+          path: compiler/cpp
+
+      - name: Run thrift-compiler
+        run: |
+          chmod a+x compiler/cpp/thrift
+          compiler/cpp/thrift -version
+
+      - name: Run make for python
+        run: make -C lib/py
+
+      - name: Run make install for python
+        run: sudo make -C lib/py install
+
+      - name: Run make check for python
+        run: make -C lib/py check
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
 
       - name: Run configure
         run: |
-          ./configure $(echo $CONFIG_ARGS_FOR_LIBS | sed 's/without-rust/with-rust/')
+          ./configure $(echo $CONFIG_ARGS_FOR_LIBS | sed 's/without-rs/with-rs/')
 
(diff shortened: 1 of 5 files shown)


fishy:
I don't know why this didn't trigger github actions, probably there are something wrong in the yaml makes it failed to parse the config?

Jimexist:
> I don't know why this didn't trigger github actions, probably there are something wrong in the yaml makes it failed to parse the config?

yes it was an indentation but which is now fixed

fishy:
the 3.x one failed, we probably should also set [`fail-fast`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to false so that one failure won't auto cancel other jobs.

Jimexist:
> the 3.x one failed, we probably should also set [`fail-fast`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to false so that one failure won't auto cancel other jobs.

> ModuleNotFoundError: No module named 'thrift'

could it be because of some path handling logic not being mirrored in the GitHub workflow code?

Jimexist:
@fishy i have no knowledge of python lib tests, can you tell the cause here? https://github.com/apache/thrift/actions/runs/4707202819/jobs/8349103781?pr=2787

Jimexist:
> @fishy i have no knowledge of python lib tests, can you tell the cause here? https://github.com/apache/thrift/actions/runs/4707202819/jobs/8349103781?pr=2787

turns out i need to backport a list of libraries for python 2

.github/workflows/build.yml
```diff
@@ -344,12 +344,79 @@ jobs:
       - name: Run make test_recursive for rust
         run: make -C lib/rs/test_recursive check
 
+  lib-python:
+    needs: compiler
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ["2.x", "3.x"]
```
Fokko, line 352: I would drop Python v2. Python 2 has been sunsetted 1st of January 2020: https://www.python.org/doc/sunset-python-2/

Fokko: It would be nice to have a matrix to test against different versions of Python 3 as well:

https://github.com/apache/iceberg/blob/9dd00fab1dc160bb531091cbdda8f19ef5b3a32e/.github/workflows/python-ci.yml#L39-L42

This way you can also be more explicit in the supported Python versions (sometimes stuff breaks). Instead of:
https://github.com/apache/thrift/blob/1e3d90d8fd4160d538b7a4d902169eae5155e08a/lib/py/setup.py#L136

You can be more explicit in versions:
https://github.com/apache/airflow/blob/8d81963c014398a7ab14505fd8e27e432f1aaf5c/setup.cfg#L39-L42

Fokko (commented):
I have some small comments, great to have this in the CI as well 👍🏻

.github/workflows/build.yml
Jimexist: > I would drop Python v2. Python 2 has been sunsetted 1st of January 2020: https://www.python.org/doc/sunset-python-2/

feel free to review https://github.com/apache/thrift/pull/2588

fishy:
>>ModuleNotFoundError: No module named 'thrift'

>could it be because of some path handling logic not being mirrored in the GitHub workflow code?

possibly. I'm getting the same error locally when trying to run the tests for python3 as well, so I guess some setup steps are missing but I don't know which (I'm not familiar with the python bootstraps either).

Jimexist:
@fishy how do you think i can mark the failing tests skipped and merge this pull request first. this pull request is at least marginally better than the status quote which is without **any** CI.

.github/workflows/build.yml
```diff
@@ -344,12 +344,79 @@ jobs:
...
+        run: make -C lib/py
+
+      - name: Run make install for python
+        run: sudo make -C lib/py install
+
+      # - name: Run make install-exec-hook for python
+      #   run: sudo make -C lib/py install-exec-hook
+
+      - name: Run make check for python
+        run: make -C lib/py check
```
fishy, line 411: we actually have two sets of tests, one under `lib/py` (which is the one here) and another under `test/py`. can you also add a step to run `make -C test/py check`?

.github/workflows/build.yml
```diff
@@ -344,12 +344,79 @@ jobs:
...
+
   cross-test:
     needs:
       - lib-java-kotlin
       - lib-swift
       - lib-rust
       - lib-go
+      - lib-python
```
fishy, line 419: we actually also need to add python (I'm not sure if it's gonna be `py3` or `python`) to line 483 (`PRECROSS_LANGS`) to actually run the cross tests with python.

lib/py/src/server/TNonblockingServer.py
```diff
@@ -268,7 +268,7 @@ def prepare(self):
         self.socket.listen()
         for _ in range(self.threads):
             thread = Worker(self.tasks)
-            thread.setDaemon(True)
+            thread.daemon = True
```
fishy, line 271: 🔕 oh I see `setDaemon` was deprecated since python 3.10.

fishy (approved):
my comments can totally be done in a follow up PR instead. I certainly agree that this is already much better than what we have today.

PR #2805, opened 2023-05-11
Stop sharing write headers across all instances of transports

ngavalas:
There is an issue with the header transport implementation in `nodejs`. All TBufferedTransport and TFramedTransport instances across all connections and clients share a single object for their read and write headers. There is a different copy for the two transport types, but within one type, they are shared.

This means that there are both data races and incorrect results possible: 
* You can write headers to a request that has written headers but hasn't flushed yet. This is especially troubling if the headers are used for auth, because you're able to mix up requests and auth as the wrong entity.
* You can read headers from other requests. They only clobber each other if they have the same name, but the union of all seen headers is returned in `getReadHeaders `.

The issue is that we aren't calling the header transport constructor with the new `this` from the concrete implementations, so the `TBufferedTransport.prototype = new THeaderTransport();` line is not properly binding `this` and causes the object sharing.

You can see this in the two test cases that fail before this PR and succeed now. Two **completely unrelated** `TFramedTransport` objects shared read and write results.

I did not create a JIRA ticket yet because I am still waiting on approval, but I wanted to get the PR up ASAP; we are currently experiencing this bug in production.


lib/nodejs/test/header.test.js
```diff
@@ -39,6 +39,19 @@ const cases = {
...
   },
+  "Should read different headers from different payload": function(assert) {
+    const transport = new TFramedTransport();
+    const buf = Buffer.from(headerPayload);
+    buf[24] = 115; // Change Parent to Parens
+    buf[32] = 122; // Change shoobar to shoobaz
+    transport.inBuf = buf;
+
+    const headers = transport.readHeaders();
+    assert.equals(headers.Parent, undefined);
```
ngavalas, line 50: Just to be 100% clear, before this fix, `headers` would have contained `{ Parent: "shoobar", Parens: "shoobaz:", Trace: "abcde" }` because it was reusing the same object as the first test. This is not test-only though and happens in real life.

Similar story for the write test.

ngavalas:
@Jens-G I'm going to assume that the cross-test failures are not from me since none of them involve nodejs. Safe to assume?

Jens-G:
Yes, seems so.

ngavalas:
Sounds good. Thoughts on the bug and the PR? We're blocked on using nodejs Thrift on this, so I'm eager to get this one in.

ngavalas:
@Jens-G Anything you need from me to review this? Happy to do whatever it takes!

ngavalas:
Thanks @Jens-G! Can we cut a new npm package as well?

Jens-G:
Sure, after release as usual there will be a new npm package.

PR #2806, opened 2023-05-12
THRIFT-5709: Drastically improve to_string() performance.

tinloaf:
Currently, the `to_string()` overloads in `TToString.h` call the `std::locale` constructor for every single call. Creating locales is surprisingly expensive. We have an application where we - especially during tests - write large amounts of Thrift dumps to disk, and is this application we currently spend around 17% of total CPU time in std::locale's constructor (building with MSVC, in a Release build - other compilers might optimize this away?). With this change, it's basically down to zero.

Since we always create the exact same locale anyways, using a `const static` one does not hurt.

### DIFF
--- a/lib/cpp/src/thrift/TToString.h
+++ b/lib/cpp/src/thrift/TToString.h
@@ -35,7 +35,8 @@ namespace thrift {
 template <typename T>
 std::string to_string(const T& t) {
   std::ostringstream o;
-  o.imbue(std::locale("C"));
+  const static auto locale = std::locale("C");
+  o.imbue(locale);
   o << t;
   return o.str();
 }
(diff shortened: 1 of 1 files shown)


Jens-G:
Why do we need 4 instances?

tinloaf:
> Why do we need 4 instances?

I don't think we do, I just wanted to make the change as small as possible. Would you prefer a global constant (in an anonymous namspace probably)?

Jens-G:
Well, I just thought if we are going to optimize, why don't we do it right ... so yes, would make sense to me, if that's not causing other overhead, too much confusion or work, then I'm for it.

lib/cpp/src/thrift/TToString.h
```diff
@@ -32,10 +32,15 @@
 namespace apache {
 namespace thrift {
 
+// unnamed namespace to enforce internal linkage - could be done with 'inline' when once have C++17
```
tinloaf, line 35: TODO: `when`/`once` is duplicate here.

PR #2810, opened 2023-05-16
THRIFT-5711: Fix FutureClient implementation when service extends from another service

bogdandrutu:
The `FutureClient` should extend the parent service `FutureClient` because the `FutureIface` does that too. Without this change the `FutureClient` will get a compilation error that does not implement all the methods.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -3369,6 +3369,7 @@ void t_java_generator::generate_service_future_client(t_service* tservice) {
   string extends_client = "";
   if (tservice->get_extends() != nullptr) {
     extends_client = "extends " + type_name(tservice->get_extends()) + ".FutureClient ";
+  }
 
   static string adapter_class = "org.apache.thrift.async.AsyncMethodFutureAdapter";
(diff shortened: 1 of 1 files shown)


ctubbsii (approved):
LGTM

Jimexist (approved):

Jimexist:
change looks good to me but please fix the compile error first

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -3366,8 +3366,13 @@ void t_java_generator::generate_service_client(t_service* tservice) {
 }
 
 void t_java_generator::generate_service_future_client(t_service* tservice) {
+  string extends_client = "";
+  if (tservice->get_extends() != nullptr) {
+    extends_client = "extends " + type_name(tservice->get_extends()) + ".FutureClient ";
```
ctubbsii, line 3371:
```suggestion
  if (tservice->get_extends() != nullptr) {
    extends_client = "extends " + type_name(tservice->get_extends()) + ".FutureClient ";
  }
```

ctubbsii:
I went ahead and fixed the compile error for them, since it was just a missing close brace. I think it's good now.

bogdandrutu:
Thanks @Jimexist and @ctubbsii

PR #2811, opened 2023-05-19
THRIFT-5706: lib: cpp: Fix C++ SecurityTest compilation on OpenSSL1.x

joemcdonnell:
A recent change modified SecurityTest.cpp/SecurityFromBufferTest.cpp to handle OpenSSL3 by using OPENSSL_VERSION_MAJOR to detect the different OpenSSL versions. OPENSSL_VERSION_MAJOR is not available on OpenSSL 1.x, so these tests no longer compile on Ubuntu 20, etc.

The underlying reason for the OpenSSL3 change is that older versions of TLS (1.0, 1.1) are only accessible when specifying @SECLEVEL=0. Some distributions with OpenSSL 1.1.1 (Ubuntu 20) make them accessible at @SECLEVEL=1, so the tests can also fail in that circumstance.

This removes the reference to OPENSSL_VERSION_MAJOR and changes the test to specify @SECLEVEL=0 for OpenSSL versions that have @SECLEVEL support (> 1.1.0).

Testing:
 - Tested SecurityTest/SecurityFromBufferTest on Ubuntu 20 and 22
 - Tested with OpenSSL 1.0.2

 This fixes the build/functionality for SecurityTest/SecurityFromBufferTest after the changes in https://github.com/apache/thrift/commit/05604e261455f1d85a5d04c4364a21d2a7e417b2

### DIFF
--- a/lib/cpp/test/SecurityFromBufferTest.cpp
+++ b/lib/cpp/test/SecurityFromBufferTest.cpp
@@ -109,7 +109,13 @@ struct SecurityFromBufferFixture {
       shared_ptr<TSSLServerSocket> pServerSocket;
 
       pServerSocketFactory.reset(new TSSLSocketFactory(static_cast<apache::thrift::transport::SSLProtocol>(protocol)));
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+      // OpenSSL 1.1.0 introduced @SECLEVEL. Modern distributions limit TLS 1.0/1.1
+      // to @SECLEVEL=0 or 1, so specify it to test all combinations.
+      pServerSocketFactory->ciphers("ALL:!ADH:!LOW:!EXP:!MD5:@SECLEVEL=0:@STRENGTH");
+#else
       pServerSocketFactory->ciphers("ALL:!ADH:!LOW:!EXP:!MD5:@STRENGTH");
+#endif
       pServerSocketFactory->loadCertificateFromBuffer(certString("server.crt").c_str());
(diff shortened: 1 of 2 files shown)


emmenlau:
@joemcdonnell can you check if this issue persists with latest thrift master? I think I fixed it, without seeing your PR before. But I did not check the details, so I may be wrong. I can say that my thrift build works with openSSL 1.x. But I may be using different options than you are. Kindly help and clarify?

emmenlau:
Thanks a lot @thomasbruggink , I think your work here is very valid and I would consider this the correct way forward. However, making backwards-incompatible changes is always hard in wide-spread projects like thrift. Therefore I have a slight preference to combine this work here with your changes suggested in https://github.com/apache/thrift/pull/2940, so that the old SSL/TLS versions continue to be supported.

Could you kindly combine the work into a single PR here, and fix the minor conflicts? Then ping me again, and I'll try to merge it ASAP!

thomasbruggink:
@emmenlau thanks for the response, pushed a merged fix :bow: https://github.com/apache/thrift/pull/2940/commits/3ea763e068fbedf7b7b33b73ca294952543e7930

thomasbruggink:
@emmenlau any concern if we merge this or my PR? I think at this point the implementation is the same.
It would be great if the test suite of cpp can pass again.

emmenlau:
I think this is now superseded by https://github.com/apache/thrift/pull/2940. Closing.

Kindly re-open if something is missing here...

PR #2815, opened 2023-06-06
THRIFT-5714 add TJSONProtocol support for swift

cozality:
The code contains the changes required to support TJSONProtocol in the Thrift-Swift. The codes are ported from C#.NET Thrift official library

Add support TJSONProtocol via implementing it.
Jira Ticket: https://issues.apache.org/jira/browse/THRIFT-5714

- [x ] Did you squash your changes to a single commit?  (not required, but preferred)
- [ x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?

### DIFF
--- a/lib/swift/Sources/TBase64Utils.swift
+++ b/lib/swift/Sources/TBase64Utils.swift
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import Foundation
+
+public class TBase64Utils {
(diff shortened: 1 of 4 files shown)


lib/swift/Tests/ThriftTests/TJSONProtocolTests.swift
```diff
@@ -0,0 +1,212 @@
+//
+//  File.swift
+//
+//  Created by Khodabakhsh, Hojjat on 5/18/23.
+//
+
+import XCTest
```
Jens-G, line 7: please replace the file header by std ASF license header

lib/swift/Sources/TBase64Utils.swift
```diff
@@ -0,0 +1,89 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import Foundation
+
+public class TBase64Utils {
+    private static let EncodeTable: String = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
```
Jens-G, line 23: Do we really need to implement this ourselves? Does Swift have no standard runtime library support for this that we can leverage?

Jens-G (changes requested):

lib/swift/Tests/ThriftTests/TJSONProtocolTests.swift
cozality: will do.

lib/swift/Sources/TBase64Utils.swift
cozality: I imported the code from C# library and the way it has been implemented is like above, so I wanted to be consistent with that.

see https://github.com/apache/thrift/blob/master/lib/netstd/Thrift/Protocol/Utilities/TBase64Utils.cs

PR #2816, opened 2023-06-12
THRIFT-5715: Python non-user defined fields mutable with slots

KTAtkinson:
In Python 3.11 exceptions generated by the compiler can't be used with a
context manager because they are immutable. As of Python 3.11
`contextlib.contextmanager` sets `exc.__traceback__` in the event that
the code in the context manager errors.

As of Thrift v0.18.1 exceptions are generated as immutable by default.
See [PR#1835](https://github.com/apache/thrift/pull/1835) for more information about why exceptions were made
immutable by default.

This change makes all non-Thrift fields mutable when slots is used  without dynamic. This
will allow exceptions to be re-raised properly by the contextmanager in
Python 3.11.

With tutorial code generated from `head` this raises a `TypeError` not 
an `InvalidOperation`:

```
import contextlib
import sys

sys.path.append('gen-py')
from tutorial.ttypes import InvalidOperation

@contextlib.contextmanager
def example():
    yield
    return

def main():
    with example():
        raise InvalidOperation

if __name__ == "__main__":
    main()
```

It raises this error:

```
root@b81016c354b4:/src/thrift/tutorial/py# python3 contextmanager.py 
  File "/src/thrift/tutorial/py/contextmanager.py", line 5, in <module>
    from tutorial.ttypes import InvalidOperation
ModuleNotFoundError: No module named 'tutorial.ttypes'
root@b81016c354b4:/src/thrift/tutorial/py# python3 contextmanager.py 
  File "/usr/local/lib/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/src/thrift/tutorial/py/contextmanager.py", line 9, in example
    yield
  File "/src/thrift/tutorial/py/contextmanager.py", line 14, in main
    raise InvalidOperation
tutorial.ttypes.InvalidOperation: InvalidOperation(whatOp=None, why=None)

During handling of the above exception, another exception occurred:

  File "/src/thrift/tutorial/py/contextmanager.py", line 17, in <module>
    main()
  File "/src/thrift/tutorial/py/contextmanager.py", line 13, in main
    with example():
  File "/usr/local/lib/python3.11/contextlib.py", line 188, in __exit__
    exc.__traceback__ = traceback
    ^^^^^^^^^^^^^^^^^
  File "/src/thrift/tutorial/py/gen-py/tutorial/ttypes.py", line 160, in __setattr__
```

The error after this change without slots, this error still happens:

```
root@daf9a41282a7:/src/thrift/tutorial/py# python3 ./contextmanager.py 
  File "/src/thrift/tutorial/py/./contextmanager.py", line 17, in <module>
    main()
  File "/src/thrift/tutorial/py/./contextmanager.py", line 14, in main
    raise InvalidOperation
tutorial.ttypes.InvalidOperation: InvalidOperation(whatOp=None, why=None)
root@daf9a41282a7:/src/thrift/tutorial/py# python3 ./contextmanager.py 
  File "/usr/local/lib/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/src/thrift/tutorial/py/./contextmanager.py", line 9, in example
    yield
  File "/src/thrift/tutorial/py/./contextmanager.py", line 14, in main
    raise InvalidOperation
tutorial.ttypes.InvalidOperation: InvalidOperation(whatOp=None, why=None)

During handling of the above exception, another exception occurred:

  File "/src/thrift/tutorial/py/./contextmanager.py", line 17, in <module>
    main()
  File "/src/thrift/tutorial/py/./contextmanager.py", line 13, in main
    with example():
  File "/usr/local/lib/python3.11/contextlib.py", line 188, in __exit__
    exc.__traceback__ = traceback
    ^^^^^^^^^^^^^^^^^
  File "/src/thrift/tutorial/py/gen-py/tutorial/ttypes.py", line 160, in __setattr__
```

After change with slots:

```
root@b81016c354b4:/src/thrift/tutorial/py# python3 contextmanager.py 
  File "/src/thrift/tutorial/py/contextmanager.py", line 17, in <module>
    main()
  File "/src/thrift/tutorial/py/contextmanager.py", line 14, in main
    raise InvalidOperation
tutorial.ttypes.InvalidOperation: InvalidOperation(whatOp=None, why=None)
```

There is no diff in the generated code when slots is not used.

Here is the diff between `thrift` at head and this branch generating with `slots`:

```
src# diff -C 5 -r tutorial/gen-py-pr/head-slots/ tutorial/gen-py-pr/dev-slots/
diff -C 5 -r tutorial/gen-py-pr/head-slots/tutorial/ttypes.py tutorial/gen-py-pr/dev-slots/tutorial/ttypes.py
*** tutorial/gen-py-pr/head-slots/tutorial/ttypes.py    Thu Jun 29 16:03:16 2023
--- tutorial/gen-py-pr/dev-slots/tutorial/ttypes.py     Thu Jun 29 16:19:11 2023
***************
*** 174,186 ****
--- 174,192 ----
      def __init__(self, whatOp=None, why=None,):
          super(InvalidOperation, self).__setattr__('whatOp', whatOp)
          super(InvalidOperation, self).__setattr__('why', why)
  
      def __setattr__(self, *args):
+         if args[0] not in self.__slots__:
+             super().__setattr__(*args)
+             return
  
      def __delattr__(self, *args):
+         if args[0] not in self.__slots__:
+             super().__delattr__(*args)
+             return
  
      def __hash__(self):
          return hash(self.__class__) ^ hash((self.whatOp, self.why, ))
```

  

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -893,12 +893,42 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
 
   if (is_immutable(tstruct)) {
     out << endl;
-    out << indent() << "def __setattr__(self, *args):" << endl
-        << indent() << indent_str() << "raise TypeError(\"can't modify immutable instance\")" << endl
-        << endl;
-    out << indent() << "def __delattr__(self, *args):" << endl
-        << indent() << indent_str() << "raise TypeError(\"can't modify immutable instance\")" << endl
-        << endl;
+    out << indent() << "def __setattr__(self, *args):" << endl;
+    indent_up();
+
+    // Not user-provided fields should be editable so that the Python Standard Library can edit
+    // internal fields of std library base classes. For example, in Python 3.11 ContextManager
+    // edits the `__traceback__` field on Exceptions. Allowing this to work with `__slots__` is
+    // trivial because we know which fields are user-provided, without slots we need to build a
+    // way to know which fields are user-provided.
+    if (gen_slots_ && !gen_dynamic_) {
+        out << indent() << "if args[0] not in self.__slots__:" << endl;
+        indent_up();
+        out << indent() << "super().__setattr__(*args)" << endl
+            << indent() << "return" << endl;
+        indent_down();
+    }
+    out << indent() << "raise TypeError(\"can't modify immutable instance\")" << endl;
+    indent_down();
+    out << endl;
+    out << indent() << "def __delattr__(self, *args):" << endl;
+    indent_up();
+
+    // Not user-provided fields should be editable so that the Python Standard Library can edit
+    // internal fields of std library base classes. For example, in Python 3.11 ContextManager
+    // edits the `__traceback__` field on Exceptions. Allowing this to work with `__slots__` is
+    // trivial because we know which fields are user-provided, without slots we need to build a
+    // way to know which fields are user-provided.
+    if (gen_slots_ && !gen_dynamic_) {
+        out << indent() << "if args[0] not in self.__slots__:" << endl;
+        indent_up();
+        out << indent() << "super().__delattr__(*args)" << endl
+            << indent() << "return" << endl;
+        indent_down();
+    }
+    out << indent() << "raise TypeError(\"can't modify immutable instance\")" << endl;
+    indent_down();
+    out << endl;
 
     // Hash all of the members in order, and also hash in the class
     // to avoid collisions for stuff like single-field structures.
--- a/test/py/TestClient.py
+++ b/test/py/TestClient.py
@@ -256,6 +256,35 @@ class AbstractTest(unittest.TestCase):
         y = self.client.testMultiException('success', 'foobar')
         self.assertEqual(y.string_thing, 'foobar')
 
+    def testException__traceback__(self):
+        print('testException__traceback__')
+        self.client.testException('Safe')
+        expect_slots = uses_slots = False
+        expect_dynamic = uses_dynamic = False
+        try:
+            self.client.testException('Xception')
+            self.fail("should have gotten exception")
+        except Xception as x:
+            uses_slots = hasattr(x, '__slots__')
+            uses_dynamic = (not isinstance(x, TException))
+            # We set expected values here so that we get clean tracebackes when
+            # the assertions fail.
+            try:
+                x.__traceback__ = x.__traceback__
+                # If `__traceback__` was set without errors than we expect that
+                # the slots option was used and that dynamic classes were not.
+                expect_slots = True
+                expect_dynamic = False
+            except Exception as e:
+                self.assertTrue(isinstance(e, TypeError))
+                # There are no other meaningful tests we can preform because we
+                # are unable to determine the desired state of either `__slots__`
+                # or `dynamic`.
+                return
+
+        self.assertEqual(expect_slots, uses_slots)
+        self.assertEqual(expect_dynamic, uses_dynamic)
+
     def testOneway(self):
         print('testOneway')
         start = time.time()


KTAtkinson:
appvayor is currently failing, but it's also failing on all other PRs. Is this expected?

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -105,7 +106,9 @@ class t_py_generator : public t_generator {
         }
         if( import_dynbase_.empty()) {
           import_dynbase_ = "from thrift.protocol.TBase import TBase, TFrozenBase, TExceptionBase, TFrozenExceptionBase, TTransport\n";
-        }
+        } 
+      } else if( iter->first.compare("immutable_exc") == 0) {
```
fishy, line 110: shouldn't this be called `mutable_exc` instead? otherwise you are making `gen_immutable_exc_` to false when user pass in `immutable_exc` arg to the compiler.

fishy:
> appvayor is currently failing, but it's also failing on all other PRs. Is this expected?

kind of.

but I think a better fix is to limit the scope of immutability of exceptions. we only need to limit the fields used by `__hash__` and `__eq__` to be immutable, and allow mutations of all other fields.

compiler/cpp/src/thrift/generate/t_py_generator.cc
KTAtkinson: Yea this naming is a bit confusing, I'll update it :)

KTAtkinson:
I updated this so that the generated python only considers fields in the `__dict__` or `__slots__` field when Thrift types are immutable.

fishy (commented):
@KTAtkinson can you please also paste what the new generated python exception code look like (with and without slots)?

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -888,12 +888,33 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
...
-        << endl;
-    out << indent() << "def __delattr__(self, *args):" << endl
-        << indent() << indent_str() << "raise TypeError(\"can't modify immutable instance\")" << endl
-        << endl;
+    out << indent() << "def __setattr__(self, attr, value):" << endl;
+    indent_up();
+    if (gen_slots_) {
+      out << indent() << "if attr not in self.__slots__:" << endl;
+    } else {
+      out << indent() << "if not self.__dict__.has_key(attr):" << endl;
```
fishy, line 896: this might not work as expected:
```
$ python3
Python 3.11.4 (main, Jun  7 2023, 10:13:09) [GCC 12.2.0] on linux
>>> class Foo:
...   _bar = None
... 
>>> foo = Foo()
>>> foo.__dict__
{}
>>> foo._bar = 1
>>> foo.__dict__
{'_bar': 1}
```
it seems that the `__dict__` would only have the key if it's used at least once?

test/py/TestClient.py
```diff
@@ -256,6 +257,19 @@ def testMultiException(self):
         y = self.client.testMultiException('success', 'foobar')
         self.assertEqual(y.string_thing, 'foobar')
 
+    def testExceptionInContextManager(self):
```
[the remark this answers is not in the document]
KTAtkinson: I think there should be some sort of test here. I originally wrote the test testing the functionality directly, assigning and deleting particular attributes. I wasn't able to write a test to ensure that user defined attributes are not editable as in some cases they are editable. 

My thinking with this test case is that it will at least catch the specific failure when Thrift is updated to use Python 3.11. I did try to run this test with 3.11 but there are parts of the test script that aren't compatible with 3.11.

Thoughts on how to effectively write at test here?

compiler/cpp/src/thrift/generate/t_py_generator.cc
KTAtkinson: Currently, it looks like [each use defined field](https://github.com/apache/thrift/pull/2816/files#diff-ad110290fb2055999f700940287b9f8db58007832084247e6ffb0bd2b14b897dR850-R887) is explicitly set in the `__init__`. I believe this would avoid this issue. Is that not correct?

Alternately I could keep track of user defined fields similar to how we do for `__slots__`, thoughts?

fishy: my main concern here is that after contextmanager changes the `__traceback__` it will start to appear in `__dict__` and then it can no longer be changed again

KTAtkinson: Long story short, `__traceback__` does not appear in `__dict__` from what I could track down it's because the implementation of `Exception` is in C and `__traceback__` is not technically an official property.

Fishy and I discussed off PR and decided that it was acceptable to only support context managers (in Python 3.11) with when using slots. This means any user that runs into this issue should use slots to get around their issue.

KTAtkinson:
@fishy Have time to take a look at this again?

test/py/TestClient.py
```diff
@@ -256,6 +256,26 @@ def testMultiException(self):
...
+            uses_slots = hasattr(x, '__slots__')
+            try:
+                x.__traceback__ = x.__traceback__
+                self.assertTrue(
+                    uses_slots,
+                    'expected editable exception to use slots, no slots defined',
+                )
+            except Exception as e:
+                self.assertTrue(
+                    isinstance(e, TypeError),
```
fishy, line 275: this should be `isinstance(e, TypeError) and not uses_slots` instead?

fishy:
>it was acceptable to only support context managers (in Python 3.11) with when using slots.

also to clarify, I said it's acceptable to limit the scope of this PR to only fix the slots use-case, but that does not fully fix the issue (so we cannot close the issue with this PR merged) and we should still fix the non-slots use-case later.

test/py/TestClient.py
fishy: or probably better split into 2 separated asserts.

fishy:
please also rebase your change on top of latest master. there are 2 commits related to py compiler merged recently and I want to make sure that this does not break those changes.

fishy:
Please also change the title of this PR accordingly. You are no longer adding a flag to allow mutable exceptions.

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -893,12 +893,42 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
...
-    out << indent() << "def __setattr__(self, *args):" << endl
-        << indent() << indent_str() << "raise TypeError(\"can't modify immutable instance\")" << endl
-        << endl;
-    out << indent() << "def __delattr__(self, *args):" << endl
-        << indent() << indent_str() << "raise TypeError(\"can't modify immutable instance\")" << endl
-        << endl;
+    out << indent() << "def __setattr__(self, *args):" << endl;
+    indent_up();
+
+    // Not user-provided fields should be editable so that the Python Standard L`ibrary can edit
```
fishy, line 899:
```suggestion
    // Not user-provided fields should be editable so that the Python Standard Library can edit
```

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -893,12 +893,42 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
...
+    // internal fields of std library base classes. For example, in Python 3.11 context managers
+    // edit the `__traceback__` field on Exceptions. Allowing this to work with `__slots__` is
```
fishy, line 901:
```suggestion
    // internal fields of std library base classes. For example, in Python 3.11 ContextManager
    // edits the `__traceback__` field on Exceptions. Allowing this to work with `__slots__` is
```

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -893,12 +893,42 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
...
+    }
+    out << indent() << "raise TypeError(\"can't modify immutable instance\")" << endl;
+    indent_down();
+    out << endl;
+    out << indent() << "def __delattr__(self, *args):" << endl;
+    indent_up();
+
+    // Not user-provided fields should be editable so that the Python Standard L`ibrary can edit
+    // internal fields of std library base classes. For example, in Python 3.11 context managers
+    // edit the `__traceback__` field on Exceptions. Allowing this to work with `__slots__` is
```
fishy, line 919: same here

fishy (approved):

PR #2818, opened 2023-06-16
THRIFT-5716: lib: cpp: Fix crash in TMemoryBuffer due to uint32_t overflow

stiga-huang:
This fixes the overflow in `TMemoryBuffer::ensureCanWrite()` when the current usage plus the length of new content exceeds the max of uint32_t (e.g. 4GB). The buggy code is
```cpp
  const uint32_t current_used = bufferSize_ - avail;
  const uint32_t required_buffer_size = len + current_used;
  if (required_buffer_size > maxBufferSize_) {
    throw TTransportException(...)
  }
```
`len + current_used` could overflow in `uint32_t`, which results in not rejecting the write, i.e. not throwing TTransportException. The follow-up write on this TMemoryBuffer will write to invalid memory address and leads to a crash (e.g. [IMPALA-12223](https://issues.apache.org/jira/browse/IMPALA-12223))

We should define `current_used` as `uint64_t` so the result of the expression is also `uint64_t`. Changed the type of `required_buffer_size` to `uint64_t` as well.

Added a unit test for this.

### DIFF
--- a/lib/cpp/src/thrift/transport/TBufferTransports.cpp
+++ b/lib/cpp/src/thrift/transport/TBufferTransports.cpp
@@ -363,9 +363,9 @@ void TMemoryBuffer::ensureCanWrite(uint32_t len) {
     throw TTransportException("Insufficient space in external MemoryBuffer");
   }
 
-  // Grow the buffer as necessary.
-  const uint32_t current_used = bufferSize_ - avail;
-  const uint32_t required_buffer_size = len + current_used;
+  // Grow the buffer as necessary. Use uint64_t to avoid overflow.
+  const uint64_t current_used = bufferSize_ - avail;
+  const uint64_t required_buffer_size = len + current_used;
   if (required_buffer_size > maxBufferSize_) {
     throw TTransportException(TTransportException::BAD_ARGS,
                               "Internal buffer size overflow when requesting a buffer of size " + std::to_string(required_buffer_size));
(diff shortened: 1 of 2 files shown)


emmenlau:
Looks reasonable. But a naiive question: is thrift currently supporting any data size above 4GB?

fishy:
> Looks reasonable. But a naiive question: is thrift currently supporting any data size above 4GB?

From what I understand, on protocol level, all of [TBinaryProtocol](https://github.com/apache/thrift/blob/master/doc/specs/thrift-binary-protocol.md), [TCompactProtocol](https://github.com/apache/thrift/blob/master/doc/specs/thrift-compact-protocol.md), and [THeaderProtocol](https://github.com/apache/thrift/blob/master/doc/specs/HeaderFormat.md) only support 32-bit lengths.

TFramedTransport does not have a spec (as far as I could tell) but the implementations should also all be limiting to 32-bit.

fishy:
Both TBinary and TCompact specs also explicitly stated that it's _signed_ 32-bit int. THeader spec doesn't explicitly call out signedness, but in general we are using signed ints across thrift because java's lack of support of unsigned ints.

stiga-huang:
Thanks for the quick review!

IIUC, `TMemoryBuffer` can be used to serialize any thrift object and the thrift object can exceed 4GB in the total size. E.g. in Impala, we have such a thrift object that contains `map<string, string>` and `list<string>`. The total length could be arbitrarily long:
https://github.com/apache/impala/blob/06eb62d3efa1c94810c4276f90896fa62205a49b/common/thrift/RuntimeProfile.thrift#L239-L242

We need to correctly reject such serialization, i.e. throw `TTransportException` about "Internal buffer size overflow". Currently, due to the overflow of `uint32_t`, `TMemoryBuffer::ensureCanWrite()` thought the write is legal and proceed without actually expanding the buffer (since `new_size` is smaller than the current size). It then writes to invalid memory and leads to a crash.

PR #2820, opened 2023-06-21
Dependabot alerts

Jens-G:
Trying to address a whole bunch of these https://github.com/apache/thrift/security/dependabot


contrib/thrift-maven-plugin/pom.xml
```diff
@@ -94,7 +94,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>32.0.0-jre</version>
+        <version>32.0.1</version>
```
ctubbsii, line 97: Not sure what's up with the change in spacing, but this needs `-jre` in the version name to resolve.

```suggestion
      <version>32.0.1-jre</version>
```

ctubbsii (changes requested):

Jens-G:
Also nice ...

> Warning: The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672

PR #2824, opened 2023-06-28
Use Enum or string in Struct serialization/deserialization for py:enum gen code

roshanjrajan-zip:
When serializing and deserializing Enums for Structs, we convert the Enum to a string instead of keeping it as an Enum. What this means is that the thrift file shows that an Enum is passed in the struct, when actually a string is passed in. This is very confusing can arbitrarily lead to KeyError and in the deserialization logic.

However, changing this behavior is potentially breaking. To avoid breaking existing users, check if the type is an Enum by using the `hasattr` check to see if type has attribute value. If so, parse as an enum. Otherwise parse as a string.

Given a thrift file like the following:
```python
enum TestEnum {
  TestEnum0 = 0,
  TestEnum1 = 1,
}

struct TestStruct {
    1: optional string param1
    2: optional TestEnum param2
}
```

this is how you would create a struct before and after the change but both are valid.
```python
# Before
TestStruct(param1="test_string", param2=TestEnum.TestEnum1.name) # or pass in "TestEnum1"

# After
TestStruct(param1="test_string", param2=TestEnum.TestEnum1)
```

Serialization/Deserialization Code Change Example- 
```python
class TestStruct(object):
    """
    Attributes:
     - param1
     - param2

    """


    def __init__(self, param1=None, param2=None,):
        self.param1 = param1
        self.param2 = param2

    def read(self, iprot):
        if iprot._fast_decode is not None and isinstance(iprot.trans, TTransport.CReadableTransport) and self.thrift_spec is not None:
            iprot._fast_decode(self, iprot, [self.__class__, self.thrift_spec])
            return
        iprot.readStructBegin()
        while True:
            (fname, ftype, fid) = iprot.readFieldBegin()
            if ftype == TType.STOP:
                break
            if fid == 1:
                if ftype == TType.STRING:
                    self.param1 = iprot.readString().decode('utf-8', errors='replace') if sys.version_info[0] == 2 else iprot.readString()
                else:
                    iprot.skip(ftype)
            elif fid == 2:
                if ftype == TType.I32:
--                    self.param2 = TestEnum(iprot.readI32()).name
++                    self.param2 = TestEnum(iprot.readI32())
                else:
                    iprot.skip(ftype)
            else:
                iprot.skip(ftype)
            iprot.readFieldEnd()
        iprot.readStructEnd()

    def write(self, oprot):
        if oprot._fast_encode is not None and self.thrift_spec is not None:
            return
        oprot.writeStructBegin('TestStruct')
        if self.param1 is not None:
            oprot.writeFieldBegin('param1', TType.STRING, 1)
            oprot.writeString(self.param1.encode('utf-8') if sys.version_info[0] == 2 else self.param1)
            oprot.writeFieldEnd()
        if self.param2 is not None:
            oprot.writeFieldBegin('param2', TType.I32, 2)
--            oprot.writeI32(TestEnum[self.param2].value)
++           oprot.writeI32(self.param2.value if hasattr(self.param2, 'value') else TestEnum[self.param2].value)
            oprot.writeFieldEnd()
        oprot.writeFieldStop()
        oprot.writeStructEnd()

    def validate(self):
        return

    def __repr__(self):
        L = ['%s=%r' % (key, value)
             for key, value in self.__dict__.items()]
        return '%s(%s)' % (self.__class__.__name__, ', '.join(L))

    def __eq__(self, other):
        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__

    def __ne__(self, other):
        return not (self == other)
all_structs.append(TestStruct)
TestStruct.thrift_spec = (
    None,  # 0
    (1, TType.STRING, 'param1', 'UTF8', None, ),  # 1
    (2, TType.I32, 'param2', None, None, ),  # 2
)
fix_spec(all_structs)
del all_structs
```

  

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -2287,7 +2287,7 @@ void t_py_generator::generate_deserialize_field(ostream& out,
     out << endl;
   } else if (type->is_enum()) {
     if (gen_enum_) {
-      indent(out) << name << " = " << type_name(type) << "(iprot.readI32()).name";
+      indent(out) << name << " = " << type_name(type) << "(iprot.readI32())";
     } else {
       indent(out) << name << " = iprot.readI32()";
     }
@@ -2477,7 +2477,8 @@ void t_py_generator::generate_serialize_field(ostream& out, t_field* tfield, str
       }
     } else if (type->is_enum()) {
       if (gen_enum_){
-        out << "writeI32(" << type_name(type) << "[" << name << "].value)";
+        out << "writeI32(" << name << ".value if hasattr(" << name << ", 'value') else "
+            << type_name(type) << "[" << name << "].value)";
       } else {
         out << "writeI32(" << name << ")";
       }
--- a/test/py/explicit_module/EnumSerializationTest.py
+++ b/test/py/explicit_module/EnumSerializationTest.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+from test5.ttypes import TestEnum, TestStruct
+from thrift.TSerialization import serialize, deserialize
+
+def serialization_deserialization_enum_test():
+    test_obj = TestStruct(param1="test_string", param2=TestEnum.TestEnum1)
+    test_obj_serialized = serialize(test_obj)
+    test_obj2 = deserialize(TestStruct(), test_obj_serialized)
+    assert test_obj.param1 == test_obj2.param1
+    assert test_obj.param2 == test_obj2.param2
+
+def serialization_deserialization_string_test():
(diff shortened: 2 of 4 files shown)


fishy:
does this only affect how the python lib is called or does this actually affect how the serialized binary blob (aka wire protocol) would look like?

if this affects how the binary blob would look like, then this most certainly would break cross-test. we currently don't run cross-tests in github actions because the lib-java-kotlin test is broken, you can temporarily remove that as the dependency of cross-test in your PR to make it run cross test.

fishy:
I see. So this is only about the "constructor", but the actual constructor doesn't do anything special and just stores the args, and all the special logic happens in other functions using the stored args.

I think it would make more sense to actually move the logic to constructor instead, e.g. convert either string or actual enum to the actual enum, so we don't have to do the same check everywhere else.

roshanjrajan-zip:
> I see. So this is only about the "constructor", but the actual constructor doesn't do anything special and just stores the args, and all the special logic happens in other functions using the stored args.
> 
> I think it would make more sense to actually move the logic to constructor instead, e.g. convert either string or actual enum to the actual enum, so we don't have to do the same check everywhere else.

The issue is that people can construct the struct and then serialize/deserialize the struct later. I'm fine with your suggestion but it could potentially be breaking behavior.
```python
struct = TestStruct(param1="test_string")
struct.param2="TestEnum1"
# serialize/deserialize logic ...
```

fishy:
right but this is python so it's very hard to force anything, I guess? we don't even have type annotation (I believe there's an open ticket for that just no one picking it up yet).

so even if we move the logic to constructor the risk of someone doing other thing to override it is. I guess the current approach makes sense in the given circumstances.

fishy (approved):
looks reasonable to me but again, give others a few days to chime in before I merge it.

roshanjrajan-zip:
> right but this is python so it's very hard to force anything, I guess? we don't even have type annotation (I believe there's an open ticket for that just no one picking it up yet).
> 
> so even if we move the logic to constructor the risk of someone doing other thing to override it is. I guess the current approach makes sense in the given circumstances.

We can force this to be an enum by refusing to parse the value if it's a string but that would be a breaking change. I think its the right thing to do and I am okay with making this change as I'm pretty certain very few people use this functionality because it doesn't really make sense when interchanging strings/enums.

roshanjrajan-zip:
@fishy I gave it another try and I think this code is a bit more complex but is safer change to take in. It will essentially force the type to become an Enum and simplifies the hasattr check.
https://github.com/apache/thrift/pull/2825

PR #2825, opened 2023-06-30
Convert string types to Enums in Struct py:enum gen code

roshanjrajan-zip:
When serializing and deserializing Enums for Structs, the current code expects Enum member variables to be a string instead of an Enum. What this means is that the thrift file shows that an Enum is passed in the struct, when actually a string is passed in. This is very confusing can arbitrarily lead to KeyError in the deserialization logic.

However, changing this behavior is potentially breaking. To avoid breaking existing users, check if the type is an Enum when setting the member variable and convert the type to an Enum if it is a string. This allows the actual type to be an Enum and simplifies the serialization/deserialization logic to use Enums instead of converting strings.

Given a thrift file like the following:
```python
enum TestEnum {
  TestEnum0 = 0,
  TestEnum1 = 1,
}

struct TestStruct {
    1: optional string param1
    2: optional TestEnum param2
}
```

this is how you would create a struct before and after the change but both are valid.
```python
# Before
TestStruct(param1="test_string", param2=TestEnum.TestEnum1.name) # or pass in "TestEnum1"

# After
TestStruct(param1="test_string", param2=TestEnum.TestEnum1)
```

Serialization/Deserialization Code Change Example- 
```python
class TestStruct(object):
    """
    Attributes:
     - param1
     - param2

    """


    def __init__(self, param1=None, param2=None,):
        self.param1 = param1
        self.param2 = param2

++    def __setattr__(self, name, value):
++		if name == "param2":
++	        super().__setattr__(name, value if hasattr(value, 'value') else TestEnum.__members__.get(value))
++			return
++        super().__setattr__(name, value)

    def read(self, iprot):
        if iprot._fast_decode is not None and isinstance(iprot.trans, TTransport.CReadableTransport) and self.thrift_spec is not None:
            iprot._fast_decode(self, iprot, [self.__class__, self.thrift_spec])
            return
        iprot.readStructBegin()
        while True:
            (fname, ftype, fid) = iprot.readFieldBegin()
            if ftype == TType.STOP:
                break
            if fid == 1:
                if ftype == TType.STRING:
                    self.param1 = iprot.readString().decode('utf-8', errors='replace') if sys.version_info[0] == 2 else iprot.readString()
                else:
                    iprot.skip(ftype)
            elif fid == 2:
                if ftype == TType.I32:
--                    self.param2 = TestEnum(iprot.readI32()).name
++                    self.param2 = TestEnum(iprot.readI32())
                else:
                    iprot.skip(ftype)
            else:
                iprot.skip(ftype)
            iprot.readFieldEnd()
        iprot.readStructEnd()

    def write(self, oprot):
        if oprot._fast_encode is not None and self.thrift_spec is not None:
            return
        oprot.writeStructBegin('TestStruct')
        if self.param1 is not None:
            oprot.writeFieldBegin('param1', TType.STRING, 1)
            oprot.writeString(self.param1.encode('utf-8') if sys.version_info[0] == 2 else self.param1)
            oprot.writeFieldEnd()
        if self.param2 is not None:
            oprot.writeFieldBegin('param2', TType.I32, 2)
--            oprot.writeI32(TestEnum[self.param2].value)
++           oprot.writeI32(self.param2.value)
            oprot.writeFieldEnd()
        oprot.writeFieldStop()
        oprot.writeStructEnd()

    def validate(self):
        return

    def __repr__(self):
        L = ['%s=%r' % (key, value)
             for key, value in self.__dict__.items()]
        return '%s(%s)' % (self.__class__.__name__, ', '.join(L))

    def __eq__(self, other):
        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__

    def __ne__(self, other):
        return not (self == other)
all_structs.append(TestStruct)
TestStruct.thrift_spec = (
    None,  # 0
    (1, TType.STRING, 'param1', 'UTF8', None, ),  # 1
    (2, TType.I32, 'param2', None, None, ),  # 2
)
fix_spec(all_structs)
del all_structs
```

  

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -871,7 +871,12 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
       }
 
       if (is_immutable(tstruct)) {
-        if (gen_newstyle_ || gen_dynamic_) {
+        if (gen_enum_ && type->is_enum()) {
+          indent(out) << "super(" << tstruct->get_name() << ", self).__setattr__('"
+                      << (*m_iter)->get_name() << "', " << (*m_iter)->get_name()
+                      << " if hasattr("  << (*m_iter)->get_name() << ", 'value') else "
+                      << type_name(type) << ".__members__.get(" << (*m_iter)->get_name() << "))" << endl;
+        } else if (gen_newstyle_ || gen_dynamic_) {
           indent(out) << "super(" << tstruct->get_name() << ", self).__setattr__('"
                       << (*m_iter)->get_name() << "', " << (*m_iter)->get_name() << ")" << endl;
         } else {
@@ -905,6 +910,32 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
     }
 
     out << "))" << endl;
+  } else if (gen_enum_) {
+    bool has_enum = false;
+    for (m_iter = members.begin(); m_iter != members.end(); ++m_iter) {
+      t_type* type = (*m_iter)->get_type();
+      if (type->is_enum()) {
+        has_enum = true;
+        break;
+      }
+    }
+
+    if (has_enum) {
+      out << endl;
+      indent(out) << "def __setattr__(self, name, value):" << endl;
+      indent_up();
+      for (m_iter = members.begin(); m_iter != members.end(); ++m_iter) {
+        t_type* type = (*m_iter)->get_type();
+        if (type->is_enum()) {
+          out << indent() << "if name == \"" << (*m_iter)->get_name() << "\":" << endl
+              << indent() << indent_str() << "super().__setattr__(name, value if hasattr(value, 'value') else "
+              << type_name(type) << ".__members__.get(value))" << endl
+              << indent() << indent_str() << "return" << endl;
+        }
+      }
+      indent(out) << "super().__setattr__(name, value)" << endl << endl;
+      indent_down();
+    }
   }
 
   if (!gen_dynamic_) {
@@ -2287,7 +2318,7 @@ void t_py_generator::generate_deserialize_field(ostream& out,
     out << endl;
   } else if (type->is_enum()) {
     if (gen_enum_) {
-      indent(out) << name << " = " << type_name(type) << "(iprot.readI32()).name";
+      indent(out) << name << " = " << type_name(type) << "(iprot.readI32())";
     } else {
       indent(out) << name << " = iprot.readI32()";
     }
@@ -2477,7 +2508,7 @@ void t_py_generator::generate_serialize_field(ostream& out, t_field* tfield, str
       }
     } else if (type->is_enum()) {
       if (gen_enum_){
-        out << "writeI32(" << type_name(type) << "[" << name << "].value)";
+        out << "writeI32(" << name << ".value)";
       } else {
         out << "writeI32(" << name << ")";
       }
--- a/test/py/explicit_module/runtest.sh
+++ b/test/py/explicit_module/runtest.sh
@@ -25,10 +25,15 @@ rm -rf gen-py
 ../../../compiler/cpp/thrift --gen py test3.thrift && exit 1  # Fail since test3.thrift has python keywords
 ../../../compiler/cpp/thrift --gen py:enum shared_types.thrift || exit 1
 ../../../compiler/cpp/thrift --gen py:enum test4.thrift || exit 1
+../../../compiler/cpp/thrift --gen py:enum test5.thrift || exit 1
+mkdir -p ./gen-py/test5_slots
+../../../compiler/cpp/thrift --gen py:enum,slots -out ./gen-py/test5_slots test5.thrift || exit 1
 PYTHONPATH=./gen-py python -c 'import foo.bar.baz' || exit 1
 PYTHONPATH=./gen-py python -c 'import test2' || exit 1
 PYTHONPATH=./gen-py python -c 'import test1' &>/dev/null && exit 1  # Should fail.
 PYTHONPATH=./gen-py python -c 'import test4.constants' || exit 1
+PYTHONPATH=./gen-py python EnumSerializationTest.py || exit 1
+PYTHONPATH=./gen-py python EnumSerializationTest.py slot|| exit 1
 cp -r gen-py simple
 ../../../compiler/cpp/thrift -r --gen py test2.thrift || exit 1
 PYTHONPATH=./gen-py python -c 'import test2' || exit 1
--- a/test/py/explicit_module/EnumSerializationTest.py
+++ b/test/py/explicit_module/EnumSerializationTest.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+from __future__ import annotations
+
+import sys
+from shared_types.ttypes import SharedEnum
+from thrift.TSerialization import serialize, deserialize
+from thrift.protocol import TBinaryProtocol
+from thrift.transport import TTransport
+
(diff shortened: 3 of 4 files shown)


compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -905,6 +905,32 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
...
+        has_enum = true;
+        break;
+      }
+    }
+
+    if (has_enum) {
+      out << endl;
+      indent(out) << "def __setattr__(self, name, value):" << endl;
+      indent_up();
+      for (m_iter = members.begin(); m_iter != members.end(); ++m_iter) {
```
fishy, line 922: can you add a test (and also paste the generated python code into PR discussion) on when a struct has 2 fields that's enum (and they are different enum types)?

roshanjrajan-zip: Updated the test! Here is the entire generated code file
```python
#
# Autogenerated by Thrift Compiler (0.19.0)
#
# DO NOT EDIT UNLESS YOU ARE SURE THAT YOU KNOW WHAT YOU ARE DOING
#
#  options string: py:enum
#

from thrift.Thrift import TType, TMessageType, TFrozenDict, TException, TApplicationException
from thrift.protocol.TProtocol import TProtocolException
from thrift.TRecursive import fix_spec
from enum import IntEnum

import sys
import shared_types.ttypes

from thrift.transport import TTransport
all_structs = []


class TestEnum(IntEnum):
    TestEnum0 = 0
    TestEnum1 = 1



class TestStruct(object):
    """
    Attributes:
     - param1
     - param2
     - param3

    """


    def __init__(self, param1=None, param2=None, param3=None,):
        self.param1 = param1
        self.param2 = param2
        self.param3 = param3

    def __setattr__(self, name, value):
        if name == "param2":
            super().__setattr__(name, value if hasattr(value, 'value') else TestEnum.__members__.get(value))
            return
        if name == "param3":
            super().__setattr__(name, value if hasattr(value, 'value') else shared_types.ttypes.SharedEnum.__members__.get(value))
            return
        super().__setattr__(name, value)


    def read(self, iprot):
        if iprot._fast_decode is not None and isinstance(iprot.trans, TTransport.CReadableTransport) and self.thrift_spec is not None:
            iprot._fast_decode(self, iprot, [self.__class__, self.thrift_spec])
            return
        iprot.readStructBegin()
        while True:
            (fname, ftype, fid) = iprot.readFieldBegin()
            if ftype == TType.STOP:
                break
            if fid == 1:
                if ftype == TType.STRING:
                    self.param1 = iprot.readString().decode('utf-8', errors='replace') if sys.version_info[0] == 2 else iprot.readString()
                else:
                    iprot.skip(ftype)
            elif fid == 2:
                if ftype == TType.I32:
                    self.param2 = TestEnum(iprot.readI32())
                else:
                    iprot.skip(ftype)
            elif fid == 3:
                if ftype == TType.I32:
                    self.param3 = shared_types.ttypes.SharedEnum(iprot.readI32())
                else:
                    iprot.skip(ftype)
            else:
                iprot.skip(ftype)
            iprot.readFieldEnd()
        iprot.readStructEnd()

    def write(self, oprot):
        if oprot._fast_encode is not None and self.thrift_spec is not None:
            return
        oprot.writeStructBegin('TestStruct')
        if self.param1 is not None:
            oprot.writeFieldBegin('param1', TType.STRING, 1)
            oprot.writeString(self.param1.encode('utf-8') if sys.version_info[0] == 2 else self.param1)
            oprot.writeFieldEnd()
        if self.param2 is not None:
            oprot.writeFieldBegin('param2', TType.I32, 2)
            oprot.writeI32(self.param2.value)
            oprot.writeFieldEnd()
        if self.param3 is not None:
            oprot.writeFieldBegin('param3', TType.I32, 3)
            oprot.writeI32(self.param3.value)
            oprot.writeFieldEnd()
        oprot.writeFieldStop()
        oprot.writeStructEnd()

    def validate(self):
        return

    def __repr__(self):
        L = ['%s=%r' % (key, value)
             for key, value in self.__dict__.items()]
        return '%s(%s)' % (self.__class__.__name__, ', '.join(L))

    def __eq__(self, other):
        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__

    def __ne__(self, other):
        return not (self == other)
all_structs.append(TestStruct)
TestStruct.thrift_spec = (
    None,  # 0
    (1, TType.STRING, 'param1', 'UTF8', None, ),  # 1
    (2, TType.I32, 'param2', None, None, ),  # 2
    (3, TType.I32, 'param3', None, None, ),  # 3
)
fix_spec(all_structs)
del all_structs
```

fishy:
actually I just remembered another corner case, we also override `__setattr__` on all exceptions, so we need to make sure that an exception with an enum field still works. please also add a test for that.

(you can check https://issues.apache.org/jira/browse/THRIFT-5715 & https://issues.apache.org/jira/browse/THRIFT-4002 for more context)

roshanjrajan-zip:
@fishy Updated the PR with handling the Immutable struct logic and added to the test case.

compiler/cpp/src/thrift/generate/t_py_generator.cc
```diff
@@ -871,7 +871,12 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
       }
 
       if (is_immutable(tstruct)) {
-        if (gen_newstyle_ || gen_dynamic_) {
+        if (gen_enum_ && type->is_enum()) {
+          indent(out) << "super(" << tstruct->get_name() << ", self).__setattr__('"
```
fishy, line 875: this is probably too much magic to my preference, makes me start to prefer the approach in #2824 again 😅

test/py/explicit_module/runtest.sh
```diff
@@ -25,10 +25,12 @@ rm -rf gen-py
 ../../../compiler/cpp/thrift --gen py test3.thrift && exit 1  # Fail since test3.thrift has python keywords
 ../../../compiler/cpp/thrift --gen py:enum shared_types.thrift || exit 1
 ../../../compiler/cpp/thrift --gen py:enum test4.thrift || exit 1
+../../../compiler/cpp/thrift --gen py:enum test5.thrift || exit 1
```
fishy, line 28: can you please also generate the code with both enum and slots (as slots also change the code significantly), you can use `--out` to put it under a different directory than `gen-py` so you can have another test around line 33 to also make sure that the combination of enum+slots still works.

compiler/cpp/src/thrift/generate/t_py_generator.cc
roshanjrajan-zip: Yeah I was thinking about this and then I realized that the fix was actually broken because I wasn't running one of the tests that I added.... 😓 . So actually this fix is needed to make sure existing users aren't broken.

```python
if __name__ == "__main__":
    serialization_deserialization_enum_test()
    serialization_deserialization_string_test
```

test/py/explicit_module/runtest.sh
roshanjrajan-zip: I am not sure what slots are and how I can generate code that has enum and slots. Is there an example how I can generate code with enum and slots? I tried variations of `py:enum py:slots` and such but no such luck.

roshanjrajan-zip: NVM I figured it out and it seems to work. Is there a specific test case you have in mind? I just changed this - `../../../compiler/cpp/thrift --gen py:enum,slots test5.thrift` and the new serialization/deserialization test  that I added works.

compiler/cpp/src/thrift/generate/t_py_generator.cc
fishy: so this code:
```py

    def __init__(self, whatOp=None, why=None, who=None,):
        super(TestException, self).__setattr__('whatOp', whatOp)
        super(TestException, self).__setattr__('why', why if hasattr(why, 'value') else shared_types.ttypes.SharedEnum.__members__.get(why))
        super(TestException, self).__setattr__('who', who if hasattr(who, 'value') else TestEnum.__members__.get(who))
```
calls the parent's `__setattr__`, doesn't that essentially set those fields on the parent instead? I guess it still "works" as when accessing them you just inherit from the parent, but this _feels_ wrong to me.

test/py/explicit_module/runtest.sh
fishy: yea can you add that into the PR so we make sure neither `enum` and `enum,slots` will break in the future?

compiler/cpp/src/thrift/generate/t_py_generator.cc
roshanjrajan-zip: The code that generates `super(TestException, self).__setattr__('whatOp', whatOp)` already exists before my change and I am extending this pattern. In python, [super](https://docs.python.org/3/library/functions.html#super) is implemented as part of the binding process for explicit dotted attribute lookups such as __setattr__ and such. So I don't think this is a concern because this is how Python does things under the hood.

test/py/explicit_module/runtest.sh
roshanjrajan-zip: Added the slots option to the existing test5 test!

fishy (approved):
it's holiday/long weekend here, I'll probably merge this later next week if no one objects.

PR #2830, opened 2023-07-06
THRIFT-5723 Php8.1 fix warnings

sveneld:
In PHP 8.1 when creating new Exception with null message you will receive warning
E_DEPRECATED: Exception::__construct(): Passing null to parameter #1 ($message) of type string is deprecated


sveneld:
@Jens-G is it possible to add this minor fix in release which you announced few days ago

PR #2831, opened 2023-07-07
THRIFT-5725: Fix the comparison check for bio_get_fd

psarkar24:
The routine _bio_get_fd_ checks for the validity of a file descriptor by comparing the fd to be less
than or equal to zero. The issue is that fd of 0 is a valid file descriptor. This PR changes the check
to be less than zero.  


Jens-G:
+1 good catch

psarkar24:
@Jens-G If the patch looks good to you, can you please merge the changes? Thanks!

PR #2832, opened 2023-07-14
Setting TCP_ NODELAY option for rust server

tsxiaofang:
Resolve TFramedWriteTransport write message size 4 bytes causing approximately 40 milliseconds delay.


allengeorge (commented):
LGTM

PR #2842, opened 2023-08-10
Apply trivial formatting improvements to kotlin

ctubbsii:
This applies trivial formatting requirements for ktfmt 0.13.0 from abandoned PR #2837, but does not bump ktfmt to 0.13.0, due to bugs in that version. These formatting improvements should be suitable for the current 0.12.0 version of ktfmt, but will avoid needing to re-apply these changes when ktfmt is eventually updated.


ctubbsii:
This isn't worth it. I thought maybe code formatted this way would be okay with 0.12.0, but no, 0.12.0 and 0.13.0 formattings are mutually exclusive. The code will just have to be reformatted if/when the ktfmt plugin gets updated, which right now it can't be because of the bug. So, the current code is fine, and I'll close this PR.

PR #2843, opened 2023-08-10
THRIFT-5653: Fix Java UUID typeid

fishy:
It should be 16 not 17 according to the spec.

Currently ENUM holds 16, it's not in TBinaryProtocol spec and seems to be a Java implementation detail somehow got mixed inside TType, move that to ~255~ -1 (java's byte is signed) for now. Someone more familiar with Java can probably remove it from TType completely in the future.

client: java

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/protocol/TType.java
+++ b/lib/java/src/main/java/org/apache/thrift/protocol/TType.java
@@ -34,6 +34,8 @@ public final class TType {
   public static final byte MAP = 13;
   public static final byte SET = 14;
   public static final byte LIST = 15;
-  public static final byte ENUM = 16;
-  public static final byte UUID = 17;
+  public static final byte UUID = 16;
+
+  /** This is not part of the TBinaryProtocol spec but Java specific implementation detail */
+  public static final byte ENUM = -1;
 }


fishy:
From the cross test log it seems that this change fixed the cross tests regarding uuid, and didn't add new failures. But I'm still not sure how `TType.ENUM` is actually used in Java code and whether -1 is going to be ok. cc @Jimexist

Jimexist:
> From the cross test log it seems that this change fixed the cross tests regarding uuid, and didn't add new failures. But I'm still not sure how `TType.ENUM` is actually used in Java code and whether -1 is going to be ok. cc @Jimexist

not sure about the `TType.ENUM` myself either. regarding cross tests, since both java and kotlin uses the same convention so either ordinal will work since they are consistent.

lib/java/src/main/java/org/apache/thrift/protocol/TType.java
```diff
@@ -34,6 +34,9 @@ public final class TType {
   public static final byte MAP = 13;
   public static final byte SET = 14;
   public static final byte LIST = 15;
-  public static final byte ENUM = 16;
-  public static final byte UUID = 17;
+  public static final byte UUID = 16;
+
+  // This is not part of the TBinaryProtocol spec but Java specific
+  // implementation detail.
```
Jimexist, line 40:
```suggestion
  /** 
   * This is not part of the TBinaryProtocol spec but Java specific
   * implementation detail
   */
```

lib/java/src/main/java/org/apache/thrift/protocol/TType.java
```diff
@@ -34,6 +34,11 @@ public final class TType {
...
   public static final byte LIST = 15;
-  public static final byte ENUM = 16;
-  public static final byte UUID = 17;
+  public static final byte UUID = 16;
+
+  /** 
+   * This is not part of the TBinaryProtocol spec but Java specific
+   * implementation detail
+   */
+  public static final byte ENUM = -1;
```
ctubbsii, line 43: I'm not sure what ENUM was used for... but it seems like changing this will break wire compatibility under some circumstances when using Thrift with Java.

I'm not even sure why there needs to be a separate UUID type at all... UUIDs are just 128-bit numbers, and often hex-encoded as Strings. They can be replaced by two 8-byte numbers, a 36 character String (or 32 without the standard, but unnecessary, dashes), or a 16 byte array.

Since you can already pass a UUID in several different ways, creating a separate dedicated type seems like it may have caused more problems than it helps.

fishy: UUID is something we added to the spec in 0.17.0: https://github.com/apache/thrift/commit/4959a92385e2a6d7a4b8419784f85e5762c714cf. Java implementation used the wrong number not in the spec.

The reason to add a new type was discussed in https://issues.apache.org/jira/browse/THRIFT-5587.

ENUM is not a TType in the spec, it's only in Java's implementation of TType as far as I'm aware of. I also don't believe it's used in the wire protocol because if so the TBinaryProtocol cross test between java and other languages would break on enum fields, which does not happen today.

ctubbsii: Maybe I should have spoken up back then. In any case, since it's too late for that, I'm just trying to figure out what this constant is actually used for now, and what the implications are for changing it in Java. Are you certain the cross tests actually have an enum? If it's not used for the wire protocol, what's it doing in the first place? What's it used for?

Jimexist (approved):
might need a reformat

lib/java/src/main/java/org/apache/thrift/protocol/TType.java
fishy: this is all the usages of it:
```
$ git grep TType.ENUM
compiler/cpp/src/thrift/generate/t_java_generator.cc:    return "org.apache.thrift.protocol.TType.ENUM";
compiler/cpp/src/thrift/generate/t_java_generator.cc:        << "new org.apache.thrift.meta_data.EnumMetaData(org.apache.thrift.protocol.TType.ENUM, "
compiler/cpp/src/thrift/generate/t_javame_generator.cc:    return "TType.ENUM";
compiler/cpp/src/thrift/generate/t_javame_generator.cc:    indent(out) << "new EnumMetaData(TType.ENUM, " << type_name(type) << ".class";
lib/java/src/main/java/org/apache/thrift/TDeserializer.java:      case TType.ENUM:
lib/java/src/main/java/org/apache/thrift/TDeserializer.java:      case TType.ENUM:
lib/java/src/main/java/org/apache/thrift/partial/PartialThriftComparer.java:      case TType.ENUM:
lib/java/src/main/java/org/apache/thrift/partial/ThriftMetadata.java:          case TType.ENUM:
lib/java/src/test/java/org/apache/thrift/partial/ThriftMetadataTest.java:    validateFieldMetadata(testStruct, 7, "enumField", TType.ENUM);
lib/java/src/test/java/org/apache/thrift/partial/ThriftMetadataTest.java:      case TType.ENUM:
lib/java/src/test/java/org/apache/thrift/partial/ThriftMetadataTest.java:      case TType.ENUM:
```
the `partial` subdirectory was added from https://issues.apache.org/jira/browse/THRIFT-5443, while the 2 usages in `TDeserializer` are both switching from the metadata object stored in memory:
1. https://github.com/apache/thrift/blob/0e872c84aaa49fe9ce72123827f666c0064f0c97/lib/java/src/main/java/org/apache/thrift/TDeserializer.java#L516
2. https://github.com/apache/thrift/blob/0e872c84aaa49fe9ce72123827f666c0064f0c97/lib/java/src/main/java/org/apache/thrift/TDeserializer.java#L596

Here is enum type in the cross test:
https://github.com/apache/thrift/blob/0e872c84aaa49fe9ce72123827f666c0064f0c97/test/ThriftTest.thrift#L257

fishy:
> might need a reformat

@Jimexist what reformats need to be done? the CI didn't complain about anything: https://github.com/apache/thrift/actions/runs/5823459440/job/15790590984

fishy:
oh it passed in my original change and failed after your suggestion :)

fixed again.

lib/java/src/main/java/org/apache/thrift/protocol/TType.java
ctubbsii: I'm not that familiar with this part of the code, but from my perspective, it looks like a Thrift Enum type serialized by the previous version of Thrift (like 0.18.1) would get incorrectly deserialized as a UUID by the newer version by the newer version (0.19.0) instead of the intended enum. I previously mentioned wire compatibility, because the situation I was thinking was if the client and server had different libthrift.jar library versions.

It is highly likely I just don't understand this code. At a high level, where is this constant being serialized? Is it only ever in memory within a single process, or is it ever persisted or show up in any user objects that they might serialize?

fishy: from my reading of the code, this is really not used in the wire protocol at all (so it's never serialized/deserialized). it's used (read) by deserialization code to decide what to do, while all the "writing" only happen on thrift compiler generated java code.

basically, compiler generated java type will use this value as some metadata, and then when deserializing the library code will use the field from the metadata to handle the actual logic. the metadata is only in a in-memory object, not on the wire protocol.

ctubbsii: If it's just used in process to direct the deserialization logic, and doesn't actually get persisted anywhere, then it should be fine... but this way of directing the deserialization logic by way of in-memory metadata is very confusing to me. Oh well. :smiley_cat:

ctubbsii (approved):

dennis-reiter:
Hi!
this constant change does indeed break things, at least for Java. Enum is used in the Thrift IDL and by the [Avro-Thrift-Library](https://github.com/apache/avro/blob/bb5b8429b7bec86a972f2f1f5ae5f18eeef332a7/lang/java/thrift/src/main/java/org/apache/avro/thrift/ThriftData.java#L233). Hopefully, avro-thrift will adapt, but I am unsure about the impact on already generated thrift files.
Regards
Dennis

PR #2846, opened 2023-08-22
UnitTest may fail when the platform does not support handling localization-related operations

jackwenshann:
Fixed the situation that the locale function encountered a situation that the platform did not support when handling localization-related operations, resulting in test failure

### DIFF
--- a/lib/cpp/test/ToStringTest.cpp
+++ b/lib/cpp/test/ToStringTest.cpp
@@ -46,23 +46,31 @@ BOOST_AUTO_TEST_CASE(base_types_to_string) {
 // Windows:
 #ifndef _WIN32
 BOOST_AUTO_TEST_CASE(locale_en_US_int_to_string) {
+  try {
 #ifdef _WIN32
-  std::locale::global(std::locale("en-US.UTF-8"));
+    std::locale::global(std::locale("en-US.UTF-8"));
 #else
-  std::locale::global(std::locale("en_US.UTF-8"));
+    std::locale::global(std::locale("en_US.UTF-8"));
 #endif
-  BOOST_CHECK_EQUAL(to_string(1000000), "1000000");
+    BOOST_CHECK_EQUAL(to_string(1000000), "1000000");
+  } catch (const std::runtime_error& e) {
+
+  }
 }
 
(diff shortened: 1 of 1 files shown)


lib/cpp/test/ToStringTest.cpp
```diff
@@ -46,23 +46,31 @@ BOOST_AUTO_TEST_CASE(base_types_to_string) {
...
-  std::locale::global(std::locale("en-US.UTF-8"));
+    std::locale::global(std::locale("en-US.UTF-8"));
 #else
-  std::locale::global(std::locale("en_US.UTF-8"));
+    std::locale::global(std::locale("en_US.UTF-8"));
 #endif
-  BOOST_CHECK_EQUAL(to_string(1000000), "1000000");
+    BOOST_CHECK_EQUAL(to_string(1000000), "1000000");
+  } catch (const std::runtime_error& e) {
+
```
Jens-G, line 57: Empty catch-all clauses are triggering some sort of "Are you sure?" reflex. Especially if the patch is called a fix. I would rather label it as "curing a symptom". Can we have at least a comment explaining why this is the optimal way to handle it?

lib/cpp/test/ToStringTest.cpp
```diff
@@ -46,23 +46,31 @@ BOOST_AUTO_TEST_CASE(base_types_to_string) {
...
+    std::locale::global(std::locale("de_DE.UTF-8"));
 #endif
-  BOOST_CHECK_EQUAL(to_string(1.5), "1.5");
-  BOOST_CHECK_EQUAL(to_string(1.5f), "1.5");
-  BOOST_CHECK_EQUAL(to_string(1.5L), "1.5");
+    BOOST_CHECK_EQUAL(to_string(1.5), "1.5");
+    BOOST_CHECK_EQUAL(to_string(1.5f), "1.5");
+    BOOST_CHECK_EQUAL(to_string(1.5L), "1.5");
+  } catch (const std::runtime_error& e) {
+
```
Jens-G, line 72: Same here.

Jens-G (changes requested):

PR #2851, opened 2023-09-02
THRIFT-5733 Building code with circular includes crashes Thrift compiler

Jens-G:


jfly (approved):
LGTM! I tried this out with my repro [documented here](https://github.com/jfly/2023-09-01-thrift-circular-import), and it now quickly errors out with a nice message. Much better!

```
$ make repro
mkdir -p generated
echo 'include "bar.thrift"' > generated/foo.thrift
printf 'include "large-enum.thrift"\ninclude "foo.thrift"' > generated/bar.thrift
python -c 'print("enum LargeEnum {"); print("\n".join(f"    FOO{i} = {i}," for i in range(2000))); print("}\n");' > generated/large-enum.thrift
(cd generated; thrift --allow-64bit-consts --gen js:ts,es6 foo.thrift)
[FAILURE:/home/jeremy/tmp/2023-09-01-thrift-circular-import/generated/large-enum.thrift:2004] Recursion detected, file: "/home/jeremy/tmp/2023-09-01-thrift-circular-import/generated/foo.thrift"
make: *** [Makefile:5: repro] Error 1
```

Thank you for the quick fix!

PR #2867, opened 2023-10-12
ApacheThrift.nuspec: Updated to netstandard 2.1

emmenlau:
Currently the nuspec file is set to netstandard 2.0. As far as I know, the LTS version is netstandard 2.1. This PR increases the version to the LTS one.

### DIFF
--- a/ApacheThrift.nuspec
+++ b/ApacheThrift.nuspec
@@ -34,12 +34,12 @@
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <summary>Apache Thrift .NET Library</summary>
     <description>
-      Contains runtime libraries from lib/netstd for netstandard2.0 framework development.
+      Contains runtime libraries from lib/netstd for netstandard2.1 framework development.
     </description>
     <repository type="GitHub" url="https://github.com/apache/thrift" branch="release/0.20.0" />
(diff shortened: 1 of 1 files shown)


Jens-G:
AFAIK it is still the case that if you want to target both NET framework and NET std the 2.0 is the version of choice. Am I wrong?

emmenlau:
Ah, this may explain my problem. I want to share code with `.NET Core 3.x` and this did not work out of the box. Setting the netstandard to 2.1 helped. But I can see how this may break other use cases.

Jens-G:
From [that same URL](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-standard-not-deprecated):

> * Use netstandard2.0 to share code between .NET Framework and all other implementations of .NET.
> * Use netstandard2.1 to share code between Mono, Xamarin, and .NET Core 3.x.

emmenlau:
Ok, maybe there is another problem on my side. I will retract this PR!

Jens-G:
BTW, I would love to get rid of that as well. But that use case is not so uncommon I think. We actually use it at our company in exactly that combination where we have certain NET Standard modules included into existing older NET Framework 4.x applications that will most likely never be converted to NET Core for the rest of their life time.

PR #2882, opened 2023-11-17
THRIFT-5744: Switch to slog for go library

fishy:
Client: go

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -22,8 +22,8 @@ package thrift
 import (
 	"context"
 	"errors"
-	"fmt"
 	"io"
+	"log/slog"
 	"net"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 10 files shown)


lib/go/thrift/simple_server.go
```diff
@@ -179,8 +179,18 @@ func (p *TSimpleServer) SetForwardHeaders(headers []string) {
...
+// Deprecated: The logging inside TSimpleServer is now done via slog on error
+// level, this does nothing now. It will be removed in a future version.
+func (p *TSimpleServer) SetLogger(_ Logger) {}
+
+// SetLogContext sets the context to be used when logging errors inside
+// TSimpleServer.
+//
+// If this is not called before calling Serve, context.Background() will be
+// used.
+func (p *TSimpleServer) SetLogContext(ctx context.Context) {
```
fishy, line 192: note: I used the term `LogContext` over `BaseContext` here as I intend this context to only be used for logging.

if we use it as base context to create ctx that we pass into processors and eventually handlers, then whatever developers set on the ctx to control the logging of TSimpleServer will also end up in their service handlers, which is most likely not intended and could be a big surprise.

dcelasun: Sounds reasonable.

dcelasun (approved):

fishy:
the go 1.22 github actions is not ready yet, will rerun github actions once that's ready.

fishy:
> the go 1.22 github actions is not ready yet, will rerun github actions once that's ready.

oh nevermind the new `setup-go` can just download it from upstream if it's not pre-built. neat.

PR #2883, opened 2023-11-17
enforce to close accepted connection after processing

jshen14:
Enforce close of accepted connection in golang simple server to avoid memory leak  

- [X ] Did you squash your changes to a single commit?  (not required, but preferred)
- [X ] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?

### DIFF
--- a/lib/go/thrift/simple_server.go
+++ b/lib/go/thrift/simple_server.go
@@ -211,6 +211,10 @@ func (p *TSimpleServer) innerAccept() (int32, error) {
 			select {
 			case <-ctx.Done():
 				// client exited, do nothing
+				if client != nil {
+					client.Close()
+					client = nil
+				}
 			case <-p.stopChan:
 				// TSimpleServer.Close called, close the client connection
 				client.Close()
@@ -270,6 +274,7 @@ func (p *TSimpleServer) Stop() error {
 	}
 
 	<-ctx.Done()
+	p.stopChan = nil
 	p.stopChan = make(chan struct{})
 	return nil
 }
@@ -356,6 +361,7 @@ func (p *TSimpleServer) processRequests(client TTransport) (err error) {
 		ok, err := processor.Process(ctx, inputProtocol, outputProtocol)
 		if errors.Is(err, ErrAbandonRequest) {
 			err := client.Close()
+			client = nil
 			if errors.Is(err, net.ErrClosed) {
 				// In this case, it's kinda expected to get
 				// net.ErrClosed, treat that as no-error


lib/go/thrift/simple_server.go
```diff
@@ -211,6 +211,10 @@ func (p *TSimpleServer) innerAccept() (int32, error) {
 			select {
 			case <-ctx.Done():
 				// client exited, do nothing
+				if client != nil {
+					client.Close()
```
fishy, line 215: can you explain more why do we need to close the client here? is there any connection leak otherwise?

jshen14: Yes, if the call to Close is not called, then any dynamically allocated data in conn object is lost forever and hence the caller's golang runtime keeps eating memory.

Jens-G: How would that affect overall performance, e.g. when doing multiple calls?

jshen14: Could you help clarify what is your overall performance concern? 

Below is my understanding of how the simpleServer works with its serverTransport regarding accept an incoming request. Please correct me if I miss anything.
A client object is created by calling the underlying serverTransport.Accept(). I think client.Close() is to signal serverTransport that *thrift server* has finished using this client. So serverTransport could do whatever is required on its part, either to cleanup or recycle the object for reuse. 

Without the signal from thrift server, what is expected from serverTransport to manage *client* object created by itself? The reference to client is lost after processing the request.

I guess thrift simple server has some assumption on how serverTransport to manage client created by serverTransport itself. Could you help advise where I could find the assumption?

Jens-G: I simply asked a question. If your answer is a validated "No it will not affect perf because ..." that's fine with me.

lib/go/thrift/simple_server.go
```diff
@@ -270,6 +274,7 @@ func (p *TSimpleServer) Stop() error {
 	}
 
 	<-ctx.Done()
+	p.stopChan = nil
```
fishy, line 277: this does nothing, it's immediately overwritten by the next line

lib/go/thrift/simple_server.go
```diff
@@ -356,6 +361,7 @@ func (p *TSimpleServer) processRequests(client TTransport) (err error) {
 		ok, err := processor.Process(ctx, inputProtocol, outputProtocol)
 		if errors.Is(err, ErrAbandonRequest) {
 			err := client.Close()
+			client = nil
```
fishy, line 364: this does not work as you expected.

here `client` is the local arg of `processRequests` function, setting it to `nil` only affects that copy of `client`. it does not affect the `client` you are checking on line 214, which is at the caller of `processRequests`.

the whole nil checking regarding client is also unnecessary, if `client` is already closed, closing it again will only give you an already closed error, which you ignored anyways (on line 215).

fishy (changes requested):

lib/go/thrift/simple_server.go
jshen14: Makes sense. Then is it good to have ```defer client.Close``` in processRequests()?

fishy: I would prefer to defer it in `innerAccept` instead.

fishy:
please fix the failed unit test.

jshen14:
@fishy I updated to fix golang unit test. Now I see there is build error due to python test. I dont understand how that is related to my changes. Could you help me understand what I need to fix?

Jens-G:
Any progress here?

PR #2884, opened 2023-11-22
THRIFT-5745: Implement slog.LogValuer on go TStructs

fishy:
Client: go

Implement slog.LogValuer for all TStruct and TException generated by the compiler for go code. Also add SlogTStructWrapper in the library so we don't have to repeat it in the compiler generated go code.


dcelasun (approved):
LGTM

PR #2898, opened 2023-12-04
Github Actions: Make java spotless step optional

fishy:
Add `continue-on-error: true` so it does not block cross-tests from running.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -169,11 +169,13 @@ jobs:
         run: |
           cd lib/java
           gradle spotlessCheck
+        continue-on-error: true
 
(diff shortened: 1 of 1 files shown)


fishy:
ok this doesn't work, it breaks the whole gradle so java build is also broken.

Jimexist:
> Add `continue-on-error: true` so it does not block cross-tests from running.

wonder why it broke the cross-tests runs?

fishy:
> > Add `continue-on-error: true` so it does not block cross-tests from running.
> 
> wonder why it broke the cross-tests runs?

because cross tests depend on all other tests to pass.

PR #2914, opened 2024-01-02
Fix github actions for python3 tests

fishy:
Add a dummy test in test_sslsocket.py to workaround an issue in Python
3.12.


fishy:
filed upstream bug: https://github.com/python/cpython/issues/113661

Jimexist (approved):

PR #2917, opened 2024-01-06
[THRIFT-5752] Add TTransportFactoryInterface

sveneld:
I'm offer to add TTransportFactoryInterface for having a possibility to use own implementation of TTransportFactory in TServer.

I have a problem when I'm tried to make a TSimpleServer for client who used a 
TFramedTransport. Current realization of TTransportFactory does not give me a possibility to add a TFramedTransport as  a decorator for use TTransport
  
- [x] https://issues.apache.org/jira/browse/THRIFT-5752

### DIFF
--- a/lib/php/Makefile.am
+++ b/lib/php/Makefile.am
@@ -64,7 +64,9 @@ phpfactory_DATA = \
 	lib/Factory/TJSONProtocolFactory.php \
 	lib/Factory/TProtocolFactory.php \
 	lib/Factory/TStringFuncFactory.php \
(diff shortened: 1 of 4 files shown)


Jens-G:
How would compatibility affected? Does that change break existing code? If so, can we prevent that?

sveneld:
Combability should not be affected. Class TTransportFactory used only for creation of TServer.
I am just added an Interface for it, and replace all places of usage from TTransportFactory to TTransportFactoryInterface.
It should work fine.

PR #2919, opened 2024-01-11
THRIFT-5753: PHP 8.1 deprecated warning about return type in jsonSerialize functions

pkvach:
In PHP 8.1 a "mixed" return type was added to the "JsonSerializable::jsonSerialize()" method.

This triggers a deprecation warning for generated PHP classes:

`Return type of TestClass::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in ...`

This will not affect previous versions of PHP, meaning these changes are backwards compatible.


pkvach:
> Can you also fix such error PHP Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in Thrift/StringFunc/Core.php on line 38 ?

Sure.
But it's probably best to do it in a separate task/PR.

PR #2928, opened 2024-02-08
THRIFT-5688: Add PyPI publishing github actions

fishy:
This is tested with https://github.com/apache/thrift/pull/2927, which published to https://test.pypi.org/project/thrift-test/. I tested locally with:

    (venv) fishy@penguin:~/work/test$ pip install -i https://test.pypi.org/simple/ thrift-test
    Looking in indexes: https://test.pypi.org/simple/
    Collecting thrift-test
      Downloading https://test-files.pythonhosted.org/packages/e6/02/5885ea1406f560d0a23351f68acc2892d7f6495b16bfc2eeee8de4649777/thrift-test-0.21.0.tar.gz (62 kB)
         ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 62.3/62.3 kB 1.4 MB/s eta 0:00:00
      Preparing metadata (setup.py) ... done
    Collecting six>=1.7.2 (from thrift-test)
      Downloading https://test-files.pythonhosted.org/packages/b3/b2/238e2590826bfdd113244a40d9d3eb26918bd798fc187e2360a8367068db/six-1.10.0.tar.gz (29 kB)
      Preparing metadata (setup.py) ... done
    Building wheels for collected packages: thrift-test, six
      Building wheel for thrift-test (setup.py) ... done
      Created wheel for thrift-test: filename=thrift_test-0.21.0-cp311-cp311-linux_x86_64.whl size=416914 sha256=3a972bc562be7ed19cb37399e444ed0d373cde5319023974080b625e550901d4
      Stored in directory: /home/fishy/.cache/pip/wheels/45/20/1f/d3e1b869ac068d63ca2b2c13a2f4e33a80b360fae7091c8a9b
      Building wheel for six (setup.py) ... done
      Created wheel for six: filename=six-1.10.0-py2.py3-none-any.whl size=9942 sha256=74014380446ccf331366316cec0b1aaf40e0162e70307622b493e38e8451115f
      Stored in directory: /home/fishy/.cache/pip/wheels/e4/18/d0/e02474c90dcf14c511c0f52145d7e72e41ff3fb80b330ba58e
    Successfully built thrift-test six
    Installing collected packages: six, thrift-test
    Successfully installed six-1.10.0 thrift-test-0.21.0
    (venv) fishy@penguin:~/work/test$ python3
    Python 3.11.7 (main, Dec  8 2023, 14:22:46) [GCC 13.2.0] on linux
[...]

PR #2931, opened 2024-02-19
THRIFT-5688: Add PyPI publishing github actions

fishy:
Cherry-picking https://github.com/apache/thrift/pull/2928 & https://github.com/apache/thrift/pull/2933 into `0.20.0` branch.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,3 +1,22 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi.yml
```diff
@@ -0,0 +1,32 @@
+name: "PyPI publishing"
```
Jens-G, line 1: License header?

fishy: we don't have license header on other github action workflow files currently either:
- https://github.com/apache/thrift/blob/master/.github/workflows/build.yml
- https://github.com/apache/thrift/blob/master/.github/workflows/cmake.yml

since this is a cherry-pick of #2928 we already merged, if we deem it necessary to add license header to all 3 files I'd suggest we do them in separated PR (probably after 0.20.0 release), what do you think?

fishy: created https://github.com/apache/thrift/pull/2933 and also added to this PR.

PR #2939, opened 2024-02-26
THRIFT-5762 Expose service result objects in Java

thomasbruggink:
Some libraries want to bypass the TServer class and handle the full service startup manually. For example when building a service that hosts multiple thrift services where the IFace type is unknown when handling a request.

For example when you host multiple services on top of netty and through an HTTP path you want to route to the correct thrift service. In this situation you treat can treat an IFace as an Object and use the `getProcessMapView()` method to parse a byte array into a thrift message and pass let the `AsyncProcessFunction` handle the invocation.

To return a correct thrift response it's necessary to write the `{service_name}_result` that contains the response args. While it is possible to get an incoming args object from the (Async)ProcessFunction its unfortunately not possible to get a result object without using reflection.

This PR extends the (Async)ProcessFunction by adding a `getEmptyResultInstance` method that returns a new generic `A` (answer) that matches the `{service_name}_result` object.

This allows thrift users to write the following processing code:
```java
<I> void handleRequest(
        TProtocol in,
        TProtocol out,
        TBaseAsyncProcessor<I> processor,
        I asyncIface
) throws TException {
    final Map<String, AsyncProcessFunction<Object, TBase<?, ?>, TBase<?, ?>, TBase<?, ?>>> processMap = (Map) processor.getProcessMapView();
    final var msg = in.readMessageBegin();
    final var fn = processMap.get(msg.name);

    final var args = fn.getEmptyArgsInstance();
    args.read(in);
    in.readMessageEnd();

    if (fn.isOneway()) {
        return;
    }

    fn.start(asyncIface, args, new AsyncMethodCallback<>() {
        @Override
        public void onComplete(TBase<?, ?> o) {
            try {
                out.writeMessageBegin(new TMessage(fn.getMethodName(), TMessageType.REPLY, msg.getSeqid()));
                final var response_result = fn.getEmptyResultInstance();
                final var success_field = response_result.fieldForId(SUCCESS_ID);
                ((TBase) response_result).setFieldValue(success_field, o);
                response_result.write(out);
                out.writeMessageEnd();
                out.getTransport().flush();
            } catch (TException e) {
                throw new RuntimeException(e);
            }
        }

        @Override
        public void onError(Exception e) {
            try {
                out.writeMessageBegin(new TMessage(fn.getMethodName(), TMessageType.EXCEPTION, msg.getSeqid()));
                ((TApplicationException) e).write(out);
                out.writeMessageEnd();
                out.getTransport().flush();
            } catch (TException ex) {
                throw new RuntimeException(ex);
            }
        }
    });
}
```
The above example code doesn't need any reference to the original types and can dynamically create the correct objects to return a correct response.

Please let me know what you think.

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/ProcessFunction.java
+++ b/lib/java/src/main/java/org/apache/thrift/ProcessFunction.java
@@ -8,86 +8,91 @@ import org.apache.thrift.transport.TTransportException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public abstract class ProcessFunction<I, T extends TBase> {
-  private final String methodName;
+public abstract class ProcessFunction<I, T extends TBase, A extends TBase> {
+    private final String methodName;
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(ProcessFunction.class.getName());
+    private static final Logger LOGGER = LoggerFactory.getLogger(ProcessFunction.class.getName());
 
-  public ProcessFunction(String methodName) {
-    this.methodName = methodName;
-  }
-
-  public final void process(int seqid, TProtocol iprot, TProtocol oprot, I iface)
-      throws TException {
-    T args = getEmptyArgsInstance();
-    try {
-      args.read(iprot);
-    } catch (TProtocolException e) {
-      iprot.readMessageEnd();
-      TApplicationException x =
-          new TApplicationException(TApplicationException.PROTOCOL_ERROR, e.getMessage());
-      oprot.writeMessageBegin(new TMessage(getMethodName(), TMessageType.EXCEPTION, seqid));
-      x.write(oprot);
-      oprot.writeMessageEnd();
-      oprot.getTransport().flush();
-      return;
-    }
-    iprot.readMessageEnd();
-    TSerializable result = null;
-    byte msgType = TMessageType.REPLY;
-
-    try {
-      result = getResult(iface, args);
-    } catch (TTransportException ex) {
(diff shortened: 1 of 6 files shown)


lib/java/src/main/java/org/apache/thrift/ProcessFunction.java
```diff
@@ -8,86 +8,91 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public abstract class ProcessFunction<I, T extends TBase> {
-  private final String methodName;
+public abstract class ProcessFunction<I, T extends TBase, A extends TBase> {
+    private final String methodName;
```
fishy, line 12: I believe this is the commit that breaks spotless check for java in CI (and thus stops running all cross tests): https://github.com/apache/thrift/actions/runs/8048152173/job/22320319518

@thomasbruggink can you please run `./gradlew :spotlessApply` and commit the format changes it made in a new PR? thank you.

thomasbruggink: Yes, let me make a new PR today.

thomasbruggink: PR: https://github.com/apache/thrift/pull/2945

fishy:
@Jens-G just to double check: This is _not_ in 0.20.0 branch, is that correct?

Jens-G:
No, its not.

Merged early in March:
https://github.com/apache/thrift/commits/master/

But branch 0.20.2 created about a month earlier, and not merged into
https://github.com/apache/thrift/commits/0.20.0/

PR #2940, opened 2024-02-26
THRIFT-5706: lib/cpp Fix the Security tests on openssl 1.1 and 3.0

thomasbruggink:
This PR fixes the Security tests to build on a clean install of ubuntu
20.04 and ubuntu 22.04 without modifications to the systems openssl
configuration.

    * Enable TLS 1.0 and TLS 1.1 on OpenSSL 1.1 with the seclevel=0 flag
    * Disable TLS 1.0 and TLS 1.1 on OpenSSL 3.0

While its technically possible to enable it on OpenSSL 3 I think because
of all the issues with these old TLS versions dropping support for it is
better.

This PR builds forth on the work done here: https://github.com/apache/thrift/pull/2811

Tested with the ubuntu 20.04 (OpenSSL 1.1) and 22.04 (OpenSSL 3.0) docker containers.
All lib/cpp tests succeed in both.

### DIFF
--- a/lib/cpp/test/SecurityFromBufferTest.cpp
+++ b/lib/cpp/test/SecurityFromBufferTest.cpp
@@ -109,7 +109,13 @@ struct SecurityFromBufferFixture {
       shared_ptr<TSSLServerSocket> pServerSocket;
 
       pServerSocketFactory.reset(new TSSLSocketFactory(static_cast<apache::thrift::transport::SSLProtocol>(protocol)));
-      pServerSocketFactory->ciphers("ALL:!ADH:!LOW:!EXP:!MD5:@STRENGTH");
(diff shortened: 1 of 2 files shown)


thomasbruggink:
@emmenlau any update on the review of this?
cpp test suite is still broken without this patch on modern systems.

emmenlau:
Sorry for the long delay! Looks good to me, lets bring this in!

emmenlau (commented):
I'm not sufficiently deep in the details of OpenSSL to know why these changes are correct. But they look thorough and to quite some extent self-explanatory. Also, I understand that tests have *increased* rather than *decreased* by the changes (see the changed check of openssl1 vs openssl3 at the end. Therefore: Looks good to me.

emmenlau (approved):

PR #2943, opened 2024-03-08
THRIFT-5766: Replace std::endl with "\n"

CJCombrink:
Replaced usage of `endl` with `\n` directly

  - Using endl for linebreaks is bad practice
- This is confirmed by the `static const string endl = "\n"` (removed)

Reference to my comment [here](https://github.com/CJCombrink/thrift/commit/f02ac2fb573bed72e9a2d1875807c6ff7ac19ec8#r139564157)


CJCombrink:
(_Personally_)I don't think it is a matter of taste.
`std::endl`'s claim to fame (if I can phrase it like that) is that it forces a newline and flush the stream.
At some point in the past someone decided that the "flush the stream" is undesirable and _redefined_ `endl` to just force a new line.
Thus the usage of `endl` enforces a known bad habit, or rather misplaced/misguided habit.
The more appropriate thing to do at that time was make this change, instead of enforcing/entertaining/hiding a bad habit.

I am also bringing up [C++ Core Guidelines SL.io.50](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slio50-avoid-endl) (PS Which Bjarne [commented](https://github.com/isocpp/CppCoreGuidelines/issues/357#issuecomment-153515258) on himself).
And then if you pay attention to the following comment on SL.io.50:
> Note Apart from the (occasionally important) issue of performance, the choice between '\n' and endl is almost completely aesthetic.

I would argue if it was only aesthetic why was the _overload_ added in the first place...

PS: I am probably passionate about the wrong things, shoot down the PR if this is 'out of scope' for this repo

CJCombrink:
I _doubled down_ and applied the changes to the whole codebase in my fork @ https://github.com/CJCombrink/thrift/pull/1
If there is any appetite for this type of change, please indicate a preference between baby steps (just "t_cpp_generator.cc") or all or nothing as per the linked PR.

emmenlau:
> (_Personally_)I don't think it is a matter of taste. `std::endl`'s claim to fame (if I can phrase it like that) is that it forces a newline and flush the stream.

Well we are certainly on the same page when it comes to `std::endl`, as this has a number of significant drawbacks! But the code in question is already using `\n` rather than `std::endl`. The main difference is that it is concatenating every time with `<<` which may have a negative performance impact. The impact on readability is (in my humble opinion) debatable, because I can see also advantages of recognizing `<< endl <<` as a clear sign of newline, whereas "hello\nworld" may not jump as much into the eye of the reader.

In any case, I'm ok with the PR, but without a strong opinion pro or con on my side :-)

Jens-G:
> static const string endl = "\n"; // avoid ostream << std::endl flushes

That line was introduced intentionally in all generators. You may look up the git history and find out why, before removing it.

CJCombrink:
@Jens-G 
Sorry I did not do the research before making the PR, but the linked ticket confirms my suspicion 100%:
A bad practice was followed and instead of fixing the actual issue, by not using `std::endl`, the PR decided to entertain a bad practice with the motivation of "minimizing code churn" by changing expected behaviour with hidden unexpected behaviour (PS I was fully aware of that constant before making this PR). 

9 Years later I am hoping that at least we can get to a point where sanity is restored and bad practices are not encouraged/entertained in such important code bases. 

PS: I come from a background where juniors used `std::endl` on embedded devices that flushed over serial. On confrontation the response was "_we were taught that way_". We should be better in teaching and educating on these matters, not hide it with `consts` that change documented behaviour and make bad practices acceptable. 

I am saying this again, I realise I am passionate about the wrong things thus I will not take offence if this/these PR(s) is/are declined, but I might keep stating the point :P 

@emmenlau 
>  advantages of recognizing `<< endl <<` as a clear sign of newline

Depends on perspective: I see it as: "why do you need flushing" and my editor (VS Code with no custom modification) highlights `\n` differently in string literals, thus I can argue it stands out even more and communicates the intend clearly (no hidden oh but we manually changed behaviour).

Jens-G:
+1 Go on then, change all the files and add a comment to the ticket, to put an end on this fruitless discussion. Question remains, why bad practice can't be fixed where it belongs - by properly deprecating `std::endl` 

> We should be better in teaching and educating on these matters,

Usually [deprecation](https://en.cppreference.com/w/cpp/language/attributes/deprecated) and [removal](https://github.com/isocpp/CppCoreGuidelines/issues/357) teaches quite well. 

> It is bad, and most people know it, 

[Agree, I wasn't fully aware of it](https://github.com/search?q=%2Fendl%2F+language%3Acpp&type=code). 


___
PS: Waiting for the great Rust rewrite, to save more walls.

CJCombrink:
I Updated this PR with replacing `std::endl` in most of the `std::cout` calls.
I did go through most of the changes not in the generators folder manually. 
For the generators I did spot check and I did test the new output vs the old output for at least C++, there were no changes in the generated files.

Do I still need to create a JIRA ticket for this change? I did request a user already thus waiting for that.

compiler/cpp/src/thrift/generate/t_cl_generator.cc
```diff
@@ -221,11 +221,11 @@ void t_cl_generator::package_def(std::ostream &out) {
     }
     out << ")";
   }
-  out << ")" << endl << endl;
+  out << ")\n\n";
```
fishy, line 224: in cases like this, the usage of `endl`/`std::endl` makes it much more obvious how many lines we are adding while `\n` being part of the longer string just meshes everything together.

I'd suggest still explicitly adding every `\n` individually, like this:
```suggestion
  out << ")" << "\n" << "\n";
```

but I don't feel strongly either way

CJCombrink: @fishy Thanks for the review and feedback.

Although I do understand the suggestion I specifically went through the effort to combine these. It does sort of stand out, but some editors color the `\n` different and I deem that good enough. I find myself looking at the generated code much more often than the generator code thus there I quickly see how much (or few) newlines are added. 

If there is enough pushback against the combination I can revert that combination change and (force)push the updates. 

I don't think speed matters that much but I guess calling the `operator<<` multiple times, more than the needed amount of time, would have a performance and speed implication.

emmenlau: I'm slightly with @fishy here, as my comment in https://github.com/apache/thrift/pull/2943#issuecomment-1986363604 was actually going in the same direction: Having dedicated calls to `<< "\n"` jumps much more to the eye and conveyes how many newlines are added.

If you would be willing to change that, to make every newline jump out more, I'm two thumbs up for this PR! But without that change, I'm still ok to merge, albeit with minor hesitation...

CJCombrink: @emmenlau Thank you again for the feedback.
I think there is enough reason to put back the `<< "\n` thus I will put in the effort (tonight) to 'revert' that change.

CJCombrink:
@fishy @emmenlau 
Is there anything else that I need to address in this PR, something else that stands out as 'not so nice' or blockers?

Also, I am not sure that I am responsible for the build failures, is there a way to know?

emmenlau:
@CJCombrink from my point of view, everything else is good, thanks a lot for the nice work!

The build failures are a continuous nightmare. The best you could do (and this is what I typically do) is check them individually, and see if the error seems related. Often builds fail because some third party dependency package can not be installed in some build machine, and those are the obvious ones to ignore. The others are harder, for those you can just compare the latest master build status with yours, to see if "you" broke it, or the person before you...

In any case, we will also take a look at those before merging, but its great if you validate as much as you can upfront...

CJCombrink:
I updated the branch to use `<< '\n'` instead.
TBH I am starting to appreciate the clarity it brings on many of the places. I still need more time to process `<< '\n' << '\n'`...

fishy:
FYI the lib-java-kotlin failure in CI is being addressed in https://github.com/apache/thrift/pull/2945

CJCombrink:
@fishy I think I have done what I can. 

I did check the test, master vs my changes for the parts that my environment is set up for and I don't see any regression.
I did run into #2937 when I started this work, which would enable more test to be run probably. 

I have also compared the generated code for the unit tests with those against master and see no difference.

What else can I do and what are the next steps for this PR?

fishy:
@CJCombrink sorry I'm in some family emergency and won't have time to review it any further for the next few weeks. None of my comment is blocking anyways :)

Mario, Jens, and/or others should be able to guide you through the process.

CJCombrink:
I did some thinking...
For a major project like this, a PR  touching 56 files with more than 10,135 changes from a random person might not be the easiest thing to consider. 
Thus, would it not be better if we keep the changes local to one generator and then maybe in future consider this type of change?
Obviously I would want the changes in `t_cpp_generator.cc`

Or maybe just make the changes in the generators. I did go all the way with all other files, but baby steps might be the best or safest course of action.

PS: I started a new ticket: [THRIFT-5772](https://issues.apache.org/jira/browse/THRIFT-5772) and if I have to choose I would rather have energy put there compared to this change, although this would be nice to have (in the C++ generator :P )

Jens-G:
> For a major project like this, a PR touching 56 files with more than 10,135 changes from a random person might not be the easiest thing to consider.

That's right but I'm fine in this particular case. If you could just add the commits re above changes at the end, that would make followup review easier for me. I will squash them before merge myself.

Jens-G:
Any news here?

CJCombrink:
> Any news here?

Hi Sorry, I was busy with a few other things.
Will work through the feedback today.

compiler/cpp/src/thrift/generate/t_dart_generator.cc
```diff
@@ -37,8 +37,7 @@ using std::string;
 using std::stringstream;
 using std::vector;
 
-static const string endl = "\n"; // avoid ostream << std::endl flushes
-static const string endl2 = "\n\n";
+static const string br = "\n"; // Line break
```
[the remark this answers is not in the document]
CJCombrink: The rationale here was that at some places there are `"literal" + endl` and at the time I thought it good to keep the `endl` as a string but rename to steer away from `endl` usage. 
But the correct solution would be not to do addition on literals but instead strings, I am fixing this with the next commit.

compiler/cpp/src/thrift/generate/t_cpp_generator.cc
```diff
@@ -2902,187 +2900,187 @@ void t_cpp_generator::generate_service_client(t_service* tservice, string style)
...
-            indent() << "  // this will temporarily unlock the readMutex, and let other clients get work done" << endl <<
-            indent() << "  this->sync_->waitForWork(seqid);" << endl <<
-            indent() << "} // end while(true)" << endl;
+          out << indent() << "  }" << '\n'
+              << indent() << "  // seqid != rseqid" << '\n'
+              << indent() << "  this->sync_->updatePending(fname, mtype, rseqid);" << '\n'
+              << '\n'
+              << indent()
+              << "  // this will temporarily unlock the readMutex, and let other clients get work "
+                 "done"  << '\n'
```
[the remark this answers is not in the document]
CJCombrink: This one slipped through... apologies (probably my auto formatter...)

compiler/cpp/src/thrift/generate/t_js_generator.cc
```diff
@@ -43,7 +43,7 @@ using std::stringstream;
 using std::unordered_map;
 using std::vector;
 
-static const string endl = "\n"; // avoid ostream << std::endl flushes
+static const string br = "\n"; // Line break
```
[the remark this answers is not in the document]
CJCombrink: See previous comment, also updated

compiler/cpp/src/thrift/generate/t_kotlin_generator.cc
```diff
@@ -1972,16 +1970,15 @@ bool t_kotlin_generator::is_enum_map(t_type* ttype) {
  */
 string t_kotlin_generator::kotlin_package() {
   if (!package_name_.empty()) {
-    return string("package ") + package_name_ + endl + endl;
+    return string("package ") + package_name_ + string{"\n\n"};
   }
```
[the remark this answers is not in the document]
CJCombrink: Habit... Updated to ` ... + "\n" + "\n"`
See [ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax)
In the spirit of the PR and keeping with the format of the surrounding code I updated as mentioned above

CJCombrink:
@Jens-G 

> Most of the missing single lines are `cerr` outputs. I mean, if we want to save walls, we should be more consequent and include `cerr `outputs as well, maybe doing the flush explicitly if needed. Or would `cerr `be the exception from the rule that `endl `is generally evil?

Initially I thought it might be good to keep cerr to flush. After your comment I  did some thinking and reading and I see `std::cerr` is unbuffered, thus every `<<` will flush and `endl` will then be a double flush, thus I think it is safe to conclude it is evil there as well.

compiler/cpp/src/thrift/generate/t_json_generator.cc
```diff
@@ -537,7 +535,7 @@ void t_json_generator::write_const_value(t_const_value* value, bool should_force
 }
 
 string t_json_generator::json_str(const string& str) {
-  return quot + escape_json_string(str) + quot;
+  return string("\"") + escape_json_string(str) + string("\n");
 }
```
Jens-G, line 539: Typo?  quot should be  string("\"")

compiler/cpp/src/thrift/generate/t_kotlin_generator.cc
```diff
@@ -1601,43 +1599,43 @@ void t_kotlin_generator::generate_service_processor(t_service* tservice) {
...
+  indent(out) << "private val scope: kotlinx.coroutines.CoroutineScope," << '\n';
   indent(out) << "private val processMap: kotlin.collections.Map<kotlin.String, "
                  "org.apache.thrift.AsyncProcessFunction<"
               << tservice->get_name()
               << ", out org.apache.thrift.TBase<*, "
-                 "*>, out kotlin.Any, out org.apache.thrift.TBase<*, *>>> = mapOf("
-              << endl;
+                 "*>, out kotlin.Any>> = mapOf("
+              << '\n';
   indent_up();
```
Jens-G, line 1618: Probably merge failure ... that came in with 2826098 THRIFT-5762 Fix spotless errors
 `, out org.apache.thrift.TBase<*, *>`

Jens-G (changes requested):
Great, thanks! Two minor bugs left though.

CJCombrink:
@Jens-G Thank you for your effort and _meticulous attention to the details_

emmenlau:
nice work @CJCombrink and (as usually) @Jens-G !

PR #2946, opened 2024-03-15
THRIFT-5767: Go Simple JSON Protocol re-allocates memory for every escaped quote

k-walton:
The code in `ParseQuotedStringBody()` was recursing in our application every time there was an escaped quote in strings  and generated a new string buffer for every recursive call. This was a real issue for escaped JSON in our application in particular.

So this pull requests moves that logic into a loop, and uses a string builder to avoid re-allocating buffers for every iteration.

We observed an ~40x decrease in p99 latency from 2 seconds -> 50 milliseconds on our servers when implementing this change. So it is a substantial performance improvement for this particular use case.

### DIFF
--- a/lib/go/test/tests/string_parse_allocation_test.go
+++ b/lib/go/test/tests/string_parse_allocation_test.go
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package tests
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/apache/thrift/lib/go/test/gopath/src/stringparseallocationtest"
+	"github.com/apache/thrift/lib/go/thrift"
+)
+
+func TestSimpleJsonStringParse_Allocations(t *testing.T) {
+	byteAllocationLimit := 100 * 1024 // 100 KB
+	res := testing.Benchmark(BenchmarkSimpleJsonStringParse_Allocations)
+	if res.AllocedBytesPerOp() > int64(byteAllocationLimit) {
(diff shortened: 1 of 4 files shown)


fishy:
The code looks good to me, but some suggestions:

Can you please add a benchmark test with the example you gave in the jira ticket, and post the before/after result here, with `b.ReportAllocs()` called? since this benchmark test will require a thrift definition, you likely cannot have it under `lib/go/thrift` and will need it to be under `lib/go/test` instead.

Depending on the result, I would also suggest you to write an unit test (also likely need to be under `lib/go/test`, in the same file as the benchmark test), that runs the benchmark test (via https://pkg.go.dev/testing#Benchmark), and checks the result, and verify that `AllocsPerOp` is below a certain threshold.

k-walton:
> The code looks good to me, but some suggestions:
> 
> Can you please add a benchmark test with the example you gave in the jira ticket, and post the before/after result here, with `b.ReportAllocs()` called? since this benchmark test will require a thrift definition, you likely cannot have it under `lib/go/thrift` and will need it to be under `lib/go/test` instead.
> 
> Depending on the result, I would also suggest you to write an unit test (also likely need to be under `lib/go/test`, in the same file as the benchmark test), that runs the benchmark test (via https://pkg.go.dev/testing#Benchmark), and checks the result, and verify that `AllocsPerOp` is below a certain threshold.

Sure! I was trying to keep the change set small, but those sound like a great idea! I'll go ahead and take some time to add those.

lib/go/test/tests/string_parse_allocation_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/apache/thrift/lib/go/test/gopath/src/stringparseallocationtest"
+	"github.com/apache/thrift/lib/go/thrift"
+)
+
+func TestSimpleJsonStringParse_Allocations(t *testing.T) {
+	byteAllocationLimit := 100 * 1024 // 100 KB
+	res := testing.Benchmark(BenchmarkSimpleJsonStringParse_Allocations)
+	if res.AllocedBytesPerOp() > int64(byteAllocationLimit) {
+		t.Errorf("Total memory allocation size too high: %d", res.AllocedBytesPerOp())
```
fishy, line 17: nit: also output what's the limit:
```suggestion
		t.Errorf("Total memory allocation size too high: %d (> %d)", res.AllocedBytesPerOp(), byteAllocationLimit)
```

fishy (approved):

k-walton:
@fishy I've added the test to validate that the benchmark's memory allocation is below a threshold.  I chose the `AllocedBytesPerOp()` stat to make a limit, due to the normal protocol buffer reads using allocations as well making the results less clear.

I'll add the before and after below.

Before change:
```
=== RUN   BenchmarkSimpleJsonStringParse_Allocations
BenchmarkSimpleJsonStringParse_Allocations
BenchmarkSimpleJsonStringParse_Allocations-10               5881            222369 ns/op         1073930 B/op       2016 allocs/op
```

After change:
```
=== RUN   BenchmarkSimpleJsonStringParse_Allocations
BenchmarkSimpleJsonStringParse_Allocations
BenchmarkSimpleJsonStringParse_Allocations-10              28100             41955 ns/op           15440 B/op       1027 allocs/op
```

fishy (approved):

PR #2952, opened 2024-03-27
THRIFT-5772: UUID support for c++

CJCombrink:
[THRIFT-5772: Add UUID support for C++](https://issues.apache.org/jira/browse/THRIFT-5772)

 UUID support for C++.
See description and questions on the linked ticket.

-- Not yet, want the history on the review, can squash before merge if needed


CJCombrink:
See https://github.com/CJCombrink/thrift/pull/2 for a UUID wrapper instead of using `std::string`. 
Personally I prefer the wrapper solution

xsc351072127:
I got a compile error with my CentOS 7 dev.
```
libtool: compile:  g++ -std=c++11 -DHAVE_CONFIG_H -I/home/dev/workspace/boost/boost_1_56_0 -I./src -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -Wall -Wextra -pedantic -g -O2 -MT src/thrift/TUuid.lo -MD -MP -MF src/thrift/.deps/TUuid.Tpo -c src/thrift/TUuid.cpp  -fPIC -DPIC -o src/thrift/.libs/TUuid.o
src/thrift/TUuid.cpp: In constructor 'apache::thrift::TUuid::TUuid(const string&)':
src/thrift/TUuid.cpp:40:43: error: cannot convert 'boost::uuids::uuid' to 'uint8_t {aka unsigned char}' in initialization
     const boost::uuids::uuid uuid{gen(str)};
```

My gcc version is:
```
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC)
```
And I got some warring
```
src/thrift/TUuid.cpp: In member function 'bool apache::thrift::TUuid::is_nil() const':
src/thrift/TUuid.cpp:48:31: warning: missing initializer for member 'boost::uuids::uuid::data' [-Wmissing-field-initializers]
   boost::uuids::uuid uuid_tmp{};
```

CJCombrink:
I can look into this. 
I see the Thrift [Language requirements](https://thrift.apache.org/docs/install/#language-requirements) state g++ 4.2 and boost 1.53.0 thus I will have to look at it from that side, if it is still valid (but assuming it is) since you are well within that spec.

As for the OS, Centos 7 reached end of life June 30th, 2024 (https://www.redhat.com/en/blog/centos-linux-going-end-life-what-does-mean-me) thus I am going to assume support from that angle will not be possible (And I don't have the means to consider RHEL).

It should not be too difficult to fix. 
Will you please create an issue in the Thrift JIRA and just post it here?

xsc351072127:
Thanks for reply, I will try to create a issue in Thrift JIRA.
I can provide more infomation about these isse.

1.Yes my production environment is RHEL7 for an old financial system. So I have no choose to use Centos7 and old version gcc 
 toolchain because libc and other runtime in OS.  I  have strictly followed the "Language requirements" page to build Thtift.

2.I struggle try to know what happen with my compile. 
I read a lots of cpp's initialization syntax:
https://en.cppreference.com/w/cpp/language/initialization
https://en.cppreference.com/w/cpp/language/aggregate_initialization
https://en.cppreference.com/w/cpp/language/copy_constructor
(....WTF)

And do some search with some key world, I got a gcc buy bugzilla may be related, but I am not sure:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53361

I do the minimum reproduction. try the godbolt to get the IR tree. with the compile options:--std=c++11 -Wall -fno-elide-constructors
```
class uuid
{
public:
    //should not declared user constructors
    uuid() = default;
    int data[16];
};

int main()
{
    uuid uuu;
    uuid u{uuu};
}
```
https://godbolt.org/z/ehdr1bj5P
For  gcc5 and later version compile, will call the default copy constructor create by compiler, but for gcc4.x will get a compile error.
I do not why, but this is the phenomenon.

PR #2958, opened 2024-04-08
THRIFT-5773: Strong UUID for C++

CJCombrink:
[THRIFT-5773: UUID wrapper for C++](https://issues.apache.org/jira/browse/THRIFT-5773)

  Adds a strong UUID wrapper for C++.
The use of a `std::string` can be misleading and since C++ is a strong typed language, this makes the most sense. 


lib/cpp/src/thrift/TUuid.h
```diff
@@ -0,0 +1,148 @@
...
+   */
+  explicit TUuid(const std::string& str) noexcept;
+
+  /**
+   * Copy assignment from a UUID string.
+   *
+   * This function will throw an exception if the string is not
+   * a valid UUID.
+   */
+  TUuid& operator=(const std::string& str) noexcept(false);
```
CJCombrink, line 71: Any comment on this function? 
I am still very much in two minds if this function should ever exist.
Perhaps not since there is an `explicit` constructor taking a string.
I am concerned about the porting issue for people that already used a string like object (byte[16]) but wants to upgrade to using the strong UUID.

CJCombrink:
**NB**: This PR also includes #2968 (THRIFT-5776) since I had to fix those errors before I could make progress on the makefile related unit tests. I have rebased on that branch

CJCombrink:
I see there are MSVC build errors related to this, I will have to look into this

CJCombrink:
@Jens-G I see there are some python tests failing in the MSVC build, would my changes be the cause (I can't think that it would be but I have been wrong in the past)

PR #2959, opened 2024-04-11
THRIFT-5774: Add remote client's IP address to ServerContext in TServ…

yoelyang:
[THRIFT-5774: Add remote client's IP address to ServerContext in TServerEventHandler](https://issues.apache.org/jira/browse/THRIFT-5774)

  Add the remote client's IP address to the ServerContext in TServerEventHandler when creating ServerContext.


ctubbsii (commented):
I'm not familiar enough with the implementation to know if this is good or not, but I will say that I think this feature would be great to have. My project, Apache Accumulo, has had to do workarounds for years to get the client address, just to log it. It would be great to not have to do that anymore.

Jens-G:
What happens if the transport is not socket based?

lib/java/src/main/java/org/apache/thrift/server/TServerEventHandler.java
```diff
@@ -37,8 +38,15 @@ public interface TServerEventHandler {
   /** Called before the server begins. */
   void preServe();
 
-  /** Called when a new client has connected and is about to being processing. */
-  ServerContext createContext(TProtocol input, TProtocol output);
+  /**
+   * Called when a new client has connected and is about to being processing. The
+   * remoteSocketAddress and localSocketAddress are null when transport is not socket based
+   */
+  ServerContext createContext(
```
dengzhhu653, line 45: how add a default method in ServerContext to set the address, instead of a new method in the event handler?

yoelyang: This is a good idea, I'll modify the implementation.

Jens-G:
* What went wrong:
Execution failed for task ':cross-test-server:ktfmtCheckMain'.
> [ktfmt] Found 1 files that are not properly formatted:
  src/main/kotlin/org/apache/thrift/test/TestServer.kt

yoelyang:
> * What went wrong:
>   Execution failed for task ':cross-test-server:ktfmtCheckMain'.
> 
> > [ktfmt] Found 1 files that are not properly formatted:
> > src/main/kotlin/org/apache/thrift/test/TestServer.kt

I'm sorry, I forgot to format the Kotlin code.

lib/java/src/main/java/org/apache/thrift/server/ServerContext.java
```diff
@@ -42,4 +44,11 @@ public interface ServerContext {
    *     unwrapped from this context.
    */
   boolean isWrapperFor(Class<?> iface);
+
+  /**
+   * Apply the transport to save some connection information.
+   *
+   * @param transport The base transport
+   */
+  default void applyTransport(TTransport transport) {}
```
dengzhhu653, line 53: To make it simpler, how about make this method accept the SocketAddress? something like
`default void setRemoteAddress(SocketAddress remoteAddress) {}`
By this way, we don't expose the `TTransport` directly to the application side, and the old ServerContext doesn't need to extend the SocketAddressServerContext just for getting this address.

yoelyang: Indeed, exposing the transport to the application layer is not a good practice; I overlooked this due to considerations of extensibility.

Jens-G:
What's the status here? Are there still any open issues with this?

yoelyang:
> What's the status here? Are there still any open issues with this?

The issues raised so far have been resolved.

PR #2961, opened 2024-04-15
THRIFT-5777: python fix mismatched timeout exceptions.

bwangelme:
## What error occurred?

When the Python client times out while reading content from the server, the exception thrown is not "read timeout," but rather "unexpected exception."

### the socket.timeout error doesn't includes the errno parameter in its args attribute.

I try to make a timeout error on my pc, it only contains the error string.

```
Python 3.11.1 (main, Jan  5 2023, 14:25:08) [GCC 11.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.13.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import socket
   ...: from urllib.request import urlopen
   ...:
   ...: url = 'http://httpbin.org/get'
   ...:
   ...: socket.setdefaulttimeout(0.01)
   ...: try:
   ...:     urlopen(url)
   ...: except Exception as e:
   ...:     err1 = e.reason
   ...:
   ...:

In [2]: type(err1)
Out[2]: TimeoutError

In [3]: err1.args
Out[3]: ('timed out',)
```

And my system is ubuntu 22.04

```
ø> uname -a
Linux Macmini 6.2.0-36-generic #37~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Oct  9 15:34:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

### TSocket.read handle exception error

Because `socket.error.args == ('timed out',)`, the condition `elif e.args[0] == errno.ETIMEDOUT` cannot be True, resulting in the throwing of the wrong exception `TTransportException(message="unexpected exception", inner=e)`.

## How to fix it

I determine whether a timeout has occurred by checking if the exception is a socket.timeout, no longer relying on e.args.

    - https://issues.apache.org/jira/browse/THRIFT-5777

### DIFF
--- a/lib/py/src/transport/TSocket.py
+++ b/lib/py/src/transport/TSocket.py
@@ -151,6 +151,8 @@ class TSocket(TSocketBase):
     def read(self, sz):
         try:
             buff = self.handle.recv(sz)
+        except socket.timeout as e:
+            raise TTransportException(type=TTransportException.TIMED_OUT, message="read timeout", inner=e)
         except socket.error as e:
             if (e.args[0] == errno.ECONNRESET and
                     (sys.platform == 'darwin' or sys.platform.startswith('freebsd'))):
@@ -161,8 +163,6 @@ class TSocket(TSocketBase):
                 self.close()
                 # Trigger the check to raise the END_OF_FILE exception below.
                 buff = ''
-            elif e.args[0] == errno.ETIMEDOUT:
-                raise TTransportException(type=TTransportException.TIMED_OUT, message="read timeout", inner=e)
             else:
                 raise TTransportException(message="unexpected exception", inner=e)
         if len(buff) == 0:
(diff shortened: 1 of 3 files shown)


lib/py/src/transport/TSocket.py
```diff
@@ -151,6 +151,8 @@ def open(self):
     def read(self, sz):
         try:
             buff = self.handle.recv(sz)
+        except socket.timeout as e:
```
fishy, line 154: starting from python 3.10, this is a _deprecated_ alias to `TimeoutError`: https://docs.python.org/3.12/library/socket.html#socket.timeout

while `TimeoutError` also exists in 3.8 (the earliest version of python still supported): https://docs.python.org/3.8/library/exceptions.html#TimeoutError

so I wonder if we should be more future proof and handle both `socket.timeout` and `TimeoutError` here, and then we can drop `socket.timeout` when 3.10 becomes the earliest version of python supported.

bwangelme: modified in https://github.com/apache/thrift/pull/2991

PR #2962, opened 2024-04-16
Use pip install instead of deprecated python setup.py install command

ccoulombe:
Use pip install instead of deprecated python setup.py install command.

Egg formats will be deprecated in pip v24.3 (planned for October). Users are required to use `pip install` instead.

For reference : https://github.com/pypa/pip/issues/12330

### DIFF
--- a/lib/py/Makefile.am
+++ b/lib/py/Makefile.am
@@ -45,7 +45,7 @@ all-local: py3-build
 # Old version (can't put inline because it's not portable).
 #$(PYTHON) setup.py install --prefix=$(prefix) --root=$(DESTDIR) $(PYTHON_SETUPUTIL_ARGS)
 install-exec-hook:
-	$(PYTHON) setup.py install --root=$(DESTDIR) --prefix=$(PY_PREFIX) $(PYTHON_SETUPUTIL_ARGS)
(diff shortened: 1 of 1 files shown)


lib/py/Makefile.am
```diff
@@ -45,7 +45,7 @@ all-local: py3-build
 # Old version (can't put inline because it's not portable).
 #$(PYTHON) setup.py install --prefix=$(prefix) --root=$(DESTDIR) $(PYTHON_SETUPUTIL_ARGS)
 install-exec-hook:
-	$(PYTHON) setup.py install --root=$(DESTDIR) --prefix=$(PY_PREFIX) $(PYTHON_SETUPUTIL_ARGS)
```
fishy, line 48: do we still need `setup.py`? or is it still used by pip?

ctubbsii: Here's more background for it: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

There is a link to https://setuptools.pypa.io/en/latest/userguide/declarative_config.html 

The short answer is: no, you probably don't need setup.py, and should probably be using setuptools' declarative config file, setup.cfg (if you need one at all).

PR #2964, opened 2024-04-23
THRIFT-5779: Fix Thrift server getting killed for incomplete requests

anshulmgupta:
The thrift server is getting killed when using security port scan tools in the hosts running the thrift server. These tools try to connect to the open ports by sending requests to the ports, and the error can happen when accept syscall call, waiting for an incoming connection, or receiving a connection that terminates before the accept process completes, hence killing the thrift server. This can cause potential DoS (Denial of service) attacks on the applications running the thrift server, causing them to become unresponsive. Sometimes, even just running the netcat (nc -zvvvw2 <hostname> <thrift server port>) on the port remote can kill the entire thrift server, making it unresponsive.

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -593,6 +593,7 @@ bool TServerSocket::isUnixDomainSocket() const {
 }
 
 shared_ptr<TTransport> TServerSocket::acceptImpl() {
+try_again:
   if (serverSocket_ == THRIFT_INVALID_SOCKET) {
     throw TTransportException(TTransportException::NOT_OPEN, "TServerSocket not listening");
   }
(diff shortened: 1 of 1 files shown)


anshulmgupta:
@emmenlau, I added the comment on JIRA. Can you please take a look? Thanks!

lib/cpp/src/thrift/transport/TServerSocket.cpp
```diff
@@ -654,8 +655,15 @@ shared_ptr<TTransport> TServerSocket::acceptImpl() {
...
-    throw TTransportException(TTransportException::UNKNOWN, "accept()", errno_copy);
+    // In case of an ECONNABORTED error or any other error with the accept
+    // call retry the accept again instead of raising the exception which
+    // might kill the thrift server and hence making the thrift server
+    // unresponsive. The error can happen when accept syscall call, waiting
+    // for an incoming connection, or receiving a connection that terminates
+    // before the accept process completes, hence killing the thrift serve
+    GlobalOutput.perror("TServerSocket::acceptImpl() ::accept() Retrying ",
+                        errno_copy);
+    goto try_again;
```
CJCombrink, line 666: Generally `goto` should be avoided: [ES.76: Avoid goto](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-goto)

This looks like it can be rewritten with a simple for loop

Jens-G:
Any progress here?

Jens-G (changes requested):
@CJCombrink requested some changes here:
https://github.com/apache/thrift/pull/2964/files#r1648505818

Jens-G:
Feel free to reopen with changes appplied.

PR #2973, opened 2024-05-02
THRIFT-5784: Add THeaderTransforms to TConfiguration

fishy:
Client: go

While I'm here, also auto add compression transforms read (currently only zlib is supported) to writeTransforms so that a server will auto use the same compression on the responses as the client chose to use in the requests.

### DIFF
--- a/lib/go/thrift/header_transport.go
+++ b/lib/go/thrift/header_transport.go
@@ -151,6 +151,11 @@ func (tr *TransformReader) Close() error {
 }
 
 // AddTransform adds a transform.
+//
+// Deprecated: This only applies to the next message written, and the next read
+// message will cause write transforms to be reset from what's configured in
+// TConfiguration. For sticky transforms, use TConfiguration.THeaderTransforms
+// instead.
 func (tr *TransformReader) AddTransform(id THeaderTransformID) error {
 	switch id {
 	default:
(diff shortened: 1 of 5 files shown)


lib/go/thrift/header_transport.go
```diff
@@ -449,6 +455,11 @@ func (t *THeaderTransport) parseHeaders(ctx context.Context, frameSize uint32) e
 	}
 	t.protocolID = THeaderProtocolID(protoID)
 
+	// Reset writeTransforms to the ones from cfg, as we are going to add
+	// compression transforms from what we read, we don't want to accumulate
+	// different transforms read from different requests
+	t.writeTransforms = t.cfg.GetTHeaderTransforms()
```
fishy, line 461: a few notes here:
1. in TSimpleServer we use the same THeaderProtocol instance for incoming and outgoing, while in TStandardClient we use different THeaderProtocol instances, so this behavior should only affect servers (clients would not use the THeaderProtocol read for writing)
2. the consideration here (and the NOTE added in `AddTransform`) is for a hypothetical scenario: if a server have a long client connection, but the long client connection is actually multiple clients (e.g. it's a proxy), and those multiple clients use different compression transforms (we currently only support zlib, but the spec also defined snappy so we might support that some day in the future), we don't want to add different compression transforms cumulatively and doing both, but just use the same compression used by the request in the response. Maybe we should mark `AddTransform` as deprecated instead? the new feature in TConfiguration should be enough to replace that.

dcelasun: > Maybe we should mark AddTransform as deprecated instead?

Agreed.

fishy: Done. Also added a real test for req and resp compression.

fishy:
@dcelasun mind taking another look?

dcelasun (approved):
Looks good!

PR #2976, opened 2024-05-14
go: Proper indent in compiler

fishy:
This is a "trivial" change for go compiler to always use the combination of indent_up, indent_down, and indent, over manual indentation (by adding 2 spaces at the beginning of the string). Also change go compiler's indent_str to tab over 2 spaces.

While I'm here, also made a few minor tweaks on generated go code.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -785,46 +785,54 @@ void t_go_generator::generate_enum(t_enum* tenum) {
   std::ostringstream to_string_mapping, from_string_mapping;
   std::string tenum_name(publicize(tenum->get_name()));
   generate_go_docstring(f_types_, tenum);
-  f_types_ << "type " << tenum_name << " int64" << '\n' << "const (" << '\n';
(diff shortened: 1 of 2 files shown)


dcelasun (approved):
Great idea. Does this make `go fmt` a no-op?

fishy:
> Great idea. Does this make `go fmt` a no-op?

Not really. There are still quite a few things not there yet:

* imports need to be sorted alphabetically
* vertical alignment (for example, for field names of a struct, need to figure out the longest one and align everything else)
* proper blank lines (no two blank lines between functions, etc.)
* ...

Jens-G:
Just to mention it ... there once was a call to the net effect of system("go fmt") in the go generator so the maintainer did not have to care at all. That might explain why it is at it is. That call was later thrown out because some automated code scanner labeled it as the pure evil. Removing it was the most effective solution. So just in case someone comes up with that same idea, you can immediately give the -1 yourself. ;-)

fishy:
Thanks for the context! It might not be that bad an idea if for each language we allow the compiler to define a command to run optionally after the file is generated to format it, and configured the "official" formatter for the languages if available.

But there are definitely challenges to make it work on both windows and posix :)

PR #2977, opened 2024-05-17
THRIFT-5786: Full deprecation support in go

fishy:
Client: go

This should got most of them. Also fixed an indentation bug from 4930ca.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -768,9 +768,11 @@ void t_go_generator::generate_typedef(t_typedef* ttypedef) {
     return;
   }
 
+  generate_deprecation_comment(f_types_, ttypedef->annotations_);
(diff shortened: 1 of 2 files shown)


jimexist:
maybe you can share a code snippet with before and after when generated?

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -4286,6 +4311,32 @@ void t_go_generator::parse_go_tags(map<string,string>* tags, const string in) {
   }
 }
 
+void t_go_generator::generate_deprecation_comment(ostream& out, map<string, vector<string>>& annotations) {
```
CJCombrink, line 4314: Is there a reason why `annotations` are not const? is there an expectation that it will be modified inside this function at some time?
```suggestion
void t_go_generator::generate_deprecation_comment(ostream& out, const map<string, vector<string>>& annotations) {
```

CJCombrink (approved):

PR #2991, opened 2024-07-02
fix: make timeouterror more compatible with future versions of Python

bwangelme:
see reason in https://github.com/apache/thrift/pull/2961

> starting from python 3.10, this is a deprecated alias to TimeoutError: https://docs.python.org/3.12/library/socket.html#socket.timeout
> while TimeoutError also exists in 3.8 (the earliest version of python still supported): https://docs.python.org/3.8/library/exceptions.html#TimeoutError
> so I wonder if we should be more future proof and handle both socket.timeout and TimeoutError here, and then we can drop socket.timeout when 3.10 becomes the earliest version of python supported.
  

### DIFF
--- a/lib/py/src/transport/TSocket.py
+++ b/lib/py/src/transport/TSocket.py
@@ -151,7 +151,8 @@ class TSocket(TSocketBase):
     def read(self, sz):
         try:
             buff = self.handle.recv(sz)
-        except socket.timeout as e:
+        # TODO: remove socket.timeout when 3.10 becomes the earliest version of python supported.
+        except (socket.timeout, TimeoutError) as e:
             raise TTransportException(type=TTransportException.TIMED_OUT, message="read timeout", inner=e)
         except socket.error as e:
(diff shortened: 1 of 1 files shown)


lib/py/src/transport/TSocket.py
```diff
@@ -151,7 +151,8 @@ def open(self):
     def read(self, sz):
         try:
             buff = self.handle.recv(sz)
-        except socket.timeout as e:
+        # TODO: remove socket.timeout when 3.10 becomes the earliest version of python supported.
+        except (socket.timeout, TimeoutError) as e:
             raise TTransportException(type=TTransportException.TIMED_OUT, message="read timeout", inner=e)
         except socket.error as e:
```
fishy, line 157: similarly `socket.error` is also going to be a deprecated alias to [`OSError`](https://docs.python.org/3.12/library/exceptions.html#OSError) so we probably want to do something similar here.

fishy (approved):

PR #2993, opened 2024-07-04
netstd: fix syntax error in deepcopy() method for v0.20.0

SvenRoederer:
The ending ";" was missing when creating instance of class which caused a syntax error.

This was introduced when adding .net8 support ([4115e952b5bed2887113af053b63acd3a03c6e19](https://github.com/apache/thrift/commit/4115e952b5bed2887113af053b63acd3a03c6e19#diff-171cba564f81e70ad05350c5b6be4131a21c4a4c0b1f608954d9cb3f55ffc6deR1100))

example of generated broken code

~~~
    public CIELab DeepCopy()
    {
      var tmp5 = new CIELab()    // <----- here the closing ";" is missing
      if(__isset.L)
      {
        tmp5.L = this.L;
      }
~~~

This issue was fixed in master inside 4f1839575f3af168f960110414114255bd344203. This PR only picks the relevant fix into v0.20.0 without adding the feature of this commit.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_netstd_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_netstd_generator.cc
@@ -1103,7 +1103,7 @@ void t_netstd_generator::generate_netstd_deepcopy_method(ostream& out, t_struct*
       out << endl << indent() << "{" << endl;
       indent_up();
     } else {
-      out << endl;
+      out << ";" << endl;
     }
 
     for (m_iter = members.begin(); m_iter != members.end(); ++m_iter) {
--- a/lib/netstd/Tests/Thrift.PublicInterfaces.Compile.Tests/Thrift.PublicInterfaces.Compile.Tests.csproj
+++ b/lib/netstd/Tests/Thrift.PublicInterfaces.Compile.Tests/Thrift.PublicInterfaces.Compile.Tests.csproj
@@ -77,6 +77,7 @@
     <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./Thrift5253.thrift" />
     <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./Thrift5320.thrift" />
     <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./Thrift5382.thrift" />
+    <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./git-pr2993.thrift" />
   </Target>
 
 </Project>
--- a/lib/netstd/Tests/Thrift.PublicInterfaces.Compile.Tests/Thrift.PublicInterfaces.Compile.Tests.csproj
+++ b/lib/netstd/Tests/Thrift.PublicInterfaces.Compile.Tests/Thrift.PublicInterfaces.Compile.Tests.csproj
@@ -77,7 +77,7 @@
     <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./Thrift5253.thrift" />
     <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./Thrift5320.thrift" />
     <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./Thrift5382.thrift" />
-    <Exec Command="$(PathToThrift) -gen netstd:wcf,union,serial,net8 -r ./git-pr2993.thrift" />
+    <Exec Command="$(PathToThrift) -gen netstd -r ./git-pr2993.thrift" />
   </Target>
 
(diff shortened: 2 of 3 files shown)


Jens-G:
Testcase?

Jens-G:
image
Not so sure if that really works. Plus I still cannot repro the original issue after wasting 30 minutes on it.

SvenRoederer:
@Jens-G you looking at the wrong place. I updated the PR text,to point to the error-location. It's the 1st instruction after the method-header.

regarding testcase: I need to check what test are running for the project / how tests are organized. But I wonder why this syntax-error was not seen by tests already. As it's syntax error it pops up directly, as long ad deepcopy() is not disabled.

Jens-G:
Well, it is hard not to look at the right place. I applied your patch and got the errors shown above. The compiler stops there. So what other place would that be? that I should tell the C# compiler to look instead? 

> But I wonder why this syntax-error was not seen by tests already.

Possibly because they had the same reproducible test case as I do - still none. Please provide one or I close the PR.

SvenRoederer:
Sorry for the confusion, but I sorted it out now.
The issue only affects v0.20.0, where I found it. I fixed the code on my system for v0.20.0 and tested the generator. When adapting my fix to current master, I broke the code as pointed out by @Jens-G.
Looking in depth I realized, that current master has already a fix for the issue. The trick is, that this fix was included to [4f1839575f3af168f960110414114255bd344203](https://github.com/apache/thrift/commit/4f1839575f3af168f960110414114255bd344203#diff-171cba564f81e70ad05350c5b6be4131a21c4a4c0b1f608954d9cb3f55ffc6deR1194) ("implement full deprecation support"). I did not expect this change to fix the syntax-error, while just following the commit messages.

Finally I just updated the PR to mark the change only relevant for v0.20.0.

Jens-G:
And yet we still don't have a test case ... ok, thanks anyway for your efforts

SvenRoederer:
Yes, I put some hours in adding GitHubAction for netstd (#2995) and added a testcase with separate commit. 
I made it also on top of v0.20.0 to see that it actually detects the problem seen.

Feel free to cherry-pick these commits into master.


Is it planned to make a v0.20.1 for this fix  or will this version stay not-usable for us and we need to wait for 0.21.0?

Jens-G:
> Is it planned to make a v0.20.1 for this fix or will this version stay not-usable for us and we need to wait for 0.21.0?

I integrated the changes for convenience into 0.20.0, however there will be most likely no 0.20.1 release just because of this. There is a workaround, you have a patch and it is all open source. The Windows compiler EXE is a convenience build anyway. 

PS: Unusable to me is when an update of some closed-source OS leads you into a state where the entire docker engine effectively and reproducibly stalls when you try to run two docker compose sets on it, and you can't get out of that lockup state until you reboot the entire machine and waste four hours on trying to spot the issue in your own code - oh wait, exactly that just happened (still unsolved BTW).

SvenRoederer:
Just seen that @Jens-G created a Jira issue for this. 
I squashed all commits into a sinlge one and added updated references to THRIFT-5794. Feel free to cherry-pick the result to master (https://github.com/SvenRoederer/thrift/commit/06d37f239b153853aa3fc80b3daca10ed4242af9)

raulcd:
> I integrated the changes for convenience into the 0.20.0 branch

FYI for the future, maybe you don't want to act on it as this is a valid workflow for thrift but merging this to an already released branch has caused some upstream issues: https://github.com/microsoft/vcpkg/pull/39787 and for the Apache Arrow project: https://github.com/apache/arrow/issues/43204

raulcd:
Sorry @Jens-G this was not my intent, it was only to let know the Thrift committers and to make them aware. I as an Apache Arrow PMC would like to know if this happens to my project.

Jens-G:
I really wonder, if this is whole thing so super-urgent and critical an issue, why on earth has nobody filed a ticket? The 0.20.0 relaese is out for months now. This PR is the first ever that even tried to address/mention the issue. So how urgent can it really be? And why am I responsible for someone who wants to merge that commit elsewhere? And how should I know all of this everybody only comes around and finds out a way to communicate **after** the problem manifests? I should have closed this from the beginning, would have upset someone else (= no change to above) but saved me a lot of my own time.

Jens-G:
> would like to know if this happens to my project.

Appreciated. You are only the messenger. Since that tag did not change I wonder why that SHA changed? I have to admit that I did not read all of it.

SvenRoederer:
@Jens-G @raulcd sorry for the trouble and frustration caused by this fix.

But I agree that it's the best solution that other projects rely on release tags or release packages instead of HEAD of branches. At least vcpkg will be future-save with the fix they implemented. 
As a sidenote: Conan is using "http://archive.apache.org/dist" to avoid similar problems.

SvenRoederer:
final test for this added by Github PR #3000

PR #2994, opened 2024-07-06
CI bump versions of GitHubActions and Ubuntu

SvenRoederer:
upgrade artifact-actions from v3 to v4
* v3 uses Node.js 16, which is deprecated and causes warning messages
* < v4 will get deprecated in Nov 2024 (https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)

switch to ubuntu-22.04 by default
* Ubuntu 20.04 will go EoL in Aril 2025
* build the compiler on ubuntu-24.04, ubuntu-22.04 and ubuntu-20.04

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,7 +78,7 @@ jobs:
 
   lib-php:
     needs: compiler
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         php-version: [7.1, 7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3]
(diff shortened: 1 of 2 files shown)


sveneld:
It will be great to add a dockerFile for Ubuntu 24.04 in folder build/docker, not just in github actions.

jimexist (approved):

jimexist:
how about just dropping 20.04?

sveneld:
Ubuntu 20.04 is still active, my opinion is we can remove its support only after EOL.

And I suggest you add support for Ubuntu 24.04 instead of removing support for Ubuntu 20.04

PR #2995, opened 2024-07-06
make GitHubActions also run tests for netstd

SvenRoederer:
Add a job to build and run tests for netstd. Also upload the binaries as artifact.
This gives basic testing, but leaves including to cross-test for the future.

Most job-steps are copied from the other jobs.

In relation to #2994 it already uses artifact-action v4 and ubuntu-22.04 (ubuntu-20.04 has issues with dotnet-support).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,7 +103,7 @@ jobs:
         run: |
           ./configure $(echo $CONFIG_ARGS_FOR_LIBS | sed 's/without-php/with-php/' | sed 's/without-php_extension/with-php_extension/' )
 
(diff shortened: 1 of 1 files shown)


SvenRoederer:
This change was made to run tests for #2993 issue (which only affects v0.20.0) and therefore base on v0.20 branch. The current PR also relies on some changes of #2994 ans it should probably merged after it.
If it's more convenient to merge this PR before #2994 I can update it to apply on master directly as a single commit with linear history.

Jens-G:
@SvenRoederer could you send another PR with the added tests only targeting `master` branch?

PR #2998, opened 2024-07-10
make GitHubActions also run tests for netstd (updated)

SvenRoederer:
Add a job to build and run tests for netstd. Also upload the binaries as artifact.
This gives basic testing, but leaves including to cross-test for the future.

Most job-steps are copied from the other jobs.

In relation to https://github.com/apache/thrift/pull/2994 it already uses ubuntu-22.04 (ubuntu-20.04 has issues with dotnet-support).

This PR supersedes #2995 to apply straight forward on master.


SvenRoederer:
Test failing with "python_test" for msvc builds - quite sure this does not relate to to the GitHubAction changes here.

.github/workflows/build.yml
```diff
@@ -297,6 +297,67 @@ jobs:
...
+
+      - name: Run make check for test/netstd
+        run: make -C test/netstd check
+
+      - name: Run make precross for test/netstd
+        run: make -C test/netstd precross
+
+      - name: Upload netstd precross artifacts
+        uses: actions/upload-artifact@v3
+        with:
```
Jens-G, line 353: https://github.com/actions/upload-artifact is scheduled for deprecation on November 30, 2024

SvenRoederer: upgrading this is part of #2994 (89506d34b3901483f7f78d0ea22d92ef090f44bf), as all uses need to be of the same version to work with each other.

SvenRoederer: ... but when after merging the above PR, this commit needs to be updated to `upload-artifact@v3`. I can change and push here

SvenRoederer: done. all artifact-actions are v4 now.

PR #3003, opened 2024-07-18
CI/appveyor: fix download location for flex-sdk

SvenRoederer:
Current build seem to fail on downloading flex-SDK. Use new location according to https://stackoverflow.com/a/21269691/19761878


SvenRoederer:
For the moment the current URL is fine (again), so download issues might have been caused by a glitch in the internet. But probably worth to keep this URL in mind a alternative.

PR #3007, opened 2024-07-19
lib: py: remove distutils in favor of setuptools

pyrox0:
As distutils is removed in python 3.12, and setuptools provides the same interface, it can be switched in place with only minor changes. This allows using the thrift bindings on versions of Python 3.12 or greater, and in python environments without distutils installed(such as NixOS, which does not expose setuptools' internal distutils distribution)

- [ x Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?

### DIFF
--- a/lib/py/setup.py
+++ b/lib/py/setup.py
@@ -25,8 +25,8 @@ try:
 except Exception:
     from distutils.core import setup, Extension
 
-from distutils.command.build_ext import build_ext
-from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
+from setuptools.command.build_ext import build_ext
+from setuptools.errors import CCompilerError, ExecError, PlatformError
 
 # Fix to build sdist under vagrant
 import os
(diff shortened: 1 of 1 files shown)


ctubbsii (commented):
This is a small incremental improvement. However, it should also be done with other modernizations, like those described in https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

Jens-G:
Breaks build: https://ci.appveyor.com/project/ApacheSoftwareFoundation/thrift/builds/53103767
Reverted via ead302baf8

pyrox0:
The CI runs on python 3.5(which has been deprecated since sep 2020) which does not have support for setuptools 59.0.0(released nov 2021), which is why this fails. CI even seems to use swtuptools 28.8.0, which was released in 2016!

pyrox0:
The CI runs on python 3.5(which has been deprecated since sep 2020) which does not have support for setuptools 59.0.0(released nov 2021), which is why this fails. CI even seems to use swtuptools 28.8.0, which was released in 2016!

PR #3010, opened 2024-07-25
THRIFT-4847: CancelledKeyException causes TThreadedSelectorServer to fail

mrcjkb:
Adds validity checks before trying to use the `SelectionKey` in `AbstractNonBlockingServer.FrameBuffer`.

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
+++ b/lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
@@ -248,7 +248,7 @@ public abstract class AbstractNonblockingServer extends TServer {
     protected final TNonblockingTransport trans_;
 
     // the SelectionKey that corresponds to our transport
-    protected final SelectionKey selectionKey_;
+    protected SelectionKey selectionKey_;
 
     // the SelectThread that owns the registration of our transport
     protected final AbstractSelectThread selectThread_;
@@ -300,6 +300,14 @@ public abstract class AbstractNonblockingServer extends TServer {
       } else {
         context_ = null;
       }
+    }
+
+    /**
+     * Sets the selection key (this is not thread safe).
+     * @param selectionKey the new key to set.
+     */
+    public void setSelectionKey(SelectionKey selectionKey) {
+      selectionKey_ = selectionKey;
     }
 
     /**
@@ -375,7 +383,11 @@ public abstract class AbstractNonblockingServer extends TServer {
         // modify our selection key directly.
         if (buffer_.remaining() == 0) {
           // get rid of the read select interests
-          selectionKey_.interestOps(0);
+          if (selectionKey_.isValid()) {
+            selectionKey_.interestOps(0);
+          } else {
+            LOGGER.warn("SelectionKey was invalidated during read");
+          }
           state_ = FrameBufferState.READ_FRAME_COMPLETE;
         }
 
@@ -415,8 +427,12 @@ public abstract class AbstractNonblockingServer extends TServer {
       switch (state_) {
         case AWAITING_REGISTER_WRITE:
           // set the OP_WRITE interest
-          selectionKey_.interestOps(SelectionKey.OP_WRITE);
-          state_ = FrameBufferState.WRITING;
+          if (selectionKey_.isValid()) {
+            selectionKey_.interestOps(SelectionKey.OP_WRITE);
+            state_ = FrameBufferState.WRITING;
+          } else {
+            LOGGER.warn("SelectionKey was invalidated before write");
+          }
           break;
         case AWAITING_REGISTER_READ:
           prepareRead();
@@ -520,7 +536,11 @@ public abstract class AbstractNonblockingServer extends TServer {
     private void prepareRead() {
       // we can set our interest directly without using the queue because
       // we're in the select thread.
-      selectionKey_.interestOps(SelectionKey.OP_READ);
+      if (selectionKey_.isValid()) {
+        selectionKey_.interestOps(SelectionKey.OP_READ);
+      } else {
+        LOGGER.warn("SelectionKey was invalidated before read");
+      }
       // get ready for another go-around
       buffer_ = ByteBuffer.allocate(4);
       state_ = FrameBufferState.READING_FRAME_SIZE;
--- a/lib/java/src/main/java/org/apache/thrift/server/TThreadedSelectorServer.java
+++ b/lib/java/src/main/java/org/apache/thrift/server/TThreadedSelectorServer.java
@@ -626,16 +626,23 @@ public class TThreadedSelectorServer extends AbstractNonblockingServer {
         LOGGER.error("Create new Selector error.", e);
       }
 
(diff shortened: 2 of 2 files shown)


lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
```diff
@@ -518,9 +522,11 @@ private boolean internalRead() {
...
-      // we're in the select thread.
-      selectionKey_.interestOps(SelectionKey.OP_READ);
+      if (selectionKey_.isValid()) {
+        // we can set our interest directly without using the queue because
+        // we're in the select thread.
+        selectionKey_.interestOps(SelectionKey.OP_READ);
+      }
       // get ready for another go-around
       buffer_ = ByteBuffer.allocate(4);
       state_ = FrameBufferState.READING_FRAME_SIZE;
```
mrcjkb, line 532: Not sure if the rest of `prepareRead` should go in the conditional block :thinking:

mrcjkb:
CI failure looks unrelated (?)

OneSizeFitsQuorum:
Relatively complete solution exists in https://github.com/apache/thrift/pull/2171, but I don't know why this PR was Close off

mrcjkb:
> Relatively complete solution exists in #2171, but I don't know why this PR was Close off

Looks like it was never reviewed and closed automatically.

Jens-G:
Reopened. So this is now a duplicate?

OneSizeFitsQuorum:
> now a duplicate?

I believe so~

mrcjkb:
Yes, this can be closed as duplicate.
Is there anything that needs to be done to get #2171 merged?

mrcjkb:
Actually, it might still be beneficial to add the `isValid` checks in this PR, in addition to the fix in #2171.

mrcjkb:
Since @zengji doesn't appear to be active anymore, I've incorporated their changes here.

lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
```diff
@@ -302,6 +302,14 @@ public FrameBuffer(
       }
     }
 
+    /**
+     * Sets the selection key.
+     * @param selectionKey the new key to set.
+     */
+    public void setSelectionKey(SelectionKey selectionKey) {
```
mrcjkb, line 309: Not too sure about this. It feels a bit hacky. Maybe someone who is familiar with the code base can suggest a better solution?

OneSizeFitsQuorum:
> Actually, it might still be beneficial to add the `isValid` checks in this PR, in addition to the fix in #2171.

This is a bit like defensive programming, which is good for stability

lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
OneSizeFitsQuorum: I checked this part of the code, and in the current Reactor architecture, all the operations of the selector thread will only happen in the current thread, Inspecting the code shows that only the selector thread processes the selectionKey_ in the frameBuffer and reads and writes to transport, so there are no read or write requests in progress when rebuilding and no visibility issues (you don't need to change SelectionKey to AtomicReference when you strip it of final)

lib/java/src/main/java/org/apache/thrift/server/TThreadedSelectorServer.java
```diff
@@ -626,17 +626,23 @@ private synchronized void rebuildSelector() {
         LOGGER.error("Create new Selector error.", e);
       }
 
-      for (SelectionKey key : oldSelector.selectedKeys()) {
-        if (!key.isValid() && key.readyOps() == 0) continue;
+      for (SelectionKey key : oldSelector.keys()) {
+        if (!key.isValid() || key.channel().keyFor(newSelector) != null) continue;
```
OneSizeFitsQuorum, line 630: if (!key.isValid() || key.interestOps() == 0 || key.channel().keyFor(newSelector) != null)

lib/java/src/main/java/org/apache/thrift/server/TThreadedSelectorServer.java
```diff
@@ -626,17 +626,23 @@ private synchronized void rebuildSelector() {
...
-            channel.register(newSelector, key.readyOps());
+            newKey = channel.register(newSelector, interestOpts);
           } else {
-            channel.register(newSelector, key.readyOps(), attachment);
+            newKey = channel.register(newSelector, interestOpts, attachment);
           }
+          if (attachment instanceof FrameBuffer) {
+            ((FrameBuffer) attachment.setSelectionKey(newKey);
+          }
+          key.cancel();
```
OneSizeFitsQuorum, line 645: Seems redundant  as we will call oldSelector.close() later

OneSizeFitsQuorum (commented):
Leave a few thoughts, welcome to discuss, let's move forward with this PR

lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
mrcjkb: My concern is that both `FrameBuffer` and `setSelectionKey` are public, so this could be a footgun in the future.
But based on your comment, I'd say YAGNI for now. I've added a note that it's not tread safe to the javadoc.

lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
```diff
@@ -518,9 +534,13 @@ private boolean internalRead() {
 
     /** We're done writing, so reset our interest ops and change state accordingly. */
     private void prepareRead() {
-      // we can set our interest directly without using the queue because
-      // we're in the select thread.
-      selectionKey_.interestOps(SelectionKey.OP_READ);
+      if (selectionKey_.isValid()) {
+        // we can set our interest directly without using the queue because
```
OneSizeFitsQuorum, line 538: move these comments to line 537?

lib/java/src/main/java/org/apache/thrift/server/TThreadedSelectorServer.java
```diff
@@ -626,16 +626,23 @@ private synchronized void rebuildSelector() {
...
+        SelectionKey newKey;
         try {
           if (attachment == null) {
-            channel.register(newSelector, key.readyOps());
+            newKey = channel.register(newSelector, interestOpts);
           } else {
-            channel.register(newSelector, key.readyOps(), attachment);
+            newKey = channel.register(newSelector, interestOpts, attachment);
+          }
+          if (attachment instanceof FrameBuffer) {
```
OneSizeFitsQuorum, line 644: maybe we can move to line 643 as we do not need do anything when attachment is null

OneSizeFitsQuorum:
@jeking3 Could you please review this? Many thanks!

mrcjkb:
Thanks for the reviews :pray: :smile: 

@Jens-G is there an ETA on the next release?

PR #3011, opened 2024-07-25
THRIFT-5800: Make non-exist include a failure when -strict is on

fishy:


compiler/cpp/src/thrift/main.cc
```diff
@@ -373,7 +373,12 @@ string include_file(string filename) {
   }
 
   // Uh oh
-  pwarning(0, "Could not find include file %s\n", filename.c_str());
+  if (g_strict >= 192) {
```
dcelasun, line 376: Shouldn't this be just 255? https://github.com/apache/thrift/blob/3d02c0f94e858828e9251a57979172ae48d71b9d/compiler/cpp/src/thrift/main.cc#L1115-L1116

fishy: so I did a `git grep -w g_strict` and the only place we check the value is this:
```
$ git grep -w g_strict
src/thrift/globals.h:extern int g_strict;
src/thrift/main.cc:int g_strict = 127;
src/thrift/main.cc:  if (g_strict >= 192) {
src/thrift/main.cc:        g_strict = 255;
src/thrift/thrifty.yy:        if (g_strict >= 192) {
```
so I opted with follow the existing convention :)

dcelasun (approved):

Jens-G (approved):

Jens-G:
Merge?

PR #3012, opened 2024-07-28
THRIFT-4386 Add Lua 5.3/5.4 support

thomasbruggink:
This PR adds support for running with Lua 5.3 and 5.4 while also keeping backwards compatibility with Lua 5.1.
Updated ubuntu-jammy dockerfile to install Lua 5.4 for testing.

Fixed cross-testing in a [follow up PR](https://github.com/thomasbruggink/thrift/commit/f42bc81f3945834531d8080bb941393d5717e643) to confirm this works with for both Lua versions with the Java client.

### DIFF
--- a/lib/lua/Thrift.lua
+++ b/lib/lua/Thrift.lua
@@ -23,6 +23,9 @@
 --setfenv(1, thrift)
 
 package.cpath = package.cpath .. ';bin/?.so' -- TODO FIX
+
+local libluabitwise = require 'libluabitwise'
+
 function ttype(obj)
   if type(obj) == 'table' and
     obj.__type and
@@ -66,8 +69,7 @@ TType = {
   MAP    = 13,
   SET    = 14,
   LIST   = 15,
-  UTF8   = 16,
-  UTF16  = 17
+  UUID   = 16
 }
 
 TMessageType = {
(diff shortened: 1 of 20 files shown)


thomasbruggink:
@Jens-G do you know anyone who might be able to review this?

If there is currently no active Lua maintainer would it also be acceptable to have the cross tests pass to proof the existing implementation works as expected? If yes I'll move the cross test fixes into this PR :)

Jens-G:
+1 agreed

thomasbruggink:
Added 4 more commits to ensure cross testing works as expected.
UUID tests are required to confirm anything so I added THRIFT-5804 to this PR as well.

The follow cross tests now work:
 - lua client -> lua server (100%)
 - java client -> lua server (100%)
 - netstd -> lua server (66%)

The failing netstd tests are the result of 15mb Json data which also times out on netstd client -> netstd server.

Note that to get Http working I needed to slightly modify the oneway implementation in the compiler and added a new `flushOneway` method. There is a default NOOP handler in `TTransportBase` which ensures that custom implementations of `TTransportBase` will keep working, but users that upgrade to the thrift version that contains this commit will also need to update their `TTransportBase.lua`.

lib/lua/Thrift.lua
```diff
@@ -233,6 +235,35 @@ function TException:write(oprot)
...
+}
+
+TUUID = __TObject:new{
+  __type = 'TUUID'
+}
+
+function TUUIDfromString(str)
+  local iterator = string.gmatch(str, "[A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9]")
+  return TUUID:new {
+    zero = libluabitwise.buor(libluabitwise.ushiftl(tonumber(iterator(), 16), 16), tonumber(iterator(), 16)),
```
thomasbruggink, line 252: note that bitshifting was added in Lua 5.3, so if support for Lua 5.2 is dropped this implementation and the entire `libluabitwise` C module can be deleted.

thomasbruggink:
> Whats the status with this? Still work in progress?

sorry for the delayed response.
The current status is already good enough to merge. It fixes both THRIFT-4386 and THRIFT-5804.

All tests still work for Java<->Lua. I was looking into getting Swift to work but there is some strange network connection problem that caused it to fail. I'm still looking into this when time is available.

@Jens-G do you want Java<->Lua tests added as github actions? Otherwise I think its ok to merge as is :bow:

Jens-G:
> do you want Java<->Lua tests added as github actions?

What should I say when you ask this way .... 😆  
Can be done separately.

thomasbruggink:
> What should I say when you ask this way .... 😆

feel free to require this to be in the PR before accepting it to be merged ^^

let me follow up with a github action PR then :bow:

PR #3014, opened 2024-07-31
Fix JS property/method collisions for fields named read/write

adrianhelvikspond:
If a thrift struct has a member named `read`, it causes a naming collision. This change renames `read` to `thriftRead` and `write` to `thriftWrite` on generated JavaScript classes.

I reviewed the NodeJS library code to see if read or write was called directly on the generated classes:

<details>
<summary>`read` calls in NodeJS library</summary>

```
❯ rg '\.read\('
binary_protocol.js
193:    name = this.trans.read(sz);
271:  var buff = this.trans.read(8);
288:  return this.trans.read(len);

compact_protocol.js
727:  var buff = this.trans.read(8);
769:  return this.trans.read(size);

header_transport.js
280:  this.read(endOfHeaders);
```
</details>

So `.read` is only called on the transports.

<details>
<summary>write calls in NodeJS library</summary>

```
server.js
49:            stream.write(buf); // TCP connection

connection.js
187:    self.write(data);
^ Connection.prototype
196:  this.connection.write(data);
^ Connection.prototype
387:    self.write(data);
^ StdIOConnection.prototype
396:  this.connection.write(data);
^ StdIOConnection.prototype

http_connection.js
227:  req.write(data);
^ The http request

multiplexed_protocol.js
56:    connection.write(buf,seqid);
^ Third argument `connection` on Multiplexer.prototype.createClient. Assume this is Connection

compact_protocol.js
353:  this.trans.write(new Buffer([b]));
428:  this.trans.write(buff);
434:    this.trans.write(new Buffer(arg, encoding));
441:    this.trans.write(arg);
505:  this.trans.write(wbuf);
539:  this.trans.write(wbuf);
^ Transport only

header_transport.js
89:  this.transport.write(str);
316:  varintWriter.write(paddingBuffer);
^ Transport and VarIntHelper

binary_protocol.js
130:  this.trans.write(new Buffer([b]));
134:  this.trans.write(binary.writeI16(new Buffer(2), i16));
138:  this.trans.write(binary.writeI32(new Buffer(4), i32));
143:    this.trans.write(i64.buffer);
145:    this.trans.write(new Int64(i64).buffer);
150:  this.trans.write(binary.writeDouble(new Buffer(8), dub));
156:    this.trans.write(new Buffer(arg, encoding));
163:    this.trans.write(arg);
^ Transport only

create_client.js
47:    connection.write(buf, seqid);
^ Connection

web_server.js
453:        response.write(file, "binary");
^ Http response
468:          socket.write(frame);
^ Websocket
518:      socket.write("HTTP/1.1 403 No Apache Thrift Service available\r\n\r\n");
^ Http socket
524:    socket.write("HTTP/1.1 101 Switching Protocols\r\n" +
^ Http socket

json_protocol.js
99:    this.trans.write(this.tstack.pop());
125:  this.trans.write(this.wbuf);
^ Transport
```
</details>

So `.write` is not called on generated classes here either.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_js_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_js_generator.cc
@@ -1061,10 +1061,10 @@ void t_js_generator::generate_js_struct_reader(ostream& out, t_struct* tstruct) 
   vector<t_field*>::const_iterator f_iter;
 
   if (gen_es6_) {
-    indent(out) << "read (input) {" << '\n';
+    indent(out) << "[Symbol.for(\"read\")] (input) {" << '\n';
   } else {
     indent(out) << js_namespace(tstruct->get_program()) << tstruct->get_name()
-        << ".prototype.read = function(input) {" << '\n';
+        << ".prototype[Symbol.for(\"read\")] = function(input) {" << '\n';
   }
 
   indent_up();
@@ -1151,10 +1151,10 @@ void t_js_generator::generate_js_struct_writer(ostream& out, t_struct* tstruct) 
   vector<t_field*>::const_iterator f_iter;
 
   if (gen_es6_) {
-    indent(out) << "write (output) {" << '\n';
+    indent(out) << "[Symbol.for(\"write\")] (output) {" << '\n';
   } else {
     indent(out) << js_namespace(tstruct->get_program()) << tstruct->get_name()
-        << ".prototype.write = function(output) {" << '\n';
+        << ".prototype[Symbol.for(\"write\")] = function(output) {" << '\n';
   }
 
   indent_up();
@@ -1395,7 +1395,7 @@ void t_js_generator::generate_service_processor(t_service* tservice) {
                 "Thrift.TApplicationException(Thrift.TApplicationExceptionType.UNKNOWN_METHOD, "
                 "'Unknown function ' + r.fname);" << '\n' << indent()
              << "  output.writeMessageBegin(r.fname, Thrift.MessageType.EXCEPTION, r.rseqid);"
-             << '\n' << indent() << "  x.write(output);" << '\n' << indent()
+             << '\n' << indent() << "  x[Symbol.for(\"write\")](output);" << '\n' << indent()
              << "  output.writeMessageEnd();" << '\n' << indent() << "  output.flush();" << '\n'
              << indent() << "}" << '\n';
 
@@ -1447,7 +1447,7 @@ void t_js_generator::generate_process_function(t_service* tservice, t_function* 
                       + "_result";
 
   indent(f_service_) << js_const_type_ << "args = new " << argsname << "();" << '\n' << indent()
-             << "args.read(input);" << '\n' << indent() << "input.readMessageEnd();" << '\n';
+             << "args[Symbol.for(\"read\")](input);" << '\n' << indent() << "input.readMessageEnd();" << '\n';
 
   // Generate the function call
   t_struct* arg_struct = tfunction->get_arglist();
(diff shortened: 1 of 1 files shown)


Jens-G:
>  This change renames read to thriftRead and write to thriftWrite on generated JavaScript classes.

Ok, but what if 

> a thrift struct has a membe named 

`thriftRead` or `thriftWrite`?

adrianhelvikspond:
It absolutely could, but this would make it less likely. An alternative approach could be to use symbols instead.

```javascript
  [Symbol.for("read")] (input) {
    // ...
  }

  [Symbol.for("write")] (output) {
    // ...
  }
```

This would prevent this problem entirely.

adrianhelvikspond:
I've updated the PR to use symbols instead so there is zero chance of name collisions.

Jens-G:
Ok, nice. How does that runtime lookup affect performance?

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for) the symbols are global and the recommendaion is to prefix them.  Especially with names like "read" or "write" that sounds like a potential problem to me. What do you think?

adrianhelvikspond:
**1. How does that runtime lookup affect performance?**

I did a quick benchmark here, and using a pre-constructed symbol was equally fast as using strings, while constructing the symbol on the fly like I did take about 10s/billion calls.

<details>
<summary>Benchmark code</summary>
const ONE_BILLION = 1_000_000_000;

const uniqueSymbol = Symbol.for("read");
const symbol = Symbol("read")

const symbolEntity = {
	[Symbol.for("read")]() {
		return "Hello world";
	},
};

const constSymbolEntity = {
	[symbol]() {
		return "Hello world";
	},
};

const constUniqueSymbolEntity = {
	[uniqueSymbol]() {
		return "Hello world";
	},
};

const stringEntity = {
	read: () => {
		return "Hello world";
	},
};

for (let i = 0; i < 100; i++) {
	time("symbol constructed every time", () => {
		for (let i = 0; i < ONE_BILLION; i++) {
			symbolEntity[Symbol.for("read")]();
		}
	});
	time("const symbol", () => {
		for (let i = 0; i < ONE_BILLION; i++) {
			constSymbolEntity[symbol]();
		}
	});
	time("const unique symbol", () => {
		for (let i = 0; i < ONE_BILLION; i++) {
			constUniqueSymbolEntity[uniqueSymbol]();
		}
	});
	time("string", () => {
		for (let i = 0; i < ONE_BILLION; i++) {
			stringEntity["read"]();
		}
	});
}

function time(name, fn) {
	const start = process.hrtime.bigint();
	fn();
	const end = process.hrtime.bigint();
	console.log(name, format(end - start));
}

function format(time) {
	// Nanos to micros
	time /= 1000n;

	if (time < 1000n) {
		return `${time}µs`;
	}

	// Micros to millis
	time /= 1000n;

	return `${time}ms`;
}
</details>

```
symbol constructed every time 9876ms
const symbol 471ms
const unique symbol 486ms
string 478ms
```

So while I doubt it's going to be the performance bottleneck of most applications, it's easy enough to store the symbol in a variable before using it. So I'll fix that.

Regarding prefixing, the way JavaScript introduces standardised symbols is that they use `Symbol("...")` (globally unique) instead of `Symbol.for("...")` (unique to the passed in string).

`Symbol.iterator` or `Symbol.asyncIterator` are examples of that. You categorically cannot construct these symbols yourself. So while somebody may create a symbol that is equal to `Symbol.for("read")` it will never be a problem. I'd liken it to somebody subclassing a Java thrift class and overriding the `.read` method vs not being able to do that. Possibly useful to be able to override the method for testing purposes, but not in any other case.

Security-wise I'd argue that symbols are a teeny tiny bit safer than strings as you cannot deserialise a value into a symbol with any of the thrifty serialisers or even JSON. But there's no real difference. Symbols are essentially just a different class of strings that can be globally unique if you use `Symbol("...")` instead of `Symbol.for("...")`.

cameron-martin:
This seemed to break the NodeJS tests (`lib/nodejs/test/testAll.sh`). Note these do not currently run on CI, which is why this likely slipped through the net. I'm hoping to put up a PR to make these run on CI, but in the meantime can this be reverted?

PR #3017, opened 2024-08-13
Resolve the occupation of java transmission object content

afuaide:
The content of the transmitted object will not be immediately released after the request is processed. The object content will only be released in the next request. If the next request requires a long wait, the object will continue to occupy memory until a new request arrives.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_java_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_java_generator.cc
@@ -3328,6 +3328,7 @@ void t_java_generator::generate_service_client(t_service* tservice) {
       // Careful, only return _result if not a void function
       if (!(*f_iter)->get_returntype()->is_void()) {
         f_service_ << indent() << "if (result." << generate_isset_check("success") << ") {" << '\n'
+                   << indent() << "  after();" << '\n'
                    << indent() << "  return result.success;" << '\n'
                    << indent() << "}" << '\n';
       }
--- a/lib/java/src/main/java/org/apache/thrift/TServiceClient.java
+++ b/lib/java/src/main/java/org/apache/thrift/TServiceClient.java
@@ -22,6 +22,8 @@ package org.apache.thrift;
 import org.apache.thrift.protocol.TMessage;
 import org.apache.thrift.protocol.TMessageType;
 import org.apache.thrift.protocol.TProtocol;
+import org.apache.thrift.transport.TTransport;
+import org.apache.thrift.transport.layered.TFramedTransport;
 
 /**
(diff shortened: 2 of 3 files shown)


ctubbsii (dismissed):
This looks okay to me, but the commit message needs a much better description of what this change is doing. "make buffer gc early" is not really a very good message, nor is "optimize gc" or "optimize buff". It also seems to fix a compilation bug introduced in d4503a1f0e62d38f7a12ba9748744c5490c915c4

lib/java/src/main/java/org/apache/thrift/TServiceClient.java
```diff
@@ -93,4 +95,12 @@ protected void receiveBase(TBase<?, ?> result, String methodName) throws TExcept
     result.read(iprot_);
     iprot_.readMessageEnd();
   }
+
+  protected void after() {
+    TTransport transport = this.iprot_.getTransport();
+    if (transport instanceof TFramedTransport) {
+      TFramedTransport framedTransport = (TFramedTransport) transport;
+      framedTransport.clear();
```
ctubbsii, line 103: This can be simplified

```suggestion
      ((TFramedTransport) transport).clear();
```

ctubbsii:
I resolved the merge conflicts in this, but it needs a better commit message/description, and a second pair of eyes to see if the PR is needed at all. Without a better description, I'm not really sure what this is intended to solve.

afuaide:
> I resolved the merge conflicts in this, but it needs a better commit message/description, and a second pair of eyes to see if the PR is needed at all. Without a better description, I'm not really sure what this is intended to solve.

@ctubbsii Updated

compiler/cpp/src/thrift/generate/t_java_generator.cc
```diff
@@ -3328,6 +3328,7 @@ void t_java_generator::generate_service_client(t_service* tservice) {
       // Careful, only return _result if not a void function
       if (!(*f_iter)->get_returntype()->is_void()) {
         f_service_ << indent() << "if (result." << generate_isset_check("success") << ") {" << '\n'
+                   << indent() << "  after();" << '\n'
```
ctubbsii, line 3331: Wouldn't you want this to be called even if it is a void function? Adding this call here only calls it when it isn't a void function.

afuaide: My previous idea was that it would only be called when the data is returned

afuaide: If the contents of the byte array are confirmed not to be used again after this method is called, it can be called directly

PR #3019, opened 2024-08-14
Fix failed java build

fishy:
I think this was a bug from d4503a1f0e62d38f7a12ba9748744c5490c915c4.

  

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
+++ b/lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
@@ -304,6 +304,7 @@ public abstract class AbstractNonblockingServer extends TServer {
 
     /**
      * Sets the selection key (this is not thread safe).
+     *
      * @param selectionKey the new key to set.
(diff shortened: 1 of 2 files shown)


ctubbsii (approved):

ctubbsii:
It looks like that commit also broke something else which is fixed in https://github.com/apache/thrift/pull/3017/files#diff-dac985b5f3c62d122459132ffb79349f124eaf7f54eea6eba1d3b9d19b4d6a33R640-R644

fishy:
> It looks like that commit also broke something else which is fixed in https://github.com/apache/thrift/pull/3017/files#diff-dac985b5f3c62d122459132ffb79349f124eaf7f54eea6eba1d3b9d19b4d6a33R640-R644

Thanks. Since that PR is not merged yet, I'm going to incorporate that part of the fix into this PR instead.

fishy:
cc @emmenlau I listed you as co-author as you did the part of fix in 1a4bf78ad09841c3c84bfed7826b871f4dfc9876

PR #3021, opened 2024-08-14
go: Fix cross test server

fishy:
Client: go

This was a bug introduced by 91565d490e98306ac6797dd6ed4f72c0e8222e78 that broke go's cross-test server, but because other CI issues we didn't run cross-test so we didn't notice the issue.

### DIFF
--- a/test/go/src/bin/testserver/main.go
+++ b/test/go/src/bin/testserver/main.go
@@ -59,7 +59,10 @@ func main() {
 				return
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


fishy:
testing in https://github.com/apache/thrift/actions/runs/10394909931?pr=3020

fishy:
all the go-go cross tests are passing now here: https://github.com/apache/thrift/actions/runs/10395209364/job/28787065515?pr=3020

PR #3030, opened 2024-08-28
remove dependency on Six

a-detiste:
https://wiki.debian.org/Python3-six-removal

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -483,7 +483,7 @@ jobs:
 
       - name: Python setup
         run: |
-          python -m pip install --upgrade pip setuptools wheel flake8 six tornado twisted zope.interface
+          python -m pip install --upgrade pip setuptools wheel flake8 tornado twisted zope.interface
           python --version
           pip --version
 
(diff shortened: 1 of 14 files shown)


fishy:
There is an open ticket (https://issues.apache.org/jira/browse/THRIFT-5537) and open PRs to drop Python 2 support, which is not done yet.

I don't think we can remove the dependency of six without removing python 2 support.

Jens-G:
Unfortunately #2588 seems a bit rotten ... it still features `endl`and the number of whitespace changes is as huge as the time when auto-formatting was the new hot thing 🤣 so that seems not an easy task. Any ideas how to solve that mess?

a-detiste:
>  solve that mess?

Well, first declare Python2 officialy unsupported and let vultures tear it appart apiece ;-)

From what I've seen the "one mega PR" approach usualy never completes. At some point people suggests to add typign annotation ;-)

fishy (approved):
>Well, first declare Python2 officialy unsupported and let vultures tear it appart apiece ;-)

sounds good to me :)

fishy:
we probably should consider this PR part of THRIFT-5537 (add that to PR/commit message title)?

jimexist:
would love to see python2 removed but don't have time or bandwidth to actually test it to make it solid.
- https://github.com/apache/thrift/pull/2588
if anyone's interested feel free to take over

PR #3035, opened 2024-09-03
THRIFT-5816: Fix UUID for boost 1.86.0 (change in {{data}} member usage)

CJCombrink:
Tries to fix compiler issues with latest boost (1.86)

 [THRIFT-5816](https://issues.apache.org/jira/browse/THRIFT-5816)

### DIFF
--- a/lib/c_glib/test/testthrifttestclient.cpp
+++ b/lib/c_glib/test/testthrifttestclient.cpp
@@ -112,8 +112,8 @@ class TestHandler : public ThriftTestIf {
     out = thing;
   }
 
-  std::string testUuid(const std::string thing) override {
-    cout << "[C -> C++] testUuid(\"" << std::hex << thing << "\")" << '\n';
+  apache::thrift::TUuid testUuid(const apache::thrift::TUuid thing) override {
+    cout << "[C -> C++] testUuid(\"" << thing << "\")" << '\n';
     return thing;
   }
 
--- a/lib/c_glib/test/testthrifttestzlibclient.cpp
+++ b/lib/c_glib/test/testthrifttestzlibclient.cpp
@@ -107,8 +107,8 @@ class TestHandler : public ThriftTestIf {
     out = thing;
   }
 
-  std::string testUuid(const std::string thing) override {
-    cout << "[C -> C++] testUuid(\"" << std::hex << thing << "\")" << '\n';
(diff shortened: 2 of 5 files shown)


CJCombrink:
@Jens-G 
I have tested with Boost 1.86 and the following changes does fix the compilation and the unit tests still pass as expected

lib/c_glib/test/testthrifttestclient.cpp
```diff
@@ -112,8 +112,8 @@ class TestHandler : public ThriftTestIf {
     out = thing;
   }
 
-  std::string testUuid(const std::string thing) override {
-    cout << "[C -> C++] testUuid(\"" << std::hex << thing << "\")" << '\n';
+  apache::thrift::TUuid testUuid(const apache::thrift::TUuid thing) override {
```
jeking3, line 115: If you use const ref on the parameter and return value instead, it avoids copies.

lib/c_glib/test/testthrifttestzlibclient.cpp
```diff
@@ -107,8 +107,8 @@ class TestHandler : public ThriftTestIf {
     out = thing;
   }
 
-  std::string testUuid(const std::string thing) override {
-    cout << "[C -> C++] testUuid(\"" << std::hex << thing << "\")" << '\n';
+  apache::thrift::TUuid testUuid(const apache::thrift::TUuid thing) override {
```
jeking3, line 110: If you use const ref on the parameter and return value instead, it avoids copies.

lib/c_glib/test/testthrifttestclient.cpp
CJCombrink: Unfortunately thirft generates the interface with a copy. Perhaps it is something I missed when I added the new type because I see the other functions uses const ref on some arguments

CJCombrink: @Jens-G Should I create a ticket for this and fix asap?
It would change how the method is generated if I change it to be a reference:
It appears to change to
```
virtual void testUuid(apache::thrift::TUuid& _return, const apache::thrift::TUuid& thing) = 0;
```
but I can't work on this more at this moment, probably only tonight again

CJCombrink: @jeking3  I made a new PR related to this, would you be so kind to look at it please:
https://github.com/apache/thrift/pull/3038

PR #3037, opened 2024-09-04
THRIFT-5637: Add `no_constructors` option to C++ compiler

alkis:
This PR adds the option `no_constructors` in the C++ generator. This option elides declaring and defining:
* constructor/destructor
* move/copy constructor/assignment

In addition this PR fixes some minor issues:
1. `gen_no_skeleton_` documentation was incorrect
2. when `templates` are requested, avoid any `virtual` functions

Allowing the compiler to emit the default set of constructor/destructor/assignment operators makes the code faster as it allows inlining when they are trivial. This improves creating deep/elaborate thrift objects.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_cpp_generator.cc
@@ -66,6 +66,7 @@ public:
     gen_moveable_ = false;
     gen_no_ostream_operators_ = false;
     gen_no_skeleton_ = false;
+    gen_no_constructors_ = false;
     has_members_ = false;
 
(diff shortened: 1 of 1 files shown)


alkis:
Perhaps the option should be `aggregates` instead of `no_constructor` and `class`es be made into `struct`s instead?

alkis:
@Jens-G could you advise how to move this forward?

Jens-G:
> @Jens-G could you advise how to move this forward?

The usual way would be that someone fom the C++ folks looks at it and decides how to continue.

alkis:
Awesome, thank you @Jens-G!

PR #3038, opened 2024-09-05
THRIFT-5817: [C++] Avoid copy of TUuid

CJCombrink:
[THRIFT-5817: [C++] Avoid copy of TUuid](https://issues.apache.org/jira/browse/THRIFT-5817)
Make TYPE_UUID as complex for C and C++ generator
- Generate code that does not make a copy for TUuid for C and C++

    **This can be breaking, but not sure if it applies since I don't think UUID support has been released for C++**

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_cpp_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_cpp_generator.cc
@@ -274,9 +274,12 @@ public:
   bool is_complex_type(t_type* ttype) {
     ttype = get_true_type(ttype);
 
-    return ttype->is_container() || ttype->is_struct() || ttype->is_xception()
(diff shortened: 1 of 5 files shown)


Jens-G:
Happy to merge into 0.21.0 but I want a second opinion here.

jeking3:
The return value optimization to use a reference is probably a wash, since the value is going to sent through the protocol anyway.  Maybe it provides a tiny memory savings.

Avoiding a copy on the input is a performance improvement and that's the one I thought would be useful.

Jens-G:
@CJCombrink  Could you please rebase so we have another CI run?

CJCombrink:
@Jens-G Successfully rebased

PR #3039, opened 2024-09-06
Remove some gradle deprecation warnings

ctubbsii:
* Specify JUnit platform explicitly
* Remove unused system property (build.test) from unit test environment that relied on a deprecated task property (destinationDir)
* Replace kotlinOptions.jvmTarget with compilerOptions replacement
* Update README to avoid incorrect specification of gradle 8.0, when other gradle 8 versions are acceptable

  

### DIFF
--- a/lib/java/README.md
+++ b/lib/java/README.md
@@ -42,7 +42,7 @@ The Thrift Java source is not build using the GNU tools, but rather uses
 the Gradle build system, which tends to be predominant amongst Java
 developers.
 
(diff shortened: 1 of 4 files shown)


ctubbsii:
I spent a long time trying to figure out how to replace `destinationDir` with its replacement, `destinationDirectory`, which does not work here because it's a complex type and can't be serialized as a system property. But it looks like it doesn't matter anyway, because the `build.test` property didn't even look like it was used anywhere.

jimexist (approved):

PR #3040, opened 2024-09-17
Fix python compiler generated all_structs with type_hints

fishy:
Currently with type_hints enabled, we does not explicitly declare the type of generated all_structs. It's meant to be a list of different types, but that will cause problems, as the generated code is usually like this:

    all_structs = []
    ...
    all_structs.append(Type1)
    ...
    all_structs.append(Type2)
    ...

The first append infers the type of all_struct to be list[Type1], so mypy would start complaining from the 2nd append as the type is unexpected.

Explicitly declare all_struct to be list[typing.Any] instead, which fixes the issue.


fishy:
NOTE: this fixes a new feature added to the 0.21.0 release, so it should ideally be part of the 0.21.0 release.

SalomonSmeke:
Thank You! One can skip type checking the import or get around this with implicit Any but its much nicer to have this declared :)

salomon-smekecohen:
Working on adding more to this.

PR #3042, opened 2024-09-25
THRIFT-5818: Add AIX support

jtroy:
This PR adds support for AIX by splitting the `syscall.Recvfrom()` call into AIX and non-AIX variants. This split is necessary because AIX does not have a `MSG_DONTWAIT` flag, but has an equivalent called `MSG_NONBLOCK`.

### DIFF
--- a/lib/go/thrift/socket_aix_syscall.go
+++ b/lib/go/thrift/socket_aix_syscall.go
@@ -1,0 +1,29 @@
+//go:build aix
+// +build aix
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package thrift
+
+import "syscall"
+
+func peekNonblocking(fd int, p []byte) (int, syscall.Sockaddr, error) {
+	return syscall.Recvfrom(fd, p, syscall.MSG_PEEK|syscall.MSG_NONBLOCK)
+}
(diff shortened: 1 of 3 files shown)


lib/go/thrift/socket_non_aix_syscall.go
```diff
@@ -0,0 +1,29 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package thrift
+
+import "syscall"
+
+func recvfromNonblocking(fd int, p []byte) (int, syscall.Sockaddr, error) {
+	return syscall.Recvfrom(fd, p, syscall.MSG_PEEK|syscall.MSG_DONTWAIT)
```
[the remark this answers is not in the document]
jtroy: Done, sorry about that.

lib/go/thrift/socket_aix_syscall.go
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package thrift
+
+import "syscall"
+
+func recvfromNonblocking(fd int, p []byte) (int, syscall.Sockaddr, error) {
```
fishy, line 27: looks good to me, but I have some nitpick on the  naming here.

this function is not just recvfrom nonblocking, we also always apply the peek flag, so a more accurate name is probably `peekNonblocking`. what do you think?

jtroy: Yes, good point, I think you're right. I've renamed it to `peekNonblocking()`.

fishy (approved):
while I don't have the hardwares, I tested with go's crossbuilding feature:
```
GOOS=windows go build
GOOS=js GOARCH=wasm go build
GOOS=aix GOARCH=ppc64 go build
```
so this _should_ be working

Dylan-M:
This is blocking work I'm doing on an OpenTelemetry build for AIX. Your README claims you don't have a release schedule, but try to release twice a year. As this was merged 5 days after the latest release, I'm wondering if a patch release making it available would be possible?

(Yes, I know I could do a local build and use path redirection to that local build to bypass, but that isn't realistic for a production ready product that needs to be buildable by many others)

Dylan-M:
> @Dylan-M since this is purely a library change, you can always get it via `go get github.com/apache/thrift@master` before the next release happens.

I could, but we don't really do that for software released. We need a pinned version. Hence why I asked if a patch release would be possible, because then we can use it as a pinned version.

That and the, "Never hurts to ask, worst you can get is to be told no" ;)

Dylan-M:
TIL: We can pin versions in our collector builder manifest by forcing replacements, even for indirect dependencies of dependencies.

`- github.com/apache/thrift => github.com/apache/thrift v0.21.1-0.20240926171305-90af876c67be`

Got me what I needed. Glad I didn't give up on this (I'm not a golang developer, I'm a DevOps engineer who happens to be assisting with making this work on AIX)

Jens-G:
Glad it works for you & thanks for using Thrift. 👍
There will be another release, just not immediately.

PR #3051, opened 2024-10-11
THRIFT-5823 Fix illegal uses of exceptions as normal struct type

Jens-G:
Patch: Jens Geyer

### DIFF
--- a/lib/go/test/ClientMiddlewareExceptionTest.thrift
+++ b/lib/go/test/ClientMiddlewareExceptionTest.thrift
@@ -25,11 +25,11 @@ exception Exception2 {
 
 // This is a special case, we want to make sure that the middleware don't
 // accidentally pull result as error.
-exception FooResponse {
+struct/*exception*/ FooResponse {      // returning an exception by any means other than "throws" is illegal
 }
 
 service ClientMiddlewareExceptionTest {
-  FooResponse foo() throws(
+  FooResponse foo() throws(            // returning an exception by any means other than "throws" is illegal
       1: Exception1 error1,
       2: Exception2 error2,
   )
--- a/compiler/cpp/src/thrift/generate/t_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_generator.cc
@@ -101,35 +101,43 @@ void t_generator::validate(const vector<T>& list) const{
 
 void t_generator::validate(t_function const* f) const {
   validate_id(f->get_name());
+  f->validate();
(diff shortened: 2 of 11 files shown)


lib/go/test/ClientMiddlewareExceptionTest.thrift
```diff
@@ -25,11 +25,11 @@ exception Exception2 {
 
 // This is a special case, we want to make sure that the middleware don't
 // accidentally pull result as error.
-exception FooResponse {
+struct/*exception*/ FooResponse {      // returning an exception by any means other than "throws" is illegal
```
fishy, line 28: the purpose of this test is to verify that the library code does not have a bug to accidentally pull the exception response out as the thrown exception, so this change will actually make the test always pass even if someone introduces such a bug in the future.

but if your plan is to forbid that on the compiler level so this scenario would never happen in the first place, then I think this is fine.

fishy: I think I can construct the test differently (construct the go struct directly instead of via thrift IDL) to keep the test to do what it does, though.

fishy (approved):

lib/go/test/ClientMiddlewareExceptionTest.thrift
Jens-G: > I think I can construct the test differently [...] to keep the test to do what it does, though.

Great, thanks!

> if your plan is to forbid that on the compiler level 

I am actually thinking about this, indeed

Jens-G:
> if your plan is to forbid that on the compiler level

implemented

cameron-martin:
Not sure what's gone on here in terms of history, but presumably this was merged as commit 8b6d29d3ee7fb1bece7be1eb728e2e88e867cc2b.

I will repeat what I've said on the commit for visibility:

This change makes the thrift compiler segfault during execution of the NodeJS tests. Note these tests are not run on CI, which is why this will have slipped through the cracks.

Jens-G:
> Should these changes be reverted?

No but we accept pull reuiqests., Would that work for you? 
Alternatively, a minimal viable example how to reproduce could help too.

cameron-martin:
> > Should these changes be reverted?
> 
> No but we accept pull requests., Would that work for you? Alternatively, a minimal viable example how to reproduce could help too.

This was actually an issue with incremental compilation. Ignore the above.

PR #3063, opened 2024-11-14
THRIFT-5833: Add ProcessorError

fishy:
Client: go

Provide ProcessorError as the combined error of write i/o error and original error returned by the endpoint implementation, and update compiler to use that in Process functions.

  

### DIFF
--- a/lib/go/thrift/processor_factory.go
+++ b/lib/go/thrift/processor_factory.go
@@ -19,7 +19,11 @@
 
 package thrift
 
-import "context"
+import (
+	"context"
+	"fmt"
+	"strings"
+)
 
 // A processor is a generic object which operates upon an input stream and
 // writes to some output stream.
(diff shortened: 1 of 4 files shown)


lib/go/thrift/processor_factory.go
```diff
@@ -78,3 +82,47 @@ func NewTProcessorFunctionFactory(p TProcessorFunction) TProcessorFunctionFactor
 func (p *tProcessorFunctionFactory) GetProcessorFunction(trans TTransport) TProcessorFunction {
 	return p.processor
 }
+
+// TProcessorError is the combined original error returned by the endpoint
+// implementation, and I/O error when writing the response back to the client.
+//
+// This type will be returned by Process function if there's an error happend
```
dcelasun, line 89:
```suggestion
// This type will be returned by Process function if there's an error happened
```

dcelasun (approved):
LGTM

fishy:
@dcelasun I renamed it from `TProcessorError` to `ProcessorError`. I think we use the `T-` prefix for things in the spec and shared by most languages, and this is a go only thing. PTAL.

dcelasun (approved):

PR #3068, opened 2024-11-23
THRIFT-5837 Delphi implementation for THRIFT-5835

Jens-G:
Client: Delphi
Patch: Jens Geyer


ctubbsii:
The build seems to have been broken since this was merged. I can't seem to find the exact problem, but it seems it might be that some warnings are being treated as errors. This might not even be related to this change, but could be due to an update to the compiler in the GitHub Actions runner image.

Jens-G:
No idea how that slipped through - thanks, looking at it

PR #3072, opened 2024-12-09
lib/cpp/test/CMakeLists.txt: Fix building with OpenBSD

brad0:
OpenBSD does not have librt.


Jens-G:
Could you rebase and push again pls?

PR #3075, opened 2024-12-12
check overflow on getting field id

jp0317:
Add overflow checking when getting field id based on field delta.

### DIFF
--- a/lib/rs/src/protocol/compact.rs
+++ b/lib/rs/src/protocol/compact.rs
@@ -186,7 +186,16 @@ where
             ),
             _ => {
                 if field_delta != 0 {
-                    self.last_read_field_id += field_delta as i16;
+                    self.last_read_field_id = self
+                        .last_read_field_id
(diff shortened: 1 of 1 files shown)


Jens-G:
Sth is not right here

     error: use of `ok_or` followed by a function call

jp0317:
yeah, it's should be fine to use `ok_or` there. this seems to be a false alarm (found a similar issue [here](https://github.com/rust-lang/rust-clippy/issues/1609))

Jens-G:
Maybe but it breaks the build.

jimexist:
ok or else saves one allocation, it's trivial to change to that

Jens-G:
> trivial to change to that

Seems not the case ... or is it? 😅

Jens-G:
Any progress here?

jp0317:
sorry for the late reply. I've updated to use `ok_or_else`

PR #3076, opened 2024-12-13
Fix documentation to match implementation

mpoeter:
Boolean values are encoded as `BOOLEAN_TRUE` and `BOOLEAN_FALSE` which are defined as 1 and 2 instead of 1 and 0.

The element-type of a bool is encoded as `BOOLEAN_TRUE` so 1 instead of 2.

[skip ci]


doc/specs/thrift-compact-protocol.md
```diff
@@ -249,7 +249,7 @@ The short form should be used when the length is in the range 0 - 14 (inclusive)
 
 The following element-types are used (see note below):
 
-* `BOOL`, encoded as `2`
+* `BOOL`, encoded as `1`
```
Jens-G, line 252: That is debatable. A better word could be inconcistent.

* thrift_compact_protocol.erl does it acciotrding to the spec and writes 2
* anyone else writes 1

PR #3079, opened 2024-12-26
remove more Python2 compatibility

a-detiste:
Hi,

This is a follow up to https://github.com/apache/thrift/pull/3030 
(which itself took over https://github.com/apache/thrift/pull/2588)


a-detiste:
https://issues.apache.org/jira/browse/THRIFT-5537 is still open

lib/py/src/compat.py
```diff
@@ -17,30 +17,13 @@
 # under the License.
 #
 
-import sys
+from io import BytesIO as BufferIO  # noqa
```
fishy, line 20: why is this import needed? neither `BytesIO` nor `BufferIO` seems to be used in this file.

also, why the rename here?

a-detiste: The two compat.py should go.

I planned to do this in the next PR

fishy: ok looks like this is from old code

fishy (approved):
looks like the python unit tests passed. cross-tests were not run because swift/netstd/rust are currently broken.

I'll give others some time to chime in before merging this.

PR #3082, opened 2025-01-06
THRIFT-5564: Add nodejs tests to github actions

cameron-martin:
These tests exist, but don't currently run on github actions. This adds a new job to run these.

This also fixes the regression in the tests caused by https://github.com/apache/thrift/pull/3014.

- [ ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes)
- [ ] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [x] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #3083, opened 2025-01-07
THRIFT-5811: Add ESM support to nodejs codegen

cameron-martin:
Client: nodejs

This adds a flag to the JS generator to output ES modules instead of CommonJS. This is only valid when targeting node. A lot of the changes here are to test this.

The `testAll.sh` script now generates an ES module version of the services and types, and tests the client and the server with these. This has a few knock-on effects. Firstly, any module that imports a generated ES module must itself be an ES module, since CommonJS modules cannot import ES modules. ES modules also do not support `NODE_PATH`, so instead the tests directory is converted into a node package with a `file:` dependency on the root thrift package.

  

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_js_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_js_generator.cc
@@ -68,6 +68,7 @@ public:
     gen_jquery_ = false;
     gen_ts_ = false;
     gen_es6_ = false;
+    gen_esm_ = false;
     gen_episode_file_ = false;
 
     bool with_ns_ = false;
@@ -83,6 +84,8 @@ public:
         with_ns_ = true;
       } else if( iter->first.compare("es6") == 0) {
         gen_es6_ = true;
+      } else if ( iter->first.compare("esm") == 0) {
+        gen_esm_ = true;
       } else if( iter->first.compare("imports") == 0) {
         parse_imports(program, iter->second);
       } else if (iter->first.compare("thrift_package_output_directory") == 0) {
@@ -103,6 +106,10 @@ public:
 
     if (!gen_node_ && with_ns_) {
       throw std::invalid_argument("invalid switch: [-gen js:with_ns] is only valid when using node.js");
+    }
+
+    if (!gen_node_ && gen_esm_) {
+      throw std::invalid_argument("invalid switch: [-gen js:esm] is only valid when using node.js");
     }
 
     // Depending on the processing flags, we will update these to be ES6 compatible
@@ -278,13 +285,6 @@ public:
     return js_namespace(p);
   }
 
-  std::string js_export_namespace(t_program* p) {
-    if (gen_node_) {
-      return "exports.";
-    }
-    return js_namespace(p);
-  }
-
   bool has_js_namespace(t_program* p) {
     if (no_ns_) {
       return false;
@@ -375,6 +375,11 @@ private:
   bool gen_es6_;
 
   /**
+   * True if we should generate ES modules, instead of CommonJS.
+   */
+  bool gen_esm_;
+
+  /**
    * True if we will generate an episode file.
    */
   bool gen_episode_file_;
@@ -452,7 +457,7 @@ void t_js_generator::init_generator() {
     f_episode_.open(f_episode_file_path);
   }
 
-  const auto f_types_name = outdir + program_->get_name() + "_types.js";
+  const auto f_types_name = outdir + program_->get_name() + "_types" + (gen_esm_ ? ".mjs" : ".js");
   f_types_.open(f_types_name.c_str());
   if (gen_episode_file_) {
     const auto types_module = program_->get_name() + "_types";
@@ -478,7 +483,13 @@ void t_js_generator::init_generator() {
   }
 
   if (gen_node_) {
-    f_types_ << js_const_type_ << "ttypes = module.exports = {};" << '\n';
+    if (gen_esm_) {
+      // Import the current module, so we can reference it as ttypes. This is
+      // fine in ESM, because it allows circular imports.
+      f_types_ << "import * as ttypes from './" + program_->get_name() + "_types.mjs';" << '\n';
+    } else {
+      f_types_ << js_const_type_ << "ttypes = module.exports = {};" << '\n';
+    }
   }
 
   string pns;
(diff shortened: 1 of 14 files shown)


thomasbruggink (commented):
Thanks for this, we were actually planning to add something similar but for the browser. We can make a follow up PR to add support for `esm` on the browser side as well.

While testing this locally we noticed import statements are not changed to the esm format.
The existing PR keeps the `require(type)` style instead of changing it to the esm format of:
```
import * as <Identifier>_ttypes from <Identifier>.mjs 
```
(in t_js_generator.cc line 584)

Other than that it worked as expected :+1:

cc: @tri613

cameron-martin:
> Thanks for this, we were actually planning to add something similar but for the browser. We can make a follow up PR to add support for esm on the browser side as well.

That's interesting, because this is our exact use case, but we use the nodejs package in the browser. Is your plan here to support es modules that can directly execute in the browser, with no build tools at all?

> While testing this locally we noticed import statements are not changed to the esm format.

I must have missed this, I'll push an update. It also will need syncing with the nodejs CI changes (#3082).

cameron-martin:
> While testing this locally we noticed import statements are not changed to the esm format.

Which import statements are not changed? All of the tests files inside `gen-nodejs-esm` use `import` rather than `require`.

tri613:
Hello! Thanks for this great PR again. 👍 

> import statements are not changed to the esm format

We tried to use this PR's changes to build our own service, and when looking inside the generated files, we noticed that the import statements are like this:
```
// FooService.mjs
import { Thrift } from 'thrift';
import { Q } from 'thrift';
import Int64 from 'node-int64';
var Foo_ttypes = require('./Foo_types');
var Bar_ttypes = require('./Bar_types');
```

As @thomasbruggink suggested we think this is related to the `render_includes` function at line 584.

> Is your plan here to support es modules that can directly execute in the browser, with no build tools at all?

We had some issues when integrating the nodejs thrift build into our project, so we thought it would be nice if we could decouple the esm build from the nodejs dependencies. And of course, supporting native esm browser packages would be beneficial too :D 
However, it is good to know that the nodejs version worked for you - I will take another look and check if we are missing anything in our project's build tool!

cameron-martin:
> We tried to use this PR's changes to build our own service, and when looking inside the generated files, we noticed that the import statements are like this:

Does this happen when a thrift file includes another thrift file? I don't think this is something the tests are covering atm. I'll investigate and add a test for it.

> We had some issues when integrating the nodejs thrift build into our project, so we thought it would be nice if we could decouple the esm build from the nodejs dependencies. And of course, supporting native esm browser packages would be beneficial too :D
However, it is good to know that the nodejs version worked for you - I will take another look and check if we are missing anything in our project's build tool!

It worked for us by polyfilling a few NodeJS dependencies (such as `util`, `Buffer`, etc), but I've been wanting to remove the dependencies on these since in modern JS environments there are cross-runtime alternatives. See [THRIFT-5844](https://issues.apache.org/jira/browse/THRIFT-5844). In fact we have a few PRs in the works that should remove these.

Another part that is missing is browser tests for the NodeJS package, but I also have plans to contribute these too.

This does call into question the necessity of a separate nodejs and js package. Does anyone do web development without a bundler currently?

cameron-martin:
@tri613 @thomasbruggink I've updated this to render includes correctly, plus added a test (`include.test.mjs`) for this.

However, CI will fail on this until #3087 is merged, since the current version of eslint and prettier do not support the dynamic import syntax, and the nodejs tests now rely on this.

thomasbruggink:
> This does call into question the necessity of a separate nodejs and js package. Does anyone do web development without a bundler currently?

At the moment we rely on the web version while also having a bundler in place. It looks to me like the web version is a leaner but less feature rich (although using some older standards like xmlhttprequest).
I haven't checked this into detail but I wonder if there is a benefit to keeping the more lightweight web version around :thinking:

thomasbruggink (approved):
i have no power in the repository but we've tested this and it works for us ^^
thanks @cameron-martin for making these changes

Jens-G:
could you prebase this, please?

cameron-martin:
@Jens-G done

Jens-G:
../../node_modules/typescript/bin/tsc --outDir test-compiled/ --project test/tsconfig.json
        ../../node_modules/@types/estree/index.d.ts(137,38): error TS2304: Cannot find name 'Omit'.
        ../../node_modules/@types/estree/index.d.ts(452,13): error TS2304: Cannot find name 'bigint'.
        ../../node_modules/@types/estree/index.d.ts(664,42): error TS2304: Cannot find name 'Omit'.
        make: *** [Makefile:634: ts-compile] Error 2
        make: Leaving directory '/home/runner/work/thrift/thrift/lib/nodets'

Jens-G:
> error TS2304: Cannot find name 'Omit'.

That seems not directly related but caused byo some 3rtd party package. At least that's what I concluded from the action logs. Any chance we get that fixed too?

cameron-martin:
It looks like this is a conflict between https://github.com/apache/thrift/pull/3087 and https://github.com/apache/thrift/pull/3089.

The issue here is that upgrading eslint updated `@types/estree`, and this package uses features of the typescript standard library (introduced in [3.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#bigint) and [3.5](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type)) that aren't in the version that we use.

My suggested fix here would be to upgrade typescript to a newer version. I already have a PR up for this (https://github.com/apache/thrift/pull/3084). This PR upgrades it to the latest version. However, I see that typescript version is in the [language support document](https://github.com/apache/thrift/blob/master/LANGUAGES.md), so do we want to upgrade this to the latest or just a slightly newer version, to unsure that we support older versions too?

Jens-G:
Great! I'd say that newest in master is fine. Feel free to update the lang document accordingly 

BTW seems #3084 needs a rebase too.

cameron-martin:
@Jens-G I've rebased #3084.

cameron-martin:
@Jens-G This should pass CI now.

katamyra:
@cameron-martin I'm trying to use thrift in the browser but am encountering lots of errors even after generating using the flag `es6`. I was just wondering if es6 is only for server code rather than for the browser? Thanks!

cameron-martin:
With the necessary polyfills it should work in the browser. There's quite a lot more to do to make this better though. See [THRIFT-5844](https://issues.apache.org/jira/browse/THRIFT-5844).

katamyra:
Thanks! I was thinking of building from source on this [commit](https://github.com/apache/thrift/commit/21ed4a2f690fb8547593c64df376b97a877b1675), do you think that would be easier than the polyfills? I tried the polyfills but had lots of small messy errors

Edit: nvm it looks like you still need to polyfill

PR #3088, opened 2025-01-15
test/crossrunner: merge in old Py2vs3 compat.py

a-detiste:
as promised in previous PR


fishy (approved):

fishy:
Currently some of the language tests are failing so we don't really run cross tests in CI so this cannot be verified in CI.

PR #3094, opened 2025-02-02
Support previously and currently documented encodings for boolean lists in rust compact protocol

jhorstmann:
In https://github.com/apache/datafusion/issues/14281 it was noticed that the go parquet implementation writes metadata containing boolean lists in an unexpected encoding. The collection type was encoded as `0x02`, which the documentation now specifies as an allowed alternative value to `0x01`. The values were encoded as 0 (false) and 1 (true), which matches a previous version of the documentation (see documentation update in https://github.com/apache/thrift/commit/2c29c5665bc442e703480bb0ee60fe925ffe02e8). This change makes the rust `TCompactInputProtocol` support both variations.

Happy to create a jira ticket if needed.

### DIFF
--- a/lib/rs/src/protocol/compact.rs
+++ b/lib/rs/src/protocol/compact.rs
@@ -210,6 +210,10 @@ where
             None => {
                 let b = self.read_byte()?;
                 match b {
+                    // Previous versions of the thrift compact protocol specification said to use 0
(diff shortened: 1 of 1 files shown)


alamb:
Related ticket in thrift:
 -https://github.com/apache/thrift/pull/3076 (a few weeks ago)

PR #3095, opened 2025-02-03
Various build fixes for clang and rust builds

mhlakhani:
This PR fixes the clang build (for the thrift compiler) and makes the Rust tutorial build pass as well.

The C++ changes fix `inconsistent override` compiler warnings, and the Rust change fixes a clippy warning about an unnecessary dereference.

I tested this with the below, which now passes (previously failed):

```
./configure --disable-dependency-tracking --without-python --without-ruby --without-perl  --without-py3 --without-swift --without-nodejs
make -j16
```

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), **not required for trivial changes**)

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.h
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.h
@@ -288,7 +288,7 @@ public:
 
   static bool is_pointer_field(t_field* tfield, bool in_container = false);
 
-  std::string indent_str() const {
+  std::string indent_str() const override {
     return "\t";
   }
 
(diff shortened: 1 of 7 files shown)


emmenlau (commented):
I have not compiled this myself, but if your compiler happily accepts the C++ changes, they should be good. Adding override can only help, or break the build, so if the build still works, it should only have helped :-)

So: C++ side looks good to me. Review of the Rust change, anyone?

PR #3101, opened 2025-02-11
Update supported go versions

fishy:
With the release of go 1.24.0, go 1.22.x is no longer supported. Update supported go versions to go 1.23 and go 1.24.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,8 +135,8 @@ jobs:
     strategy:
       matrix:
         go:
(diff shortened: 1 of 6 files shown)


dcelasun (approved):
> @dcelasun please let me know if you can think of anything we can improve with new features from 1.23 :)

I did a quick skim of the [release notes](https://go.dev/doc/go1.23). We might be able to use some of the slices/maps functions in a few places, but it would mostly be cosmetic, so let's go ahead as is :)

PR #3109, opened 2025-03-09
THRIFT-1482: Unix domain socket support under PHP

balping:
To open a unix socket, the port number supplied to `fsockopen` (and `pfsockopen`) has to be `-1`.

We use `substr` instead of `str_starts_with` to keep compatibility with php 7.

Now a UDS can be created like this:

```php
new TSocket('unix:///tmp/ipc.sock', -1)
```

### DIFF
--- a/lib/php/lib/Transport/TSocket.php
+++ b/lib/php/lib/Transport/TSocket.php
@@ -218,7 +218,7 @@ class TSocket extends TTransport
             throw new TTransportException('Cannot open null host', TTransportException::NOT_OPEN);
         }
 
-        if ($this->port_ <= 0) {
+        if ($this->port_ <= 0 && substr($this->host_, 0, strlen('unix://')) !== 'unix://') {
             throw new TTransportException('Cannot open without port', TTransportException::NOT_OPEN);
         }
 
--- a/lib/php/lib/Transport/TSocket.php
+++ b/lib/php/lib/Transport/TSocket.php
@@ -218,7 +218,7 @@ class TSocket extends TTransport
             throw new TTransportException('Cannot open null host', TTransportException::NOT_OPEN);
         }
 
-        if ($this->port_ <= 0 && substr($this->host_, 0, strlen('unix://')) !== 'unix://') {
+        if ($this->port_ <= 0 && strpos($this->host_, 'unix://') !== 0) {
             throw new TTransportException('Cannot open without port', TTransportException::NOT_OPEN);
         }
 
--- a/lib/php/test/Unit/Lib/Transport/TSocketTest.php
+++ b/lib/php/test/Unit/Lib/Transport/TSocketTest.php
@@ -246,6 +246,33 @@ class TSocketTest extends TestCase
         $this->assertTrue($transport->isOpen());
     }
 
+    public function testOpenUnixSocket()
+    {
+        $host = 'unix:///tmp/ipc.sock';
+        $port = -1;
+        $persist = false;
+        $debugHandler = null;
+
+        $this->getFunctionMock('Thrift\Transport', 'fsockopen')
+             ->expects($this->once())
+             ->with(
+                 $host,
+                 $port,
+                 $this->anything(), #$errno,
+                 $this->anything(), #$errstr,
+                 $this->anything() #$this->sendTimeoutSec_ + ($this->sendTimeoutUsec_ / 1000000),
+             );
+
+        $transport = new TSocket(
+            $host,
+            $port,
+            $persist,
+            $debugHandler
+        );
+
+        $transport->open();
+    }
+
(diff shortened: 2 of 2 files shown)


lib/php/lib/Transport/TSocket.php
```diff
@@ -218,7 +218,7 @@ public function open()
             throw new TTransportException('Cannot open null host', TTransportException::NOT_OPEN);
         }
 
-        if ($this->port_ <= 0) {
+        if ($this->port_ <= 0 && substr($this->host_, 0, strlen('unix://')) !== 'unix://') {
```
sveneld, line 221: you can make this check like this
```
if ($this->port_ <= 0 && strpos($this->host_, 'unix://') !== 0) {
```
update a unit test lib/php/test/Unit/Lib/Transport/TSocketTest.php

sveneld (changes requested):

lib/php/lib/Transport/TSocket.php
balping: `strpos` is less efficient as it tries to scan the whole string. See discussion here: https://stackoverflow.com/a/2790912

I've added a unit test

sveneld: Thank you for updating a unit test.
I don't agree with using substr
instead of using one function you offer to use two substr and strlen, and even if remove strlen and relace it with const value it will be less efficient 
just run this code for compare, the most productive will be strpos 
https://3v4l.org/rv7pl
```
<?php

$host = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
//$host = 'unix://unixhost';

$start = microtime(true);

for ($i=0; $i<=1000; $i++) {
    substr($host, 0, strlen('unix://')) !== 'unix://';
}
echo (microtime(true) - $start) . PHP_EOL;


$start = microtime(true);

for ($i=0; $i<=1000; $i++) {
    strpos($host, 'unix://') !== 0;
}
echo (microtime(true) - $start) . PHP_EOL;


$start = microtime(true);

for ($i=0; $i<=1000; $i++) {
    preg_match('~^unix://~', $host);
}
echo (microtime(true) - $start) . PHP_EOL;
```

sveneld (changes requested):

lib/php/lib/Transport/TSocket.php
balping: I changed it to the strpos implementation

sveneld (approved):

sveneld:
@balping  apply this patch for test.

```
Index: lib/php/test/Unit/Lib/Transport/TSocketTest.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/php/test/Unit/Lib/Transport/TSocketTest.php b/lib/php/test/Unit/Lib/Transport/TSocketTest.php
--- a/lib/php/test/Unit/Lib/Transport/TSocketTest.php	(revision 19081602c35311ee13e6ba9ab2fa5001d182048b)
+++ b/lib/php/test/Unit/Lib/Transport/TSocketTest.php	(revision b1d4b05f352ded2ba5b78cb0a4f56dde01a11d41)
@@ -252,6 +252,7 @@
         $port = -1;
         $persist = false;
         $debugHandler = null;
+        $handle = fopen('php://memory', 'r+');
 
         $this->getFunctionMock('Thrift\Transport', 'fsockopen')
              ->expects($this->once())
@@ -261,7 +262,22 @@
                  $this->anything(), #$errno,
                  $this->anything(), #$errstr,
                  $this->anything() #$this->sendTimeoutSec_ + ($this->sendTimeoutUsec_ / 1000000),
-             );
+             )->willReturn($handle);
+
+        $this->getFunctionMock('Thrift\Transport', 'socket_import_stream')
+            ->expects($this->once())
+            ->with($handle)
+            ->willReturn(true);
+
+        $this->getFunctionMock('Thrift\Transport', 'socket_set_option')
+            ->expects($this->once())
+            ->with(
+                $this->anything(), #$socket,
+                SOL_TCP, #$level
+                TCP_NODELAY, #$option
+                1 #$value
+            )
+            ->willReturn(true);
 
         $transport = new TSocket(
             $host,

```

sveneld:
@Jens-G  what about failing tests?
patch for them should be applied...

Jens-G:
> @Jens-G what about failing tests? patch for them should be applied...

Just noticed and reverted.

Jens-G:
Any news here?

sveneld:
@Jens-G I add patch for test and created a new pull request https://github.com/apache/thrift/pull/3130

PR #3113, opened 2025-03-20
Log the exception when TThreadPoolServer breaks the work loop

dengzhhu653:

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/server/TThreadPoolServer.java
+++ b/lib/java/src/main/java/org/apache/thrift/server/TThreadPoolServer.java
@@ -277,7 +277,7 @@ public class TThreadPoolServer extends TServer {
     private void logException(Exception x) {
       // We'll usually receive RuntimeException types here
       // Need to unwrap to ascertain real causing exception before we choose to ignore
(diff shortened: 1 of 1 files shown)


Jens-G:
The exception to break the work loop occurs rather often. It is essentially part of the normal expected workflow. 

Whether this is a clever design is a different question, but that's the way it is.

jeking3 (approved):

jeking3:
Given it's a debug log message seems reasonable, so I approved it.

If a specific exception is used to break it out "normally" then maybe that case should be handled without the log (like InterruptedException).

jeking3:
Are we running java build and test in CI?

dengzhhu653:
> Are we running java build and test in CI?

Thank you @jeking3!
Not sure if I'm right, the java build: https://github.com/apache/thrift/actions/runs/13964934545/job/39093292609?pr=3113 looks successful, while the cross-test is skipped, don't know how to tigger it myself.

jimexist (approved):

jimexist:
i can confirm that the java suite was run and successful

PR #3116, opened 2025-03-26
THRIFT-5859: Generate a values iterator for enum types for go

fishy:
Client: go

The generated code looks like this:

    type Foo int64

    const (
    	Foo_One Foo = 1
    	Foo_Two Foo = 2
    )

    var knownFooValues = []Foo{
    	Foo_One,
    	Foo_Two,
    }

    func FooValues() iter.Seq[Foo] {
    	return func(yield func(Foo) bool) {
    		for _, v := range knownFooValues {
    			if !yield(v) {
    				return
    			}
    		}
    	}
    }

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -706,6 +706,7 @@ string t_go_generator::go_imports_begin(bool consts) {
   }
   system_packages.push_back("errors");
   system_packages.push_back("fmt");
+  system_packages.push_back("iter");
   system_packages.push_back("log/slog");
   system_packages.push_back("time");
   // For the thrift import, always do rename import to make sure it's called thrift.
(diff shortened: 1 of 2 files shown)


dcelasun:
Copying over from Jira: I think a `func FooValues() []Foo` would be better. If you prefer a map though, at least make the values `struct{}` :)

fishy:
@dcelasun changed to return an iterator instead, please take another look.

iterator has the advantage of immutability over slice. it does have the disadvantage of lack of length info. if people need that, a one-time O(N) operation of `slices.Collect` or similar should be trivial. we can also provide another function to return length info, for example `func FooValuesN() int`, but the naming is kinda awkward, and I think in most use cases that info is needed, they probably would need to build a map anyways, so they need the one-time O(N) operation anyways. we can add it later if really needed.

dcelasun (approved):
Deferring to `len(slices.Collect(FooValues()))` is fine by me.

PR #3120, opened 2025-04-04
Erlang generator improvements

kpy3:
Add some minor improvements to Erlang generator:

- Mark records generated from union as 'union' instead of 'struct'
- Allow choose type for string, can be 'string', 'binary' or 'both' (default)
- Allow sets to be initialised with new(fast) backend (default is old)

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_erl_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_erl_generator.cc
@@ -40,6 +40,16 @@ using std::stringstream;
 using std::vector;
 
 /**
+ * Helper enum for string mapping
+ */
+enum class StringTo {String, Binary, Both};
+
+/**
+ * Helper enum for sets mapping
+ */
+enum class SetsTo {V1, V2};
+
+/**
  * Erlang code generator.
  *
  */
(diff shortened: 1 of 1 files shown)


compiler/cpp/src/thrift/generate/t_erl_generator.cc
```diff
@@ -715,21 +755,30 @@ string t_erl_generator::render_default_value(t_field* field) {
...
 }
 
+string t_erl_generator::render_default_sets_value() {
+  switch (sets_to_) {
+  case SetsTo::V1:
+    return "sets:new()";
+  case SetsTo::V2:
+    return "sets:new([{version,2}])";
+  }
+}
```
Jens-G, line 773: there's something missing obviosuly

compiler/cpp/src/thrift/generate/t_erl_generator.cc
```diff
@@ -761,6 +810,17 @@ string t_erl_generator::render_member_type(t_field* field) {
...
+string t_erl_generator::render_string_type() {
+  switch (string_to_) {
+  case StringTo::String:
+    return "string()";
+  case StringTo::Binary:
+    return "binary()";
+  case StringTo::Both:
+    return "string() | binary()";
+  }
+}
```
Jens-G, line 822: same here

Jens-G (changes requested):

compiler/cpp/src/thrift/generate/t_erl_generator.cc
kpy3: I've added default branch to switch here

compiler/cpp/src/thrift/generate/t_erl_generator.cc
kpy3: And here :)

PR #3125, opened 2025-04-18
Fix field stop read in duplicate_protocol.go

winmillwill:
When generated code reads a struct, it runs a `for` loop calling `ReadFieldBegin` at the top, but breaks if the field type ID is `thrift.STOP`.

With TDuplicateToProtocol naively performing an equivalent write for every read, this results in extra bytes, which breaks just about any protocol in the `DuplicateTo` struct field.

The proposed fix is to simply add special handling for `thrift.STOP` to `ReadFieldBegin`.

I'm no thrift expert, so I have no idea how other libraries handle this concern.  Ideally, it seems like each protocol should understand and enforce the invariant that an attempt to call `WriteFieldBegin` with type ID 0 either isn't valid or is a misguided attempt to call `WriteFieldStop`, which could be performed instead, perhaps with the additional guardrail that a sequence of stops (with length > 1) never makes any sense.  

- [ x ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes)
- [ x ] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [ x ] Did you squash your changes to a single commit?  (not required, but preferred)
- [ x ] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ x ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

^^ hopefully we can agree this change is trivial, I very badly do not want to look at anyone's Jira, especially if it doesn't belong to my employer.

### DIFF
--- a/lib/go/thrift/duplicate_protocol_test.go
+++ b/lib/go/thrift/duplicate_protocol_test.go
@@ -1,0 +1,52 @@
+package thrift
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"testing"
+)
+
+func TestDuplicateToJSONReadMakesEqualJSON(t *testing.T) {
+	delegateBuf := NewTMemoryBuffer()
+	iprot := NewTJSONProtocolFactory().GetProtocol(delegateBuf)
+
+	s := NewMyTestStruct()
+	ctx := context.Background()
+	s.Write(ctx, iprot)
+
+	iprot.Flush(ctx)
+	jsonBytesWritten := delegateBuf.Bytes()
+
+	if err := json.Unmarshal(jsonBytesWritten, NewMyTestStruct()); err != nil {
+		t.Errorf("error unmarshaling written bytes: %s", err)
+	}
+
+	duplicateBuf := NewTMemoryBuffer()
+	dupProto := NewTJSONProtocolFactory().GetProtocol(duplicateBuf)
+	proto := &TDuplicateToProtocol{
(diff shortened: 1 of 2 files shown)


KostenetskyiAndrii (approved):

fishy (commented):
Can you please add a test (since this needs to use a struct, the test needs to be under `lib/go/test`) that would catch the bug?

winmillwill:
> Can you please add a test (since this needs to use a struct, the test needs to be under `lib/go/test`) that would catch the bug?

I have written a test in the way that made sense to me.

lib/go/thrift/duplicate_protocol_test.go
```diff
@@ -0,0 +1,52 @@
...
+	dupProto.Flush(ctx)
+
+	jsonBytesRead := duplicateBuf.Bytes()
+
+	if !bytes.Equal(jsonBytesWritten, jsonBytesRead) {
+		t.Errorf(`bytes read into duplicate do not equal bytes written
+		read:
+		%+v
+		written:
+		%+v
```
fishy, line 44: since we are using TJSONProtocol here, output the string will be more useful:
```suggestion
		%s
		written:
		%s
```

fishy (approved):

PR #3131, opened 2025-04-27
THRIFT-5863: Add method for customizing the TConfiguration in TServer…

dengzhhu653:
…Transport

  


lib/java/src/main/java/org/apache/thrift/transport/TServerSocket.java
```diff
@@ -122,6 +126,7 @@ public TSocket accept() throws TTransportException {
       throw new TTransportException("Blocking server's accept() may not return NULL");
     }
     TSocket socket = new TSocket(result);
+    socket.getConfiguration().deepCopy(configuration_);
```
Jens-G, line 129: Question: Why is it considered good to have multiple configuration instances, in opposite to use a shared instance through the whole transport/protocol stack? What's the added value exactly?

dengzhhu653: the main idea is to avoid another new constructor in `TSocket`, the `deepCopy` will copy the threshold from the `configuration_` into `socket.getConfiguration()`, the original `socket.getConfiguration()` is still shared across the whole transport/protocol stack.

dengzhhu653: Add a new constructor in `TSocket`, this constructor might also help the client create the instance over the SSLSocket. Thank you for the comment!

fishy: I think in general we want to propagate a single `TConfiguration` instance across the stack, vs. copying. This is at least implied in the spec of TConfiguration (https://github.com/apache/thrift/blob/master/doc/specs/thrift-tconfiguration.md, it mentioned "...we want to pass the instance by reference rather than by value.").

That's also how I implemented it in go. see the NOTE part of the doc of TConfiguration for some explanation: https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TConfiguration

When implementing it for go we also deprecated most of the old constructors doing configuration without TConfiguration.

dengzhhu653: Hi @fishy, I'm confused here, do you mean this line: https://github.com/apache/thrift/blob/aa5f183312201bda890ab0065ec26d9e8a23b9f9/lib/java/src/main/java/org/apache/thrift/transport/TServerTransport.java#L62
I create a new `TConfiguration` for the compatibility of `maxFrameSize(int maxFrameSize) ` method(for pushing the maxFrameSize to the TConfiguration directly), and this calls at the service startup, there is no transport pipeline built successfully yet.
After the server accepting the connection, the transport pipeline should reuse the TSocket's configuration, the new change is say the same story per my understanding, as I don't change how to propagate the `TConfiguration` through the stack.

To make the propagation and validation clear, I raised another PR https://github.com/apache/thrift/pull/3127, for a stack like: TSocket -> TSaslTransport(TMemoryInputTransport) -> TUGIAssumingTransport
Before: 
    1, TSocket, TSaslTransport and TMemoryInputTransport, have a local reference to the same `TConfiguration`.
    2, TSocket, TSaslTransport and TMemoryInputTransport have it own private `knownMessageSize` and `remainingMessageSize`, updated and checked separately.
After:
   1, TSaslTransport and TMemoryInputTransport don't have such local reference, while TSaslTransport can use `getConfiguration` to retrieve the `TConfiguration` from end TSocket, same to TUGIAssumingTransport, which ensures the same `TConfiguration` is shared across the stack.
   2, Only TSocket updates and checks the message size.

dengzhhu653:
Hi @Jens-G @fishy, any idea about this PR? this gives the server a chance to change the max message/frame size of the accepted transport, and it doesn't block anything.

Jens-G:
I'm not sure what the status here is ... to me it looks like there are open questions and/or change requests.

dengzhhu653:
> I'm not sure what the status here is ... to me it looks like there are open questions and/or change requests.

Thank you @Jens-G. I guess I have fixed the change request:
>  have multiple configuration instances, in opposite to use a shared instance through the whole transport/protocol stack

> I think in general we want to propagate a single TConfiguration instance across the stack, vs. copying

Done, a single TConfiguration is shared across the stack now

Jens-G:
Could you force-push please to trigger another CI run? @dengzhhu653

PR #3135, opened 2025-05-14
Adds trace logging for java non blocking server memory limit

keith-turner:
Added a trace log and a method to help get information about the java AbstractNonblockingServer memory limit for reading data from clients.  This memory limit is configurable, but there is no mechanism to help someone understand how that limits impacts a thrift server at runtime.  These two simple changes attempt to provide some runtime insight in the memory usage and when the limit is hit.

  

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
+++ b/lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
@@ -312,9 +312,10 @@ public abstract class AbstractNonblockingServer extends TServer {
     }
 
     /**
-     * @return the amount of memory currently used to read data from clients. This information can be useful for debugging, metrics, and configuring the maximum memory limit.
+     * @return the amount of memory currently used to read data from clients. This information can
+     *     be useful for debugging, metrics, and configuring the maximum memory limit.
      */
-    public long getReadBufferBytesAllocated(){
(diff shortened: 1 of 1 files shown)


Jens-G:
Spotless has som,e complaints

lib/java/src/main/java/org/apache/thrift/server/AbstractNonblockingServer.java
```diff
@@ -351,7 +359,13 @@ public boolean read() {
 
           // if this frame will push us over the memory limit, then return.
           // with luck, more memory will free up the next time around.
-          if (readBufferBytesAllocated.get() + frameSize > MAX_READ_BUFFER_BYTES) {
+          long currentAllocated = readBufferBytesAllocated.get();
```
ctubbsii, line 362: In case the method is overridden, should probably use the new method to get this instead.

```suggestion
          long currentAllocated = getReadBufferBytesAllocated();
```

ctubbsii: Alternatively, just make the method final, since it depends so much on the internal implementation.

keith-turner: made it final which makes it safer to call the getter

PR #3139, opened 2025-05-18
[nit] Throw exception instead of asserting in TJSONProtocol::readByte

mhlakhani:
Code structure wise, I think this is a bit nicer as it lets a caller catch this and return their own specific error messages.

Change is trivial, so I didn't file a ticket

### DIFF
--- a/lib/cpp/src/thrift/protocol/TJSONProtocol.cpp
+++ b/lib/cpp/src/thrift/protocol/TJSONProtocol.cpp
@@ -1087,7 +1087,10 @@ uint32_t TJSONProtocol::readBool(bool& value) {
 uint32_t TJSONProtocol::readByte(int8_t& byte) {
   auto tmp = (int16_t)byte;
   uint32_t result = readJSONInteger(tmp);
-  assert(tmp < 256);
+  if (tmp > 255 || tmp < -128) {
+    throw TProtocolException(TProtocolException::INVALID_DATA,
+                            "Expected byte value; got " + to_string(tmp));
+  }
   byte = (int8_t)tmp;
   return result;
 }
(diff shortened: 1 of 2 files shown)


lib/cpp/src/thrift/protocol/TJSONProtocol.cpp
```diff
@@ -1087,7 +1087,10 @@ uint32_t TJSONProtocol::readBool(bool& value) {
 uint32_t TJSONProtocol::readByte(int8_t& byte) {
   auto tmp = (int16_t)byte;
   uint32_t result = readJSONInteger(tmp);
-  assert(tmp < 256);
+  if (tmp > 255 || tmp < -128) {
```
mhlakhani, line 1090: I tried to guess the right bounds here based on the original assert and int8_t bounds, but happy to adjust this based on feedback.

Jens-G: The `byte` type is like all integer types in Thrift a **signed** type. So not really sure what the intention of the test against < 256 was. Only David knows.

mhlakhani: thanks for the info, @Jens-G!

What do you suggest? I see two options:

1. make this check be as per the suggestion below. This may break existing code potentially, though I'd argue it may be buggy.
2. Just remove the assert.

Happy to do whatever you prefer.

```suggestion
  if (tmp > 127 || tmp < -128) {
```

mhlakhani: note that this is a debug assert so probably not relevant to code in production (I'm not sure how many users of thrift compile with asserts on)

mhlakhani: Sounds good - I think that sounds like option 1, so I will put up a change accordingly.

PR #3151, opened 2025-05-28
[THRIFT-5871] Improve MAX_MESSAGE_SIZE_CHECK and friends

mhlakhani:
This makes a number of changes, following the spec laid out in the ticket. The high level goal is to ensure that these checks are consistent throughout the various implementations.

Concretely:

* Some implementations checked for numBytes < 0 in the checkReadBytesAvailable() call. Do this everywhere  
* Always ensure we call checkReadBytesAvailable() with the number of bytes to read, not just container size
* The C glib binary protocol was missing some checks the compact protocol had (I just wholesale copied them, replacing 'compact' with 'binary')
* Some string length checks weren't consistently applied
* getMinSerializedSize() should generally return a minimum of 1, if my understanding of the spec and the context in which it's called in is correct.

I made one PR touching all languages, I can split into separate PRs if desired but I figured this gives a more holistic picture of the changes.

### DIFF
--- a/lib/go/thrift/json_protocol.go
+++ b/lib/go/thrift/json_protocol.go
@@ -314,11 +314,14 @@ func (p *TJSONProtocol) ReadMapBegin(ctx context.Context) (keyType TType, valueT
 	if err != nil {
 		return keyType, valueType, size, err
 	}
-	err = checkSizeForProtocol(int32(iSize), p.cfg)
+	size = int(iSize)
+
+	minElemSize := p.getMinSerializedSize(keyType) + p.getMinSerializedSize(valueType)
+	totalMinSize := int32(iSize) * minElemSize
+	err = checkSizeForProtocol(totalMinSize, p.cfg)
 	if err != nil {
 		return keyType, valueType, 0, err
 	}
-	size = int(iSize)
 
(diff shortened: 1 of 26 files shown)


lib/go/thrift/json_protocol.go
```diff
@@ -314,11 +314,14 @@ func (p *TJSONProtocol) ReadMapBegin(ctx context.Context) (keyType TType, valueT
 	if err != nil {
 		return keyType, valueType, size, err
 	}
-	err = checkSizeForProtocol(int32(iSize), p.cfg)
+	size = int(iSize)
+
+	minElemSize := p.getMinSerializedSize(keyType) + p.getMinSerializedSize(valueType)
```
fishy, line 319: for json map, there's at least another `+1` for the `:`, I think? not that that's super important though.

mhlakhani: ahh that's a good catch. Open to interpretation on how to fix this, I think I didn't want to deviate too much from existing code structure. This is tightening the lower bound. Because, at this time it also doesn't consider e.g. quoting/spacing.

fishy: yea that's my "not that that's super important" part :)

PR #3153, opened 2025-05-28
go: Add a zlib reader pool

fishy:
We implemented a zlib writer pool for default level when implementing THeader, this change also add a zlib reader pool to help speed up things when zlib is used.

Also make TZlibTransport to use the zlib writer pool when it's using the default compression level.

  

### DIFF
--- a/lib/go/thrift/zlib_pool.go
+++ b/lib/go/thrift/zlib_pool.go
@@ -1,0 +1,109 @@
+/*
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements. See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership. The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License. You may obtain a copy of the License at
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied. See the License for the
+* specific language governing permissions and limitations
+* under the License.
+ */
+
+package thrift
+
+import (
+	"compress/zlib"
+	"io"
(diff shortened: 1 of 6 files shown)


lib/go/thrift/zlib_pool.go
```diff
@@ -0,0 +1,90 @@
...
+	defer func() {
+		z.writer.Reset(nil)
+		zlibDefaultLevelWriterPool.put(&z.writer)
+	}()
+	return z.writer.Close()
+}
+
+func newZlibWriterCloserLevel(w io.Writer, level int) (*zlib.Writer, io.Closer, error) {
+	if level != zlib.DefaultCompression {
+		// not pooled
```
dcelasun, line 80: There are only 10 levels, so it might be worth doing per-level pools like:
```go
var zlibLevelWriterPool = func() (pool [10]*pool[zlib.Writer]) {
	for i := 0; i < 10; i++ {
		pool[i] = newPool(
			func() *zlib.Writer {
				return zlib.NewWriter(nil)
			},
			nil,
		)
	}
	return
}()
```

dcelasun (approved):
LGTM with nit.

lib/go/thrift/zlib_pool.go
fishy: for other levels you need to use `zlib.NewWriterLevel` over `zlib.NewWriter` in the pool generator, and `zlib.NewWriterLevel` can return error (although the error is likely only for invalid level).

also besides levels 0 though 9, there's also HuffmanOnly at -2: https://pkg.go.dev/compress/flate#HuffmanOnly

dcelasun: Well, that's Go's fault for reusing the compression *level* field for compression [strategy](https://zlib.net/manual.html#:~:text=The%20strategy%20parameter%20is%20used%20to%20tune%20the%20compression%20algorithm.) :) 

But anyway, it's fine like this.

fishy: Added. PTAL.

dcelasun: LGTM.

dcelasun (approved):

PR #3154, opened 2025-05-28
Fix cross-test

fishy:
1. Comment in the middle of multi-line shell command joined by \ does not work, change to multiple individual commands instead.
2. Fix a py 2vs3 compat issue missed in 24df0a5 used in cross runner.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -619,13 +619,12 @@ jobs:
 
       - name: Set back executable flags
         run: |
-          chmod a+x \
-            lib/java/build/run* \
-            lib/kotlin/cross-test-client/build/install/TestClient/bin/* \
-            lib/kotlin/cross-test-server/build/install/TestServer/bin/* \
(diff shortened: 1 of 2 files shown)


fishy:
> Would it make sense to also include [48b0e1c](https://github.com/apache/thrift/commit/48b0e1cb889fbe0f332742657a3e9dcb648b43fb) as part of this PR?
> 
> I am not sure how the infrastructure is set up but it does not appear that cpp cross tests are being included, perhaps after the above commit it can also be enabled if my observation is correct

@CJCombrink no that only changes which cross test for cpp to run (with or without uuid), not how it's run

to include cpp into cross test in CI, you need to add a lib-cpp workflow into https://github.com/apache/thrift/blob/master/.github/workflows/build.yml, then add cpp to servers and clients at https://github.com/apache/thrift/blob/63b7a263fc669c56fedca5d9a7310902d98df335/.github/workflows/build.yml#L564-L568

PR #3160, opened 2025-06-05
Protocol conformity and serialization tests for UUIDs

Jens-G:
This implements tests for C# and Delphi (Haxe to folllow later) to support this https://github.com/apache/thrift/pull/3144#issuecomment-2941215847 idea


CJCombrink:
Perhaps include 9bf9e2faf3a3c8f4152aa3ee681083bf56ee9cf1 since you are touching netstd. From my side it appears to be needed to make the cross tests run for it

PR #3161, opened 2025-06-05
Fix build- and dialyzer-warnings in Erlang

bjosv:
Client: Erlang

- Correcting type info in `thrift_processor` state record.
    
    The '[Support for Multiplexing Services](https://github.com/apache/thrift/commit/ae971ce)' changed so that the processor could be initiated with a multiplexed map wrapper. Fixes dialyzer warnings:
  ```
  src/thrift_processor.erl
  Line 52 Column 31: Guard test is_list(HandlerModules::atom()) can never succeed
  Line 66 Column 87: The call thrift_multiplexed_map_wrapper:fetch(ServiceName::nonempty_string(), Service::atom()) will never return since the success typing is (any(), [{_,_},...]) -> any() and the contract is (ServiceHandler, Map) -> Module when ServiceHandler :: service_handler(), Module :: module(), Map :: service_handler_map()
  ```    
 - Add type information to `thrift_reconnecting_client` state.    
    The module only handles socket transports, which uses gen_tcp.
    Fixes the build warning:
     ```
    ===> Compiling thrift
        ┌─ src/thrift_reconnecting_client.erl:
        │
     43 │  -record(state, {
        │   ╰── Warning: record state has field(s) without type information
     ```

### DIFF
--- a/lib/erl/src/thrift_processor.erl
+++ b/lib/erl/src/thrift_processor.erl
@@ -25,9 +25,9 @@
 -include("thrift_protocol.hrl").
 
 -record(thrift_processor, {
-    handler :: module(),
+    handler :: module() | thrift_multiplexed_map_wrapper:service_handler_map(),
     protocol :: term(),
-    service :: atom()
+    service :: atom() | thrift_multiplexed_map_wrapper:service_handler_map()
 }).
 
 init({_Server, ProtoGen, Service, Handler}) when is_function(ProtoGen, 0) ->


bjosv:
> The second part of this PR, the compile warning "Warning: record state has field(s) without type information", is it a fairly new thing in recent Erlang versions? Just curious. I'm a bit surprised because it's not an error to omit type information.

I've seen that there is an erl_opts `warn_untyped_record` in [rebar.conf](https://github.com/apache/thrift/blob/master/lib/erl/rebar.config#L7).
Toggling this seems also "fix" it :) i.e. nothing new in recent Erlang versions.

PR #3164, opened 2025-06-06
Try to add cpp cross tests

CJCombrink:
This PR tries to enable cpp cross tests

I am not sure if all of this is needed but I need help getting past the ssl issues, I have tried [many, many times (Build#25 already) ](https://github.com/CJCombrink/thrift/actions/runs/15484246919) and SSL is outside of my comfort zone. 

See some of my comments on my internal PR: https://github.com/CJCombrink/thrift/pull/3
It appears to be something about either expired or self signed certificates or both :shrug: 

Any help will be appreciated


.github/workflows/build.yml
```diff
@@ -577,14 +644,17 @@ jobs:
       - uses: actions/setup-java@v4
         with:
           distribution: temurin
-          # here we intentionally use an older version so that we also verify Java 17 compiles to it
```
CJCombrink, line 580: Is this comment actually true?
I know Java compiled on an old version should work with newer versions: This is not the first time I run into the issue with a old version trying something compiled with a new version.
The logs print this:
```

File: lib/cpp/test/OneWayHTTPTest.cpp [hunk 2]
```diff
@@ -202,10 +202,10 @@ BOOST_AUTO_TEST_CASE( JSON_BufferedHTTP )
 #endif
 
   {
-    std::shared_ptr<TSocket> socket(new TSocket("localhost", port));
+    std::shared_ptr<TSocket> socket(new TSocket("127.0.0.1", port));
```
CJCombrink, line 205: This appears to be an issue with `localhost` getting resolved to the IPv6 address `::1` and stuff breaks.
I am actually reconsidering this commit and instead for the pipeline purge the IPv6 `::1     localhost` from the `/etc/hosts` file in the pipeline.

test/cpp/src/TestClient.cpp
```diff
@@ -321,9 +322,15 @@ int main(int argc, char** argv) {
...
-    cout << "Client Certificate File: " << certPath << '\n';
-    cout << "Client Key         File: " << keyPath << '\n';
-    cout << "CA                 File: " << caPath << '\n';
+    auto fileExists = [](const std::string& path) {
+      std::ifstream f(path.c_str());
+      return f.good();
+    };
+
+    cout << "Client Path            : " << testDir  << '\n';
+    cout << "Client Certificate File: " << certPath << " (" << std::boolalpha << fileExists(certPath) << ")"<< '\n';
```
CJCombrink, line 331: Trying to debug SSL and cert issues I though the certs are missing in the pipeline.
Concluded through this code that it was not the issue and decided to keep this here since it can be useful in future

.github/workflows/build.yml
fishy: cc @jimexist to chime in.

fishy (approved):
lgtm modulo the java version change

.github/workflows/build.yml
jimexist: it was not a typo.

that `UnsupportedClassVersionError` likely from kotlin which isn't configured with the same version as the java lib project?

.github/workflows/build.yml
```diff
@@ -552,20 +553,86 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -yq
+          sudo apt-get install -y --no-install-recommends g++ $BUILD_DEPS locales
+          sudo locale-gen en_US.UTF-8
+          sudo locale-gen de_DE.UTF-8
```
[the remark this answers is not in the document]
CJCombrink: There is a test in ToStringTest.cpp that is part of the lib tests that requires this, else the tests fails. 
See [ToStringTest.cpp#L57](https://github.com/apache/thrift/blob/master/lib/cpp/test/ToStringTest.cpp#L57)

CJCombrink:
> lgtm modulo the java version change

@fishy Thanks for you review thus far. 
As mentioned, I want to drop the java change but something needs to be fixed to fix the java/kotlin cross tests.

The C++ SSL cross tests all still fail and I need some help to get past that. The worst case would be to mark them as known failures but ideally I would like to solve the issue.

CJCombrink:
Specifically the part that I expect to work that is not is the java Server and C++ Client for SSL.

For example in the cross tests logs (re-ordered them a bit to match pairs)

```
...
java-cpp                binary            zlib-ip                  success
java-cpp                binary            zlib-ip-ssl              failure(-6)
...
java-cpp                binary            framed-ip                success
java-cpp                binary            framed-ip-ssl            failure(-6)
...
```

fishy:
There's probably some known issue with java cross test ssl handling. If you look at other cross tests, or history or cross tests of prior runs, it seems to only work between java-java and java-kotlin, and never worked with other languages.

(by "never worked" I actually meant "have not worked for a looooong time")

CJCombrink:
> There's probably some known issue with java cross test ssl handling

Thanks for the info, it does in deed look like a known issue.
I am trying to spend some time to figure out why does cpp <-> cpp not all work and then perhaps see where I am at and maybe mark the ssl ones as know issues.

lib/cpp/src/thrift/protocol/THeaderProtocol.cpp
```diff
@@ -146,6 +146,10 @@ uint32_t THeaderProtocol::writeBinary(const std::string& str) {
   return proto_->writeBinary(str);
 }
 
+uint32_t THeaderProtocol::writeUUID(const TUuid& uuid) {
```
CJCombrink, line 149: @Jens-G This is a bugfix for the THeaderProtocol.
Can I just create a Jira Ticket for it and we close it part of this PR (once I get there) or do I need to pull this out into a new PR for the issue?

CJCombrink:
I am starting to clean this up and split out the relevant parts to relevant PRs.

@jimexist I still need some guidance on what is the correct fix for the kotlin error (https://github.com/apache/thrift/pull/3164#discussion_r2131618892)

@Jens-G Can I bundle the two fixes into one PR that also adds the cross tests (https://github.com/apache/thrift/pull/3164#discussion_r2142395247, https://github.com/apache/thrift/pull/3164#discussion_r2142390570)

CJCombrink:
Hoping for some answers but closing this PR for the more cleaned up https://github.com/apache/thrift/pull/3167

.github/workflows/build.yml
CJCombrink: I created a ticket: [THRIFT-5879: java and kotlin cross tests fail in the GitHub action](https://issues.apache.org/jira/browse/THRIFT-5879)

PR #3168, opened 2025-06-16
THRIFT-5880: ipv6 only linux systems cannot resolve 127.0.0.1

jeking3:
getaddrinfo() resolving 127.0.0.1 as a hostname on a linux system with ipv6 only defined on eth0 results in EAI_ADDRFAMILY.  This means AI_ADDRCONFIG needs to be removed to properly resolve it.  This is similar to the fix for EAI_NODATA for ipv6.


jeking3:
This change was tested live elsewhere; merging.

PR #3170, opened 2025-06-23
specify Go toolchain version

zheguang:
When building from source, `go.mod` is missing the toolchain version, resulting in incorrect Makefile generation:
```Makefile
# In Makefile line 747
golang_version = go: downloading go1.23 (darwin/arm64)
go: download go1.23 for darwin/arm64: toolchain not available
```
The error can be seen from `go` as well:
```bash
$ go verison
go: downloading go1.23 (darwin/arm64)
go: download go1.23 for darwin/arm64: toolchain not available
```
This is due to the separation of the language version and the release version of Go [1,2].  

To fix the build, this patch adds the toolchain version `1.23.0`.  

**Testing**
```bash
$ go version
go version go1.23.0 darwin/arm64
```

**References**
1. https://github.com/golang/go/issues/62278#issuecomment-1693538776
2. https://go.dev/doc/toolchain

- [ x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes) -- Trivial change
- [ x ] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"? -- N/A
- [ x ] Did you squash your changes to a single commit?  (not required, but preferred)
- [ x ] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ x ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.


go.mod
```diff
@@ -1,3 +1,4 @@
 module github.com/apache/thrift
 
 go 1.23
```
CJCombrink, line 3: I have not much experience with go but during my hackings (unrelated) found that if I set this version to 1.23.0 I get a valid makefile (See this commit: https://github.com/apache/thrift/pull/3144/commits/5d2d4adaea237b118ea8ef01bea0741635826941) 
If your change is the correct one then thanks :)

go.mod
```diff
@@ -1,3 +1,4 @@
...
+toolchain go1.23.0
```
CJCombrink, line 4: See my comment above, is this something that should not be done in all the files my commit touched?

zheguang: Good catch, thanks!  I will add them.

go.mod
zheguang: Yes, I think specifying minimum release version 1.23.0 in `go` line or `toolchain` line works.  Having toolchain line seems a bit more specific.

zheguang:
Thanks for reviewing, @fishy!

To your point though, having `toolchain` line (or release version `1.23.R` within the `go` line) will accomplish what we need here, which is the minimum required version to compile the module:  Older version will need to go through language compatibility check by Go, and if compatible, auto switch to the minimum release version [3].  

For this auto-switching to work though, we'll need specify not just the language version `1.23` but also the release version `1.23.R`.

One sticky point is whether we want this auto-switching...  Here is what I see:
1. Without auto-switching, we need to document the minimum version requirement in text
2. Or with auto-switching, we codify the minimum version, and let Go check for compatibility

I think (2) seems better.

Plan B - if we don't want to go as far as auto-switching as in (1), one idea is to improve the makefile generation to catch somewhat cryptic the toolchain-not-found error, and suggest a hard upgrade, rather than chucking up a garbled makefile:
https://github.com/apache/thrift/blob/67f828012710d05495ec240467c3750f99ddfe27/configure.ac#L380

Feel free to let me know which way we prefer.

References:
3. https://go.dev/blog/toolchain

fishy:
the problem with auto-switching is that you need to specify the version to be used when user is below that version. so if an user's local go toolchain is 1.22.x, auto switch will switch to 1.23.0.

and the problem is that we don't want them to use 1.23.0, there are several bugs, a lot of them security related, fixed in later 1.23.x versions. we want them to use the latest version of 1.23.x, or even the latest version of 1.24.x. and no one is going to update that line every time a new patch version is released.

and also, as I have said before, any version before 1.23.0 is already no longer supported and not receiving any security updates. so people should not be using those versions, and any supported version will work with thrift's go code (see our go version support policy at https://github.com/apache/thrift/blob/master/lib/go/README.md#suppored-go-releases)

Jens-G:
Rebase needed

Jens-G:
Canhes requested and no progress since June. Feel free to reopen as needed or send a new PR.

PR #3173, opened 2025-06-30
Add various/minor changes to some TProtocols for consistency and convenience

jonahaapala:
I noticed that when partial deserialization was added a new method `readFieldBeginData()` was added to elide the TField object creation. However, only the **TBinaryProtocol** takes advantage of this. This PR updates **TCompactProtocol** to do the same.

Also, while using the **TProtocol** helper method `readMap()` that takes a **ReadMapEntryCallback** I found myself wishing I could passing a separate keyCallback and valueCallback so my calling code can use method references/lambdas.

Another minor tweak in this area was the default `mapCreator` for maps and sets. In later versions of Java they added the `HashMap.newHashMap()` factory method which will size the table to prevent resizing when given an initial capacity that won't be exceeded. I don't know of an equivalent for HashSet, though the HashSet constructor that takes in a Collection also performs this optimization. I have added them here for convenience. I have noticed in the past that the Java thrift compiler initially sizes maps to 2x their size likely for this reason, so this is like an analogue to that. Perhaps the compiler should also be doing this, but I suspect it's of minor consequence either way.

While trying to utilize the **TProtocolDecorator** I was sad that none of the `skip*()` methods were delegated to. I mostly wanted `skipBinary()`, but added them all for completeness. Again, looks like these were added for partial deserialization so I assume it was just missed. I added a few more overrides to it as well to really treat it as a delegating wrapper.

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/protocol/TCompactProtocol.java
+++ b/lib/java/src/main/java/org/apache/thrift/protocol/TCompactProtocol.java
@@ -23,6 +23,7 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.UUID;
 import org.apache.thrift.TException;
+import org.apache.thrift.partial.TFieldData;
 import org.apache.thrift.transport.TTransport;
 import org.apache.thrift.transport.TTransportException;
 
@@ -536,29 +537,7 @@ public class TCompactProtocol extends TProtocol {
       return TSTOP;
     }
 
-    short fieldId;
-
-    // mask off the 4 MSB of the type header. it could contain a field id delta.
-    short modifier = (short) ((type & 0xf0) >> 4);
-    if (modifier == 0) {
-      // not a delta. look ahead for the zigzag varint field id.
-      fieldId = readI16();
-    } else {
-      // has a delta. add the delta to the last read field id.
-      fieldId = (short) (lastFieldId_ + modifier);
-    }
-
-    TField field = new TField("", getTType((byte) (type & 0x0f)), fieldId);
-
-    // if this happens to be a boolean field, the value is encoded in the type
-    if (isBoolType(type)) {
-      // save the boolean value in a special instance variable.
-      boolValue_ = (byte) (type & 0x0f) == Types.BOOLEAN_TRUE ? Boolean.TRUE : Boolean.FALSE;
-    }
-
-    // push the new field onto the field stack so we can keep the deltas going.
-    lastFieldId_ = field.id;
-    return field;
+    return new TField("", getTType((byte) (type & 0x0f)), readFieldId(type));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


fishy:
>In later versions of Java they added...

Which version is that first added? I _could_ be wrong but I believe we need to maintain compatibility with JDK 8.

jonahaapala:
> Which version is that first added

The HashMap.newHashMap factory was add in Java 19, but I assumed we couldn't use it here for backwards compatibility so I basically just copied what it does here.

fishy:
> > Which version is that first added
> 
> The HashMap.newHashMap factory was add in Java 19, but I assumed we couldn't use it here for backwards compatibility so I basically just copied what it does here.

I don't think that's in compliance with JDK's nor Apache's license

jonahaapala:
> I don't think that's in compliance with JDK's nor Apache's license

I'm just dividing the size by the load factor. It's not really ground breaking stuff...

jonahaapala:
I'll undo it if it's really a concern, but it feels like fairly intuitive code

jonahaapala:
I will just switch to 2*map size which is done elsewhere in the thrift code base

jonahaapala:
It seems like all the failures are CPP related. I'm not sure how to resolve them or why these changes cause them to fail.

jonahaapala:
hm now it's showing cross test failures.. didn't show that before 👀

lib/java/src/main/java/org/apache/thrift/protocol/TCompactProtocol.java
```diff
@@ -944,6 +923,43 @@ public int getMinSerializedSize(byte type) throws TTransportException {
...
+    // if this happens to be a boolean field, the value is encoded in the type
+    if (isBoolType(type)) {
+      // save the boolean value in a special instance variable.
+      boolValue_ = (byte) (type & 0x0f) == Types.BOOLEAN_TRUE ? Boolean.TRUE : Boolean.FALSE;
+    }
+
+    // push the new field onto the field stack so we can keep the deltas going.
+    lastFieldId_ = fieldId;
+    return fieldId;
+  }
```
jonahaapala, line 961: Just a copy of the existing code minus the TField allocation. But this should be logically equivalent.

image

jonahaapala:
still trying to get the cross tests running locally, but I don't logically see what I broke :/

fishy:
most of the cross tests are already broken, you just want to check against a previous run to see whether there are new broken cross tests.

jonahaapala:
https://github.com/apache/thrift/pull/3169

image

vs this PR:
image

Looks like the same failing cross tests

jonahaapala:
> most of the cross tests are already broken

Is there any context you can share on this? Seems concerning, no?

fishy:
> > most of the cross tests are already broken
> 
> Is there any context you can share on this? Seems concerning, no?

the context is complicated. for a while (almost a year I think) at least one of the language build/tests are failing, which prevents cross tests to run (we only run cross tests after all language builds are succeeded). when we finally fixed those and start running cross tests again, it's already failing.

for java related cross tests in particular, I _think_ it's mostly that there are some issue with java client's ssl setup in cross tests (for example, all the failures with go server and java client are with ssl).

jonahaapala:
anything else you need from me before a final review?

fishy:
> anything else you need from me before a final review?

someone more familiar with java code probably should review it instead.

jonahaapala:
@fishy is there just a queue of PR reviews that I am behind, or how can I attract a Java code reviewer?

fishy:
cc @jimexist can you take a look?

jonahaapala:
@fishy sorry to be a pest, but this is my first PR to the Thrift project so I'm wondering if you could set some expectations on how long reviews usually take? I'm not trying to be pushy, I'm just wondering.

jonahaapala:
@fishy @jimexist any updates here?

fishy (approved):

PR #3180, opened 2025-07-25
THRIFT-5888: Do not dynamically enable the GIL under free-threaded Python

lysnikolaou:
I've completed an auditing of the C extension modules and it does not appear to have any global state. Thus, we can signify that the extension does not rely on the GIL for thread safety, so that it's not dynamically reenabled when `thrift.protocal.fastbinary` is imported.  

### DIFF
--- a/lib/py/src/ext/module.cpp
+++ b/lib/py/src/ext/module.cpp
@@ -196,6 +196,10 @@ void initfastbinary() {
   if (module == nullptr)
     INITERROR;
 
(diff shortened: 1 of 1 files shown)


lysnikolaou:
Friendly ping on this. All of the CI failures appear to be unrelated.

ngoldbaum:
@lysnikolaou can you see if the failures go away when you rebase?

PR #3167, opened 2025-06-12
THRIFT-5877: Add cpp cross tests

CJCombrink:
Add cross tests to the GitHub pipeline builds

During the work two other issues were identified and implemented/fixed: Missing UUID support for THeaderProtocol and TProtocolTap. This PR includes the fixes for both.

Updated the cross tests to include cpp.
Also updated the known_issues list by adding all java<->cpp ssl related tests

The tests also identified an issue due to https://github.com/apache/thrift/pull/3165 and should be fixed as well

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? 
    - [THRIFT-5877: Add cpp cross tests](https://issues.apache.org/jira/browse/THRIFT-5877)
    - [THRIFT-5878: [c++] Add UUID support for THeaderProtocol and TProtocolTap](https://issues.apache.org/jira/browse/THRIFT-5878)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [ ] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #3165, opened 2025-06-10
THRIFT-5698: remove use of deprecated std::iterator

kou:
Client: cpp

C++17 deprecated std::iterator.

See
https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

Prior to this change, compiling while targeting C++17 or higher results in warnings.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [x] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [x] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #3182, opened 2025-07-27
Fix C++ build

mhlakhani:
https://github.com/apache/thrift/pull/3165 broke the build with c++11.

I split out the compile time fix from https://github.com/apache/thrift/pull/3167/files#diff-d8d7400cdf01511c2ad2df7caf958173c2e96e8633271709c9b0b5cd8e89215c to ensure we have a clean trunk (so I can have other changes on top). I can't take credit for the fixes.

I verified this compiles with gcc, clang (on c++11) and also on clang with c++1z  

### DIFF
--- a/lib/cpp/src/thrift/Thrift.h
+++ b/lib/cpp/src/thrift/Thrift.h
@@ -54,6 +54,9 @@ class TEnumIterator {
 public:
   using iterator_category = std::forward_iterator_tag;
   using value_type = std::pair<int, const char*>;
+  using difference_type = std::ptrdiff_t;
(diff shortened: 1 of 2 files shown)


CJCombrink:
Thanks, I am not sure what the holdup is on #3167
Nonetheless you will also need the change in "lib/cpp/src/thrift/transport/TSocketUtils.h" Same idea different file

mhlakhani:
thanks @CJCombrink ! I've pulled in those changes -- though I wasn't able to reproduce that build failure.

PR #3186, opened 2025-08-03
THRIFT-5855: Add FUZZING.md and test structures

mhlakhani:
This PR adds a basic readme describing the upcoming fuzzing support (it's forward looking, I have local commits that add support to all the languages that I'll re-test and submit once this merges).

It also adds 2 files with test structures - the main difference just being one is for languages with UUID support and one without.

I'm gonna [skip ci] here as this doesn't need CI.

### DIFF
--- a/FUZZING.md
+++ b/FUZZING.md
@@ -1,0 +1,78 @@
+# Fuzzing Apache Thrift
+
+This document describes the fuzzing infrastructure and goals for Apache Thrift. 
+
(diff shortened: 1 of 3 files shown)


mhlakhani:
@Jens-G I would appreciate your review here -- in particular, looking to see if there's any thrift feature I missed in the fuzzing files that we should test. The fuzzers mostly test serialization/deserialization right now (not networking) so I did leave out service definitions, but want to ensure I'm not missing anything else.

PR #3055, opened 2024-10-24
Add support for boringssl

layus:
This adds support for boringssl where OPENSSL_thread_stop is not implemented/defined and CONF_modules_unload is also missing. Fixed by removing these calls when boringssl is in use.

- [ ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes)
- [ ] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [ ] Did you squash your changes to a single commit?  (not required, but preferred)
- [X] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [X] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #3194, opened 2025-08-25
Add support for boringssl

Dev-iL:
This is a rebased copy of #3055 (an older PR which appears to be abandoned) with a couple of additional comments.


Dev-iL:
What does the CI tell us exactly? The cpp (cmake) task seems to fine, but what else?

PR #3210, opened 2025-09-15
THRIFT-5892: Always define PY_SSIZE_T_CLEAN before including Python.h

csringhofer:
The error THRIFT-5488 tried to solve resurfaced in a spefic environment: Python3.11.13, chainguard based docker image, GCC 15

Defining PY_SSIZE_T_CLEAN in module.cpp fixed the issue in that environment. Also defined it in types.cpp to consistently define it before including Python.h, which seems to be the recommended way: https://docs.python.org/3.11/c-api/intro.html#include-files

It is not clear to me why this was a problem in some environments and not in others. My guess is that including Python.h with different defines leads to different behavior based on GCC version and options.

### DIFF
--- a/lib/py/src/ext/module.cpp
+++ b/lib/py/src/ext/module.cpp
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+#define PY_SSIZE_T_CLEAN
 #include <Python.h>
 #include "types.h"
 #include "binary.h"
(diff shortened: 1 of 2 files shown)


fishy (approved):
(unrelated to this PR) it looks like python is not in any crosstests and we probably should fix that.

PR #3211, opened 2025-09-16
Support socket activation by  fd passing

fedegiova:
This change adds another constructor to the TServerSocket class to support passing a socket that's already bound with the goal to support socket-activation by systemd. If a service is configured to use socket-activation systemd will bind the socket and pass an FD that's already bound as an environment variable when the service is started with the goal to parallelize boot startup and remove ordering dependencies between server ( bind ) and client ( connect).

The new enum type introduced is used mostly to disambiguate constructors.

### DIFF
--- a/lib/cpp/src/thrift/transport/TServerSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TServerSocket.cpp
@@ -117,7 +117,8 @@ TServerSocket::TServerSocket(int port)
     listening_(false),
     interruptSockWriter_(THRIFT_INVALID_SOCKET),
     interruptSockReader_(THRIFT_INVALID_SOCKET),
-    childInterruptSockWriter_(THRIFT_INVALID_SOCKET) {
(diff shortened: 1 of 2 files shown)


fedegiova:
I'll create the ticket and update the MR title. Just waiting for account approval

fedegiova:
Sorry for not having created the Jira ticket, but the request for the account has not been approved yet.

Jens-G (changes requested):
The enum partially mirrors data we already have. i understand the idea but I would love to see some tests for the whole construct to make sure not only it works as intended but also does not break other things.

fedegiova:
@Jens-G  Hello! I've added some tests. The only point that's not clear to me is the "websocket" entry in the "trasport" protocols in `tests.json`. 

If I leave it and checkout master it's not used (TesteClient is not invoked with it). If I add the "domain-sockeactivated" in the "socket" options, then it's used, but some tests fails since TestClient does not supports websocket.

PR #3212, opened 2025-09-25
go: reduce memory consumption of small fields

JamesGuthrie:
In the context of THRIFT-5322 commit 37c2ceb7 introduced the safeReadBytes method, which prevents allocating very large buffers on malformed input by using a bytes.Buffer as the backing buffer to contain data.

Due to how io.CopyN and bytes.Buffer interact, and bytes.Buffer's grow behaviour, the smallest buffer size is 512 bytes, but typically buffer sizes are at least 1024 bytes. In degenerate cases (reading many small fields) this can cause significant bloat, e.g. a value of 8 bytes in length sits in a buffer of 1024 bytes.

This change eliminates bloat by allocating an array of exactly the desired size, when the desired size is below bytes.Buffer's minimum buffer size.

### DIFF
--- a/lib/go/thrift/binary_protocol.go
+++ b/lib/go/thrift/binary_protocol.go
@@ -603,33 +603,8 @@ func safeReadBytes(size int32, trans io.Reader) ([]byte, error) {
 		_, err := io.CopyN(buf, trans, int64(size))
 		return buf.Bytes(), err
 	}
-	return readExactBytes(size, trans)
-}
-
-// readExactBytes reads a maximum of size bytes from r until an error or EOF
-// and returns the data it read.
-// The implementation is similar to io.ReadAll, but it uses a buffer of
-// exactly size bytes to prevent bloat when reading short sequences.
-// A successful call returns err == nil, not err == EOF. Because readExactBytes
-// is defined to read from src until EOF, it does not treat an EOF from Read
-// as an error to be reported.
-func readExactBytes(size int32, r io.Reader) ([]byte, error) {
 	// Allocate size bytes
-	b := make([]byte, 0, size)
-	for {
-		n, err := r.Read(b[len(b):cap(b)])
-		b = b[:len(b)+n]
-		if err != nil {
-			if err == io.EOF {
-				err = nil
-			}
-			return b, err
-		}
-
-		if len(b) == cap(b) {
-			break
-		}
-	}
-	return b, nil
-
+	b := make([]byte, size)
+	n, err := io.ReadFull(trans, b)
+	return b[:n], err
 }


JamesGuthrie:
For additional context and real-world anchoring, the memory bloat was discovered in the context of analysing memory usage when reading parquet files with the apache/arrow-go library.

The following screenshots show the heap profile of opening a ~1GB parquet file with 18300 row groups using [apache/arrow-go/v18/parquet/file.NewParquetReader](https://pkg.go.dev/github.com/apache/arrow-go/v18/parquet/file#NewParquetReader), both before and after this change.

The code used to obtain these results:

```go
package main

import (
	"os"
	"runtime"
	"runtime/pprof"

	"github.com/apache/arrow-go/v18/parquet"
	"github.com/apache/arrow-go/v18/parquet/file"
)

func main() {
	f, err := os.Open("/path/to/sensor_readings_1gb.parquet")
	if err != nil {
		panic(err.Error())
	}
	r, err := file.NewParquetReader(f, file.WithReadProps(&parquet.ReaderProperties{
		BufferSize:            4 * 1024 * 1024,
		BufferedStreamEnabled: true,
	}))
	if err != nil {
		panic(err.Error())
	}
	p, err := os.Create("/path/to/heap.pprof")
	if err != nil {
		panic(err.Error())
	}
	defer p.Close()
	runtime.GC()
	err = pprof.WriteHeapProfile(p)
	if err != nil {
		panic(err.Error())
	}
	runtime.GC()
	runtime.KeepAlive(r)
}
```

The test file is available at this link: https://jg-test-document-public-bucket.s3.eu-central-1.amazonaws.com/sensor_readings_1gb.parquet

Before:

Screenshot 2025-09-25 at 17 49 38

After:

Screenshot 2025-09-25 at 17 49 49

Note how memory usage of the parquet file comes primarily from metadata, the size of which is reduced from ~460MB to ~60MB.

lib/go/thrift/binary_protocol.go
```diff
@@ -588,18 +588,48 @@ func (p *TBinaryProtocol) getMinSerializedSize(ttype TType) int32 {
...
+// and returns the data it read.
+// The implementation is similar to io.ReadAll, but it uses a buffer of
+// exactly size bytes to prevent bloat when reading short sequences.
+// A successful call returns err == nil, not err == EOF. Because readExactBytes
+// is defined to read from src until EOF, it does not treat an EOF from Read
+// as an error to be reported.
+func readExactBytes(size int32, r io.Reader) ([]byte, error) {
+	// Allocate size bytes
+	b := make([]byte, 0, size)
+	for {
```
fishy, line 619: use [io.ReadFull](https://pkg.go.dev/io#ReadFull) instead of this for loop?

JamesGuthrie: Yes, that is a better option. Thank you.

fishy (approved):

JamesGuthrie:
Thanks for the prompt review feedback and merge! I appreciate it.

PR #3220, opened 2025-10-17
THRIFT-5896: Fix race condition in TServerSocket.Addr()

AtharvaChiplunkar12:
Client: go

  ## Problem
The `Addr()` method in `server_socket.go` was accessing the `p.listener` field without mutex protection, which could cause data races when multiple goroutines access it concurrently (e.g., one goroutine calling `Addr()` while another calls `Close()`, `Open()`, or `Listen()`).

## Solution
- Added `RLock()`/`RUnlock()` protection around listener access for thread-safety
- Changed direct nil check to use `IsListening()` method for consistency with the codebase
- Follows the same locking pattern used in other methods (`Accept()`, `Close()`, `Listen()`, `Open()`)

### DIFF
--- a/lib/go/thrift/server_socket.go
+++ b/lib/go/thrift/server_socket.go
@@ -26,12 +26,12 @@ import (
 )
 
 type TServerSocket struct {
-	listener      net.Listener
 	addr          net.Addr
 	clientTimeout time.Duration
 
-	// Protects the interrupted value to make it thread safe.
+	// Protects the listener and interrupted fields to make them thread safe.
 	mu          sync.RWMutex
+	listener    net.Listener
 	interrupted bool
 }
 
@@ -110,7 +110,9 @@ func (p *TServerSocket) Open() error {
 }
 
 func (p *TServerSocket) Addr() net.Addr {
-	if p.listener != nil {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	if p.IsListening() {
 		return p.listener.Addr()
 	}
 	return p.addr


lib/go/thrift/server_socket.go
```diff
@@ -110,7 +110,9 @@ func (p *TServerSocket) Open() error {
 }
 
 func (p *TServerSocket) Addr() net.Addr {
-	if p.listener != nil {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	if p.IsListening() {
```
fishy: If you look at the comment about mu at https://github.com/apache/thrift/blob/d9a97c1610372eeb6db9f1a54e590a0a89067aea/lib/go/thrift/server_socket.go#L33-L34, it's only supposed to protect `interrupted` field.

I guess that has evolved over the years, and we already have existing code to use `mu` to also protect `listener`.

so this is the time we should update the comment to reflect that. please move the `listener` field to the same group with `mu` and `interrupted`, and also update the comment.

fishy (approved):

glaubitz:
Does this PR have a CVE assigned? And is there any pointer to which commit fixed CVE-2026-41602?

fishy:
I don't believe this has a CVE assigned, and this is unlikely to be related to CVE-2026-41602.

glaubitz:
Can you point me to the commit that fixes CVE-2026-41602? I can’t find it anywhere in the changelog.Message ID: ***@***.***>

Jens-G:
The commits are not marked primarily for the [technical reason that it would break git history](https://www.apache.org/security/committers.html#complete). Also, binding a CVE to single commit ids is not always possible. Finally, there are opinions on both end of the spectrum why this is a great or a bad idea (pick your side) at all. The CVE you mentioned is fixed in 0.23.0 and master.

glaubitz:
> The commits are not marked primarily for the [technical reason that it would break git history](https://www.apache.org/security/committers.html#complete). Also, binding a CVE to single commit ids is not always possible. Finally, there are opinions on both end of the spectrum why this is a great or a bad idea (pick your side) at all. The CVE you mentioned is fixed in 0.23.0 and master.

The problem is that enterprise distributions can't just easily update to the latest and greatest upstream versions. We have to cherry-pick patches and that usually works for 99% of the upstream projects. I'm quite surprised that the Apache Foundation handles it differently and expects downstream just to keep updating the whole software package to fix a CVE.

I'll try to find the commit myself then. It fixes a buffer overflow, so it should be a single identifiable patch.

PR #3328, opened 2026-03-03
THRIFT-5922: made http header lookup case insensitive

Nighmared:
Client: lua

  
The http transport client breaks when the traffic is routed through a proxy that changes the casing of the header keys. This is not only incorrect but also different from the implementation in all other languages.
This pull request changes this, so http headers are now stored/retrieved in a case insensitive matter.
Since there are no unit tests for lua (that i could find?) I did not add tests that isolate this bug, but it's quite obvious from looking at the source.

### DIFF
--- a/lib/lua/THttpTransport.lua
+++ b/lib/lua/THttpTransport.lua
@@ -42,6 +42,15 @@ function THttpTransport:new(obj)
   return TTransportBase.new(self, obj)
 end
 
+local function THttpHeaders()
+    local data = {}
+    return setmetatable({}, {
+        __index = function(_, key) return data[string.lower(key)] end,
+        __newindex = function(_, key, value) data[string.lower(key)] = value end,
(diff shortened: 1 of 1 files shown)


Jens-G:
Reference: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2

lib/lua/THttpTransport.lua
thomasbruggink: should this line also be comparing lowercase? 
now all headers are stored lower but this check is case sensitive

Nighmared: @thomasbruggink what is the cleanest way to update this? should I open a new issue on Jira?

thomasbruggink: I'm not a thrift core developer but since this has already been released I think making a new ticket is a good idea

PR #3347, opened 2026-03-17
Build & Publish Thrift Maven Plugin

jmkeyes:
This change builds the `thrift-maven-plugin` in the `contrib` folder with Maven by default.

The intention is to keep [`org.apache.thrift:thrift-maven-plugin`](https://mvnrepository.com/artifact/org.apache.thrift/thrift-maven-plugin) updated as Thrift evolves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,12 @@ if(BUILD_PYTHON)
     endif()
 endif()
 
+if(BUILD_CONTRIB)
+    if(BUILD_MAVEN_PLUGIN)
+        add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/contrib/thrift-maven-plugin)
+    endif()
+endif()
+
 # Create the uninstall target
 add_custom_target(uninstall "${CMAKE_COMMAND}" -P "${PROJECT_SOURCE_DIR}/build/cmake/uninstall.cmake")
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,6 +29,10 @@ if WITH_TUTORIAL
 SUBDIRS += tutorial
 endif
 
+if WITH_CONTRIB
+SUBDIRS += contrib
+endif
+
 clean-local:
 	$(RM) -r vendor/
 
--- a/aclocal/ax_prog_maven.m4
+++ b/aclocal/ax_prog_maven.m4
@@ -1,0 +1,54 @@
+# SYNOPSIS
+#
+#   AX_PROG_MAVEN_VERSION([VERSION],[ACTION-IF-TRUE],[ACTION-IF-FALSE])
+#
+# DESCRIPTION
+#
+#   Makes sure that maven supports the version indicated. If true the shell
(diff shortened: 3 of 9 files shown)


Jens-G:
1. What means "keep updated as Thrift evolves"?

2. Would the patch be compatible with what the [ASF ReleasePolicy](https://www.apache.org/legal/release-policy.html#publication) has to say about what can and what cannot be published?

jmkeyes:
> 1. What means "keep updated as Thrift evolves"?

I would like for the `thrift-maven-plugin` to be published along with new versions of Thrift so I can generate Java bindings for my applications with a native Maven plugin.

> 2. Would the patch be compatible with what the [ASF ReleasePolicy](https://www.apache.org/legal/release-policy.html#publication) has to say about what can and what cannot be published?

I believe so but I don't know for sure. I saw that the Thrift Maven plugin was available within the Thrift project's source code so I created a patch to build it in a way similar to other Java parts of Thrift.

If the Maven plugin in Thrift's source code is not compliant with the Apache Foundation's release policy, then I'd like to know what I could change to make it acceptable, as the Maven plugin provides a lot of value.

Jens-G:
I just wonder because we usually publish the Java **packages** via maven. Als I'm being told the maven **plugin** did not change for a long time. Disclaimer: I'm no Java user myself, so pls bear with me if I misrepresent sth and/or my knowledge is outdated. I might add that the Java publishing is always kind of a hassle, so if we can make it better, I'm open to it.

@jeking3  thoughts?

Jens-G:
```
Total Test time (real) = 239.14 sec
The following tests FAILED:
	 58 - ThriftMavenPluginTest (BAD_COMMAND)
Errors while running CTest
Output from these tests are in: C:/projects/build/MSVC2022/x64/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
C:\projects\thrift>set lastexitcode=8 
C:\projects\thrift>set  1>C:\Users\appveyor\AppData\Local\Temp\1\tmpF416.tmp 
```

All of the AppVeyor builds fail?

ctubbsii:
As an alternative, I would propose deleting the thrift-maven-plugin, or moving it to its own separate repository, instead of shipping it in the contrib directory. Its functions are trivially replaced by exec-maven-plugin and build-helper-maven-plugin, so its existence doesn't really offer much value and its build probably should not complicate the main thrift build at all.

Jens-G:
Set to draft. Fix errors and/or decide how to proceed.

slachiewicz:
#3789 takes the deletion route @ctubbsii proposed above: it removes `contrib/thrift-maven-plugin` and documents the `exec-maven-plugin` + `build-helper-maven-plugin` equivalent in `lib/java/README.md`. That also removes the `ThriftMavenPluginTest` MSVC failure this PR has been stuck on, since there is no plugin left to build.

Tracked as [THRIFT-6196](https://issues.apache.org/jira/browse/THRIFT-6196).

PR #3379, opened 2026-04-07
Replace addr with factory in TServerSocket

qr243vbi:

### DIFF
--- a/lib/go/thrift/server_socket.go
+++ b/lib/go/thrift/server_socket.go
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+
 package thrift
 
 import (
@@ -26,6 +27,8 @@ import (
 )
 
 type TServerSocket struct {
+	// TServerSocketListenerFactory abstracts how listeners are created.
+	listenerFactory      func(net.Addr) (net.Listener, error)
 	addr          net.Addr
 	clientTimeout time.Duration
 
@@ -44,26 +47,60 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
 	if err != nil {
 		return nil, err
 	}
-	return &TServerSocket{addr: addr, clientTimeout: clientTimeout}, nil
+
+	return NewTServerSocketFromAddrTimeout(addr, clientTimeout), nil
 }
 
+// NewTServerSocketFromAddrTimeout returns TServerSocket
 // Creates a TServerSocket from a net.Addr
 func NewTServerSocketFromAddrTimeout(addr net.Addr, clientTimeout time.Duration) *TServerSocket {
-	return &TServerSocket{addr: addr, clientTimeout: clientTimeout}
+	listenerFactory := func(addr net.Addr) (net.Listener, error) {
+		return net.Listen(addr.Network(), addr.String())
+	}
+
+	return NewTServerSocketFromFactoryTimeout(listenerFactory, addr, clientTimeout)
 }
 
-func (p *TServerSocket) Listen() error {
+// NewTServerSocketFromFactoryTimeout creates TServerSocket via a listener factory.
+//
+// Allows full customization (TLS, mocks, unix sockets, windows named pipes, etc.)
+func NewTServerSocketFromFactoryTimeout(listenerFactory func(addr net.Addr) (listener net.Listener, err error), addr net.Addr, clientTimeout time.Duration) *TServerSocket {
+	return &TServerSocket{
+		listenerFactory:       listenerFactory,
+		addr:          addr,
+		clientTimeout: clientTimeout,
+	}
+}
+
+func (p *TServerSocket) try_listen(raise bool) error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
-	if p.IsListening() {
+
+	if p.listener != nil {
+		if (raise) {
+			return NewTTransportException(ALREADY_OPEN, "Server socket already open")
+		}
 		return nil
 	}
-	l, err := net.Listen(p.addr.Network(), p.addr.String())
+
+	l, err := p.listenerFactory(p.addr)
 	if err != nil {
 		return err
 	}
+
 	p.listener = l
+	p.interrupted = false
 	return nil
+}
+
+// Open does try to listen and return on failure
+// Connects the socket, creating a new socket object if necessary.
+func (p *TServerSocket) Open() error {
+	return p.try_listen(true /* raise error if listening */)
+}
+
+func (p *TServerSocket) Listen() error {
+	return p.try_listen(false /* do not raise error if listening */)
 }
 
 func (p *TServerSocket) Accept() (TTransport, error) {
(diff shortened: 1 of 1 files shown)


qr243vbi:
This allows to use named pipes from go-winio.

lib/go/thrift/server_socket.go
```diff
@@ -25,16 +26,20 @@ import (
 	"time"
 )
 
+// TServerSocketListenerFactory abstracts how listeners are created.
+type TServerSocketListenerFactory func(listen bool) (net.Addr, net.Listener, error)
```
fishy: I kind of get your idea but I still find this confusing as a public API. I think you are just trying to squash 2 functions into 1? e.g. either:
1. start listening, return _configured_ listener address, listener, and error
or
2. just return the _configured_ listener address

note that `Listener` interface also has api to return the address, but that can be different from the configured address (`:0` vs. the actually randomly picked port, for example).

I also don't like boolean args, as they almost always force you to dig into function signature to understand what `true` and `false` means when you read the code.

qr243vbi: Another solution is to add a method that accepts a simplified signature function and an address, and then build a factory function.

lib/go/thrift/server_socket.go
```diff
@@ -26,15 +27,17 @@ import (
 )
 
 type TServerSocket struct {
-	addr          net.Addr
+	// TServerSocketListenerFactory abstracts how listeners are created.
+	factory       func(bool) (net.Addr, net.Listener, error)
```
fishy: now this is no longer _public_ api, but just `func(bool) (net.Addr, net.Listener, error)` is still hard to read (figure out what's going on). please typedef this into an internal type (uncapitalized), and write doc comments to explain it.

lib/go/thrift/server_socket.go
```diff
@@ -44,28 +47,75 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
...
+		}
+		return addr, listener, err
+	}
+
+	return NewTServerSocketFromFactoryTimeout(factory, clientTimeout)
 }
 
-func (p *TServerSocket) Listen() error {
+// Allows full customization (TLS, mocks, unix sockets, windows named pipes, etc.)
+func NewTServerSocketFromFactoryAddrTimeout(proc func(addr net.Addr) (listener net.Listener, err error),addr net.Addr, clientTimeout time.Duration) *TServerSocket {
```
fishy: 1. maybe rename `proc` to `listen`? you are essentially asking for a custom listener implementation here, not for a processor.
2. space between `,` and `addr net.Addr`.
3. we probably should also reword `NewTServerSocketFromFactoryAddrTimeout` slightly. is this considered a factory in traditional/conventional sense? I don't think we are actually going to create a lot of listeners on the fly so probably not. maybe `NewTServerSocketFromListner`?

lib/go/thrift/server_socket.go
```diff
@@ -44,28 +47,75 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
...
+			listener, err = net.Listen(addr.Network(), addr.String())
+		}
+		return addr, listener, err
+	}
+
+	return NewTServerSocketFromFactoryTimeout(factory, clientTimeout)
 }
 
-func (p *TServerSocket) Listen() error {
+// Allows full customization (TLS, mocks, unix sockets, windows named pipes, etc.)
```
fishy: also, I know the old code don't always follow the go's doc comment convention, but please follow that for new code: https://go.dev/doc/comment

lib/go/thrift/server_socket.go
```diff
@@ -44,28 +47,75 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
...
 		return err
 	}
+
 	p.listener = l
+	p.interrupted = false
 	return nil
 }
 
+func (p *TServerSocket) Open() error {
+	return p.try_listen(true)
```
fishy: also, to mitigate the readability issues of boolean args I mentioned earlier, when we still end up using boolean args, I'd strongly prefer to always document the arg name at the callsite:
```suggestion
	return p.try_listen(true /* raise */)
```
same for all other callsites with boolean args in this PR.

lib/go/thrift/server_socket.go
qr243vbi: @fishy, suggest changes, please

lib/go/thrift/server_socket.go
qr243vbi: @fishy, suggest changes, please

lib/go/thrift/server_socket.go
```diff
@@ -44,28 +47,62 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
...
+		}
+		return addr, listener, err
+	}
+
+	return NewTServerSocketFromFactoryTimeout(factory, clientTimeout)
 }
 
-func (p *TServerSocket) Listen() error {
+// Allows full customization (TLS, mocks, unix sockets, windows named pipes, etc.)
+func NewTServerSocketFromFactoryTimeout(factory func(listen bool) (addr net.Addr, listener net.Listener, err error), clientTimeout time.Duration) *TServerSocket {
```
fishy: you are changing this back. the previous version passing `addr` and `func() (net.Listener, error)` as two separate args is so much better.

lib/go/thrift/server_socket.go
fishy: similarly here, just store `addr` and `func() (net.Listener, error)` separately

lib/go/thrift/server_socket.go
fishy: go's doc comment require the comment starts with `// NewTServerSocketFromFactoryTimeout ...`

lib/go/thrift/server_socket.go
```diff
@@ -89,49 +122,41 @@ func (p *TServerSocket) Accept() (TTransport, error) {
...
 func (p *TServerSocket) Addr() net.Addr {
 	p.mu.RLock()
 	defer p.mu.RUnlock()
-	if p.IsListening() {
+
+	if p.listener != nil {
 		return p.listener.Addr()
 	}
-	return p.addr
+	addr, _, _ := p.factory(false)
```
fishy: this line no longer compiles?

lib/go/thrift/server_socket.go
```diff
@@ -44,28 +47,58 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
...
+
+	if p.listener != nil {
+		if (raise) {
+		        return NewTTransportException(ALREADY_OPEN, "Server socket already open")
+		}
 		return nil
 	}
-	l, err := net.Listen(p.addr.Network(), p.addr.String())
+
+	_, l, err := p.factory(true)
```
fishy: this line also doesn't compile any more.

lib/go/thrift/server_socket.go
```diff
@@ -44,28 +47,61 @@ func NewTServerSocketTimeout(listenAddr string, clientTimeout time.Duration) (*T
...
+	factory := func(addr net.Addr) (net.Listener, error) {
+		return net.Listen(addr.Network(), addr.String())
+	}
+
+	return NewTServerSocketFromFactoryTimeout(factory, addr, clientTimeout)
 }
 
-func (p *TServerSocket) Listen() error {
+// NewTServerSocketFromFactoryTimeout returns TServerSocket
+// Allows full customization (TLS, mocks, unix sockets, windows named pipes, etc.)
```
fishy: Something like:
```suggestion
// NewTServerSocketFromFactoryTimeout creates TServerSocket via a listener factory.
//
// Allows full customization (TLS, mocks, unix sockets, windows named pipes, etc.)
```

fishy (commented):
I still don't think "factory" is the best term here, maybe "listener" or something like that?

Also please squash your commits.

qr243vbi:
> I still don't think "factory" is the best term here, maybe "listener" or something like that?
> 
> Also please squash your commits.

Right

fishy:
the code doesn't compile: https://github.com/apache/thrift/actions/runs/24448228720/job/71489703825?pr=3379

qr243vbi:
> the code doesn't compile: https://github.com/apache/thrift/actions/runs/24448228720/job/71489703825?pr=3379

fixed

fishy:
> > the code doesn't compile: https://github.com/apache/thrift/actions/runs/24448228720/job/71489703825?pr=3379
> 
> fixed

still doesn't work: https://github.com/apache/thrift/actions/runs/24713839666/job/72357334569?pr=3379

please test it locally before asking for review.

qr243vbi:
> > > the code doesn't compile: https://github.com/apache/thrift/actions/runs/24448228720/job/71489703825?pr=3379
> > 
> > 
> > fixed
> 
> still doesn't work: https://github.com/apache/thrift/actions/runs/24713839666/job/72357334569?pr=3379
> 
> please test it locally before asking for review.

Fixed again

fishy (approved):

PR #3387, opened 2026-04-09
THRIFT-5941: Add Ruby ext cppcheck coverage

kpumuk:
Introduce `cppcheck` SCA for the Ruby native extension under `lib/rb/ext` and fix the existing findings so it can run in CI without noise.

Offenses addressed:

```
lib/rb/ext/compact_protocol.c:128:24: error: Shifting signed 32-bit value by 31 bits is undefined behaviour [shiftTooManyBitsSigned]
  return (n << 1) ^ (n >> 31);
                       ^
lib/rb/ext/compact_protocol.c:132:24: error: Shifting signed 64-bit value by 63 bits is undefined behaviour [shiftTooManyBitsSigned]
  return (n << 1) ^ (n >> 63);
                       ^
lib/rb/ext/compact_protocol.c:427:33: error: Signed integer overflow for expression '-(n&1)'. [integerOverflow]
  return (((uint32_t)n) >> 1) ^ -(n & 1);
                                ^
lib/rb/ext/struct.c:255:11: style: The scope of the variable 'key' can be reduced. [variableScope]
    VALUE key;
          ^
lib/rb/ext/struct.c:256:11: style: The scope of the variable 'val' can be reduced. [variableScope]
    VALUE val;
          ^
lib/rb/ext/struct.c:492:9: style: The scope of the variable 'i' can be reduced. [variableScope]
    int i;
        ^
lib/rb/ext/struct.c:531:9: style: The scope of the variable 'i' can be reduced. [variableScope]
    int i;
        ^
lib/rb/ext/struct.c:558:9: style: The scope of the variable 'i' can be reduced. [variableScope]
    int i;
        ^
lib/rb/ext/thrift_native.c:124:0: style: The function 'Init_thrift_native' is never used. [unusedFunction]
RUBY_FUNC_EXPORTED void Init_thrift_native(void) {
^
```

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-5941](https://issues.apache.org/jira/browse/THRIFT-5941)

### DIFF
--- a/lib/rb/ext/compact_protocol.c
+++ b/lib/rb/ext/compact_protocol.c
@@ -124,21 +124,29 @@ static void write_field_begin_internal(VALUE self, VALUE type, VALUE id_value, V
   SET_LAST_ID(self, id_value);
 }
 
-static int32_t int_to_zig_zag(int32_t n) {
-  return (n << 1) ^ (n >> 31);
+static uint32_t int_to_zig_zag(int32_t n) {
+  return (((uint32_t)n) << 1) ^ (0U - (uint32_t)(n < 0));
 }
 
 static uint64_t ll_to_zig_zag(int64_t n) {
-  return (n << 1) ^ (n >> 63);
+  return (((uint64_t)n) << 1) ^ (0ULL - (uint64_t)(n < 0));
+}
+
+static uint32_t message_seqid_to_varint32(int32_t seqid) {
+  return seqid < 0 ? (uint32_t)((int64_t)seqid + (INT64_C(1) << 32)) : (uint32_t)seqid;
+}
+
+static int32_t message_seqid_from_varint32(uint32_t seqid) {
+  return seqid > INT32_MAX ? (int32_t)((int64_t)seqid - (INT64_C(1) << 32)) : (int32_t)seqid;
 }
 
 static void write_varint32(VALUE transport, uint32_t n) {
   while (true) {
-    if ((n & ~0x7F) == 0) {
-      write_byte_direct(transport, n & 0x7f);
+    if ((n & ~0x7FU) == 0U) {
+      write_byte_direct(transport, n & 0x7FU);
       break;
     } else {
-      write_byte_direct(transport, (n & 0x7F) | 0x80);
+      write_byte_direct(transport, (n & 0x7FU) | 0x80U);
       n = n >> 7;
     }
   }
@@ -146,11 +154,11 @@ static void write_varint32(VALUE transport, uint32_t n) {
 
 static void write_varint64(VALUE transport, uint64_t n) {
   while (true) {
-    if ((n & ~0x7F) == 0) {
-      write_byte_direct(transport, n & 0x7f);
+    if ((n & ~0x7FULL) == 0ULL) {
+      write_byte_direct(transport, n & 0x7FULL);
       break;
     } else {
-      write_byte_direct(transport, (n & 0x7F) | 0x80);
+      write_byte_direct(transport, (n & 0x7FULL) | 0x80ULL);
       n = n >> 7;
     }
   }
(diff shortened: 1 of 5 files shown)


lib/rb/ext/compact_protocol.c
```diff
@@ -419,17 +419,17 @@ static char read_byte_direct(VALUE self) {
...
 }
 
-static int32_t zig_zag_to_int(int32_t n) {
-  return (((uint32_t)n) >> 1) ^ -(n & 1);
+static int32_t zig_zag_to_int(uint32_t n) {
+  return (int32_t)((n >> 1) ^ (0U - (n & 1U)));
 }
 
-static int64_t read_varint64(VALUE self) {
+static uint64_t read_varint64(VALUE self) {
```
Jens-G, line 439: Why do we change that to `uint64_t`? That introduces a lot of implicit things going on. Do we really want that?

lib/rb/ext/compact_protocol.c
Jens-G: cast to int32_t seems stale: 
- read_varint64() returns uint64_t
- zig_zag_to_int() expects uint32_t

lib/rb/ext/compact_protocol.c
kpumuk: `read_varint64()` should return `uint64_t` because it reads raw varint bits, not a signed Thrift value yet. For values like the ZigZag encoding of `INT64_MIN`, the wire value is `0xffffffffffffffff`, which is representable as `uint64_t` but not as a meaningful positive `int64_t`.

UBSan reports the problem in the 64-bit decode path when `read_varint64()` returns `int64_t` and the raw unsigned value is forced through a signed type too early:

```
runtime error: implicit conversion from type 'uint64_t' ... value 18446744073709551615 to type 'int64_t' ... changed the value to -1
```

* `read_varint64() -> uint64_t` for raw varint bits
* `zig_zag_to_ll(uint64_t) -> int64_t` for the actual signed decode

lib/rb/ext/compact_protocol.c
kpumuk: Nice catch. I was following the cppcheck complaints and missed this :-(
I also noticed that we do not validate the size of varint in Ruby (cpp fails on >10 bytes), but will leave it for the future.

lib/rb/ext/compact_protocol.c
kpumuk: With the last amend:

* All conversions are explicit, including (potentially ~overkill~ unnecessary `seqid` overflow, mirroring logic in Ruby)
* Separated int32 from int64 path via `read_varint32` / `read_varint64`
* ZigZag decode uses unsigned ints

Still lacking:

* Size enforcement (C++ enforces 5 bytes for int32, 10 bytes for int64, binary/name length to `INT32_MAX` via signed out `int32_t` and `>= 0` comparison). Semantically, we should fail on `> INT32_MAX`

lib/rb/spec/compact_protocol_spec.rb
```diff
@@ -71,6 +81,46 @@
...
+      trans = Thrift::MemoryBufferTransport.new
+      proto = Thrift::CompactProtocol.new(trans)
+
+      proto.send(writer(primitive_type), INTEGER_BOUNDARY_TESTS.fetch(primitive_type).first)
+      expect(trans.read(trans.available).bytes).to eq(expected_bytes)
+    end
+  end
+
+  it "should decode i32 minima from direct canonical zigzag bytes" do
+    trans = Thrift::MemoryBufferTransport.new
```
Jens-G, line 107: Can we have that for i64 too?

Jens-G (approved):
+1 LGTM with one extra request

lib/rb/spec/compact_protocol_spec.rb
kpumuk: Added

PR #3389, opened 2026-04-11
Add recursion depth limit to Node.js protocol skip()

Jens-G:
Client: nodejs

Add a depth parameter to skip() in TBinaryProtocol, TCompactProtocol, and TJSONProtocol to prevent stack exhaustion from deeply nested structures. Limit set to 64, matching the C++ and Java implementations.

### DIFF
--- a/lib/nodejs/lib/thrift/binary_protocol.js
+++ b/lib/nodejs/lib/thrift/binary_protocol.js
@@ -317,7 +317,14 @@ TBinaryProtocol.prototype.getTransport = function () {
   return this.trans;
 };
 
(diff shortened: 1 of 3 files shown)


jimexist (approved):
not sure if 64 is meant to be made configurable?

Jens-G:
The limit of 64 matches `TConfiguration.DEFAULT_RECURSION_DEPTH` which is the standard default across languages. Node.js currently has no `TConfiguration` infrastructure, so making it configurable would require adding that first — feels offtopic here. But yes, we should keep that in mind.

PR #3390, opened 2026-04-11
Enable TLS hostname verification in TSSLTransportFactory

Jens-G:
Client: java

Set EndpointIdentificationAlgorithm to HTTPS on SSLSocket parameters in createClient() so that the server certificate CN/SAN is validated against the target hostname during TLS handshake.

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/transport/TSSLTransportFactory.java
+++ b/lib/java/src/main/java/org/apache/thrift/transport/TSSLTransportFactory.java
@@ -30,6 +30,7 @@ import java.security.KeyStore;
 import java.util.Arrays;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLServerSocket;
 import javax.net.ssl.SSLServerSocketFactory;
 import javax.net.ssl.SSLSocket;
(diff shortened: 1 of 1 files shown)


lib/java/src/main/java/org/apache/thrift/transport/TSSLTransportFactory.java
```diff
@@ -343,6 +344,9 @@ private static TSocket createClient(SSLSocketFactory factory, String host, int p
     try {
       SSLSocket socket = (SSLSocket) factory.createSocket(host, port);
       socket.setSoTimeout(timeout);
+      SSLParameters sslParams = socket.getSSLParameters();
+      sslParams.setEndpointIdentificationAlgorithm("HTTPS");
+      socket.setSSLParameters(sslParams);
```
jimexist, line 349: this is back compactible?

jimexist (approved):

Jens-G:
Yes — SSLParameters.setEndpointIdentificationAlgorithm("HTTPS") is available since Java 1.7. The only behavioral change is that connections where the server certificate doesn't match the target hostname will now be rejected. IOW, correct deployments are unaffected.

https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLParameters.html#setEndpointIdentificationAlgorithm(java.lang.String)

HTHou:
If enabled the TLS hostname verification in Java sync client, how about adding it in aync client as well?

PR #3391, opened 2026-04-11
Harden Node.js WebSocket server handling

Jens-G:
Client: nodejs

- Validate origin on WebSocket upgrade using the same CORS rules as HTTP requests
- Fix path containment check to use trailing separator
- Replace deprecated new Buffer() with Buffer.alloc()
- Sanitize interpolated header values in upgrade response

### DIFF
--- a/lib/nodejs/lib/thrift/web_server.js
+++ b/lib/nodejs/lib/thrift/web_server.js
@@ -27,6 +27,10 @@ var log = require("./log");
 var MultiplexedProcessor =
   require("./multiplexed_processor").MultiplexedProcessor;
 
+function sanitizeHeader(value) {
+  return (value || "").replace(/[\r\n]/g, "");
+}
+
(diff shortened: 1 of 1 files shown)


lib/nodejs/lib/thrift/web_server.js
```diff
@@ -557,6 +566,10 @@ exports.createWebServer = function (options) {
         request.headers["sec-websocket-key"] +
           "258EAFA5-E914-47DA-95CA-C5AB0DC85B11",
       );
+      //Sanitize header values to prevent CRLF injection
+      var origin = (request.headers.origin || "").replace(/[\r\n]/g, "");
+      var host = (request.headers.host || "").replace(/[\r\n]/g, "");
+      var reqUrl = (request.url || "").replace(/[\r\n]/g, "");
```
jimexist: can have a dedicated function to do this

jimexist (approved):

PR #3392, opened 2026-04-11
Add input validation to Swift protocol layer

Jens-G:
Client: swift

- Add recursion depth limit (64) and negative size checks to skip() in TProtocol for map, set, and list types
- Add range validation for fieldId before UInt8 cast in TCompactProtocol.readFieldBegin()

### DIFF
--- a/lib/swift/Sources/TCompactProtocol.swift
+++ b/lib/swift/Sources/TCompactProtocol.swift
@@ -313,11 +313,11 @@ public class TCompactProtocol: TProtocol {
       booleanValue = type == .boolean_TRUE
     }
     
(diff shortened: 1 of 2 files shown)


lib/swift/Sources/TCompactProtocol.swift
```diff
@@ -314,6 +314,10 @@ public class TCompactProtocol: TProtocol {
     }
     
     // push the new field onto the field stack so we can keep the deltas going
```
jimexist: this comment belongs to the below line

jimexist (approved):

PR #3399, opened 2026-04-15
THRIFT-5954:Add read/write timeout support to Rust TTcpChannel

ShemShadrack:
## Summary

Add timeout configuration support to the Rust TTcpChannel.

This change adds:
- set_read_timeout
- set_write_timeout
- set_timeouts

Timeouts are applied immediately when the channel already wraps an open
TcpStream, and cached for later application when the channel is opened
after configuration.

## Testing

Verified with:
- cargo fmt --manifest-path lib/rs/Cargo.toml --all -- --check
- cargo test --manifest-path lib/rs/Cargo.toml

Added unit tests covering:
- setting timeouts before open()
- applying cached timeouts during open()
- setting timeouts on wrapped streams
- clearing timeouts

### DIFF
--- a/lib/rs/src/transport/socket.rs
+++ b/lib/rs/src/transport/socket.rs
@@ -19,6 +19,7 @@ use std::convert::From;
 use std::io;
 use std::io::{ErrorKind, Read, Write};
 use std::net::{Shutdown, TcpStream, ToSocketAddrs};
+use std::time::Duration;
 
 #[cfg(unix)]
 use std::os::unix::net::UnixStream;
@@ -64,6 +65,8 @@ use crate::{new_transport_error, TransportErrorKind};
 #[derive(Debug, Default)]
 pub struct TTcpChannel {
     stream: Option<TcpStream>,
+    read_timeout: Option<Duration>,
+    write_timeout: Option<Duration>,
 }
 
 impl TTcpChannel {
@@ -72,7 +75,11 @@ impl TTcpChannel {
     /// The returned instance must be opened using `TTcpChannel::open(...)`
     /// before it can be used.
     pub fn new() -> TTcpChannel {
-        TTcpChannel { stream: None }
+        TTcpChannel {
+            stream: None,
+            read_timeout: None,
+            write_timeout: None,
+        }
     }
 
     /// Create a `TTcpChannel` that wraps an existing `TcpStream`.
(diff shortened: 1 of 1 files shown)


Jens-G:
Can we have a [JIRA ticket](https://thrift.apache.org/docs/HowToContribute) for this one pls?

lib/rs/test/Cargo.toml
```diff
@@ -10,8 +10,7 @@ publish = false
 clap = "~2.33"
 bitflags = "=1.2"
 log = "0.4"
-uuid = "1"
+uuid = ">=1, <1.21"
```
Jens-G, line 14: Why do we want to do that? If there is an incompatibility with >= 1.21, the proper approach would be to fix the incompatibility. 

-1 I'm against it. Even more as the change is not explained at all.

lib/rs/src/transport/socket.rs
```diff
@@ -82,9 +89,42 @@ impl TTcpChannel {
     pub fn with_stream(stream: TcpStream) -> TTcpChannel {
         TTcpChannel {
             stream: Some(stream),
+            read_timeout: None,
+            write_timeout: None,
         }
```
Jens-G, line 94: The incoming stream may have timeouts configured. This gives an inconsistent state.

lib/rs/src/transport/socket.rs
```diff
@@ -82,9 +89,42 @@ impl TTcpChannel {
...
+            stream.set_write_timeout(timeout)?;
+        }
+
+        self.write_timeout = timeout;
+        Ok(())
+    }
+
+    /// Set the read and write timeouts for this channel.
+    pub fn set_timeouts(
+        &mut self,
```
Jens-G, line 119: Should we also have getters to make it complete?

Jens-G (changes requested):

Jens-G:
BTW [We still look for Rust maintainers](https://issues.apache.org/jira/browse/THRIFT-5917) as the previous one got lost somehow.

lib/rs/src/transport/socket.rs
ShemShadrack: Good point. I’ll add read_timeout() and write_timeout() getters so callers can inspect the configured values.

lib/rs/src/transport/socket.rs
ShemShadrack: I’ll update `with_stream` so the cached timeout state matches the wrapped `TcpStream` instead of always defaulting to None.

lib/rs/test/Cargo.toml
ShemShadrack: My reason was concern about the newer uuid release and the lack of active Rust maintainers mentioned in THRIFT-5917. But I agree this PR should stay focused on TTcpChannel timeouts, so I’ll revert the uuid change

lib/rs/src/transport/socket.rs
```diff
@@ -141,9 +196,13 @@ impl TIoChannel for TTcpChannel {
             .map(|cloned| {
                 let read_half = ReadHalf::new(TTcpChannel {
                     stream: s.stream.take(),
+                    read_timeout: s.read_timeout,
```
kpumuk: There is a weird issue caused by the `split()` creating two `TTcpChannel` wrappers for the same underlying TCP socket. The socket timeout is shared OS state, but the new `read_timeout` / `write_timeout` fields are copied by value into each wrapper. After one half changes a timeout, the other half still reports the old cached value. If that stale half later calls `set_timeouts(...)`, it writes the old timeout back to the shared socket.

```rust
let initial = Some(Duration::from_millis(80));
let updated = Some(Duration::from_millis(250));
let updated_write = Some(Duration::from_millis(500));
let (mut channel, _server) = wrapped_channel();

channel.set_timeouts(initial, None).unwrap();
let (mut read_half, mut write_half) = channel.split().unwrap();

read_half.set_read_timeout(updated).unwrap();

let stale = write_half.read_timeout();
write_half.set_timeouts(stale, updated_write).unwrap();

assert_eq!(
    (stale, read_half.stream.as_ref().unwrap().read_timeout().unwrap()),
    (updated, updated)
);
```

ShemShadrack: on the design. Worth noting though: read_timeout() already reads directly from the OS socket when the stream is open, so stale in your example would actually equal `updated` the getter never returns the cached field while connected. Yeah still, having diverged caches after split is still bad, so I fixed `split()` to read both timeouts from the live socket rather than copying the struct fields. Caches are now consistent from the start and the regression test covers the scenario shared .

ShemShadrack:
@Jens-G  The `lib-ruby (head)` and `lib-ruby (4.0)` CI failures appear to be a pre-existing issue unrelated to this PR. The `BUNDLED WITH 2.2.34` in `Gemfile.lock` is incompatible with Ruby 4.0+, causing `bundle install` to fail. A possible fix would be to remove the bundler dev dependency from `thrift.gemspec` and bump BUNDLED WITH in `Gemfile.lock` to a current 2.6.x release. Should I submit a separate fix for this?

kpumuk:
Ruby issues are fixed in master, just rebase (https://github.com/apache/thrift/pull/3454 for reference)

ShemShadrack:
> Ruby issues are fixed in master, just rebase (#3454 for reference)


Ohh okaay updated

PR #3401, opened 2026-04-16
THRIFT-5942: Fix Ruby socket connection timeouts

kpumuk:
The Ruby transport has inconsistent and partially broken timeout behavior during connection establishment. Plain TCP open does not appear to hang forever, but it handles timeout and cleanup poorly. SSL open is worse: the handshake loop can retry indefinitely when the readiness wait times out, so a caller-provided timeout is not reliably enforced.

In this change connection logic changes to bind the timeout to the whole connect operation, including TCP connect and TLS handshake.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-5942](https://issues.apache.org/jira/browse/THRIFT-5942)

### DIFF
--- a/lib/rb/README.md
+++ b/lib/rb/README.md
@@ -111,6 +111,30 @@ directly from your application.
 
 ## Breaking Changes
 
+### 0.24.0
+
+Connect timeout handling changed for both `Thrift::Socket` and
+`Thrift::SSLSocket`.
+
+- `timeout == nil` and `timeout == 0` now use blocking connect/open semantics.
+  Older releases already treated `nil` that way, but treated `0` differently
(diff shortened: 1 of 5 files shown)


kpumuk:
Addressed a regression: When `addrinfo.connect(timeout: remaining)` times out for one candidate address, this helper raised `TIMED_OUT` immediately instead of trying the next `Addrinfo` within the still-remaining budget. That breaked multi-homed hosts and dual-stack fallbacks: if the first address blackholes but a later one is reachable before the overall deadline, both `Thrift::Socket#open` and `Thrift::SSLSocket#open` would fail even though the connection could still succeed.

Jens-G (approved):

PR #3405, opened 2026-04-17
THRIFT-5951: Add PHP runtime coverage tests

sveneld:
## Summary
- add 12 PHP runtime coverage tests for previously uncovered classes in `lib/php/lib`
- cover `TBase`, `TProtocol` skip paths, `TMultiplexedProcessor`, `StoredMessageProtocol`, base server/transport abstractions, JSON/SimpleJSON helper contexts, `TApplicationException`, and constant/type classes
- keep production PHP runtime code unchanged

## Notes
- JIRA: THRIFT-5951
- this PR adds tests only
- no third-party dependencies were added
- `LICENSE` and `NOTICE` are unchanged

## Validation
- `docker run --rm -v "$PWD:/src" -w /src php:8.3-cli php vendor/bin/phpunit -c lib/php/phpunit.xml --filter '/(TBaseTest|TProtocolTest|TMultiplexedProcessorTest|StoredMessageProtocolTest|TServerTest|TServerTransportTest|TTransportTest|JsonContextTest|SimpleJsonContextTest|TApplicationExceptionTest|ExceptionTypesTest|TypeConstantsTest)/' lib/php/test/Unit/Lib`
- `docker run --rm -v "$PWD:/src" -w /src php:8.3-cli sh -lc 'for f in lib/php/test/Unit/Lib/Base/TBaseTest.php lib/php/test/Unit/Lib/Protocol/TProtocolTest.php lib/php/test/Unit/Lib/TMultiplexedProcessorTest.php lib/php/test/Unit/Lib/StoredMessageProtocolTest.php lib/php/test/Unit/Lib/Server/TServerTest.php lib/php/test/Unit/Lib/Server/TServerTransportTest.php lib/php/test/Unit/Lib/Transport/TTransportTest.php lib/php/test/Unit/Lib/Protocol/JsonContextTest.php lib/php/test/Unit/Lib/Protocol/SimpleJsonContextTest.php lib/php/test/Unit/Lib/Exception/TApplicationExceptionTest.php lib/php/test/Unit/Lib/Exception/ExceptionTypesTest.php lib/php/test/Unit/Lib/Type/TypeConstantsTest.php; do php -l "$f" || exit 1; done'`

Generated-by: Codex GPT-5.4

### DIFF
--- a/lib/php/test/Unit/Lib/Protocol/TProtocolTest.php
+++ b/lib/php/test/Unit/Lib/Protocol/TProtocolTest.php
@@ -26,7 +26,6 @@ use Thrift\Exception\TProtocolException;
 use Thrift\Protocol\TBinaryProtocol;
 use Thrift\Protocol\TProtocol;
 use Thrift\Transport\TMemoryBuffer;
-use Thrift\Transport\TTransport;
 use Thrift\Type\TType;
 
 class TProtocolTest extends TestCase
@@ -36,9 +35,11 @@ class TProtocolTest extends TestCase
      */
     public function testSkipScalarValues(int $type, string $writerMethod, $value): void
     {
-        $buffer = $this->buildBinaryBuffer(function (TBinaryProtocol $protocol) use ($writerMethod, $value): void {
-            $protocol->{$writerMethod}($value);
-        });
+        $buffer = $this->buildBinaryBuffer(
+            function (TBinaryProtocol $protocol) use ($writerMethod, $value): void {
+                $protocol->{$writerMethod}($value);
+            }
+        );
         $transport = new TMemoryBuffer($buffer);
         $protocol = new TBinaryProtocol($transport);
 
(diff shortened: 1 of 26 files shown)


lib/php/test/Unit/Lib/Protocol/TProtocolTest.php
```diff
@@ -0,0 +1,275 @@
...
+        TProtocol::skipBinary($transport, TType::BOOL);
+    }
+
+    public function testSkipBinarySkipsString(): void
+    {
+        $transport = $this->createMock(TTransport::class);
+        $transport->expects($this->exactly(2))
+            ->method('readAll')
+            ->withConsecutive([4], [2])
+            ->willReturnOnConsecutiveCalls(pack('N', 2), '12');
```
kpumuk, line 191: There is something weird going on here with mocks. `readAll` returns strings, but `skipBinary()` does math on it:

```php
return 4 + $itrans->readAll($len);
```

In this case since we return `'12'` as a string, `4 + '12'` will be 16. Curious how does it even work...

kpumuk: What happens if you return `'aa'` instead of `'12'`? 

Edit:

```
There was 1 error:

1) Test\Thrift\Unit\Lib\Protocol\TProtocolTest::testSkipBinarySkipsPrimitiveAndContainerValues
TypeError: Unsupported operand types: int + string

/thrift/src/lib/php/lib/Protocol/TProtocol.php:289
/thrift/src/lib/php/test/Unit/Lib/Protocol/TProtocolTest.php:190

ERRORS!
Tests: 222, Assertions: 462, Errors: 1, Skipped: 5.
```

I think we have a bug in the protocol implementation, and that the test does not actually test the right thing due to mocks and inputs creatively crafted around the bug.

sveneld:
You were right, this turned out to be a real issue in the `php:inlined` path.

  `TProtocol::skipBinary()` was using `readAll()` as if it returned a byte count, but in PHP it returns raw string data. The previous tests did not expose this because the mocks
  returned numeric strings.

  I fixed the implementation and updated the coverage:
  - unit tests now use real `TMemoryBuffer` payloads
  - added an integration test for generated `php:inlined` code to verify skipping unknown/mismatched fields and continuing to parse following fields correctly

lib/php/test/Unit/Lib/Protocol/TProtocolTest.php
sveneld: In PHP, almost anything is possible :)

  For example:
  - `4 + '12'` => `16`
  - `4 + '12abc'` => `16` (with a warning / deprecation depending on version)
  - `4 + 'abc12'` => TypeError in modern PHP
  - `4 + '1e2'` => `104`

  So in the old test this “worked” only because the mock happened to return a numeric-looking string, which accidentally hide the actual bug.

Jens-G (commented):
+1 LGTM assumed TProtocolInlinedTest actually runs during CI and the build errors are solved

sveneld:
forgot to add phpi generation to CI workflow; fixed now

PR #3402, opened 2026-04-16
THRIFT-5855: Add netstd fuzzers

Jens-G:
## Summary

Revive and update the netstd fuzzer infrastructure originally submitted in PR #3203 by @mhlakhani. This adds SharpFuzz-based protocol fuzzing for the Thrift netstd library, covering all three protocols (Binary, Compact, JSON) with both parse and roundtrip fuzzer types, for both AFL and libfuzzer engines.

Key changes on top of the original PR:

- **net10 bump**: Target framework updated from `net9.0` to `net10.0`, matching the current Thrift build matrix. The `--gen netstd:net10` flag is used for code generation (verified via `t_netstd_generator.cc` — using `net9` would actively break the build on net10 targets due to `#if NET10_0_OR_GREATER` guards).
- **SharpFuzz version pin**: `<PackageReference Include="SharpFuzz" Version="2.2.0" />` — pinned to a known-good release instead of wildcard `*`.
- **`DOTNET_ROLL_FORWARD=Major`**: Temporary workaround in both `buildfuzzers.sh` and `runfuzzer.sh`. SharpFuzz.CommandLine 2.2.0 pins its runtimeconfig to net9.0; this env var allows it to run on a net10-only host. Tracked for removal: upstream fix merged as [SharpFuzz PR #72](https://github.com/Metalnem/sharpfuzz/pull/72), pending release as SharpFuzz 2.3.0.
- **Opt-in build**: `build-fuzzers` target removed from `check-local` in `Makefile.am`. It requires the SharpFuzz.CommandLine global tool and libfuzzer-dotnet binary, which are dev-only dependencies not present in CI. Run manually with `make build-fuzzers` or `./buildfuzzers.sh`.
- **`sharpfuzz` CLI preflight**: `buildfuzzers.sh` now checks for the `sharpfuzz` command before building 12 assemblies, with actionable install instructions on failure.
- **Exception narrowing**: Catch blocks in `ProtocolFuzzerBase` and `ProtocolRoundtripFuzzerBase` narrowed from `catch (Exception)` to specific types (`TProtocolException`, `TTransportException`, `TException`, `EndOfStreamException`, `IOException`) for improved fuzzer effectiveness.
- **Path fix**: `FUZZERS_DIR` and related paths in `buildfuzzers.sh` resolved via `realpath` so the script works when invoked from the repo root, not just from `lib/netstd/`.
- **Documentation**: README fuzzing section rewritten with exact install commands, PATH setup, rollforward workaround explanation, and usage examples.

[...]

PR #3407, opened 2026-04-17
Build netstd fuzzers during make check (without instrumentation)

Jens-G:
## Summary

Follow-up to #3402 addressing @mhlakhani's request:
> lowpri: is it possible to have a mode where they are built by make check (without fuzz instrumentation)?
> the reason I ask is e.g. to avoid cases where we make some change to the code that breaks the fuzzer build, and then we don't realize it because compilation doesn't run in CI

With this PR, `make check` in `lib/netstd` now compiles all 12 fuzzer variants (3 protocols × 2 fuzzer types × 2 engines) but skips SharpFuzz IL rewriting. Any source change that breaks the fuzzer projects will now fail CI. The fully instrumented build suitable for actually running a fuzzer stays opt-in.

## Changes

- **`buildfuzzers.sh`**: new `--no-instrument` flag that skips the `SHARPFUZZ_DIR` / `libfuzzer-dotnet` / `sharpfuzz` CLI preflight checks and the post-build IL rewriting. Code generation and the 12 `dotnet build` steps still run.
- **`lib/netstd/Makefile.am`**:
  - `check-local` gains a `precompile-fuzzers` dependency that invokes `./buildfuzzers.sh --no-instrument`.
  - The existing `build-fuzzers` target is rewritten to delegate to `./buildfuzzers.sh` (it previously inlined 12 `dotnet build` commands and skipped code generation, which made it fail on a clean tree).
  - `gen-netstd` added to `clean-local`; `buildfuzzers.sh` and `runfuzzer.sh` added to `EXTRA_DIST` so `make dist` includes them.
- **`lib/netstd/README.md`**: Fuzzing section updated to describe the new CI behavior.

## Verification

Tested end-to-end inside the `thrift:jammy` docker image:

- `make check` in `lib/netstd` with `SHARPFUZZ_DIR` unset: 12 fuzzer variants build cleanly, existing tests still pass (`Thrift.Tests`: 32 pass, `Thrift.IntegrationTests`: 57 pass).
- Deliberately breaking a fuzzer source file (e.g. changing `class` to `clazz` in `TBinaryProtocolAFL.cs`) now fails `make check` at the `precompile-fuzzers` step.
- `./buildfuzzers.sh --foo` rejects unknown flags; `./buildfuzzers.sh --no-instrument` exits 0 without needing `SHARPFUZZ_DIR` / `libfuzzer-dotnet` / `sharpfuzz`.

## Test plan

- [x] `make check` builds the 12 fuzzer variants without SharpFuzz tooling installed
- [x] Broken fuzzer source fails `make check`
- [x] Existing netstd tests still pass
- [x] `./buildfuzzers.sh` (full flow) unchanged when invoked without `--no-instrument`

---

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/lib/netstd/Makefile.am
+++ b/lib/netstd/Makefile.am
@@ -22,29 +22,26 @@ SUBDIRS = .
 all-local:
 	$(DOTNETCORE) build -c Release
 
-check-local:
+check-local: precompile-fuzzers
 	$(DOTNETCORE) test Tests/Thrift.Compile.Tests/Thrift.Compile.net8/Thrift.Compile.net8.csproj
 	$(DOTNETCORE) test Tests/Thrift.Compile.Tests/Thrift.Compile.net9/Thrift.Compile.net9.csproj
 	$(DOTNETCORE) test Tests/Thrift.Compile.Tests/Thrift.Compile.netstd2/Thrift.Compile.netstd2.csproj
 	$(DOTNETCORE) test Tests/Thrift.Tests/Thrift.Tests.csproj
 	$(DOTNETCORE) test Tests/Thrift.IntegrationTests/Thrift.IntegrationTests.csproj
 
-# Opt-in. Not wired into check-local because it requires the
-# SharpFuzz.CommandLine global tool and libfuzzer-dotnet binary,
-# which are dev-only dependencies. Run manually: `make build-fuzzers`.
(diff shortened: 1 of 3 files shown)


mhlakhani (approved):
thank you!

PR #3408, opened 2026-04-19
Add decompressed payload size limit to Python THeaderTransport

Jens-G:


lib/py/src/transport/THeaderTransport.py
```diff
@@ -276,8 +282,17 @@ def _parse_header_format(self, buffer):
 
         payload = buffer.read()
         for transform_id in transforms:
-            transform_fn = READ_TRANSFORMS_BY_ID[transform_id]
-            payload = transform_fn(payload)
+            if transform_id == THeaderTransformID.ZLIB:
+                decompressor = zlib.decompressobj()
+                payload = decompressor.decompress(payload, self._max_decompressed_size)
```
kpumuk: Feels like the logic of hanlding zip files is spread across the file. We define a function to unzip the paload in READ_TRANSFORMS_BY_ID (method pointer), and then have custom implementation here. Maybe it is possible to define something like `handler_transform_zlib(payload)` and then point at it in `READ_TRANSFORMS_BY_ID`? this logic can move there and we will have zlib handler localized in a single method.

kpumuk (approved):

Jens-G:
Thanks, much appreciated. I applied some refactoring but want to keep the scope focused with this. If we feel extra refactoring could be helpful, I'd leave this to a separate patch.

kpumuk:
Looks good to me, clean and simple.

PR #3414, opened 2026-04-22
THRIFT-5953: Rust codegen forward-compatible union deserialization in structs

santiagomed:
## Problem

When a struct contains a union-typed field and the wire data has a union variant (field ID) not present in the local IDL, the generated Rust code fails to deserialize the **entire struct** — not just the unknown union field.

The root cause: the union's `read_from_in_protocol` correctly returns `Err("received empty union")` for unknown variants (fixed in PR #3336). But the parent struct's generated deserializer propagates this error via `?`, causing the whole struct to fail.

This breaks Thrift's forward-compatibility guarantee. When a server adds a new union variant, all clients running an older IDL version fail to parse any message containing that variant — even though the client doesn't need to understand the new variant to process the rest of the struct.

Java, Go, Python, and C++ Thrift all handle this gracefully: unknown union variants are silently skipped, and the struct field is left unset.

## Fix

In `render_struct_sync_read`, detect struct fields whose type is a union. For those fields, wrap the `read_from_in_protocol` call in a `match` that catches the `"received empty union"` error and treats it as `None` (field absent):

```rust
// Before (all fields):
let val = FooUnion::read_from_in_protocol(i_prot)?;
f_N = Some(val);

// After (union fields only):
match FooUnion::read_from_in_protocol(i_prot) {
    Ok(val) => { f_N = Some(val); },
    Err(thrift::Error::Protocol(ref e))
        if e.message.contains("received empty union") => {
        // forward compatibility: unknown union variant skipped
    },
    Err(e) => return Err(e),
}
```

Non-union fields are unchanged. The union's own `TSerializable` impl is unchanged — it still returns `Err` for unknown variants, preserving the contract for callers that read unions directly.

## Impact

- Fixes forward compatibility for all Rust Thrift users with evolving union schemas
- No changes to the `TSerializable` trait or the thrift runtime library
- No changes to the union`s own serialization/deserialization
- Struct fields with union types that receive unknown variants get `None` instead of causing parse failure
- All other errors (malformed data, I/O, multiple union fields) still propagate normally

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_rs_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_rs_generator.cc
@@ -1699,10 +1699,41 @@ void t_rs_generator::render_struct_sync_read(const string& struct_name,
 
     for (members_iter = members.begin(); members_iter != members.end(); ++members_iter) {
       t_field* tfield = (*members_iter);
+      t_type* resolved = get_true_type(tfield->get_type());
+      bool is_union_field = resolved->is_struct() && ((t_struct*)resolved)->is_union();
       f_gen_ << indent() << rust_safe_field_id(tfield->get_key()) << " => {" << '\n';
       indent_up();
-      render_type_sync_read("val", tfield->get_type());
-      f_gen_ << indent() << struct_field_read_temp_variable(tfield) << " = Some(val);" << '\n';
+      if (is_union_field) {
+        // Use the resolved (non-Box) type since Box<T>::method() isn't valid syntax.
+        string resolved_type = to_rust_type(resolved);
+        bool is_boxed = false;
+        {
+          t_type* t = tfield->get_type();
+          while (t->is_typedef()) {
+            if (((t_typedef*)t)->is_forward_typedef()) { is_boxed = true; break; }
+            t = ((t_typedef*)t)->get_type();
+          }
+        }
+        string read_call(resolved_type + "::read_from_in_protocol(i_prot)");
+        string val_expr = is_boxed ? "Box::new(val)" : "val";
+        bool suppress_unknown = (struct_type == T_REGULAR || struct_type == T_EXCEPTION) && is_optional(actual_field_req(tfield, struct_type));
+        if (suppress_unknown) {
+          f_gen_ << indent() << "match " << read_call << " {" << '\n';
+          indent_up();
+          f_gen_ << indent() << "Ok(val) => { " << struct_field_read_temp_variable(tfield) << " = Some(" << val_expr << "); }," << '\n';
+          f_gen_ << indent() << "Err(thrift::Error::Protocol(ref e)) if e.kind == ProtocolErrorKind::UnknownUnionVariant => {" << '\n';
+          f_gen_ << indent() << "}," << '\n';
+          f_gen_ << indent() << "Err(e) => return Err(e)," << '\n';
+          indent_down();
+          f_gen_ << indent() << "}" << '\n';
+        } else {
+          f_gen_ << indent() << "let val = " << read_call << "?;" << '\n';
+          f_gen_ << indent() << struct_field_read_temp_variable(tfield) << " = Some(" << val_expr << ");" << '\n';
+        }
+      } else {
+        render_type_sync_read("val", tfield->get_type());
+        f_gen_ << indent() << struct_field_read_temp_variable(tfield) << " = Some(val);" << '\n';
+      }
       indent_down();
       f_gen_ << indent() << "}," << '\n';
     }
@@ -1778,6 +1809,7 @@ void t_rs_generator::render_union_sync_read(const string& union_name, t_struct* 
   // completed union as well as a count of fields read
   f_gen_ << indent() << "let mut ret: Option<" << union_name << "> = None;" << '\n';
   f_gen_ << indent() << "let mut received_field_count = 0;" << '\n';
+  f_gen_ << indent() << "let mut total_field_count = 0;" << '\n';
 
   // read the struct preamble
   f_gen_ << indent() << "i_prot.read_struct_begin()?;" << '\n';
@@ -1803,16 +1835,40 @@ void t_rs_generator::render_union_sync_read(const string& union_name, t_struct* 
   vector<t_field*>::const_iterator members_iter;
   for (members_iter = members.begin(); members_iter != members.end(); ++members_iter) {
     t_field* member = (*members_iter);
+    t_type* member_resolved = get_true_type(member->get_type());
+    bool member_is_union = member_resolved->is_struct() && ((t_struct*)member_resolved)->is_union();
     f_gen_ << indent() << rust_safe_field_id(member->get_key()) << " => {" << '\n';
     indent_up();
-    render_type_sync_read("val", member->get_type());
-    f_gen_ << indent() << "if ret.is_none() {" << '\n';
-    indent_up();
-    f_gen_ << indent() << "ret = Some(" << union_name << "::" << rust_union_field_name(member)
-           << "(val));" << '\n';
-    indent_down();
-    f_gen_ << indent() << "}" << '\n';
-    f_gen_ << indent() << "received_field_count += 1;" << '\n';
+    if (member_is_union) {
+      string member_type = to_rust_type(member_resolved);
+      string member_read(member_type + "::read_from_in_protocol(i_prot)");
+      f_gen_ << indent() << "match " << member_read << " {" << '\n';
+      indent_up();
+      f_gen_ << indent() << "Ok(val) => {" << '\n';
+      indent_up();
+      f_gen_ << indent() << "if ret.is_none() {" << '\n';
+      indent_up();
+      f_gen_ << indent() << "ret = Some(" << union_name << "::" << rust_union_field_name(member)
+             << "(val));" << '\n';
+      indent_down();
+      f_gen_ << indent() << "}" << '\n';
+      f_gen_ << indent() << "received_field_count += 1;" << '\n';
+      indent_down();
+      f_gen_ << indent() << "}," << '\n';
+      f_gen_ << indent() << "Err(thrift::Error::Protocol(ref e)) if e.kind == ProtocolErrorKind::UnknownUnionVariant => {}," << '\n';
+      f_gen_ << indent() << "Err(e) => return Err(e)," << '\n';
+      indent_down();
+      f_gen_ << indent() << "}" << '\n';
+    } else {
+      render_type_sync_read("val", member->get_type());
+      f_gen_ << indent() << "if ret.is_none() {" << '\n';
+      indent_up();
+      f_gen_ << indent() << "ret = Some(" << union_name << "::" << rust_union_field_name(member)
+             << "(val));" << '\n';
+      indent_down();
+      f_gen_ << indent() << "}" << '\n';
+      f_gen_ << indent() << "received_field_count += 1;" << '\n';
+    }
     indent_down();
     f_gen_ << indent() << "}," << '\n';
   }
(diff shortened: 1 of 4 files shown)


Jens-G:
BTW [We still look for Rust maintainers](https://issues.apache.org/jira/browse/THRIFT-5917) as the previous one got lost somehow.

Jens-G:
> Commits on Apr 22, 2026
> THRIFT-XXXX: Rust codegen: forward-compatible union deserialization in structs

We have rules about AI generated content. Acrtually in the codebase. And THRIFT-XXXX is not a valid number.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
```diff
@@ -1699,10 +1699,37 @@ void t_rs_generator::render_struct_sync_read(const string& struct_name,
...
+        // Use the resolved (non-Box) type for the method call since
+        // Box<T>::method() is not valid Rust syntax for turbofish.
+        string resolved_type = to_rust_type(resolved);
+        bool is_boxed_union = to_rust_type(tfield->get_type()) != resolved_type;
+        string read_call(resolved_type + "::read_from_in_protocol(i_prot)");
+        string val_expr = is_boxed_union ? "Box::new(val)" : "val";
+        f_gen_ << indent() << "match " << read_call << " {" << '\n';
+        indent_up();
+        f_gen_ << indent() << "Ok(val) => { " << struct_field_read_temp_variable(tfield) << " = Some(" << val_expr << "); }," << '\n';
+        f_gen_ << indent() << "Err(thrift::Error::Protocol(ref e)) if e.message.contains(\"received empty union\") => {" << '\n';
```
Jens-G: I would consider that string search a code smell. What about adding a dedicated ProtocolErrorKind variant? Similar to ProtocolErrorKind::InvalidData for the "empty union" case around line 1837.

Jens-G (changes requested):
Two more suggestions:
- Having at least one integration test would be great
- This change indeed deserves a JIRA ticket. It is a non-trivial codegen change so we should have one.

Thank you for your contribution!

santiagomed:
@Jens-G thanks for the review. Will apply these changes.

Also, I am happy to take on as maintainer - we at xAI use Rust + Thrift quite a bit in production and need active support. Let me know what's needed

EDIT: I requested a Jira account for the apache thrift project to create the ticket and apply as maintainer.

compiler/cpp/src/thrift/generate/t_rs_generator.cc
santiagomed: I added `ProtocolErrorKind::EmptyUnion = 7` to the runtime library and updated both sides of the codegen to use it. The union deserializer now emits `EmptyUnion` instead of `InvalidData`, and the struct-level catch arm matches on `e.kind == ProtocolErrorKind::EmptyUnion` instead of the string check. Also added an integration test covering the unknown variant path.

santiagomed:
@Jens-G created a Jira ticket: https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-5953?filter=allopenissues

kpumuk:
This looks like a large change, and the fact we have tests disabled on CI does not inspire confidence. @santiagomed would it make sense to create a separate pull request to enable back `lib-rust` and integration tests for Rust (making sure they pass) before merging this one?

See https://github.com/apache/thrift/commit/e4666f293478e3457898867095105ff2d23b155e

santiagomed:
@kpumuk Rust CI fix has been opened #3417.

kpumuk:
@santiagomed Would you please rebase and squash the changes so that we can trigger the full CI run?

Jens-G:
### Code review

Found 2 issues:

1. **`is_boxed_union` false-positive for typedef-aliased union fields** — The check `to_rust_type(tfield->get_type()) != resolved_type` uses `to_rust_type` on both sides, but `get_true_type()` only on the right. If a field's declared type is a typedef over a union (not a `Box<>`-wrapped union), `to_rust_type` on the typedef returns the alias name while `resolved_type` is the concrete union name — `is_boxed_union` becomes `true` incorrectly, and the generator emits `Box::new(val)` into a non-boxed field, producing a type error in all generated code for typedef-aliased union fields.

https://github.com/santiagomed/thrift/blob/c63c487ab438843ca6b7602105cd0a29ae1a7764/compiler/cpp/src/thrift/generate/t_rs_generator.cc#L1707-L1712

2. **`EmptyUnion` conflates two distinct error cases** — The variant doc says "A Thrift union was deserialized with no recognized variants" (forward-compat scenario). But the same kind is also raised when `received_field_count == 0` — i.e., when the sender emitted a union with zero fields on the wire (a genuine protocol violation). The struct-level catch arm that swallows `EmptyUnion` for forward-compat will also silently suppress these genuine violations, converting hard errors into `None`.

https://github.com/santiagomed/thrift/blob/c63c487ab438843ca6b7602105cd0a29ae1a7764/compiler/cpp/src/thrift/generate/t_rs_generator.cc#L1854-L1858

kpumuk:
There are two remaining edge cases that look unsafe.

### 1. Nested forward typedefs still lose the `Box`

Minimal IDL:

```thrift
typedef FutureUnion ForwardAlias
typedef ForwardAlias NestedForwardAlias

struct Wrapper {
  1: optional NestedForwardAlias value
}

union FutureUnion {
  1: i32 n
}
```

The Rust generator emits aliases like:

```rust
pub type ForwardAlias = Box<FutureUnion>;
pub type NestedForwardAlias = ForwardAlias;
```

But the new special-case union read path reads the resolved union type directly and assigns it into the field temp:

```rust
let mut f_1: Option<NestedForwardAlias> = None;

match FutureUnion::read_from_in_protocol(i_prot) {
    Ok(val) => { f_1 = Some(val); },
    Err(thrift::Error::Protocol(ref e))
        if e.kind == ProtocolErrorKind::UnknownUnionVariant => {}
    Err(e) => return Err(e),
}
```

That is a generated-code compile error: `val` is `FutureUnion`, while `NestedForwardAlias` is `Box<FutureUnion>`.

```text
error[E0308]: mismatched types
expected `Box<FutureUnion>`, found `FutureUnion`
```

The boxing decision needs to preserve boxing through the typedef chain, not only inspect the outer field typedef.

### 2. Multiple unknown union fields are treated as ignorable

The union reader currently maps `received_field_count == 0` to `UnknownUnionVariant` before rejecting multi-field unions. That also matches a union payload with two unknown fields. The parent struct reader then catches `UnknownUnionVariant` and silently leaves the optional union field as `None`.

Regression test shape:

```rust
let mut write_buf = Vec::new();
{
    let cursor = Cursor::new(&mut write_buf);
    let mut prot = TBinaryOutputProtocol::new(cursor, false);

    prot.write_struct_begin(&TStructIdentifier::new("Wrapper")).unwrap();
    prot.write_field_begin(&TFieldIdentifier::new("u", TType::Struct, 1)).unwrap();

    prot.write_struct_begin(&TStructIdentifier::new("U")).unwrap();
    prot.write_field_begin(&TFieldIdentifier::new("future_a", TType::I32, 99)).unwrap();
    prot.write_i32(1).unwrap();
    prot.write_field_end().unwrap();

    prot.write_field_begin(&TFieldIdentifier::new("future_b", TType::I32, 100)).unwrap();
    prot.write_i32(2).unwrap();
    prot.write_field_end().unwrap();

    prot.write_field_stop().unwrap();
    prot.write_struct_end().unwrap();

    prot.write_field_end().unwrap();
    prot.write_field_stop().unwrap();
    prot.write_struct_end().unwrap();
}

let read_cursor = Cursor::new(write_buf);
let mut rprot = TBinaryInputProtocol::new(read_cursor, false);
let err = Wrapper::read_from_in_protocol(&mut rprot)
    .expect_err("multi-field union should stay invalid");
```

A union with multiple fields is invalid even if none of those fields are known locally. The checks should preserve that:

```text
total_field_count == 0                         => EmptyUnion
total_field_count > 1                          => InvalidData
total_field_count == 1 && received_count == 0  => UnknownUnionVariant
```

kpumuk:
Also, the commit message starts with a placeholder `THRIFT-XXXX`  instead of `THRIFT-5953`

santiagomed:
Thanks for the feedback. Updated, the boxing check now walks the full typedef chain to handle nested forward typedefs, and multi-field unknown unions are rejected as `InvalidData` before the forward-compat path is considered (only `total_field_count == 1 && received_field_count == 0` produces the catchable `UnknownUnionVariant`).

santiagomed:
Merge?

kpumuk:
The remaining issue is that unknown union variants are swallowed in more places than optional user-struct fields.

`render_struct_sync_read` is shared by normal structs, generated RPC args structs, and generated RPC result structs. The new branch catches `UnknownUnionVariant` unconditionally and leaves the field temp as `None`. That is correct for an optional user-struct field, but not for required fields or RPC wrapper fields.

Minimal IDL:

```thrift
union Response {
  1: i32 known
}

service Svc {
  Response get()
}
```

For a newer server response containing an unknown `Response` variant, the generated result reader leaves `result_value` unset. Then `SvcGetResult::ok_or()` reports:

```text
ApplicationErrorKind::MissingResult
```

That hides the real protocol/schema mismatch. Required union args/fields have the same shape: the union read error is converted into a later missing-field error.

Regression test template, placed inside the generated module or adapted into an integration test:

```rust
#[test]
fn unknown_union_variant_in_rpc_result_is_not_missing_result() {
    use std::io::Cursor;

    use thrift::protocol::{
        TBinaryInputProtocol, TBinaryOutputProtocol, TFieldIdentifier, TOutputProtocol,
        TSerializable, TStructIdentifier, TType,
    };

    let mut write_buf = Vec::new();
    {
        let cursor = Cursor::new(&mut write_buf);
        let mut prot = TBinaryOutputProtocol::new(cursor, false);

        prot.write_struct_begin(&TStructIdentifier::new("SvcGetResult")).unwrap();
        prot.write_field_begin(&TFieldIdentifier::new("result_value", TType::Struct, 0)).unwrap();

        prot.write_struct_begin(&TStructIdentifier::new("Response")).unwrap();
        prot.write_field_begin(&TFieldIdentifier::new("future_variant", TType::I32, 99)).unwrap();
        prot.write_i32(7).unwrap();
        prot.write_field_end().unwrap();
        prot.write_field_stop().unwrap();
        prot.write_struct_end().unwrap();

        prot.write_field_end().unwrap();
        prot.write_field_stop().unwrap();
        prot.write_struct_end().unwrap();
    }

    let read_cursor = Cursor::new(write_buf);
    let mut rprot = TBinaryInputProtocol::new(read_cursor, false);

    let result = SvcGetResult::read_from_in_protocol(&mut rprot).unwrap();
    let err = result
        .ok_or()
        .expect_err("unknown success union variant should not become MissingResult");

    assert!(matches!(
        err,
        thrift::Error::Protocol(ref e)
            if e.kind == thrift::ProtocolErrorKind::UnknownUnionVariant
    ));
}
```

The fix should only suppress `UnknownUnionVariant` where treating the field as absent is actually intended, e.g. optional regular struct fields. For required fields and RPC args/result structs, the original union error should propagate.

santiagomed:
The `UnknownUnionVariant` catch is now gated on `is_optional(actual_field_req(tfield, struct_type))`. For required fields and RPC args/result structs (where `actual_field_req` returns `T_REQUIRED`), the union read uses the normal `?` propagation path, so `UnknownUnionVariant` flows up to the caller as-is instead of being silently converted to `None` and later surfacing as `MissingResult`.

kpumuk:
Not fixed completely. Passed checks:

- Nested forward typedef union now compiles and reads the boxed value.
- Plain typedef alias to union does not get spuriously boxed.
- Optional regular struct field with an unknown union variant becomes `None`.
- Required regular struct field propagates `UnknownUnionVariant`.
- Multi-unknown-field union returns `InvalidData`.
- Empty union returns `EmptyUnion`.

Still failing:

- RPC result wrapper still swallows `UnknownUnionVariant`.

Generated code for `SvcGetResult` still has:

```rust
match Response::read_from_in_protocol(i_prot) {
    Ok(val) => { f_0 = Some(val); },
    Err(thrift::Error::Protocol(ref e)) if e.kind == ProtocolErrorKind::UnknownUnionVariant => {
    },
    Err(e) => return Err(e),
}
```

So `SvcGetResult::read_from_in_protocol()` returns:

```text
Ok(SvcGetResult { result_value: None })
```

Then `ok_or()` would still report:

```text
ApplicationErrorKind::MissingResult
```

Root cause: `suppress_unknown` is based only on field optionality. RPC result success fields are generated as optional, so result wrappers still take the suppression path.

This should be limited to optional `T_REGULAR` fields, not all optional fields.

santiagomed:
Thanks. Pushed a fix, can you check now? @kpumuk

kpumuk:
Heh, found another edge case to cover. Last change fixed for `T_REGULAR`, but user-defined exceptions are also deserialized through `render_struct_sync_read()`. If an exception has an optional union field and a newer server sends a newly-added union variant, older Rust clients will still fail the whole RPC with `UnknownUnionVariant` instead of decoding the exception with that field as `None`, so exception payloads remain non-forward-compatible.

santiagomed:
Done. Suppression now fires for `T_REGULAR` and `T_EXCEPTION`, but not for `T_ARGS` or `T_RESULT`.

santiagomed:
@kpumuk can you authorize the workflow run please.

santiagomed:
merge?

santiagomed:
@Jens-G @kpumuk could we merge this if it looks good please.

Jens-G:
> @Jens-G @kpumuk could we merge this if it looks good please.

Not quite. Details follow.

Jens-G:
### Code review

Found 2 issues:

1. `total_field_count` increments for unknown fields, reintroducing the multi-field false error that THRIFT-5926 fixed — a union with 1 known + 1 unknown field sets `total_field_count = 2` and returns `InvalidData "received multiple fields"` instead of succeeding

https://github.com/apache/thrift/blob/6e56bd55e11357608f01e51ba52e75f2a953f406/compiler/cpp/src/thrift/generate/t_rs_generator.cc#L1859-L1879

2. Wire stream corruption when a union field's value type is itself a union with an unknown variant — the inner `read_from_in_protocol` returns `Err(UnknownUnionVariant)` via `?` before the outer union consumes its Stop byte; `suppress_unknown` catches the error and continues, but the orphaned Stop byte causes the enclosing struct's next `read_field_begin` to see `0x00` and terminate the struct early, silently dropping all subsequent fields

https://github.com/apache/thrift/blob/6e56bd55e11357608f01e51ba52e75f2a953f406/compiler/cpp/src/thrift/generate/t_rs_generator.cc#L1717-L1727

Regression tests for both issues have been added as a separate commit: 24af312f5.

santiagomed:
Thanks @Jens-G. Updated: total_field_count > 1 is now received_field_count > 1 so 1-known + N-unknown succeeds again. For the stream corruption, the fix is in the union reader itself: when a member type is also a union, it catches UnknownUnionVariant internally so the outer union always fully consumes its struct before returning.

santiagomed:
@Jens-G made some formatting fixes, CI should pass now: https://github.com/santiagomed/thrift/pull/2

santiagomed:
@Jens-G @kpumuk could either of you kick off the workflow and review if it looks good please?

PR #3415, opened 2026-04-23
Connect skip() to TConfiguration recursion limit

Jens-G:
Client: java

Make sure to connect all `skip()` implementations propertly and consistently to `TConfiguration`. Also, this drops `maxSkipDepth` and deprecates `TProtocolUtil.setMaxSkipDepth()` in favour of TConfiguration.

### DIFF
--- a/lib/java/README.md
+++ b/lib/java/README.md
@@ -82,10 +82,10 @@ The default build will run the unit tests which expect a usable
 Thrift compiler to exist on the system. You have two choices for
 that.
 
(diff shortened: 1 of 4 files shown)


kpumuk (approved):
Do we want a breaking change notice as `setMaxSkipDepth` might subtly break the business logic (e.g. stop working without any indication it did so)? For the https://github.com/apache/thrift/blob/master/lib/java/README.md#breaking-changes at least

PR #3426, opened 2026-04-28
Improve ruby gemspec

Jens-G:
```
WARNING:  open-ended dependency on logger (>= 0) is not recommended
  use a bounded requirement, such as "~> x.y"
WARNING:  open-ended dependency on rack (>= 2.2.23, development) is not recommended
  if rack is semantically versioned, use:
    add_development_dependency "rack", "~> 2.2", ">= 2.2.23"
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
```

### DIFF
--- a/lib/rb/thrift.gemspec
+++ b/lib/rb/thrift.gemspec
@@ -27,13 +27,13 @@ Gem::Specification.new do |s|
 
   s.require_paths = %w[lib ext]
 
-  s.add_dependency 'logger'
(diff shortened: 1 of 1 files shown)


kpumuk:
I would honestly ignore the warnings, they are aimed at applications and not libraries. With this change people will not be able to use Thrift gem with modern popular libraries, like [Sidekiq](https://rubygems.org/gems/sidekiq).

PR #3428, opened 2026-04-29
[GO] Fix range check on 32-bit architectures

srebhan:
PR #3381 introduced range checks for integer values, however, one of the checks fails with 

```
# github.com/apache/thrift/lib/go/thrift
/go/pkg/mod/github.com/apache/thrift@v0.23.0/lib/go/thrift/framed_transport.go:206:12: math.MaxUint32 (untyped int constant 4294967295) overflows int
```

on 32-bit architectures. This is because `math.MaxUint32` is of `untyped int` which is a **signed 32-bit** integer on those architectures and thus cannot hold the maximum 32-bit **unsigned** value.

This PR fixes the issue by comparing the values as `uint64` values avoiding potential overflows.

### DIFF
--- a/lib/go/thrift/framed_transport.go
+++ b/lib/go/thrift/framed_transport.go
@@ -203,8 +203,8 @@ func (p *TFramedTransport) WriteString(s string) (n int, err error) {
 
 func (p *TFramedTransport) Flush(ctx context.Context) error {
 	size := p.writeBuf.Len()
-	if size > math.MaxUint32 {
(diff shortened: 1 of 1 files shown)


kpumuk (approved):

fishy (approved):

macabu:
Hi, would it be possible to tag a new release with this fix? We're unable to upgrade since we need to support 32-bit architectures

fishy:
> Hi, would it be possible to tag a new release with this fix? We're unable to upgrade since we need to support 32-bit architectures

You can use `go get github.com/apache/thrift@d2acd3c` to get this fixed version.

macabu:
Thank you, we will do that!

PR #3431, opened 2026-04-30
THRIFT-5785: fix boost include

CJCombrink:
Remove another boost header include from the public API
- It is only used in THeaderTransport.cpp so move it there without complicating it
- It throws a TTransportException so it should not be used in other code
- People that depends on this can use `boost::numeric_cast` or `std::in_range` (C++20) instead
- The usage in the unit test was not needed since `std::string::size()` returns an `unsigned long` on 64-bit compilers
 

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  [THRIFT-5785](https://issues.apache.org/jira/browse/THRIFT-5785)

The following tickets are releated
- [THRIFT-5660](https://issues.apache.org/jira/browse/THRIFT-5660)
- [THRIFT-5785](https://issues.apache.org/jira/browse/THRIFT-5785)

### DIFF
--- a/lib/cpp/src/thrift/transport/THeaderTransport.cpp
+++ b/lib/cpp/src/thrift/transport/THeaderTransport.cpp
@@ -23,6 +23,8 @@
 #include <thrift/protocol/TBinaryProtocol.h>
 #include <thrift/protocol/TCompactProtocol.h>
 
+#include <boost/numeric/conversion/cast.hpp>
+
 #include <limits>
 #include <utility>
 #include <string>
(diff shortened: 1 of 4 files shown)


CJCombrink:
PS: just rewriting the offending function without using boost was more complex than I have anticipated (see the linked tickets)
On closer inspection I realised that just moving the function internal removes the problem thus no complex code changes needed

emmenlau (approved):
Looks good to me. I did *not* check test results. If tests are green, please merge!

hdu-sdlzx:
lgtm
I can build my project with installed thrift and without boost now.

PR #3436, opened 2026-05-02
THRIFT-5957: Add phpstan static analysis with CI guardrail

sveneld:
The PHP runtime library at `lib/php/lib/` had no static-analysis tooling in CI. Refactors and upcoming type-modernization PRs would otherwise go in without an automated safety net.

**Changes:**

- Adds `phpstan/phpstan ^1.12` as a `require-dev` dependency, plus a `scripts.phpstan` composer alias for local dev convenience.
- Adds `lib/php/phpstan.neon` (level 1) targeting `lib/php/lib/` with `test/bootstrap.php` for autoload, and a generated `lib/php/phpstan-baseline.neon` that pins the **5 remaining issues** (all "should return X but return statement is missing" in `TJSONProtocol`/`TSimpleJSONProtocol` `writeStructBegin`/`writeStructEnd` and `ThriftClassLoader::findFile` — left for a follow-up ticket because they need a small protocol-spec decision).
- Adds a new `lib-php-quality` job in `.github/workflows/build.yml` that runs phpstan on PHP 8.3 with `--error-format=github` so any new finding shows up as an inline annotation on the PR diff.

**Trivial findings fixed in this PR** (rather than baselined — the change is purely cleanup, no behaviour shift):

- `TSocketPool::open`: replaced `extract($this->servers_[$i])` with explicit `$host` / `$port` assignments. Removes 10 "might not be defined" findings and avoids `extract()`'s well-known footgun of silently rebinding existing locals from a foreign array.
- `TBufferedTransport::readAll`: changed the final `elseif ($have > $len)` to `else` so phpstan can prove `$data` is always assigned. Semantically identical; the previous form was the only remaining case after the prior three.

Level 1 catches undefined variables and obvious type mismatches without requiring widespread code changes. Higher levels will be raised in follow-up tickets as the runtime library gains native types and PHPDoc cleanup.

**Out of scope (separate tickets, in roadmap order):**

- Fixing the remaining 5 missing-return findings (needs a decision on whether `writeStructBegin/End` returns 0 or the bytes written by the inner JSON helper).
- PSR-2 → PSR-12 migration / php-cs-fixer.
- `declare(strict_types=1)` and native parameter / return / property types in `lib/php/lib/`.
- Static analysis of generated PHP under `test/Resources/packages/`.
- PHPUnit attribute migration & PHPUnit 10/11 upgrade.
- `t_php_generator.cc` modernization.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  [THRIFT-5957](https://issues.apache.org/jira/browse/THRIFT-5957)
- [x] Did you do your best to avoid breaking changes?  *(Pure tooling addition + two equivalent refactors, no behaviour change.)*

Generated-by: Claude Opus 4.7 (1M context)

### DIFF
--- a/.github/workflows/sca.yml
+++ b/.github/workflows/sca.yml
@@ -177,6 +177,13 @@ jobs:
           composer install --quiet
           ./vendor/bin/phpcs
 
+      - name: Run phpstan
+        id: phpstan
+        continue-on-error: true
+        run: |
+          # PHP static analysis
+          ./vendor/bin/phpstan analyse -c lib/php/phpstan.neon --no-progress --error-format=github
+
       - name: Run rubocop
         id: rubocop
         continue-on-error: true
@@ -209,6 +216,7 @@ jobs:
           CPPCHECK_OUTCOME: ${{ steps.cppcheck.outcome }}
           FLAKE8_OUTCOME: ${{ steps.flake8.outcome }}
           PHPCS_OUTCOME: ${{ steps.phpcs.outcome }}
+          PHPSTAN_OUTCOME: ${{ steps.phpstan.outcome }}
           RUBOCOP_OUTCOME: ${{ steps.rubocop.outcome }}
         run: |
           failed=0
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -223,6 +223,24 @@ jobs:
             test/php/gen-php-classmap/
           retention-days: 3
 
+  lib-php-quality:
```
kpumuk: Please check sca.yml, a separate build workflow for static code analysis

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # Apache Thrift Changelog
 
+## 0.24.0
```
kpumuk: This file is auto-generated on release

sveneld: ok, I will remove this change

.github/workflows/build.yml
sveneld: ok, will move this build in sca

kpumuk (approved):

PR #3443, opened 2026-05-03
THRIFT-5967: Parallelize static code analysis run on GitHub

kpumuk:
Currently SCA job runs as a single monolythic workflow, with each job checking different libraries separately. Jobs are allowed to fail, and there is a final summary job that check the status of each other job and fails if any failed. This makes it a little bit hard to read, makes the job slower, and somewhat hard to add new languages.

This pull request follows the similar structure we have in our `build.yml` - each language gets its own job, that depends on compiler build (used to generate thrift and verify generated code to follow the same rules as the library itself).

CleanShot 2026-05-03 at 13 15 28@2x

The time went down from 7m4s to 6m45s, which is not very significant, but that's pretty much a cap - compiler + cppcheck for C++ library, and the rest of the languages will finish in parallel before that, so it will never take longer. The old workflow would keep increasing the total time.

Additionally, replaces `sloccount` with `scc`, as it does not know the modern PHP syntax and was spitting lots of noise in the output:

```
Warning! Unclosed PHP file /home/runner/work/thrift/thrift/lib/php/test/Unit/Lib/Transport/TransportErrorScenariosTest.php, mode=1
Warning! Unclosed PHP file /home/runner/work/thrift/thrift/lib/php/test/Unit/Lib/Serializer/TBinarySerializerTest.php, mode=1
```

Nice and pretty now with https://github.com/apache/thrift/actions/runs/25285416874/job/74129027301

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-5967](https://issues.apache.org/jira/browse/THRIFT-5967)

### DIFF
--- a/.github/workflows/sca.yml
+++ b/.github/workflows/sca.yml
@@ -9,7 +9,6 @@ on:
 
 env:
   BUILD_DEPS: automake bison flex git libboost-all-dev libevent-dev libssl-dev libtool make pkg-config
-  SCA_DEPS: cppcheck python3-flake8 sloccount libglib2.0-dev
   # Disable all languages for which we don't have SCA checks in place
   CONFIG_ARGS_FOR_SCA: >
     --enable-tutorial=no
@@ -37,7 +36,7 @@ permissions:
   contents: read
 
 jobs:
-  sca:
+  compiler:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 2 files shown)


.github/workflows/sca.yml
```diff
@@ -146,6 +185,137 @@ jobs:
...
+        run: |
+          chmod a+x compiler/cpp/thrift
+          compiler/cpp/thrift -version
+          make -j"$(nproc)" -C lib/php/test stubs
+          make -j"$(nproc)" -C test/php precross
+
+      - name: Run phpcs
+        run: |
+          composer install --quiet
+          ./vendor/bin/phpcs
```
sveneld, line 272: I see that generated php files get many errors, I plan to fix style for generated classes.
As temporary solution phpcs can be run like this
```
./vendor/bin/phpcs --ignore=*/test/Resources/*
```

kpumuk: Yeah, I updated path in the phpcs.xml.dist, which apparently had an outdated path. Because of how I set up `sca.yml` before, it did not generate PHP fixtures before running phpcs

kpumuk:
Created an umbrella ticket https://issues.apache.org/jira/browse/THRIFT-5966 to add SCA for other libraries.

PR #3446, opened 2026-05-03
THRIFT-5064: Introduce Perl::Critic into the SCA

kpumuk:
Introduce Perl::Critic into the SCA build. Extends on the prior art #1982, but additionally introduces Perl::Critic tags into the generated code, so it also passes validations.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-5064](https://issues.apache.org/jira/browse/THRIFT-5064)

### DIFF
--- a/.github/workflows/sca.yml
+++ b/.github/workflows/sca.yml
@@ -275,6 +275,46 @@ jobs:
         run: |
           ./vendor/bin/phpstan analyse -c lib/php/phpstan.neon --no-progress --error-format=github
 
+  lib-perl:
+    needs: compiler
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -yq
+          # shellcheck disable=SC2086
+          sudo apt-get install -y --no-install-recommends g++ $BUILD_DEPS libperl-critic-perl
+
+      - name: Configure
+        run: |
+          ./bootstrap.sh
+          # shellcheck disable=SC2086
+          ./configure $CONFIG_ARGS_FOR_SCA
+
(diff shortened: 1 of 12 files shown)


Jens-G:
### Code review

Found 3 issues:

1. `ServerSocket::accept()` returns `''` instead of `undef` on the failure path — `return undef;` is deleted rather than replaced with bare `return;`

In Perl, removing `return undef;` causes the sub to return the last evaluated expression (the `if` condition), which is `''` (empty string) on the false path. Since `defined('')` is true, callers that check `if (defined $client)` (e.g. `Server.pm`) will treat a missing handle as a valid client object. The fix is `return;` rather than deleting the line entirely.

https://github.com/apache/thrift/blob/f1989fe25eccfeea015c6e0b8c7e7f001c98c99f/lib/perl/lib/Thrift/ServerSocket.pm#L90-L94

2. Secondary `package` declarations in multi-package files are missing `## no critic (RequireFilenameMatchesPackage)` — the CI scan will fail on them

`Exception.pm` annotates `Thrift::TException` but not `Thrift::TApplicationException`. `Transport.pm` annotates `Thrift::TTransportException` but not `Thrift::TransportFactory` or `Thrift::ServerTransport`.

https://github.com/apache/thrift/blob/f1989fe25eccfeea015c6e0b8c7e7f001c98c99f/lib/perl/lib/Thrift/Exception.pm#L43-L47

https://github.com/apache/thrift/blob/f1989fe25eccfeea015c6e0b8c7e7f001c98c99f/lib/perl/lib/Thrift/Transport.pm#L131-L135

https://github.com/apache/thrift/blob/f1989fe25eccfeea015c6e0b8c7e7f001c98c99f/lib/perl/lib/Thrift/Transport.pm#L158-L162

3. Before squashing, consider adding a `Client:` line to the commit message — the project convention ([AGENTS.md §2](https://github.com/apache/thrift/blob/master/AGENTS.md#L34-L38)) is `Client: cpp,perl` for changes spanning the compiler and Perl library.

PR #3450, opened 2026-05-08
Fix CloudTruth project name in publish workflow

alekseidudchenko-osp:

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,52 @@
+#
+# Apache Thrift release publishing workflow
+#
+# When a release is created in GitHub, it should be created in pre-release
+# mode first.  This will trigger a "prereleased" event which will cause this
+# workflow to run and publish packages to various package managers.  You
+# must check the Actions tab in GitHub to see the result of the workflow.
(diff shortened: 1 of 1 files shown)


alekseidudchenko-osp:
@Jens-G this should fix the 'publish 0.23.0 to public Maven' pipeline problem.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
+#
```
Jens-G, line 1: ASF header missing

Jens-G (changes requested):
> @Jens-G this should fix the 'publish 0.23.0 to public Maven' pipeline problem.

Can you explain what the idea is here? It worked before without that, so I would like at least a bit more information than "this solves some problem" which is not even described at all. As it stands, it is not a proper description of any kind.

kpumuk:
This is a Python library release action on pypi, that mostly duplicates another one in https://github.com/apache/thrift/blob/master/.github/workflows/pypi.yml workflow. What's the purpose of this change? Also, see https://github.com/apache/thrift/pull/3440

PR #3458, opened 2026-05-11
Prevent concurrent calls to socketConn.Close() in Go

erka:
Client: go

Multiple goroutines calling Close() simultaneously could result in
sc.closed.Store() being called concurrently. Use CompareAndSwap instead
to atomically check and set the closed flag, ensuring only the first
caller closes the underlying net.Conn

- [ ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes)

### DIFF
--- a/lib/go/thrift/socket_conn.go
+++ b/lib/go/thrift/socket_conn.go
@@ -114,11 +114,16 @@ func (sc *socketConn) Read(p []byte) (n int, err error) {
 	return sc.Conn.Read(p)
 }
 
+// Close closes the connection and the underlying net.Conn.
+//
+// Only the first call will actually close the underlying connection;
+// subsequent calls return net.ErrClosed error.
 func (sc *socketConn) Close() error {
-	if !sc.isValid() {
-		// Already closed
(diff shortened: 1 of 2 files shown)


lib/go/thrift/common_test.go
```diff
@@ -103,4 +105,39 @@ func (p *mockWrappableProcessor) AddToProcessorMap(name string, processorFunc TP
 var (
 	_ TProcessor = (*mockProcessor)(nil)
 	_ TProcessor = (*mockWrappableProcessor)(nil)
+	_ net.Conn   = (*mockConn)(nil)
 )
+
+type mockConn struct{}
```
fishy: since this is only used in socket_conn_test.go, let's move it there. putting it here is a bit too broad (this mock cannot be used as generic connection alternative in tests either, it's only for very specific cases)

fishy (approved):
can you squash your commits please?

PR #3467, opened 2026-05-12
THRIFT-5987: Fix PHP protocol type-safety bugs in readBool and popContext

Jens-G:
## Summary

- **`TCompactProtocol::readBool`** passes its `?bool &$bool` reference directly to `readByte(?int &$byte)`. In `STATE_CONTAINER_READ`, `readByte` unpacks a raw byte and writes an integer back through the reference — leaving a `?bool`-declared variable holding an integer. Fix routes through a local `?int $byte` variable and casts to `(bool)` before assignment.

- **`TJSONProtocol::popContext`** assigns `array_pop($this->contextStack)` — whose return type is `BaseContext|null` — to the non-nullable `private BaseContext $context` property. On any context-stack underflow this causes a `TypeError`. Fix uses `?? new BaseContext()` as a null-safe fallback.

- **`TCompactProtocolTest`** — `testWriteListBegin` and `testWriteSettBegin` dropped the `$size` argument from the `writeCollectionBegin` mock expectation, so forwarding of the element-count parameter was no longer verified. Restored `->with(TType::STRING, 1)`.

## Test plan

- [ ] `lib/php`: `composer install && vendor/bin/phpunit`
- [ ] `make style` passes in `lib/php`

Generated-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/lib/php/lib/Protocol/TCompactProtocol.php
+++ b/lib/php/lib/Protocol/TCompactProtocol.php
@@ -568,7 +568,11 @@ class TCompactProtocol extends TProtocol
 
             return 0;
         } elseif ($this->state == TCompactProtocol::STATE_CONTAINER_READ) {
-            return $this->readByte($bool);
+            $byte = null;
+            $result = $this->readByte($byte);
+            $bool = (bool) $byte;
+
+            return $result;
         } else {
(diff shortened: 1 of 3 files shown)


Jens-G:
@sveneld same here

PR #3470, opened 2026-05-12
THRIFT-5989: Work around JWT-format GITHUB_TOKEN breaking composer install

Jens-G:
## Summary

Follow-up to #3469: the `COMPOSER_AUTH='{}'` env-var approach in that PR does not work because it only overrides Composer's *local* auth.json. The global auth.json written by `shivammathur/setup-php` (which holds the invalid JWT token) is read and validated independently, before the env var takes effect.

This PR replaces the env-var workaround with an explicit `composer config --global --unset github-oauth.github.com || true` before each `composer install` call, removing the bad token at the source.

Affected steps: `lib-php` matrix ("Install Dependencies") and `cross-test` ("Install PHP dependencies").

## Test plan

- [ ] `lib-php (8.x)` jobs pass
- [ ] `cross-test` job passes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,9 @@ jobs:
           ini-values: "error_reporting=E_ALL"
 
       - name: Install Dependencies
-        run: composer install
+        run: |
+          composer config --global --unset github-oauth.github.com || true
+          composer install
 
       - name: Run bootstrap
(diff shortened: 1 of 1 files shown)


Jens-G:
Closing — replaced by a conflict-free branch. Reopening with correct head.

PR #3475, opened 2026-05-14
THRIFT-5802: Validate set uniqueness in Node.js library

KantConnect:
A Thrift set is defined to contain unique elements, but currently different language implementations have handled duplicate elements inconsistently: Go throws, Python silently deduplicates, while Node.js passed duplicates through unchanged, breaking cross-language interop.

Add duplicate-element validation in the JavaScript code generator and runtime libraries:

  * lib/nodejs/lib/thrift/thrift.js, lib/js/src/thrift.js: new Thrift.checkSetUniqueness(arr) helper. Duplicate detection via a native JS Set; throws TProtocolException(INVALID_DATA) on the first duplicate. Both the Node.js and the browser/ES6/TS runtime libraries are updated, since the JS code generator emits the same Thrift.checkSetUniqueness call regardless of target.

  * compiler/cpp/src/thrift/generate/t_js_generator.cc: in generate_serialize_container() emit a Thrift.checkSetUniqueness call immediately before output.writeSetBegin(), so the validation runs before any bytes hit the transport. In generate_deserialize_container() for sets, allocate one parallel `new Set()` before the per-element loop; in generate_deserialize_set_element() guard the push with seen.has(elem) / seen.add(elem).

  * lib/nodejs/test/check_set_uniqueness.test.js, lib/nodejs/test/testAll.sh: new tape unit tests covering unique sets, duplicate sets across primitive types, empty/single-element edge cases, and the thrown exception type.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_js_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_js_generator.cc
@@ -193,7 +193,10 @@ public:
 
   void generate_deserialize_container(std::ostream& out, t_type* ttype, std::string prefix = "");
 
-  void generate_deserialize_set_element(std::ostream& out, t_set* tset, std::string prefix = "");
+  void generate_deserialize_set_element(std::ostream& out,
+                                        t_set* tset,
+                                        std::string prefix,
+                                        std::string seen);
 
   void generate_deserialize_map_element(std::ostream& out, t_map* tmap, std::string prefix = "");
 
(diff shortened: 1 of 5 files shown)


compiler/cpp/src/thrift/generate/t_js_generator.cc
```diff
@@ -2397,6 +2400,7 @@ void t_js_generator::generate_deserialize_struct(ostream& out, t_struct* tstruct
 void t_js_generator::generate_deserialize_container(ostream& out, t_type* ttype, string prefix) {
   string size = tmp("_size");
   string rtmp3 = tmp("_rtmp3");
+  string seen; // populated only for sets, used for O(n) duplicate detection on read
```
Jens-G: The comment seems off: O(n) or O(1) ?

Jens-G (approved):
+1 LGTM otherwise

compiler/cpp/src/thrift/generate/t_js_generator.cc
KantConnect: Done, fixed the comment.

PR #3488, opened 2026-05-15
THRIFT-5996: connection check supports TLS sockets

murfffi:
Client: go

The connection check first introduced in THRIFT-5214 now supports TLS connections using a technique from the code that initially inspired the check:
https://github.com/go-sql-driver/mysql/pull/934

Testing done: new test with alongside the existing test of conn. check

### DIFF
--- a/lib/go/thrift/socket_unix_conn.go
+++ b/lib/go/thrift/socket_unix_conn.go
@@ -22,6 +22,7 @@
 package thrift
 
 import (
+	"crypto/tls"
 	"errors"
 	"io"
 	"syscall"
@@ -38,7 +39,12 @@ func (sc *socketConn) read0() error {
 }
 
 func (sc *socketConn) checkConn() error {
-	syscallConn, ok := sc.Conn.(syscall.Conn)
+	rawConn := sc.Conn
+	if tlsConn, ok := rawConn.(*tls.Conn); ok {
+		rawConn = tlsConn.NetConn()
+	}
+
+	syscallConn, ok := rawConn.(syscall.Conn)
 	if !ok {
 		// No way to check, return nil
 		return nil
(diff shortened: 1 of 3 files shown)


lib/go/thrift/socket_unix_conn.go
```diff
@@ -38,7 +39,14 @@ func (sc *socketConn) read0() error {
 }
 
 func (sc *socketConn) checkConn() error {
-	syscallConn, ok := sc.Conn.(syscall.Conn)
+	rawConn := sc.Conn
+	syscallConn, ok := rawConn.(syscall.Conn)
+	if !ok {
+		if connTLS, isTLS := rawConn.(*tls.Conn); isTLS {
+			rawConn = connTLS.NetConn()
```
fishy: in the jira ticket your proposal was to do this before trying to get `syscallConn`. is there a reason you flipped it to this approach?

I like that approach better because that way you don't have the next line that rewrites `syscallConn, ok` which is very error-prone (e.g. if you used `ok` instead of `isTLS` on line 45 that's a bug that's hard to spot).

murfffi: Thanks for the review! I agree that the version is more convoluted and thus error prone . 

I wondered if it is likely that, in the future, Go team makes *tls.Config implement syscall.Conn directly, possibly to utilize https://docs.kernel.org/networking/tls-offload.html on Linux. 

I though about this more and I believe that if something drastic like that happens, the syscall code may not work and we should unwrap the connection anyway.

I changed it back to the variant in JIRA. Please take another look.

murfffi: I reviewed https://github.com/golang/go/issues/44506 - tracking kernel TLS support in Go . It is in state Proposal-Accepted but not implemented. A couple of people implemented crypto/tls forks as PoC. I think the point still stands - there is no value to check if tls.Conn implements syscall.Conn.

fishy (approved):

PR #3491, opened 2026-05-15
[INFRA] Set up default rulesets for default and release branches

asf-gitbox-commits:
This Pull Request enables the repository to conform with the "sane default security settings" of the Apache Software Foundation by configuring a default branch ruleset that protects the default branch and any release branches.

Note that `~DEFAULT_BRANCH` is a GitHub symbolic link to the current default branch (HEAD) of the repository and does not need changing.
If the managing project does not wish to set up these defaults, please close this Pull Request. Alternatively, the project may merge this Pull Request to apply the changes immediately.

If no action is taken, this Pull Request will be automatically merged by the Apache Infrastructure team on **2026-06-14** (30 days from now).

For any further information, please reach us on Slack or at: users@infra.apache.org

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -8,6 +8,20 @@ notifications:
   jira_options: link label worklog
 
 github:
+  rulesets:
+    - name: "Default Branch Protection"
+      type: branch
+      branches:
+        includes:
+          - "~DEFAULT_BRANCH"
+          - "release/*"
+          - "rel/*"
+          - "0.*"      # existing release branches are all 0.xx.xx
+        excludes: []
+      bypass_teams:
+        - root
+      restrict_deletion: true
+      restrict_force_push: true
   autolink_jira:
     - THRIFT
   enabled_merge_buttons:


.asf.yaml
```diff
@@ -1,8 +1,21 @@
...
+
+github:
+  rulesets:
+    - name: "Default Branch Protection"
+      type: branch
+      branches:
+        includes:
+          - "~DEFAULT_BRANCH"
+          - "release/*"
+          - "rel/*"
```
kpumuk, line 18: @Jens-G we don't seem to use `release/` or `rel/`, should we protect release tags `v*` instead? Or rather in addition to, so you can use `release/0.24.0` for release work instead of `0.24.0`

ctubbsii: Because some projects have subprojects (multiple releases, with different names), I think it's a good opportunity to start using better tag names instead of just the raw version numbers, as in `rel/thrift-0.24.0`. If there's any sub-projects, like "thrift-examples", then it could be `rel/thrift-examples-1.0.0`, for example.

Having all the releases under one tag namespace allows other kinds of tags to be used for other purposes in their own namespace that don't look like `rel/` (e.g. release candidates, tests, abandoned branches, etc.)

Jens-G: Are we talking about branches or tags? 
~Don't forget to adjust the ReleaseManagement document.~ [Done](https://github.com/apache/thrift/commit/fe86fc6e1d53be1c02a8e51e506fcf0572640ef3)

Jens-G:
Added the `0.*` rule to protect existing release branches. Otherwise I'm fine with it.

.asf.yaml
ctubbsii: @Jens-G We're talking about tags only, not branches.

I see you updated the docs to use `release/<version>` instead of the shorter `rel/<version>`. If you're willing for the tags to be longer, why not also include my earlier suggestion to use `thrift-<version>` instead of just `<version>`? So, `release/thrift-<version>`. I think having the release artifact name in addition to the version makes a lot of sense.

Jens-G: Whatever you want. I mean, really. Pull request, check, merge.

PR #3495, opened 2026-05-19
THRIFT-6012: Precompile regex patterns in Go struct validator

erka:
Client: go

Replace inline regexp.MatchString calls with precompiled
regexp.MustCompile variables for literal vt.pattern annotations.
This avoids recompiling the same regex on every Validate() call.

Also fixes two bugs in the original regexp.MatchString usage:

- Arguments were swapped: MatchString(target, pattern) instead
  of MatchString(pattern, target)
- Used ; ok instead of ; !ok, causing validation to fail when the
  pattern matched instead of when it didn't

---

- [x ] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  - THRIFT-6012
### Example

```thrift
struct StarSystem {
  1: string id ( vt.pattern = "^[a-z_-]+$", vt.pattern = ".*")
  2: string hash (vt.pattern = "$id")
}
```

### Output
```go

// Precompiled regex patterns for StarSystem vt.pattern validation
var (
	vtReStarSystem0 = regexp.MustCompile(`^[a-z_-]+$`)
	vtReStarSystem1 = regexp.MustCompile(`.*`)
)

func (p *StarSystem) Validate() error {
	if !vtReStarSystem0.MatchString(p.ID){
		return thrift.NewValidationException(thrift.VALIDATION_FAILED, "vt.pattern", "StarSystem.id", "StarSystem.id not valid, rule vt.pattern check failed")
	}
	if !vtReStarSystem1.MatchString(p.ID){
		return thrift.NewValidationException(thrift.VALIDATION_FAILED, "vt.pattern", "StarSystem.id", "StarSystem.id not valid, rule vt.pattern check failed")
	}
	if ok, _ := regexp.MatchString(string(p.ID), p.Hash); !ok {
		return thrift.NewValidationException(thrift.VALIDATION_FAILED, "vt.pattern", "StarSystem.hash", "StarSystem.hash not valid, rule vt.pattern check failed")
	}
	return nil
}
```

### DIFF
--- a/compiler/cpp/src/thrift/generate/go_validator_generator.cc
+++ b/compiler/cpp/src/thrift/generate/go_validator_generator.cc
@@ -679,15 +679,29 @@ void go_validator_generator::generate_string_field_validator(std::ostream& out,
       }
       out << ")";
     } else if (key == "vt.pattern") {
-      out << indent() << "if ok, _ := regexp.MatchString(" << target << ", ";
       if (values[0]->is_field_reference()) {
+        out << indent() << "if ok, _ := regexp.MatchString(";
         out << "string(";
         out << get_field_reference_name(values[0]->get_field_reference());
         out << ")";
+        out << ", " << target << "); !ok";
       } else {
-        out << "\"" << values[0]->get_string() << "\"";
+        std::string pattern = values[0]->get_string();
+        std::string var_name;
+        bool found = false;
+        for (auto& entry : pattern_cache_) {
+          if (entry.first == pattern) {
+            var_name = entry.second;
+            found = true;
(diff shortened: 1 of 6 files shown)


fishy (approved):
looks good to me. the lib-netstd failure doesn't seem relevant. cc @Jens-G to double check on that one.

kpumuk:
> looks good to me. the lib-netstd failure doesn't seem relevant. cc @Jens-G to double check on that one.

We have merged a fix for that https://github.com/apache/thrift/pull/3501

PR #3499, opened 2026-05-19
THRIFT-6013: Add recursion depth limit to skip() in Ruby library

Jens-G:
## Summary

- Adds a `max_depth` parameter (default 64) to `BaseProtocol#skip`
- Raises `ProtocolException::DEPTH_LIMIT` when the depth limit is exceeded
- Aligns the Ruby implementation with Java, Python, Go and netstd, which already enforce this limit
- Adds two unit tests to `base_protocol_spec.rb` covering the limit boundary

## Test plan

- [ ] `spec/base_protocol_spec.rb` — new tests: raises `DEPTH_LIMIT` at `max_depth=0`, no raise at `max_depth=1` for a leaf type
- [ ] Existing skip tests (`should skip structs/maps/sets/lists`) continue to pass

### DIFF
--- a/lib/rb/lib/thrift/protocol/base_protocol.rb
+++ b/lib/rb/lib/thrift/protocol/base_protocol.rb
@@ -342,7 +342,8 @@ module Thrift
       end
     end
 
-    def skip(type)
+    def skip(type, max_depth = 64)
+      raise ProtocolException.new(ProtocolException::DEPTH_LIMIT, 'Maximum skip depth exceeded') if max_depth <= 0
       case type
       when Types::BOOL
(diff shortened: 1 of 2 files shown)


kpumuk (approved):
note for myself: this change only covers the generic unknown-field `skip()` path. Expected-field deserialization still recurses without a limit for nested structs.

That makes this narrower than the Ruby configuration plan in [THRIFT-5938](https://issues.apache.org/jira/browse/THRIFT-5938), which calls for config-driven recursion enforcement in `skip`, generated reads, generated writes, and native paths (still in progress).

PR #3514, opened 2026-05-21
THRIFT-6021: Make C++ HTTP client work with oneway requests

kpumuk:
C++ HTTP transport did not handle `oneway` calls correctly on persistent HTTP connections.

A Thrift `oneway` method has no Thrift reply message, but over HTTP the server still needs to complete the HTTP request/response exchange. Previously, the C++ generated client sent the one-way request and returned without consuming the HTTP response. The next request on the same HTTP connection then tried to read that stale HTTP response as if it belonged to the new RPC, which poisoned the connection and caused follow-up calls to fail or time out.

The proposed fix is to add a transport-level `onewayComplete()` hook. On the client side, `THttpClient` now marks that a one-way HTTP response is pending. It does not block the one-way call waiting for the response. Instead, before the next HTTP request is sent on the same connection, it drains and discards the pending HTTP response body. This keeps one-way calls non-blocking from the caller’s perspective while preserving HTTP keep-alive instead of closing and reopening the socket.

On the server side, generated C++ processors now notify the output transport when a one-way method has finished. `THttpServer` handles that hook by sending an empty HTTP 200 response, so the client has a valid HTTP response to discard.

The hook is forwarded through wrapper transports such as buffered, framed, header, and zlib transports so layered protocol/transport combinations work as expected.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6021](https://issues.apache.org/jira/browse/THRIFT-6021)


Jens-G:
Somehow that sounds weird. Also, assumed I do a lot of oneway calls, what would be the performance impact?

kpumuk:
> Somehow that sounds weird. Also, assumed I do a lot of oneway calls, what would be the performance impact?

Oh for sure it is undeniably weird. The PR is mostly to gather ideas on how to address it, and which side must handle :-)

Here is an example: https://github.com/kpumuk/thrift/actions/runs/26243739717/job/77238383058 . From `rb-cpp_compact_http-ip_client.log`:
```
testClient.testOneway(1) =>  success - took 0.03 ms
re-test testI32(-1)terminate called after throwing an instance of 'apache::thrift::transport::TTransportException'
  what():  No more data to read.
Return code: -6 (negative values indicate kill by signal)
Test execution took 1.2 seconds.
Thu May 21 18:17:57 2026
```

kpumuk:
I highly suspect that these are related too:

* https://github.com/apache/thrift/blob/master/test/known_failures_Linux.json#L309-L316
* https://github.com/apache/thrift/blob/master/test/known_failures_Linux.json#L648 (and other `py-cpp-*-http-ip` known failures)

kpumuk:
Java suffers from the same issue https://github.com/kpumuk/thrift/actions/runs/26251658457/job/77265909596:

```
testClient.testMultiException("success", "test 3") =>  {{"test 3"}}
testOneway(3)...Oneway test took too long to execute failed: took 3004ms
oneway calls are 'fire and forget' and therefore should not cause blocking.
Some transports (HTTP) have a required response, and typically this failure
means the transport response was delayed until after the execution
of the RPC.  The server should post the transport response immediately and
before executing the RPC.
*** FAILURE ***
Total time: 3274766us
Min time: 3274766us
Max time: 3274766us
Avg time: 3274766us

Sample TSimpleJSONProtocol output:
{"userMap":{"5":5,"8":8},"xtructs":[{"string_thing":"Goodbye4","byte_thing":4,"i32_thing":4,"i64_thing":4},{"string_thing":"Hello2","byte_thing":2,"i32_thing":2,"i64_thing":2}]}
Return code: 1 (negative values indicate kill by signal)
Test execution took 3.5 seconds.
Thu May 21 20:52:21 2026
```

kpumuk:
Alright, I think there are 2 bugs:

* C++ does not drain the response, which HTTP server generates
* Ruby does not return immediately, instead running the handler synchronously

Will see how to fix it...

kpumuk:
Alright, I re-did the change following Go example, basically making HTTP client to read and ignore the response from oneway call before the next RPC call, if the connection is still open. This should preserve oneway performance, while properly handling HTTP protocol semantics. 

Updated PR description to match the change and removed known failures related to HTTP that are no longer failures.

kpumuk:
Also fixes PHP cross tests with C++ https://github.com/apache/thrift/pull/3515/changes/8489c84ffc7cf584c2bd3e2e7db9ad3984b17884 (see https://github.com/apache/thrift/pull/3515)

Jens-G:
+1 LGTM (not tested)

PR #3515, opened 2026-05-21
THRIFT-6023: Add HTTP transport support to PHP cross-tests

sveneld:
Adds HTTP transport support to the PHP entry of the cross-test matrix. PHP was the only major-language entry stuck on `buffered`/`framed` while Python, Go, Node, C++, Java, Lua, Dart, JS, D, and Haskell already declare `http` and verify HTTP interop.

**Server** — `TestServer.php` detects `--transport=http` in CLI mode and `pcntl_exec()`s into PHP's built-in web server (`php -S 127.0.0.1:$port TestServer.php`). The same script re-enters under SAPI `cli-server`, reads the request via `TPhpStream` from `php://input`, dispatches it through the existing processor, and writes the response to `php://output`. The original `-d` flags from the launcher cmdline (e.g. `-dextension=thrift_protocol.so`) are preserved by parsing `/proc/self/cmdline`, so the accel extension and sockets stay loaded under the spawned server too. The cross-runner's `socket.connect_ex` readiness probe works because `php -S` TCP-binds the port immediately.

**Client** — `TestClient.php` gains an `if ($MODE == 'http')` branch that uses `TPsrHttpClient` (added in THRIFT-6010) against `http://127.0.0.1:$port/`. `TPsrHttpClient` buffers internally, so no `TBufferedTransport`/`TFramedTransport` wrapping.

**tests.json** — adds `"http"` to PHP server and client `transports`; bumps client `timeout` from 6s → 10s to absorb HTTP overhead.

**composer.json** — adds `guzzlehttp/guzzle: ^7.8` to `require-dev` so the PSR-18 auto-discovery in `TPsrHttpClient` resolves a concrete HTTP client at cross-test time.

JIRA: [THRIFT-6023](https://issues.apache.org/jira/browse/THRIFT-6023). Draft until cross-test CI validates the new php↔php / php↔py3 / php↔go HTTP matrix cells.


kpumuk:
The last commit should hopefully be unnecessary with https://github.com/apache/thrift/pull/3514.

For Java, the issue is that it is the only cross-test that actually verifies the one-way semantics, as in that HTTP server must return the response immediately, and then run the handler afterwards (so that the client is as fast as possible).

sveneld:
Thanks, i will update PR

sveneld:
@kpumuk Thanks — addressed both points in 025863b5e:

* Implemented oneway fast-response in the cli-server router: the request is peeked for `TMessageType::ONEWAY`, and on a match we send an empty HTTP 200 (`Content-Length: 0`) and `pcntl_fork()` the handler so the client's oneway call returns immediately, matching the THttpServer pattern from THRIFT-6021. This drops the four `php-java_*_http-ip` known_failures entries.
* Removed the four `php-cpp_*_http-ip` known_failures entries on the assumption that #3514 lands before this PR — let me know if you'd prefer I re-add them as a temporary hedge until #3514 merges.

While I was in there also collapsed the `if/elseif` cascade in `TestClient.php` to two `match` expressions and dropped the dead `TSocket` assignment / unused `$hosts`/`$logger`.

sveneld:
@kpumuk Tested after #3514 merged — unfortunately it didn't fix `php-cpp_*_http-ip`. The remaining failure isn't on the PHP side; it's a C++ `THttpClient` bug: when the server emits `Connection: close` and tears down the TCP connection (which PHP's built-in `cli-server` does by design), the C++ client doesn't reopen the socket on the next call and dies with a broken-pipe / buffer-refill error on the second RPC over the same client instance.

Filed as [THRIFT-6060](https://issues.apache.org/jira/browse/THRIFT-6060). The 4 `php-cpp_*_http-ip` entries stay in `known_failures_Linux.json` until that lands.

PR #3525, opened 2026-05-24
THRIFT-6038: Markdown generator: default .md extension and @param/@return table

Jens-G:
## Summary

- Output files now use the `.md` extension by default (previously extension-less). The `suffix` option still overrides; `suffix=` (empty) explicitly suppresses any extension.
- `@param` and `@return` tags in function doc-comments are now rendered as a Markdown table instead of being buried in inline prose. Tags are only matched at the start of a line, so `@`-signs in ordinary prose (e.g. email addresses) are unaffected. Multi-line descriptions are joined; `|` characters are escaped.
- Adds test fixture `DocTest.thrift`, golden file `DocTest.md`, and 6-case Python unittest `markdown_doc_test.py` registered as `MarkdownDocTest` in CMake.

## Test plan

- [x] Compiler builds cleanly (`ninja thrift-compiler` in `cmake_build_compiler/`)
- [x] All 6 Python tests pass: `python3 compiler/cpp/test/compiler/markdown_doc_test.py <thrift-binary> -v`
- [x] Manual inspection of `ThriftTest.thrift` output: `testMulti` renders 6 param rows + Returns row; `testException` plain prose is unaffected
- [ ] `clang-format` — not available in the build environment; manual style review needed before merge

## Related

JIRA: https://issues.apache.org/jira/browse/THRIFT-6038


Jens-G:
### Code review

Found 2 issues:

1. `test_inline_at_sign_not_treated_as_tag` asserts that `"@param"` appears in the `plain` function's documentation section, but `DocTest.thrift` gives `plain` only the docstring `"Plain prose doc, no @param or @return tags."` — the generated Markdown will never contain `"@param"` there, so the `assertIn` will always fail.

https://github.com/apache/thrift/blob/871ad14aa9d0bdd290af0f06a249650a8b241c19/compiler/cpp/test/compiler/markdown_doc_test.py#L99-L103

2. The PR test plan leaves the `make style` checkbox unchecked, acknowledging `clang-format` was not run. [AGENTS.md](https://github.com/apache/thrift/blob/871ad14aa9d0bdd290af0f06a249650a8b241c19/AGENTS.md) requires "Verify coding standards: `make style`" (Section 3) and lists it as a mandatory pre-PR checklist item (Section 6).

https://github.com/apache/thrift/blob/871ad14aa9d0bdd290af0f06a249650a8b241c19/AGENTS.md#L51-L51

PR #744, opened 2015-12-15
update TSocket.php

lihanharry:
- add connect timeout
- support accurate send timeout and write timeout

PR #3531, opened 2026-05-26
THRIFT-4171: Add setConnectTimeout to PHP TSocket

sveneld:
Resolves [THRIFT-4171](https://issues.apache.org/jira/browse/THRIFT-4171).

`TSocket::setSendTimeout()` was the only way to set the timeout passed to `fsockopen()`/`pfsockopen()`, so callers had to choose between a short connect timeout (good) and a short write timeout (often unwanted). Adds a separate `setConnectTimeout(int $timeout)` method; if it is not called, `open()` falls back to the send timeout, preserving today's behaviour.

Two unit tests cover both paths: the BC fallback and the override.

This addresses the same problem as the original PR #744, restructured around the typed properties introduced in recent PHP-library refactors.

- [x] [Apache JIRA](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket: THRIFT-4171
- [x] Squashed to a single commit
- [x] No breaking change — additive method; existing callers keep current behaviour
- [ ] `[skip ci]` (n/a — code change)

### DIFF
--- a/lib/php/lib/Transport/TSocket.php
+++ b/lib/php/lib/Transport/TSocket.php
@@ -87,6 +87,13 @@ class TSocket extends TTransport
     protected int $sendTimeoutUsec = 100000;
 
     /**
+     * True once a caller invoked setSendTimeout(). Used to fire a deprecation
+     * notice in open() when the caller relies on the send-timeout-as-
+     * connect-timeout coupling that this class used to enforce.
+     */
+    private bool $sendTimeoutCustomized = false;
+
+    /**
      * Recv timeout in seconds
      *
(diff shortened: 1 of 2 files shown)


sveneld:
Pushed dfc68b3d7 adding an `E_USER_DEPRECATED` notice when `open()` falls back to the send timeout for the connect step. Gated on whether the caller actually invoked `setSendTimeout()`, so default-constructed sockets (or sockets where only the new `setConnectTimeout()` is used) stay silent. Notice wording follows the pre-1.0 convention — "will be removed in the next version".

PR #3532, opened 2026-05-26
THRIFT-1209: Regression test for PHP-reserved words as Thrift enum names

sveneld:
Resolves [THRIFT-1209](https://issues.apache.org/jira/browse/THRIFT-1209).

The original bug: an IDL `enum { GLOBAL = 1, ... }` produced PHP code (`class Foo { const GLOBAL = 1; }`) that failed to parse on older PHP. PHP 7.1+ permits reserved identifiers as class constants (except `class`), so the historical breakage is gone — but there is no test pinning that behaviour. This adds one:

- `lib/php/test/Resources/ThriftTest.thrift` gains `enum ReservedKeywordEnum { GLOBAL, STATIC, LIST, RETURN }` (PHP-only test thrift, won't affect cross-tests).
- New `lib/php/test/Unit/Compiler/ReservedKeywordEnumTest.php` references `\Basic\TestValidators\ReservedKeywordEnum::GLOBAL` etc., forcing the generated stub to be parsed by PHP. If the generator ever regresses and emits unparseable code for these names, phpunit will surface it.

- [x] [Apache JIRA](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket: THRIFT-1209
- [x] Single squashed commit
- [x] No breaking change — test-only
- [ ] `[skip ci]` (n/a — test addition)

### DIFF
--- a/lib/php/test/Resources/ThriftTest.thrift
+++ b/lib/php/test/Resources/ThriftTest.thrift
@@ -26,6 +26,17 @@ union UnionOfStrings {
   2: string bb;
 }
 
+// Regression for THRIFT-1209: PHP reserved words must round-trip through the
+// generator as class constants. PHP 7.1+ permits reserved identifiers as
+// class constants (except `class`); the generator must not mangle or reject
+// them. See lib/php/test/Unit/Compiler/ReservedKeywordEnumTest.php.
+enum ReservedKeywordEnum {
+  GLOBAL = 1,
+  STATIC = 2,
+  LIST = 3,
+  RETURN = 4,
+}
+
 service TestService {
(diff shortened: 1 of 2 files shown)


sveneld:
Closing — the test was misaimed. THRIFT-1209 was about the legacy non-namespaced PHP generator output (`final class metastore_HiveObjectType { const GLOBAL = 1; ... }`) failing on older PHP. That was resolved by two independent refactors: (1) the PHP minimum bump to 8.1 (THRIFT-5956 / PHP 7.1+ allows reserved words as class constants), and (2) the namespaced PHP generator output. Adding a regression test today tests the current behaviour rather than the original bug. Will note that on the JIRA.

PR #3533, opened 2026-05-26
THRIFT-1636: Regression test ensuring generated PHP has no closing tag

sveneld:
Resolves [THRIFT-1636](https://issues.apache.org/jira/browse/THRIFT-1636).

The original ask was to drop `?>` closing tags from generated PHP. That happened years ago (the generator emits no `?>` anywhere — verified with grep on `compiler/cpp/src/thrift/generate/t_php_generator.cc`, and the regenerated stubs under `lib/php/test/Resources/packages/` all end with `}` and a newline). No test pins that behaviour today.

This adds a phpunit data-provider that walks every generated `.php` under `lib/php/test/Resources/packages/` and asserts the file does not end with `?>`. If the generator ever regresses (or someone adds a hand-written fixture with a closing tag into one of those package dirs), the lib-php CI job surfaces it immediately.

- [x] [Apache JIRA](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket: THRIFT-1636
- [x] Single squashed commit
- [x] No breaking change — test-only
- [ ] `[skip ci]` (n/a — test addition)

### DIFF
--- a/lib/php/test/Unit/Compiler/NoClosingPhpTagTest.php
+++ b/lib/php/test/Unit/Compiler/NoClosingPhpTagTest.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


sveneld:
Closing — not adding a regression test. THRIFT-1636 was already resolved as part of the PHP 8.1 generator refactor; will note that on the JIRA.

PR #3534, opened 2026-05-27
THRIFT-6028: Harden Perl binary protocol negative sizes

Jens-G:
## Summary

- Adds negative-size guards in `readMapBegin`, `readListBegin`, `readSetBegin`, `readString`, and `readStringBody` in `Thrift::BinaryProtocol`
- Fixes pre-existing bug in `Thrift::TProtocolException::new` (was calling `SUPER::new()` without forwarding arguments, so message and code were never set)
- Adds `t/binary_protocol_negative_size.t` with 4 tests covering all guarded methods

## Test plan

- [ ] `perl -Ilib t/binary_protocol_negative_size.t` — 4/4 pass

### DIFF
--- a/lib/perl/lib/Thrift/BinaryProtocol.pm
+++ b/lib/perl/lib/Thrift/BinaryProtocol.pm
@@ -321,10 +321,14 @@ sub readMapBegin
     my $self = shift;
     my ($keyType, $valType, $size) = @_;
 
-    return
+    my $result =
         $self->readByte($keyType) +
         $self->readByte($valType) +
         $self->readI32($size);
+    if ($$size < 0) {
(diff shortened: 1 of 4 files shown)


Jens-G:
> Is this correct? @Jens-G

What exactly are you trying to say? Not too much time to solve that sort of riddles here ... ;-)

erka:
Better to ask than stay silent. It turned out that solving it needed a lot of time. Do we know when this bug was introduced?

Jens-G:
The entire history is in the repo, so that should be possible to find out. Again, if you have comments or improvement proposals or mayb want to send a followup PR - don't hesitate.

PR #3547, opened 2026-05-27
Bound decompressed size for Ruby HeaderTransport ZLIB transform

Jens-G:
Client: rb

### DIFF
--- a/lib/rb/lib/thrift/transport/header_transport.rb
+++ b/lib/rb/lib/thrift/transport/header_transport.rb
@@ -74,6 +74,12 @@ module Thrift
     # Maximum frame size (~1GB)
     MAX_FRAME_SIZE = 0x3FFFFFFF
 
+    # Default decompressed-size cap for ZLIB transform (~15.6 MB, matches other Thrift bindings)
+    DEFAULT_MAX_DECOMPRESSED_SIZE = 16_384_000
+
+    # Chunk size for streaming inflate loop
+    ZLIB_INFLATE_CHUNK_SIZE = 16_384
+
     # Binary protocol version mask and version 1
     BINARY_VERSION_MASK = 0xffff0000
     BINARY_VERSION_1 = 0x80010000
@@ -113,6 +119,7 @@ module Thrift
       @sequence_id = 0
       @flags = 0
       @max_frame_size = MAX_FRAME_SIZE
+      @max_decompressed_size = DEFAULT_MAX_DECOMPRESSED_SIZE
     end
 
     def sequence_id=(sequence_id)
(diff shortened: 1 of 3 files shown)


kpumuk (approved):

lib/rb/lib/thrift/transport/header_transport.rb
```diff
@@ -370,13 +385,48 @@ def parse_header_format(buf)
...
+    # Inflates +compressed+ with a running byte-count check against @max_decompressed_size.
+    # Raises TransportException::SIZE_LIMIT if the decompressed output would exceed the limit.
+    def bounded_inflate(compressed)
+      inflater = Zlib::Inflate.new
+      total = 0
+      chunks = []
+      offset = 0
+      begin
+        while offset < compressed.bytesize
+          out = inflater.inflate(compressed.byteslice(offset, ZLIB_INFLATE_CHUNK_SIZE))
```
kpumuk: There is a memory over-use here. I would recommend defining a `buffer = ''` variable outside of the loop, and the pass it here:


```suggestion
          out = inflater.inflate(compressed.byteslice(offset, ZLIB_INFLATE_CHUNK_SIZE), buffer:)
```

Without it, uncompressing a large payload would require double memory.

lib/rb/lib/thrift/transport/header_transport.rb
```diff
@@ -370,13 +385,48 @@ def parse_header_format(buf)
...
 
       StringIO.new(payload)
     end
 
+    # Inflates +compressed+ with a running byte-count check against @max_decompressed_size.
+    # Raises TransportException::SIZE_LIMIT if the decompressed output would exceed the limit.
+    def bounded_inflate(compressed)
+      inflater = Zlib::Inflate.new
+      total = 0
+      chunks = []
```
kpumuk: Give me some time to benchmark/memory profile this. I am curious if using a string as a result would be more efficient in terms of memory/performance.

Jens-G: Feel free to edit the patch.

lib/rb/lib/thrift/transport/header_transport.rb
Jens-G: good suggestion, thanks

lib/rb/lib/thrift/transport/header_transport.rb
kpumuk: For the 16mb field read 1000 times, the results are:

| Variant | Wall | CPU | Peak RSS |
|---|---:|---:|---:|
| pre-commit unbounded `Zlib::Inflate.inflate` | 81.9s | 83.3s | 153 MB |
| array + `join` | 103.5s | 104.0s | 192 MB |
| append with preallocated capacity | 95.7s | 96.2s | 171 MB |
| **append, no preallocation, final patch** | 103.9s | 104.1s | 131 MB |
| single inflate then size check | 80.0s | 81.3s | 169 MB |

I am currently testing shared chunk buffer (caused segfault in my Ruby, trying to figure out if I did it, or I should report it to Ruby) :-)

So far current option (empty string buffer) seems to be the best.

kpumuk: Nope, definitely segfaults and so unsafe. Current implementation is good to merge

kpumuk (approved):

PR #3550, opened 2026-05-28
THRIFT-6045: Add recursion-depth round-trip test for the Ruby library

Jens-G:
## THRIFT-6045: Add a recursion-depth round-trip test for the Ruby library

This is a **test-only** change — it adds the regression test without the implementation; the limit itself is left to a follow-up (see the discussion below).

`spec/recursion_depth_spec.rb` drives the generated read/write path (`Thrift::Struct#read/#write`, `Thrift::Union#read/#write`) over recursive `RecTree` (struct), `RecUnion` (union) and `RecError` (exception) types added to `ThriftSpec.thrift`, across `BinaryProtocol` and (when the native extension is loaded) `BinaryProtocolAccelerated`.

- **Active** (round-trip / within-limit): a chain at the limit round-trips, a wide shallow tree round-trips, a struct through a `MultiplexedProtocol`, and the exception round-trip.
- **Pending** (limit-enforcement, over-limit write/read for struct, union and exception): the Ruby library does not bound recursion depth yet, so these are expected to fail until the limit is implemented, at which point RSpec flags them to be enabled.

The over-limit read payloads are hand-serialized with the real recursive field (id 1, `list<self>`) so a future guarded reader recurses through the generated path rather than `skip()`.

Validated locally (Ruby 3.0.2, pure-Ruby): **10 examples, 0 failures, 5 pending**. Implementation note for the follow-up: the native `thrift_native` C extension rebinds `Thrift::Struct#read/#write`, so a complete fix must also guard `rb_thrift_struct_write/read` (and the union equivalents) in C, not only the pure-Ruby path.

### DIFF
--- a/lib/rb/spec/ThriftSpec.thrift
+++ b/lib/rb/spec/ThriftSpec.thrift
@@ -185,3 +185,19 @@ struct NestedMapInMapKey {
 struct NestedMapInMapValue {
   2: map<byte, map<byte, byte>> value
 }
+
+# Recursive types for the struct/union/exception recursion-depth limit specs (THRIFT-6045).
+struct RecTree {
+  1: list<RecTree> children
+  2: i16 item
+}
+
+union RecUnion {
+  1: list<RecUnion> children
+  2: i32 leaf
+}
+
+exception RecError {
+  1: list<RecError> children
+  2: i32 leaf
+}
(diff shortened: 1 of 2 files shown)


kpumuk (changes requested):
This PR is insufficient:

* It only covers pure Ruby implementation, and not the native extension
* It does not cover Union
* There are no tests

I am currently working on https://issues.apache.org/jira/browse/THRIFT-5938, with recursion depth enforcement being a part of the problem. I can extract my change to merge it ahead of the configuration class

Jens-G:
> This PR is insufficient:

It is way more than that. That#s hy I made it a draft. Different story.

>  I can extract my change to merge it ahead of the configuration class

Ok I only push the test then and leave it up to you, including the ticket.

PR #3552, opened 2026-05-28
THRIFT-6047: Limit struct/union/exception read/write recursion depth in PHP library

Jens-G:
### Change
Enforces a recursion-depth counter (`incrementRecursionDepth`/`decrementRecursionDepth` on `TProtocol`, limit `DEFAULT_RECURSION_DEPTH` = 64, `TProtocolException::DEPTH_LIMIT` on excess) at the two **mutually exclusive** generated-code call sites: the generator wraps the inline read/write loop (default `--gen php`), and `TBase::readStruct`/`writeStruct` carry the guard for **oop** mode. Previously only `skip()` was bounded.

**Also guards `TException::readStruct`/`writeStruct`.** `TException` does not extend `TBase` — it carries its own copies of those methods — so oop-mode **exceptions** (whose generated `read()`/`write()` delegate to them) remained unbounded after the `TBase` change. They now increment / `try` / `finally` / decrement identically, closing the gap. (Found while adding the exception test below.)

### Test
A generated-code round-trip integration test over a recursive struct (`RecTree`), union (`RecUnion`) and exception (`RecError`) from `RecursionDepth.thrift`, generated in **both** modes (`PhpRec` inline, `PhpRecOop` oop) and exercised over Binary, Compact and JSON:
- a chain at exactly the limit round-trips (proving the inline and oop guards do not double-count — it would reject at 32 otherwise);
- a chain past the limit is rejected on write;
- a hand-serialized over-limit payload is rejected on read, crafted with the real recursive field (id 1 = `list<self>`) so the reader recurses through the guarded struct path rather than the separate, unbounded `skip()`.

Validated locally (PHP 8.1, PHPUnit 10.5): **54/54 pass**; `phpcs` (PSR-12) and `phpstan` clean. Without the `TException` guard, the oop-mode exception write/read cases fail (the gap is real). `binary_inline` mode (no `TProtocol`) stays unguarded as before.

### DIFF
--- a/lib/php/lib/Base/TBase.php
+++ b/lib/php/lib/Base/TBase.php
@@ -216,47 +216,52 @@ abstract class TBase
         $fname = null;
         $ftype = 0;
         $fid = 0;
-        $xfer += $input->readStructBegin($fname);
-        while (true) {
-            $xfer += $input->readFieldBegin($fname, $ftype, $fid);
-            if ($ftype == TType::STOP) {
-                break;
-            }
-            if (isset($spec[$fid])) {
-                $fspec = $spec[$fid];
-                $var = $fspec['var'];
-                if ($ftype == $fspec['type']) {
-                    $xfer = 0;
-                    if (isset(TBase::$tmethod[$ftype])) {
-                        $func = 'read' . TBase::$tmethod[$ftype];
-                        $xfer += $input->$func($this->$var);
-                    } else {
-                        switch ($ftype) {
-                            case TType::STRUCT:
-                                $class = $fspec['class'];
-                                $this->$var = new $class();
-                                $xfer += $this->$var->read($input);
-                                break;
-                            case TType::MAP:
-                                $xfer += $this->readMap($this->$var, $fspec, $input);
-                                break;
-                            case TType::LST:
-                                $xfer += $this->readList($this->$var, $fspec, $input, false);
-                                break;
-                            case TType::SET:
-                                $xfer += $this->readList($this->$var, $fspec, $input, true);
-                                break;
+        $input->incrementRecursionDepth();
+        try {
+            $xfer += $input->readStructBegin($fname);
+            while (true) {
+                $xfer += $input->readFieldBegin($fname, $ftype, $fid);
+                if ($ftype == TType::STOP) {
+                    break;
+                }
+                if (isset($spec[$fid])) {
+                    $fspec = $spec[$fid];
+                    $var = $fspec['var'];
+                    if ($ftype == $fspec['type']) {
+                        $xfer = 0;
+                        if (isset(TBase::$tmethod[$ftype])) {
+                            $func = 'read' . TBase::$tmethod[$ftype];
+                            $xfer += $input->$func($this->$var);
+                        } else {
+                            switch ($ftype) {
+                                case TType::STRUCT:
+                                    $class = $fspec['class'];
+                                    $this->$var = new $class();
+                                    $xfer += $this->$var->read($input);
+                                    break;
+                                case TType::MAP:
+                                    $xfer += $this->readMap($this->$var, $fspec, $input);
+                                    break;
+                                case TType::LST:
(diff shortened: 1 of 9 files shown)


lib/php/lib/Base/TBase.php
```diff
@@ -216,6 +216,8 @@ protected function readStruct(string $class, array $spec, TProtocol $input): int
         $fname = null;
         $ftype = 0;
         $fid = 0;
+        $input->incrementRecursionDepth();
+        try {
         $xfer += $input->readStructBegin($fname);
```
sveneld: Style: the `try { ... }` block body isn't indented relative to the surrounding code, so the body sits at the same column as `try {` itself. PSR-12 expects the body to be one level deeper. The same shape is in `writeStruct` below. CI is green only because phpcs doesn't enforce `Generic.WhiteSpace.ScopeIndent` for this rule, but visually it reads as if the `try` had no body. Suggest:

```suggestion
        $input->incrementRecursionDepth();
        try {
            $xfer += $input->readStructBegin($fname);
```
…and reindent the matched closing brace + every line in between.

lib/php/lib/Base/TBase.php
```diff
@@ -380,6 +385,8 @@ private function writeList(array $var, array $spec, TProtocol $output, bool $set
     protected function writeStruct(string $class, array $spec, TProtocol $output): int
     {
         $xfer = 0;
+        $output->incrementRecursionDepth();
+        try {
         $xfer += $output->writeStructBegin($class);
```
sveneld: Same indent issue as in `readStruct` — body of `try {` is one level shallower than it should be.

lib/php/test/Unit/Lib/Protocol/TProtocolTest.php
```diff
@@ -264,6 +264,40 @@ public function testSkipBinaryThrowsForUnknownType(): void
...
+        $transport = new TMemoryBuffer();
+        $protocol = new TBinaryProtocol($transport);
+        for ($i = 0; $i < TProtocol::DEFAULT_RECURSION_DEPTH; $i++) {
+            $protocol->incrementRecursionDepth();
+        }
+        $this->assertTrue(true); // no exception thrown
+    }
+
+    public function testIncrementRecursionDepthThrowsAtLimit(): void
+    {
```
sveneld: Nit: PHPUnit canonical way to express “the code under test must not throw and has no value to assert on” is `$this->expectNotToPerformAssertions();` at the top of the test, rather than a trailing `assertTrue(true)`. Same for `testDecrementRecursionDepthRestoresCapacity` below.

lib/php/test/Unit/Lib/Protocol/TProtocolTest.php
```diff
@@ -264,6 +264,40 @@ public function testSkipBinaryThrowsForUnknownType(): void
...
+        for ($i = 0; $i < TProtocol::DEFAULT_RECURSION_DEPTH; $i++) {
+            $protocol->incrementRecursionDepth();
+        }
+        $protocol->decrementRecursionDepth();
+        $protocol->incrementRecursionDepth(); // should not throw
+        $this->assertTrue(true);
+    }
+
     private function buildBinaryBuffer(callable $writer): string
     {
```
sveneld, line 268: Same nit — prefer `$this->expectNotToPerformAssertions();` over `assertTrue(true)`.

sveneld (commented):
Functionally solid — see two style nits inline. Will follow up on the binary_inline / integration-test items separately.

lib/php/lib/Base/TBase.php
```diff
@@ -380,36 +385,41 @@ private function writeList(array $var, array $spec, TProtocol $output, bool $set
     protected function writeStruct(string $class, array $spec, TProtocol $output): int
```
sveneld, line 385: Same logic is in https://github.com/apache/thrift/blob/master/lib/php/lib/Exception/TException.php it should be also updated

PR #3555, opened 2026-05-28
THRIFT-6050: Limit struct read/write recursion depth in Swift library

Jens-G:
## Summary

- Adds recursion depth counter to Swift `TProtocol` base
- Limit is 64; throws `TProtocolError` with `.depthLimit` when exceeded
- Swift generator wraps struct read/write with `defer` to guarantee decrement

## Test plan

- [ ] Unit tests in `lib/swift/Tests/` verify depth allow/reject/restore

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_swift_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_swift_generator.cc
@@ -1044,6 +1044,8 @@ void t_swift_generator::generate_swift_union_reader(ostream& out, t_struct* tstr
   indent(out) << "public static func read(from proto: TProtocol) throws -> "
               << tstruct->get_name();
   block_open(out);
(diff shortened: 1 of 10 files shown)


Jens-G:
## Problems: 

### The generated recursive types are mostly unusable in Swift 
- pre-existing, not part of THRIFT-6050
- CoUnion/CoUnion2 don't even compile — recursive Swift enums must be marked indirect; the generator doesn't. (Confirmed with the compiler.)
- CoRec/CoError/RecList have non-optional recursive fields (var other: CoError2, required init param) → they can never be constructed (no base case to
bootstrap).
- Only RecTree (list-of-self, list can be empty) is actually constructible.

### No CI gate: broken and disabled CI 
- THRIFT-5864 — "broken, no maintainers"
- Package is swift-tools:5.1 (no build plugins), 

## Decision

Bad day for Swift. Close the PR and fugetabodit.

PR #3565, opened 2026-05-29
Fix Ruby lib CI: drop pry/byebug, incompatible with Ruby 4.1+

Jens-G:
## Summary

- `byebug 11.1.3` fails to compile its C extension under Ruby 4.1 (`ruby-head`): it uses the `Data_Get_Struct` macro that was removed in Ruby 4.x
- `pry`, `pry-byebug`, and `pry-stack_explorer` are debug/REPL tools with no role in CI test execution — remove them as development dependencies
- Relax the `bundler` constraint from `~> 2.2.34` to `>= 2.2.34` so Bundler 4 (shipped with Ruby 4.x) is accepted

Fixes: https://github.com/apache/thrift/actions/runs/26610293673/job/78455020323

## Test plan

- [ ] `lib-ruby (head)` CI job passes after this change
- [ ] `lib-ruby` jobs for other Ruby versions remain green

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -907,6 +907,7 @@ jobs:
   lib-ruby:
     needs: compiler
     runs-on: ubuntu-24.04
+    continue-on-error: ${{ matrix.ruby-version == 'head' }}
     name: lib-ruby (${{ matrix.ruby-version }}) ${{ matrix.skip-build-ext && 'noext' || '' }}
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


Jens-G:
@kpumuk because https://github.com/apache/thrift/actions/runs/26610293673/job/78455020323

kpumuk:
I added `ruby-head` to get a heads up on upcoming Ruby releases, but since it is being actively worked on, it keeps breaking CI - twice over the past month. Marking it as allow to fail, so that I can look at it without blocking anyone.

PR #3585, opened 2026-06-05
limit recursion depth in c_glib thrift_protocol_skip

dxbjavid:
thrift_protocol_skip recurses through nested structs, lists, sets and maps with no depth guard, so a server skipping an unknown field (the dispatch processor calls it on every unrecognised method) can be driven into unbounded recursion by a deeply nested message and overflow the stack. The configuration already carries recursionLimit_ with a default of 64 but nothing reads it. I noticed this reading the c_glib skip next to the C++ one, which guards the same path with TInputRecursionTracker. This routes skip through an internal helper that tracks depth and stops at the configured limit.

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c
+++ b/lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c
@@ -433,9 +433,26 @@ thrift_protocol_get_min_serialized_size (ThriftProtocol *protocol, ThriftType ty
     (_RES) += _x; \
   }
 
-gint32
-thrift_protocol_skip (ThriftProtocol *protocol, ThriftType type, GError **error)
+static gint32
+thrift_protocol_skip_impl (ThriftProtocol *protocol, ThriftType type,
+                           gint32 recursion_depth, GError **error)
 {
+  gint32 recursion_limit = DEFAULT_RECURSION_DEPTH;
+
+  if (protocol->transport != NULL &&
+      protocol->transport->configuration != NULL)
+  {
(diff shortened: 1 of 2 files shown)


Jens-G:
### Code review

The code change itself looks correct: the guard is threaded through `thrift_protocol_skip_impl`, both `transport` and `configuration` are null-checked before reading `recursionLimit_`, the `-1` return propagates via `THRIFT_SKIP_RESULT_OR_RETURN`, all five recursive sites pass `recursion_depth + 1`, and the public `thrift_protocol_skip` signature is preserved. The gap is real and still present on `master` — `recursionLimit_` has existed since THRIFT-5237 but was never consulted in `skip`.

One item to address before merge: no test accompanies the change. The analogous per-language limit changes each added one. A test under `lib/c_glib/test/` that nests past the configured limit and asserts `THRIFT_PROTOCOL_ERROR_DEPTH_LIMIT` would cover it.

https://github.com/apache/thrift/blob/6934f8e3ad8b5de855efffd110673c75573855f5/lib/c_glib/src/thrift/c_glib/protocol/thrift_protocol.c#L436-L454

dxbjavid:
good point, added a test in testbinaryprotocol.c. it stands up a memory buffer with a small recursion_limit on its configuration, writes nested struct field headers a fair way past that limit, and asserts skip returns -1 with THRIFT_PROTOCOL_ERROR_DEPTH_LIMIT. there's also a shallow case nested within the limit that still skips cleanly, so the guard doesn't trip on normal input. used the binary protocol over the buffer so it stays self-contained, no socket needed.

dxbjavid:
Thanks for the merge! Appreciated

PR #3590, opened 2026-06-09
avoid integer overflow in container size precheck

dxbjavid:
the container size precheck multiplies a wire-controlled element count by the per-element minimum size in 32-bit int, so a crafted list, set or map header can wrap the product down to a small or zero value and slip past the maxMessageSize guard, letting an oversized container through before any allocation is bounded. the same shape sits in the binary, compact and json protocols and the protocol base, plus the c_glib binary and compact readers. widen the multiplication to the long type checkReadBytesAvailable already takes so the count can no longer overflow, and add a regression test for the three c++ protocols.

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/protocol/thrift_binary_protocol.c
+++ b/lib/c_glib/src/thrift/c_glib/protocol/thrift_binary_protocol.c
@@ -577,9 +577,9 @@ thrift_binary_protocol_read_map_begin (ThriftProtocol *protocol,
     return -1;
   }
 
-  if(!ttc->checkReadBytesAvailable (THRIFT_TRANSPORT(tp->transport), 
-                                    sizei * thrift_binary_protocol_get_min_serialized_size(protocol, k, error) + 
-                                    sizei * thrift_binary_protocol_get_min_serialized_size(protocol, v, error),
+  if(!ttc->checkReadBytesAvailable (THRIFT_TRANSPORT(tp->transport),
+                                    (gint64) sizei * thrift_binary_protocol_get_min_serialized_size(protocol, k, error) +
+                                    (gint64) sizei * thrift_binary_protocol_get_min_serialized_size(protocol, v, error),
                                     error))
   {
     return -1;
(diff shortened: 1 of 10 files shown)


Jens-G:
### Code review

Found 2 issues:

1. `static_cast<long>` and `(glong)` are 32-bit on Windows/MSVC (LLP64 data model), the same width as `int`, so the multiplication still overflows there. The fix works on POSIX (LP64) where `long` is 64-bit, but not on Windows. Use `int64_t`/`gint64` instead (and update `TTransport::checkReadBytesAvailable`'s signature accordingly).

https://github.com/apache/thrift/blob/068af367eea9bc63878eae19f2a4d2ef76daefcb/lib/cpp/src/thrift/protocol/TBinaryProtocol.h#L175-L192

The same pattern appears in `TCompactProtocol.h`, `TJSONProtocol.h`, `TProtocol.h`, `thrift_binary_protocol.c`, and `thrift_compact_protocol.c`.

2. PR title and commit message are missing the required `THRIFT-NNNN:` prefix and `Client: cpp,c_glib` trailer required by [AGENTS.md §2](https://github.com/apache/thrift/blob/445c062e3ac745a2841962002a8ec1e8f41320e8/AGENTS.md#L24-L40) for non-trivial changes. Please file a JIRA ticket at https://issues.apache.org/jira/browse/THRIFT and update the PR title and commit message.

dxbjavid:
good catch on the width, you're right that long and glong stay 32-bit under LLP64 so the product would still wrap on MSVC. i've switched the casts to int64_t in the c++ protocols and gint64 in the c_glib binary and compact readers, and widened the checkReadBytesAvailable parameter to match in both TTransport and the c_glib transport vtable, since otherwise the wider cast just gets truncated back through the glong parameter when the value is passed in. rechecked the binary list case and an oversized header is still rejected against a 1 KiB maxMessageSize after the change.

on the second point, i've added the Client: cpp,c_glib trailer to the commit. i'll file a JIRA ticket for this and update the PR title and commit with the THRIFT- prefix once i have the number.

Jens-G:
### Code review

Found 1 issue:

1. Both the PR body and the commit message describe the exploitation mechanism ("a crafted list, set or map header can wrap the product down to a small or zero value and slip past the maxMessageSize guard"). [AGENTS.md §6](https://github.com/apache/thrift/blob/35c1a53dd6de205a4a7c23fc356f1fba490116c4/AGENTS.md#L87-L96) says to use neutral functional language for serialization bounds changes and to never describe what an attacker can achieve in public-facing text. A neutral rewrite would be: "widen multiplication operands in container size prechecks to prevent arithmetic overflow on LLP64 platforms."

https://github.com/apache/thrift/commit/41cb98af4d872f753b58c9663786156fa8636600

Jens-G:
Merged, thanks!

dxbjavid:
Thanks for the review and merge! Appreciated

PR #3592, opened 2026-06-09
THRIFT-6043: Limit struct read/write recursion depth in Python library

Jens-G:
## Summary

Adds a configurable recursion depth limit (default 64) to struct and exception read/write in the Python library.

**What is changed:**

- `compiler/cpp/src/thrift/generate/t_py_generator.cc`: generated `read()`/`write()` methods now wrap the body in `increment_recursion_depth()` / `try` / `finally: decrement_recursion_depth()`.
- `lib/py/src/protocol/TProtocol.py`: `TProtocolBase` gains `_recursion_depth`, `DEFAULT_RECURSION_DEPTH = 64`, `increment_recursion_depth()` (raises `TProtocolException.DEPTH_LIMIT` when exceeded), and `decrement_recursion_depth()`.
- `lib/py/src/ext/protocol.h` / `protocol.tcc`: The fastbinary C extension (`TBinaryProtocolAccelerated`, `TCompactProtocolAccelerated`) is also guarded. `ProtocolBase` gains a `recursionDepth_` counter; `readStruct()` and `encodeValue(T_STRUCT)` use a RAII `RecursionGuard` that raises `TProtocolException(DEPTH_LIMIT, …)` via the Python C API when the limit is exceeded.
- `lib/py/test/test_recursion_depth.py` + `lib/py/Makefile.am`: Generated-code round-trip tests over `test/Recursive.thrift` (`RecTree` struct, `CoError` exception) for `TBinaryProtocol`, `TBinaryProtocolAccelerated`, `TCompactProtocol`, `TCompactProtocolAccelerated`, and `TJSONProtocol`.

**Note:** The guard is placed after the fast-path early return in generated `read()`, so `TBinaryProtocol` (non-accelerated) and `TJSONProtocol` are protected via the Python path; `TBinaryProtocolAccelerated` and `TCompactProtocolAccelerated` are protected via the C extension guard. The limit (64) is hardcoded in the C extension; making it read from `TConfiguration` can be a follow-up.

**`skip()`** remains independently bounded via `TProtocolBase.skip(max_depth=64)` (unchanged).

## Test plan

- [ ] `make -C lib/py check` passes (Python 3.10–3.14, with and without `skip-build-ext`)
- [ ] 12/12 round-trip tests in `test_recursion_depth.py` pass locally (Docker `thrift:jammy`)
- [ ] Baseline proof: stripping the generator guard from the generated `RecTree.read()`/`write()` causes 8 over-limit assertions to fail

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -1119,6 +1119,9 @@ void t_py_generator::generate_py_struct_reader(ostream& out, t_struct* tstruct) 
   }
   indent_down();
 
+  indent(out) << "iprot.increment_recursion_depth()" << '\n';
+  indent(out) << "try:" << '\n';
+  indent_up();
   indent(out) << "iprot.readStructBegin()" << '\n';
 
   if (is_immutable(tstruct)) {
@@ -1194,6 +1197,12 @@ void t_py_generator::generate_py_struct_reader(ostream& out, t_struct* tstruct) 
   }
 
   indent_down();
+  indent(out) << "finally:" << '\n';
+  indent_up();
+  indent(out) << "iprot.decrement_recursion_depth()" << '\n';
+  indent_down();
+
+  indent_down();
   out << '\n';
 }
 
@@ -1214,6 +1223,9 @@ void t_py_generator::generate_py_struct_writer(ostream& out, t_struct* tstruct) 
   indent(out) << "return" << '\n';
   indent_down();
 
+  indent(out) << "oprot.increment_recursion_depth()" << '\n';
+  indent(out) << "try:" << '\n';
+  indent_up();
   indent(out) << "oprot.writeStructBegin('" << name << "')" << '\n';
 
   for (f_iter = fields.begin(); f_iter != fields.end(); ++f_iter) {
@@ -1236,6 +1248,12 @@ void t_py_generator::generate_py_struct_writer(ostream& out, t_struct* tstruct) 
   // Write the struct map
   out << indent() << "oprot.writeFieldStop()" << '\n' << indent() << "oprot.writeStructEnd()"
       << '\n';
+
+  indent_down();
+  indent(out) << "finally:" << '\n';
+  indent_up();
+  indent(out) << "oprot.decrement_recursion_depth()" << '\n';
+  indent_down();
 
   out << '\n';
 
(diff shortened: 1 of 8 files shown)


Jens-G:
### Code review

Found 5 issues:

1. `increment_recursion_depth()` is placed before the `try:` block in the generated code. When the depth limit is exceeded, it raises `TProtocolException` before control enters `try:`, so the `finally: decrement_recursion_depth()` never runs. On a reused protocol object, each caught limit violation permanently increments `_recursion_depth` by 1, shrinking the effective limit. The C extension handles this correctly by decrementing before raising inside `checkDepthLimit()`.

https://github.com/apache/thrift/blob/3cc3faa19bbe18f2bd654db74c427d3657b466b8/compiler/cpp/src/thrift/generate/t_py_generator.cc#L1121-L1125

Fix: move the increment inside the `try:` block, or mirror the C extension by decrementing before raising in `increment_recursion_depth()` itself.

2. `setUp()` sets `self.has_fastbinary` but no test method ever checks it. When fastbinary is not built, `TBinaryProtocolAccelerated` silently falls back to pure Python, so the three accelerated tests pass while exercising only the Python path rather than the C extension.

https://github.com/apache/thrift/blob/3cc3faa19bbe18f2bd654db74c427d3657b466b8/lib/py/test/test_recursion_depth.py#L114-L138

3. There are three independent hardcoded `64` constants for the recursion depth limit: `DEFAULT_RECURSION_DEPTH` in TProtocol.py, `kDefaultRecursionDepth` in protocol.h, and `LIMIT` in the test file. These are not linked — changing one leaves the others stale.

https://github.com/apache/thrift/blob/3cc3faa19bbe18f2bd654db74c427d3657b466b8/lib/py/src/protocol/TProtocol.py#L45-L47
https://github.com/apache/thrift/blob/3cc3faa19bbe18f2bd654db74c427d3657b466b8/lib/py/src/ext/protocol.h#L57-L59

4. `RecursionDepthCompactTest` and `RecursionDepthJSONTest` have no `test_read_over_limit` method. The write path is tested but not the read path for these two protocols.

https://github.com/apache/thrift/blob/3cc3faa19bbe18f2bd654db74c427d3657b466b8/lib/py/test/test_recursion_depth.py#L141-L174

5. `TApplicationException.read()` and `.write()` in Thrift.py call `readStructBegin`/`writeStructBegin` directly without `increment_recursion_depth()`/`decrement_recursion_depth()` guards, unlike all generated structs. `TApplicationException` has only flat primitive fields so there is no actual recursion risk, but the inconsistency means depth tracking is not accurate across a full message exchange.

https://github.com/apache/thrift/blob/3cc3faa19bbe18f2bd654db74c427d3657b466b8/lib/py/src/Thrift.py#L138-L171

PR #3597, opened 2026-06-13
fix off-by-ten header bounds check in readHeaderFormat

dxbjavid:
readHeaderFormat checks the declared header size against the whole frame size, but the parse pointer has already moved past the 10-byte common header before that check, so a header size landing in the last 10 bytes of the frame leaves the header boundary pointing up to 10 bytes beyond the receive buffer. The varint and string readers only stop at that boundary, so a malformed THeader frame makes them read past the end of the buffer. Comparing the header section against the remaining sz - 10 bytes keeps the boundary inside the allocation, and the added test drives the read path with such a frame.


dxbjavid:
Glad to contribute. Thanks

PR #3598, opened 2026-06-15
THRIFT-6070: Publish Python wheel distributions to PyPI

HTHou:
## What changed

This updates the PyPI release workflow to build and publish Python wheel distributions in addition to the existing source distribution.

The workflow now:
- builds the sdist with `python -m build` and validates it with `twine check`
- builds CPython 3.10-3.14 wheels with `cibuildwheel`
- covers manylinux x86_64/aarch64, musllinux x86_64/aarch64, macOS x86_64/arm64, and Windows amd64
- pins manylinux wheels to the manylinux2014 baseline for glibc 2.17+ compatibility
- keeps PyPI publishing in a separate release-only job with OIDC Trusted Publishing permissions

The release management notes now describe the automated PyPI publishing path and the expected artifacts.

## Why

PyPI releases for the Python `thrift` package currently publish only an sdist, so users without a matching prebuilt wheel must build locally. This adds wheel artifacts while keeping the sdist path intact.

## Validation

- Parsed `.github/workflows/pypi.yml` as YAML
- Ran `git diff --check`
- Built `lib/py` sdist in a temporary directory with `python -m build --sdist`
- Ran `twine check` on the generated sdist
- Verified `cibuildwheel==4.1.0 --print-build-identifiers` selects the expected CPython 3.10-3.14 targets for Linux, macOS, and Windows

Full multi-platform wheel builds are expected to run in GitHub Actions.

## Licensing

`cibuildwheel` is used as a CI build-time tool and is BSD-2-Clause licensed, which is compatible with Apache License 2.0. No new runtime dependency is added.

## AI assistance

Co-Authored-By: Codex <noreply@openai.com>

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -22,16 +22,26 @@ name: "PyPI publishing"
 on:
   release:
     types: [published]
+  pull_request:
+    branches:
+      - master
+    paths:
+      - ".github/workflows/pypi.yml"
+      - "doc/ReleaseManagement.md"
+      - "lib/py/MANIFEST.in"
+      - "lib/py/pyproject.toml"
+      - "lib/py/setup.cfg"
+      - "lib/py/setup.py"
+      - "lib/py/compat/win32/**"
+      - "lib/py/src/ext/**"
+  workflow_dispatch:
 
 jobs:
-  pypi-publish:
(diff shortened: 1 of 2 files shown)


Jens-G:
### Code review

PR #3604, opened 2026-06-18
widen container size precheck to 64-bit in go protocols

dxbjavid:
The Go binary, compact and JSON protocol readers computed the container size precheck (`size * minElemSize`) in 32-bit arithmetic. This routes the three readers through a small helper that performs the multiplication in 64-bit and range-checks the element count, so the value compared against `maxMessageSize` always stays within int32 range — matching the 64-bit precheck widening already done for the C++ and C/GLib bindings in #3590. The JSON readers read the element count as `int64`, so the helper also bounds the count before the multiply. Includes regression tests driving each reader (binary, compact, JSON) with counts whose 64-bit product is outside int32 range.

### DIFF
--- a/lib/go/thrift/binary_protocol.go
+++ b/lib/go/thrift/binary_protocol.go
@@ -356,8 +356,7 @@ func (p *TBinaryProtocol) ReadMapBegin(ctx context.Context) (kType, vType TType,
 		return
 	}
 	minElemSize := p.getMinSerializedSize(kType) + p.getMinSerializedSize(vType)
-	totalMinSize := size32 * minElemSize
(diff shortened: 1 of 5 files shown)


Jens-G:
### Code review

Reviewed and applied as a maintainer edit. The binary and compact prechecks are widened correctly. On the JSON path the element count is a wire-supplied `int64`, so I added a range check on the count before the 64-bit multiply in `checkContainerSizeForProtocol`, with regression tests (a helper unit case and a `TJSONProtocol` map case). Also tidied the wording in the description and a test comment to keep it functional.

Thanks @dxbjavid — kept you as the author of the original commit.

dxbjavid:
Thanks for the merge.

PR #3605, opened 2026-06-23
enforce stringLengthLimit in TCompactProtocol.readBinary

dxbjavid:
TCompactProtocol.readBinary only checks the binary length against the remaining message budget, so the per-string limit (stringLengthLimit) that readString applies through checkStringReadLength is never enforced for binary fields. a peer that has configured a string length limit therefore has it quietly bypassed for any binary value, which can then be sized up to the whole message. routing readBinary through checkStringReadLength applies the same limit and keeps the existing remaining-bytes check, matching how TBinaryProtocol.readBinary already behaves. added a regression test alongside the existing compact string-limit test.


Jens-G:
### Code review

PR #3606, opened 2026-06-26
THRIFT-6073: Allow injecting external SSL_CTX into C++ SSLContext

hongzhi-gao:
## Summary

Add a backend-neutral extension point to the C++ `TSSLSocket` stack: applications can inject a pre-configured OpenSSL `SSL_CTX` through the existing `SSLContextFactory` / `TSSLSocketFactory` hook.

This makes it easier for applications to supply a fully configured TLS context—protocol options, cipher suites, certificate loading, and other OpenSSL settings—in application code, without patching libthrift or replacing the transport layer.

## Background

Thrift C++ already provides SSL/TLS through `TSSLSocket` / `TSSLSocketFactory`, backed by OpenSSL. The default factory covers the common case: a single certificate/key pair via `loadCertificate()` / `loadPrivateKey()`, and protocol selection via `SSLProtocol`.

`TSSLSocketFactory` already accepts a custom `SSLContextFactory`, but `SSLContext` could previously only be created from `SSLProtocol`. Applications that need a context configured outside Thrift's helpers—multi-step cert setup, non-default methods, or options not exposed by `TSSLSocketFactory`—had to patch Thrift locally.

This PR adds a small wrapper constructor on `SSLContext` while keeping the default OpenSSL build unchanged.

## What changed

- **`SSLContext`**: `explicit SSLContext(SSL_CTX* ctx)` — wrap an application-configured `SSL_CTX`; ownership is transferred and the context is freed on destruction. Doxygen documents caller responsibilities for TLS protocol configuration (the default `SSLTLS` minimum-version options are not applied automatically).
- **`lib/cpp/README.md`**: document the injection pattern.
- **`SecurityTest`**: wrapped-context and null-validation coverage.

Threat-model cross-check against `doc/thrift-threat-model.md` is recorded in JIRA THRIFT-6073 (per AGENTS.md §6).

## Usage

**Standard TLS (unchanged):**

```cpp
TSSLSocketFactory factory;  // default SSLTLS
factory.loadCertificate("server.crt");
factory.loadPrivateKey("server.key");
```

**Injected `SSL_CTX` (application configures OpenSSL, then transfers ownership):**

```cpp
#include <thrift/transport/TSSLSocket.h>

auto factory = std::make_shared<TSSLSocketFactory>([]() {
  SSL_CTX* ctx = SSL_CTX_new(TLS_method());
  SSL_CTX_set_mode(ctx, SSL_MODE_AUTO_RETRY);
  // Configure ctx before wrapping: protocol floor, ciphers, certificates, etc.
  return std::make_shared<apache::thrift::transport::SSLContext>(ctx);
});
factory->loadTrustedCertificates("ca-bundle.crt");
factory->authenticate(true);
factory->server(true);
auto socket = factory->createSocket("127.0.0.1", port);
```

Link your application against whichever OpenSSL-compatible TLS library you choose; libthrift continues to default to system OpenSSL at build time.

## Licensing

Thrift introduces **no new build-time dependency**. Nothing is vendored; no `LICENSE` / `NOTICE` update is required.

## Test plan

- [x] Standard OpenSSL build: `ctest -R SecurityTest` (`wrapped_ssl_context`, `wrapped_ssl_context_null`)

### DIFF
--- a/lib/cpp/src/thrift/transport/TSSLSocket.h
+++ b/lib/cpp/src/thrift/transport/TSSLSocket.h
@@ -365,6 +365,21 @@ public:
 class SSLContext {
 public:
   SSLContext(const SSLProtocol& protocol = SSLTLS);
+  /**
+   * Wrap an existing OpenSSL SSL_CTX.
+   *
+   * Takes ownership of @a ctx; the caller must not call SSL_CTX_free on it.
+   * The context must remain valid for the lifetime of this SSLContext and any
+   * TSSLSocket instances created from it (see TSSLSocketFactory lifetime notes).
+   *
+   * Unlike SSLContext(SSLProtocol) with the default SSLTLS protocol, this
+   * constructor does not apply a minimum TLS version (no SSL_OP_NO_TLSv1*).
+   * Configure protocol options on @a ctx before wrapping if a version floor
+   * is required.
+   *
+   * @param ctx  OpenSSL context to wrap (ownership transferred)
+   */
+  explicit SSLContext(SSL_CTX* ctx);
   virtual ~SSLContext();
   SSL* createSSL();
   SSL_CTX* get() { return ctx_; }
--- a/lib/cpp/README.md
+++ b/lib/cpp/README.md
@@ -138,7 +138,11 @@ TSSLSocket are always created from TSSLSocketFactory.
 The default TSSLSocketFactory context uses OpenSSL's version-flexible TLS
 method and sets TLS 1.2 as the minimum negotiated protocol version. Applications
 that need a different protocol range can provide a custom SSLContext factory and
-adjust the OpenSSL context options before creating sockets.
+adjust the OpenSSL context options before creating sockets. Applications that
+link against an OpenSSL-compatible TLS library can also create and configure an
+SSL_CTX externally, wrap it with `SSLContext`, and pass it through the factory
+(for example, protocol-specific or dual-certificate setups that the default
+factory methods cannot express).
 
 ## How to use SSL APIs
 
--- a/lib/cpp/src/thrift/transport/TSSLSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLSocket.cpp
@@ -213,6 +213,12 @@ SSLContext::SSLContext(const SSLProtocol& protocol) {
   }
 }
 
+SSLContext::SSLContext(SSL_CTX* ctx) : ctx_(ctx) {
+  if (ctx_ == nullptr) {
+    throw TSSLException("SSLContext: ctx must not be null");
+  }
+}
+
 SSLContext::~SSLContext() {
   if (ctx_ != nullptr) {
     SSL_CTX_free(ctx_);
--- a/lib/cpp/test/SecurityTest.cpp
+++ b/lib/cpp/test/SecurityTest.cpp
@@ -34,6 +34,7 @@
 #endif
 
 using apache::thrift::transport::TSSLServerSocket;
+using apache::thrift::transport::SSLContext;
 using apache::thrift::transport::SSLContextFactory;
 using apache::thrift::transport::TSSLException;
 using apache::thrift::transport::TServerTransport;
@@ -272,6 +273,34 @@ BOOST_AUTO_TEST_CASE(custom_ssl_context_options)
         BOOST_CHECK((options & SSL_OP_NO_TLSv1_1) == 0);
     }
     context.reset();
+}
+
+BOOST_AUTO_TEST_CASE(wrapped_ssl_context)
+{
+    SSL_CTX* raw = SSL_CTX_new(TLS_method());
+    BOOST_REQUIRE(raw != nullptr);
+    SSL_CTX_set_mode(raw, SSL_MODE_AUTO_RETRY);
+
+    std::shared_ptr<SSLContext> context;
+    TSSLSocketFactory factory([&context, raw]() {
+        context = std::make_shared<SSLContext>(raw);
+        return context;
+    });
+    BOOST_CHECK(context->get() == raw);
+    context.reset();
+}
+
+BOOST_AUTO_TEST_CASE(wrapped_ssl_context_null)
+{
+    try
+    {
+        std::make_shared<SSLContext>(nullptr);
+        BOOST_FAIL("Expected null SSL_CTX to throw");
+    }
+    catch (const TSSLException& ex)
+    {
+        BOOST_CHECK_EQUAL("SSLContext: ctx must not be null", std::string(ex.what()));
+    }
 }
 
 BOOST_AUTO_TEST_CASE(custom_ssl_context_factory_validation)


Jens-G:
### Code review

Found 2 issues:

1. All 8 new `load*` methods call `requireNtlsSupport()` before the null-pointer argument check, reversing the established contract of every existing `load*` method in the file (`loadCertificate`, `loadPrivateKey`, etc. all validate arguments first). On non-NTLS builds, callers that pass null arguments receive `TSSLException("NTLS is not available")` instead of `TTransportException(BAD_ARGS)`, making errors undiagnosable.

https://github.com/apache/thrift/blob/0d5e69aae60c5dbf3c6c746b42705642c8362711/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L1110-L1115

2. Tongsuo is introduced as a new optional build dependency but the PR description contains no mention of license verification against the ASF Category A/X list, as required by [AGENTS.md §1](https://github.com/apache/thrift/blob/deb36fa409849de45973b04ffc3ce49d277ca90a/AGENTS.md#L13-L19): *"Before introducing any dependency … verify its license is compatible with Apache 2.0 … If in doubt … add it and flag it in the PR description for committer review."*

lib/cpp/src/thrift/transport/TSSLSocket.h
```diff
@@ -43,6 +43,7 @@ enum SSLProtocol {
   TLSv1_0 = 3,  // Supports TLSv1_0 or later.
   TLSv1_1 = 4,  // Supports TLSv1_1 or later.
   TLSv1_2 = 5,  // Supports TLSv1_2 or later.
+  NTLS    = 6,  // Tongsuo NTLS/TLCP; requires dual signing and encryption certificates.
```
HTHou: Should 6 be TLSv1_3?

HTHou:
Thanks for putting this together. Could we consider narrowing the upstream change to a generic TLS-provider compatibility layer, rather than making Tongsuo/NTLS a first-class Thrift backend?

The current patch adds Tongsuo-specific build plumbing, `SSLProtocol::NTLS`, and `loadSign*` / `loadEnc*` APIs to `TSSLSocketFactory`. That makes the Tongsuo/TLCP surface part of Thrift's public C++ API/ABI and ongoing build/test matrix.

An alternative might be to expose a small backend-neutral extension point around `SSLContext` / `SSL_CTX` creation and configuration, for example allowing applications to provide an already-created `SSLContext`/`SSL_CTX`, or a protected/public hook that creates and configures the context. Then downstream applications that need Tongsuo can link against it themselves and apply the NTLS/TLCP dual-certificate setup in their own factory/subclass, while Thrift only maintains the generic OpenSSL-compatible transport layer.

That would still let users avoid carrying a private transport fork, but keeps the provider-specific dependency and policy choices outside the core Thrift tree. Would you be open to reshaping the patch in that direction?

hongzhi-gao:
> Thanks for putting this together. Could we consider narrowing the upstream change to a generic TLS-provider compatibility layer, rather than making Tongsuo/NTLS a first-class Thrift backend?
> 
> The current patch adds Tongsuo-specific build plumbing, `SSLProtocol::NTLS`, and `loadSign*` / `loadEnc*` APIs to `TSSLSocketFactory`. That makes the Tongsuo/TLCP surface part of Thrift's public C++ API/ABI and ongoing build/test matrix.
> 
> An alternative might be to expose a small backend-neutral extension point around `SSLContext` / `SSL_CTX` creation and configuration, for example allowing applications to provide an already-created `SSLContext`/`SSL_CTX`, or a protected/public hook that creates and configures the context. Then downstream applications that need Tongsuo can link against it themselves and apply the NTLS/TLCP dual-certificate setup in their own factory/subclass, while Thrift only maintains the generic OpenSSL-compatible transport layer.
> 
> That would still let users avoid carrying a private transport fork, but keeps the provider-specific dependency and policy choices outside the core Thrift tree. Would you be open to reshaping the patch in that direction?

1. Fixed by removing the Tongsuo-specific loadSign* / loadEnc* APIs entirely. The reshaped patch no longer adds those methods.
2. Added a Licensing section: Thrift still depends only on OpenSSL by default; Tongsuo is an application-side choice (Apache 2.0, ASF Category A) and is not vendored or added to Thrift's build.

lib/cpp/src/thrift/transport/TSSLSocket.h
hongzhi-gao: NTLS = 6 was removed with the provider-specific APIs.

hongzhi-gao:
> Thanks for putting this together. Could we consider narrowing the upstream change to a generic TLS-provider compatibility layer, rather than making Tongsuo/NTLS a first-class Thrift backend?
> 
> The current patch adds Tongsuo-specific build plumbing, `SSLProtocol::NTLS`, and `loadSign*` / `loadEnc*` APIs to `TSSLSocketFactory`. That makes the Tongsuo/TLCP surface part of Thrift's public C++ API/ABI and ongoing build/test matrix.
> 
> An alternative might be to expose a small backend-neutral extension point around `SSLContext` / `SSL_CTX` creation and configuration, for example allowing applications to provide an already-created `SSLContext`/`SSL_CTX`, or a protected/public hook that creates and configures the context. Then downstream applications that need Tongsuo can link against it themselves and apply the NTLS/TLCP dual-certificate setup in their own factory/subclass, while Thrift only maintains the generic OpenSSL-compatible transport layer.
> 
> That would still let users avoid carrying a private transport fork, but keeps the provider-specific dependency and policy choices outside the core Thrift tree. Would you be open to reshaping the patch in that direction?

Done — the patch is reshaped to a backend-neutral SSLContext wrapper around an application-provided SSL_CTX, using the existing SSLContextFactory hook. Tongsuo/NTLS dual-cert setup lives in application (or local test) code; Thrift core stays OpenSSL-compatible and provider-agnostic.

HTHou (approved):

Jens-G:
### Code review

Found 6 issues (adversarially verified):

1. `SSLContext(SSL_CTX*, false)` stores the raw pointer without calling `SSL_CTX_up_ref`. If the external owner frees the context while this `SSLContext` is alive, the subsequent `SSL_new(ctx_)` call in `createSSL()` uses freed memory.

https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L216-L221

Verified: no `SSL_CTX_up_ref` call present in PR code; constructor absent in base.

2. Adding `bool takeOwnership_` to `SSLContext` changes `sizeof(SSLContext)` from 8 to 16 bytes. `SSLContext` is a concrete public class in an installed header; no soname bump or ABI versioning mechanism exists in the project.

https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L365-L374

Verified: member absent in base, no SOVERSION in CMakeLists.txt, no `.map`/`.sym` files under `lib/cpp/`.

3. Both new tests exercise `takeOwnership=true` only (`wrapped_ssl_context` passes `true` explicitly; `wrapped_ssl_context_null` throws before the ownership branch). The destructor branch where `SSL_CTX_free` is skipped (`takeOwnership_=false`) has no test.

https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/test/SecurityTest.cpp#L286-L306

Verified: `grep -n "takeOwnership\|, false" SecurityTest.cpp` returns no output on PR code; branch absent in base.

4. The new constructor's Doxygen does not warn that callers are responsible for enforcing a minimum TLS version. `SSLContext(SSLProtocol)` applies `SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1` unconditionally; the new constructor wraps the context as-is and the test creates it with `TLS_method()` and no version restrictions.

https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L367-L374

Verified: no `SSL_OP_NO_TLSv1*` in new constructor; existing floor confirmed at [`TSSLSocket.cpp:210-212`](https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L210-L212); constructor absent in base.

5. When `takeOwnership=false`, the caller must keep the `SSL_CTX*` alive at least as long as the `SSLContext` and all sockets created from it. This is not documented in the new constructor's Doxygen. The factory class has an explicit parallel warning ([`TSSLSocket.h:200-206`](https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L200-L206): *"It is the responsibility of the code using TSSLSocketFactory to ensure that the factory lifetime exceeds the lifetime of any sockets it might create"*).

https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L367-L374

Verified: `@param takeOwnership` documents only the `true` case; factory warning confirmed in base and PR; constructor absent in base.

6. `AGENTS.md` §6 and its Quick Reference checklist require TLS configuration changes to be cross-checked against `doc/thrift-threat-model.md` before merging. The PR body contains no mention of this. `doc/thrift-threat-model.md` is present in the repo. (Note: per §6, any details about trust-boundary implications belong in the JIRA ticket THRIFT-6073, not in the PR description or commit messages.)

Verified: AGENTS.md requirement confirmed; PR body grep for "threat" returns no output; `doc/thrift-threat-model.md` exists at PR head.

Jens-G:
One housekeeping note: for a change of this scope, it is worth making sure the JIRA ticket captures the key design decisions and rationale in addition to what is in the commit messages. Commit messages are squashed on merge and can be hard to find later; JIRA comments and the ticket description remain a stable reference for anyone reviewing the issue in future. If there is technical context currently spread across individual commit messages (e.g. the rationale for the `takeOwnership` parameter semantics, or the TLS version floor decision), copying the relevant parts into the JIRA ticket before this is merged would be a good investment.

hongzhi-gao:
> ### Code review
> Found 6 issues (adversarially verified):
> 
> 1. `SSLContext(SSL_CTX*, false)` stores the raw pointer without calling `SSL_CTX_up_ref`. If the external owner frees the context while this `SSLContext` is alive, the subsequent `SSL_new(ctx_)` call in `createSSL()` uses freed memory.
> 
> https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L216-L221
> 
> Verified: no `SSL_CTX_up_ref` call present in PR code; constructor absent in base.
> 
> 2. Adding `bool takeOwnership_` to `SSLContext` changes `sizeof(SSLContext)` from 8 to 16 bytes. `SSLContext` is a concrete public class in an installed header; no soname bump or ABI versioning mechanism exists in the project.
> 
> https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L365-L374
> 
> Verified: member absent in base, no SOVERSION in CMakeLists.txt, no `.map`/`.sym` files under `lib/cpp/`.
> 
> 3. Both new tests exercise `takeOwnership=true` only (`wrapped_ssl_context` passes `true` explicitly; `wrapped_ssl_context_null` throws before the ownership branch). The destructor branch where `SSL_CTX_free` is skipped (`takeOwnership_=false`) has no test.
> 
> https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/test/SecurityTest.cpp#L286-L306
> 
> Verified: `grep -n "takeOwnership\|, false" SecurityTest.cpp` returns no output on PR code; branch absent in base.
> 
> 4. The new constructor's Doxygen does not warn that callers are responsible for enforcing a minimum TLS version. `SSLContext(SSLProtocol)` applies `SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1` unconditionally; the new constructor wraps the context as-is and the test creates it with `TLS_method()` and no version restrictions.
> 
> https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L367-L374
> 
> Verified: no `SSL_OP_NO_TLSv1*` in new constructor; existing floor confirmed at [`TSSLSocket.cpp:210-212`](https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L210-L212); constructor absent in base.
> 
> 5. When `takeOwnership=false`, the caller must keep the `SSL_CTX*` alive at least as long as the `SSLContext` and all sockets created from it. This is not documented in the new constructor's Doxygen. The factory class has an explicit parallel warning ([`TSSLSocket.h:200-206`](https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L200-L206): _"It is the responsibility of the code using TSSLSocketFactory to ensure that the factory lifetime exceeds the lifetime of any sockets it might create"_).
> 
> https://github.com/apache/thrift/blob/08700779e125daaf3ab9c97fba6ef2e3ef70fa20/lib/cpp/src/thrift/transport/TSSLSocket.h#L367-L374
> 
> Verified: `@param takeOwnership` documents only the `true` case; factory warning confirmed in base and PR; constructor absent in base.
> 
> 6. `AGENTS.md` §6 and its Quick Reference checklist require TLS configuration changes to be cross-checked against `doc/thrift-threat-model.md` before merging. The PR body contains no mention of this. `doc/thrift-threat-model.md` is present in the repo. (Note: per §6, any details about trust-boundary implications belong in the JIRA ticket [THRIFT-6073](https://issues.apache.org/jira/browse/THRIFT-6073), not in the PR description or commit messages.)
> 
> Verified: AGENTS.md requirement confirmed; PR body grep for "threat" returns no output; `doc/thrift-threat-model.md` exists at PR head.
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

**#1 / #2 / #3 / #5 — ownership API:** Dropped the `takeOwnership` parameter and `takeOwnership_` member entirely. The injected constructor is now `explicit SSLContext(SSL_CTX* ctx)` and unconditionally takes ownership; the destructor always calls `SSL_CTX_free`. This removes the non-owning / UAF path, restores the pre-change object layout (pointer-only), and makes the ownership contract unambiguous.

**#4 — TLS version floor:** Expanded the constructor Doxygen to note that, unlike `SSLContext(SSLProtocol)` with default `SSLTLS`, the injected path does not apply `SSL_OP_NO_TLSv1*` and callers must configure a protocol floor on the `SSL_CTX` before wrapping if needed.

**#6 — threat model:** Cross-checked against `doc/thrift-threat-model.md` per AGENTS.md §6; details recorded in JIRA THRIFT-6073 (not in the PR description, per project guidance).

hongzhi-gao:
> One housekeeping note: for a change of this scope, it is worth making sure the JIRA ticket captures the key design decisions and rationale in addition to what is in the commit messages. Commit messages are squashed on merge and can be hard to find later; JIRA comments and the ticket description remain a stable reference for anyone reviewing the issue in future. If there is technical context currently spread across individual commit messages (e.g. the rationale for the `takeOwnership` parameter semantics, or the TLS version floor decision), copying the relevant parts into the JIRA ticket before this is merged would be a good investment.

Good point — thanks. I have updated THRIFT-6073 with a *Design decisions* section covering:

- why the patch was reshaped to a backend-neutral `SSLContext(SSL_CTX*)` hook (per review feedback)
- ownership semantics: unconditional transfer of `SSL_CTX` ownership (replacing the earlier `takeOwnership` parameter, which was dropped in dcc2a06c4 after review)
- TLS version floor: injected contexts are wrapped as-is; callers configure `SSL_OP_NO_TLSv1*` before wrapping if needed
- threat-model cross-check against `doc/thrift-threat-model.md` (per AGENTS.md §6)

I will squash the branch to a single commit before merge, as usual for Thrift PRs.

PR #3607, opened 2026-06-26
copy buffered data not the GByteArray struct in c_glib read_slow

dxbjavid:
thrift_framed_transport_read_slow copies leftover bytes from the previous frame with memcpy(buf, t->r_buf, t->r_buf->len), passing the GByteArray struct pointer as the source instead of t->r_buf->data. The correct source is the buffer's data member, as used by the other memcpy calls in the same file. thrift_buffered_transport_read_slow contains the same mistake in its leftover path. Both are updated to read from t->r_buf->data, and a regression test using a memory buffer exercises the cross-boundary read path in each transport.

### DIFF
--- a/lib/c_glib/src/thrift/c_glib/transport/thrift_buffered_transport.c
+++ b/lib/c_glib/src/thrift/c_glib/transport/thrift_buffered_transport.c
@@ -93,7 +93,7 @@ thrift_buffered_transport_read_slow (ThriftTransport *transport, gpointer buf,
   /* first copy what we have in our buffer. */
   if (have > 0)
   {
-    memcpy (buf, t->r_buf, t->r_buf->len);
+    memcpy (buf, t->r_buf->data, t->r_buf->len);
     want -= t->r_buf->len;
     t->r_buf = g_byte_array_remove_range (t->r_buf, 0, t->r_buf->len);
   }
--- a/lib/c_glib/src/thrift/c_glib/transport/thrift_framed_transport.c
+++ b/lib/c_glib/src/thrift/c_glib/transport/thrift_framed_transport.c
@@ -138,7 +138,7 @@ thrift_framed_transport_read_slow (ThriftTransport *transport, gpointer buf,
   /* first copy what we have in our buffer, if there is anything left */
   if (have > 0)
   {
-    memcpy (buf, t->r_buf, t->r_buf->len);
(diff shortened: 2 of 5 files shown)


Jens-G:
### Code review

Found 2 issues:

1. `testframedtransport_LDADD` in `Makefile.am` does not include `thrift_memory_buffer.o`, but the new `test_read_across_frames` uses `THRIFT_TYPE_MEMORY_BUFFER`. `CMakeLists.txt` masks this by linking the full `thrift_c_glib` library, but autotools builds will fail at link time with undefined references. Other tests in this directory that use `ThriftMemoryBuffer` include the `.c` file directly (e.g. `#include "../src/thrift/c_glib/transport/thrift_memory_buffer.c"`).

https://github.com/apache/thrift/blob/f20a320c2c6ba061e918751210267b7318635ff1/lib/c_glib/test/Makefile.am#L156-L163

2. The fix in `thrift_buffered_transport_read_slow` has no regression test. The equivalent fix in `thrift_framed_transport_read_slow` gets `test_read_across_frames`, but `testbufferedtransport.c` has nothing that exercises the `have > 0` branch. Per CONTRIBUTING.md, tests are required for every submitted change.

https://github.com/apache/thrift/blob/f20a320c2c6ba061e918751210267b7318635ff1/lib/c_glib/src/thrift/c_glib/transport/thrift_buffered_transport.c#L93-L100

dxbjavid:
both fixed.

for the link issue i went with adding thrift_memory_buffer.o to testframedtransport_LDADD (and testbufferedtransport_LDADD) rather than including the .c directly. the readcheck test can include thrift_memory_buffer.c because it's the only transport source it pulls in, but these two files already #include thrift_framed_transport.c / thrift_buffered_transport.c, and a second transport .c in the same translation unit collides on the file-local PROP_0 enum. linking the object keeps it in its own unit and the autotools build resolves cleanly.

for the missing coverage i added ReadAcrossBuffer to testbufferedtransport.c. the buffered read path never pre-buffers on its own, so to reach the have > 0 branch the test seeds the read buffer with 96 leftover bytes (well past sizeof(GByteArray)) and then reads across it through a memory buffer. under asan it reports a heap-buffer-overflow at thrift_buffered_transport.c:96 against the old memcpy and passes with the data-member copy.

dxbjavid:
Thank you for your time.

PR #3610, opened 2026-06-30
fix info-header string bound check in THeaderTransport::readString

dxbjavid:
The info-header string length is bounded against the header section before `ptr` moves past the length varint, so the remaining count is overstated by the varint width and a negative length is never rejected. A regression test is added to `ThrifttReadCheckTests`.

### DIFF
--- a/lib/cpp/src/thrift/transport/THeaderTransport.cpp
+++ b/lib/cpp/src/thrift/transport/THeaderTransport.cpp
@@ -199,13 +199,17 @@ void THeaderTransport::readString(uint8_t*& ptr,
   int32_t strLen;
 
   uint32_t bytes = readVarint32(ptr, &strLen, headerBoundary);
-  if (strLen > headerBoundary - ptr) {
+  // Bound the string against the header bytes that remain once the length varint
+  // itself is accounted for, and reject a negative length so the size_t
+  // conversion in assign() below stays within the buffer. ptr is only advanced
+  // once these checks pass, keeping the "advances on success" contract above.
+  uint8_t* strStart = ptr + bytes;
+  if (strLen < 0 || strLen > headerBoundary - strStart) {
     throw TTransportException(TTransportException::CORRUPTED_DATA,
                               "Info header length exceeds header size");
   }
-  ptr += bytes;
-  str.assign(reinterpret_cast<const char*>(ptr), strLen);
-  ptr += strLen;
+  str.assign(reinterpret_cast<const char*>(strStart), strLen);
+  ptr = strStart + strLen;
 }
 
 void THeaderTransport::readHeaderFormat(uint16_t headerSize, uint32_t sz) {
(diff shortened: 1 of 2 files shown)


Jens-G:
Pls read your email. Now.

Jens-G:
### Code review

Found 3 issues (adversarially verified with failing tests):

1. Exploit description in test inline comment (AGENTS.md §6)

Lines 345–347 of the new test describe the byte-level overread path in terms that name the exact exploitation mechanism. [AGENTS.md §6](https://github.com/apache/thrift/blob/master/AGENTS.md#6-security-work) requires neutral functional language in inline comments; the description of what the old code allowed to happen should be rephrased in terms of the invariant being enforced rather than the failure mode.

https://github.com/apache/thrift/blob/5c190a74e7ec3e3262ae73e63043d0cd955d1ef4/lib/cpp/test/ThrifttReadCheckTests.cpp#L345-L347

Verified: text absent from base commit `d68305a73`, present in PR code.

2. `strLen < 0` branch has no test

The new guard catches negative `strLen` values, but no test in the PR exercises this path. A 5-byte varint with bit 31 set (e.g. `0x80 0x80 0x80 0x80 0x08`) produces a negative `int32_t` from `readVarint32` and reaches this branch; without the guard, `str.assign(ptr, (size_t)(-2147483648))` propagates a `std::length_error` rather than a `TTransportException`.

https://github.com/apache/thrift/blob/5c190a74e7ec3e3262ae73e63043d0cd955d1ef4/lib/cpp/src/thrift/transport/THeaderTransport.cpp#L205-L206

Verified: Docker test confirmed the path is reachable and behaves differently on base vs. PR code.

3. `ptr` is advanced before the bounds check, contradicting the `"Advances ptr on success"` doc comment

Moving `ptr += bytes` to before the `if` means `ptr` (a non-const reference) is left pointing past the varint bytes when `readString` throws. The Doxygen comment documents that the function advances `ptr` only on success. Current callers in `readHeaderFormat` do not catch `readString` exceptions, so there is no immediate runtime effect, but the documented contract no longer holds and would mislead any subclass that catches the exception and then inspects `ptr`. Either the doc comment or the advance placement should be updated.

https://github.com/apache/thrift/blob/5c190a74e7ec3e3262ae73e63043d0cd955d1ef4/lib/cpp/src/thrift/transport/THeaderTransport.cpp#L204-L209

Verified: Docker two-state proof via subclass harness — assertion that `ptr` is unchanged after exception fails on PR code, passes on base.

dxbjavid:
all three addressed in the latest push.

1. reworded the test comment to describe the invariant (the key length has to fit within the header bytes that remain after the length varint) rather than the failure mode. did the same for the commit body.

2. added test_theadertransport_info_header_string_negative_length. it feeds a 5-byte key-length varint with the top bit set (0x80 0x80 0x80 0x80 0x08, decoding to INT32_MIN) and asserts a TTransportException, where the old code would let it reach assign and raise length_error.

3. fair point on the contract. rather than loosen the doc comment i kept the advance-on-success behaviour instead: the bounds check now runs against a local pointer past the varint and ptr is only moved once the checks pass, so a caller that catches the exception still sees ptr unchanged.

also read your email, thanks.

dxbjavid:
Thank you @Jens-G. Appreciated.

PR #3620, opened 2026-07-02
Fix/open access namedPipe security fix.

TheZaphod-Beeblebrox:


Jens-G:
### -1 Sorry, but no. 

I started to make comments but stopped, because that would just be too much of them. There is a lot of noise around a questionable and completely unexplained change. There is no problem statement, no reasoning of any kind that would remotely explain why this is a good thing to do or what the probvlem might be you're running into - just plain nothing. Also, the scripts you added have huge quality issues, not to mention the highly questionable changes to *.nuspec

Feel free to clean it up etc and **then** reopen.

PR #3621, opened 2026-07-04
THRIFT-5955 Require C extension in Python wheels

vitaly-zdanevich:
## What changed
- Keep cibuildwheel release builds from falling back to a pure-Python wheel when the fastbinary C extension fails to compile.
- Leave the upstream PyPI wheel workflow intact; it already builds manylinux aarch64 wheels after the rebase.

## Why
THRIFT-5955 is about published Linux arm64 wheels. The current upstream workflow now builds those wheels, and this PR adds a safety guard so release wheel jobs fail fast instead of uploading incomplete binary artifacts if the extension build fails.

## Validation
- Parsed `.github/workflows/pypi.yml` with Ruby YAML.
- Ran `python setup.py sdist --dist-dir /tmp/thrift-py-dist`.
- Ran `CIBUILDWHEEL=1 CXX=false python setup.py build_ext --force` and verified it fails with the new guard.
- Ran `git diff --check HEAD^ HEAD`.

---

Not my cup of tea - asked Codex to do it.

### DIFF
--- a/lib/py/setup.py
+++ b/lib/py/setup.py
@@ -149,6 +149,12 @@ except BuildFailed:
     print()
     print('*' * 80)
     print("An error occurred while trying to compile with the C extension enabled")
+
+    if os.environ.get('CIBUILDWHEEL') == '1':
+        print('Refusing to build a release wheel without the C extension')
+        print('*' * 80)
+        raise
+
     print("Attempting to build without the extension now")
     print('*' * 80)
     print()


HTHou:
This looks like the right behavior for release wheels: when `cibuildwheel` is producing binary artifacts for PyPI, silently falling back to a pure-Python wheel would make a broken fastbinary build easy to miss.

LGTM from my side.

Jens-G:
> Not my cup of tea - asked Codex to do it.

I have also nothing to do with it. WTF ...

https://www.apache.org/legal/generative-tooling.html

> [...] remind contributors that they are responsible [...]

PR #3642, opened 2026-07-19
THRIFT-6101: Use Ruby truthiness for CompactProtocol booleans

kpumuk:
The native `CompactProtocol#write_bool` implementation treated only the literal `true` as true, while the pure-Ruby implementation followed normal Ruby truthiness. Truthy values such as numbers and objects could therefore produce different wire values depending on whether `thrift_native` was loaded.

The native writer now uses Ruby’s truthiness semantics. Both implementations encode only `false` and `nil` as false, while other values encode as true.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6101](https://issues.apache.org/jira/browse/THRIFT-6101)


Jens-G:
### Code review

No issues found. Checked for bugs, AGENTS.md compliance, and security implications.

PR #3647, opened 2026-07-19
THRIFT-6106: Decode fixed-width values without undefined shifts

kpumuk:
Native Ruby protocol readers previously assembled fixed-width values through shifts on promoted signed integers. On common builds, those reads can appear to return the expected values, but UBSan exposes the undefined operation for high-bit byte patterns.

Before this change, building and running the focused examples from `lib/rb` with:

```sh
make -C ext clean all CFLAGS='-fPIC -O1 -g -fsanitize=undefined -fno-sanitize-recover=undefined' DLDFLAGS='-fsanitize=undefined'
UBSAN_OPTIONS='halt_on_error=1:print_stacktrace=1' bundle exec rspec spec/binary_protocol_spec.rb spec/compact_protocol_spec.rb --example fixed-width
```

reports:

```text
compact_protocol.c:618:42: runtime error: left shift of 128 by 24 places cannot be represented in type 'int'
```

This change builds the binary and compact protocol values with correctly sized unsigned intermediates, then converts to signed values only after assembly. It also preserves double bit patterns when the high bit is set.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6106](https://issues.apache.org/jira/browse/THRIFT-6106)

### DIFF
--- a/lib/rb/ext/binary_protocol_accelerated.c
+++ b/lib/rb/ext/binary_protocol_accelerated.c
@@ -331,33 +331,62 @@ static char read_byte_direct(VALUE self) {
   return (char)(FIX2INT(byte));
 }
 
+static int16_t uint16_to_int16(uint16_t value) {
+  if (value <= INT16_MAX) {
+    return (int16_t)value;
+  }
+  return (int16_t)((int32_t)value - ((int32_t)UINT16_MAX + 1));
+}
+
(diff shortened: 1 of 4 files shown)


Jens-G:
### Code review

No issues found. Checked for bugs, AGENTS.md compliance, and security implications.

PR #3648, opened 2026-07-20
THRIFT-6107: Preserve NonblockingServer reply order

kpumuk:
Ruby's `NonblockingServer` can dispatch multiple requests from one connection to different workers. When a later handler finishes first, it can write its reply before an earlier request’s reply, which breaks clients that intentionally pipeline requests on that connection.

This change gives each connection an ordered response queue. A completed reply waits until all earlier request sequences have been published; oneway calls, handler failures, write failures, connection removal, and shutdown advance or discard the affected queue safely. Requests on separate connections continue to run independently.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6107](https://issues.apache.org/jira/browse/THRIFT-6107)

### DIFF
--- a/lib/rb/lib/thrift/server/nonblocking_server.rb
+++ b/lib/rb/lib/thrift/server/nonblocking_server.rb
@@ -107,12 +107,17 @@ module Thrift
         @num_threads = num
         @logger = logger
         @connections = []
+        @readable_connections = []
         @buffers = Hash.new { |h, k| h[k] = Bytes.empty_byte_buffer }
+        @response_queues = {}
         @signal_queue = Queue.new
         @signal_pipes = IO.pipe
         @signal_pipes[1].sync = true
         @worker_queue = Queue.new
         @shutdown_queue = Queue.new
+        @capacity_callback = method(:response_capacity_available)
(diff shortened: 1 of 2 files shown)


Jens-G:
### Code review

Found 1 issue:

1. `ResponseQueue#complete` holds `@mutex` while performing a blocking socket write (`write_response` calls `@transport.write`/`flush` inside the lock). The same mutex is required by `reserve`, called only from the single `IOManager` thread's read/dispatch loop for every connection — so one client that stops draining its receive buffer while a reply write is in flight can stall dispatch for all other connections too, for up to `client_timeout` (default 5s; unbounded if `client_timeout: nil`). This contradicts the PR's stated goal that "requests on separate connections continue to run independently."

https://github.com/apache/thrift/blob/d8da61a6930123f8f47c9002e9709d71b5d5b161/lib/rb/lib/thrift/server/nonblocking_server.rb#L438-L461

https://github.com/apache/thrift/blob/d8da61a6930123f8f47c9002e9709d71b5d5b161/lib/rb/lib/thrift/server/nonblocking_server.rb#L497-L504

PR #3639, opened 2026-07-19
THRIFT-6098: Verify Ruby SSL peers by default

kpumuk:
This updates `Thrift::SSLSocket` to verify peers by default when no SSL context is provided or when the caller supplies a blank `OpenSSL::SSL::SSLContext`.

Caller-configured CA files, CA directories, and certificate stores are preserved. When no trust source is configured, the socket uses the system certificate store. Applications that intentionally disable peer identity checks can pass `verify_peer: false` explicitly.

This makes `Thrift::SSLSocket` consistent with `Thrift::HTTPClientTransport`, which already enables HTTPS peer verification by default.

## Compatibility

This is an intentional Ruby behavior change for 0.25.0. Applications currently relying on connections without peer identity checks must opt out explicitly with `verify_peer: false`.

The Ruby README documents the new default and its migration requirements.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? THRIFT-6098
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [x] Did you squash your changes to a single commit?  (not required, but preferred)
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"?
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

PR #3652, opened 2026-07-21
THRIFT-6098: Preserve caller-supplied Ruby SSL contexts

kpumuk:
`Thrift::SSLSocket` reconfigured a caller-supplied `OpenSSL::SSL::SSLContext` in place to apply its separate peer-verification setting. This changed the caller’s verification mode and could install a certificate store on an object the application expected to configure and reuse itself. Ruby OpenSSL also freezes a context after it is used by a socket, making later attempts to apply a different setting fail.

This change gives the context a single owner. When no context is supplied, Thrift creates one with peer verification enabled and the system certificate store. When an application supplies a context, Thrift passes it to OpenSSL without reconfiguring it; the context’s `verify_mode` and trust sources control certificate-chain verification. Thrift continues to check the server certificate against `server_hostname`, preserving hostname-mismatch detection.

The Ruby README now documents both ownership modes and shows how to supply a context configured with `OpenSSL::SSL::VERIFY_NONE` when certificate-chain verification is intentionally disabled.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6098](https://issues.apache.org/jira/browse/THRIFT-6098)

### DIFF
--- a/lib/rb/README.md
+++ b/lib/rb/README.md
@@ -157,11 +157,13 @@ Its EventMachine dependency also does not build on new Ruby development
 versions. Creating a Thin server now prints a warning. Move to
 `Thrift::RackApplication` with a supported Rack server such as Puma or Falcon.
 
-`Thrift::SSLSocket` now verifies peers by default, including when given a
-blank SSL context. It preserves configured certificate authorities and uses
-the system certificate store when no trust source is configured. Applications
-that intentionally connect without peer identity checks must pass
-`verify_peer: false` explicitly.
+`Thrift::SSLSocket` now verifies peers by default when no SSL context is
+provided. It creates a context that uses the system certificate store. A
+supplied SSL context is not reconfigured by Thrift, so the application is
+responsible for its verification mode and trust sources. A newly constructed
+`OpenSSL::SSL::SSLContext` defaults to `OpenSSL::SSL::VERIFY_NONE`; configuring
+a trust source alone does not enable certificate chain verification. The server
+certificate must still match the connection hostname.
 
 ### 0.24.0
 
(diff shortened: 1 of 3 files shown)


Jens-G:
### Code review

Found 1 issue:

1. This PR correctly removes PR #3639's in-place mutation of a caller-supplied `ssl_context`. One side effect: `configured_ssl_context`'s `@ssl_context ||= ...` now applies safe defaults (`VERIFY_PEER` + system cert store) only when no context is supplied at all — any caller-supplied context passes to `OpenSSL::SSL::SSLSocket.new` unchanged, regardless of its `verify_mode`. A plain `OpenSSL::SSL::SSLContext.new` defaults to `verify_mode == VERIFY_NONE`; setting a trust source like `ca_file` alone doesn't change that. `post_connection_check`, now called unconditionally, checks hostname match against the presented certificate but not chain trust. This PR's own diff adds `context.verify_mode = OpenSSL::SSL::VERIFY_PEER` to the "should use a configured certificate authority" spec to keep it exercising real verification, and the README's "supply a context configured with `VERIFY_NONE` to disable verification" wording doesn't state what verify_mode a supplied-but-unconfigured context actually gets. Worth confirming "verification is the caller's responsibility once any context is supplied" is the intended contract, and documenting that default explicitly if so.

https://github.com/apache/thrift/blob/4b2ed9ebecc57f87056912e6c6564e055b9012b6/lib/rb/lib/thrift/transport/ssl_socket.rb#L89-L96

https://github.com/apache/thrift/blob/4b2ed9ebecc57f87056912e6c6564e055b9012b6/lib/rb/spec/ssl_socket_spec.rb#L347-L355

PR #3656, opened 2026-07-21
THRIFT-6112: Reject duplicate Ruby socket opens

kpumuk:
The Ruby TCP transport currently replaces its live socket when `open` is called a second time. The original connection remains open but is no longer reachable through the transport, so a later `close` only closes the replacement connection.

This change rejects a duplicate open with `TransportException::ALREADY_OPEN` while preserving the existing live connection. Callers can continue using or explicitly close the original transport without leaving an orphaned TCP session.

The change is intentionally scoped to `Thrift::Socket`; transports that provide their own `open` implementation are not altered.
  

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6112](https://issues.apache.org/jira/browse/THRIFT-6112)

### DIFF
--- a/lib/rb/README.md
+++ b/lib/rb/README.md
@@ -165,6 +165,10 @@ responsible for its verification mode and trust sources. A newly constructed
 a trust source alone does not enable certificate chain verification. The server
 certificate must still match the connection hostname.
 
+`Thrift::Socket#open` now raises
+`Thrift::TransportException::ALREADY_OPEN` when the TCP transport is already
+open. Close the transport before opening it again.
+
 ### 0.24.0
 
 Connect timeout handling changed for both `Thrift::Socket` and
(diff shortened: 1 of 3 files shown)


Jens-G:
### Code review

Two suggestions, not blockers:

1. The new guard in `Thrift::Socket#open` doesn't propagate to `SSLSocket#open` or `UNIXSocket#open`, which override `open` without calling `super` or checking `open?` (the PR description notes this scoping is intentional):

https://github.com/apache/thrift/blob/70b530df05e3a5ded32d55970c97afbf3dde198a/lib/rb/lib/thrift/transport/ssl_socket.rb#L36-L39

https://github.com/apache/thrift/blob/70b530df05e3a5ded32d55970c97afbf3dde198a/lib/rb/lib/thrift/transport/unix_socket.rb#L32-L36

Moving the new spec into the shared example group both `ssl_socket_spec.rb` and `unix_socket_spec.rb` already include (`it_should_behave_like "a socket"`) would exercise the same property across all three transports:

https://github.com/apache/thrift/blob/70b530df05e3a5ded32d55970c97afbf3dde198a/lib/rb/spec/socket_spec_shared.rb#L22-L26

2. This PR changes observable behavior (a second `open` now raises `ALREADY_OPEN` instead of silently succeeding). THRIFT-5942 and THRIFT-6098 — prior changes to this same exception behavior — both added a note to `lib/rb/README.md`'s Breaking Changes / Migration Notes sections; consider doing the same here:

https://github.com/apache/thrift/commit/73123f9d44c33da5d92b9cdfdc183256beb7b889

https://github.com/apache/thrift/commit/bc2dd8fdb5b2aeecaa66b737db7de6c019a452f2

kpumuk:
Updated documentation, and created https://issues.apache.org/jira/browse/THRIFT-6117 to handle SSL and Unix sockets separately.

PR #3666, opened 2026-07-23
THRIFT-6108: implemented exception documentation generation

SLAVONchick:
Client: cpp,haxe,java,javame,kotlin,netstd,py

In this PR I've implemented the documentation generation of the exceptions thrown in the service methods.
I skipped the clients where there is no documentation is generated at all, so it only touched cpp,haxe,java,netstd,py.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes) - [Yes](https://issues.apache.org/jira/browse/THRIFT-6108)
- [x] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"? - Yes
- [x] Did you squash your changes to a single commit?  (not required, but preferred) - Yes
- [x] Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"? - Yes
- [x] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources. - Does involve code.


compiler/cpp/src/thrift/generate/t_oop_generator.h
```diff
@@ -104,6 +105,27 @@ class t_oop_generator : public t_generator {
...
+      
+      ss << '\n';
+
+      const std::vector<t_field*>& exceptions = tfunction->get_xceptions()->get_members();
+      std::vector<t_field*>::const_iterator e_iter;
+      for (e_iter = exceptions.begin(); e_iter != exceptions.end(); ++e_iter) {
+        t_field* e = *e_iter;
+        std::string namespace_str = e->get_type()->get_program()->get_namespace(lang);
+        
+        if (static_cast<std::string>(lang).compare("cpp") == 0) {
```
Jens-G: No, really, don't. We have OOP these days, like in "t_oop_generator", so we can use virtual methods to do such stuff. 

PS: Pun intended - SCNR 🤣

Jens-G (changes requested):

Jens-G:
I like the idea. Would be glad to see this move forward.

compiler/cpp/src/thrift/generate/t_oop_generator.h
```diff
@@ -20,6 +20,7 @@
 #ifndef T_OOP_GENERATOR_H
 #define T_OOP_GENERATOR_H
 
+#include <boost/algorithm/string/replace.hpp>
```
Jens-G: IIRC we don't use boost in the compiler. Can this be replaced by sth else?

compiler/cpp/src/thrift/generate/t_oop_generator.h
SLAVONchick: Totally agree, my bad. Fixed.

compiler/cpp/src/thrift/generate/t_oop_generator.h
SLAVONchick: Found identical implementations in several generators and left only one in base t_oop_generator.

SLAVONchick:
> I like the idea. Would be glad to see this move forward.

Hi! Thank you for the review! I'm really looking forward to finish this work.

lib/cpp/test/TTransportFactoryConfigTest.cpp
```diff
@@ -104,7 +104,8 @@ BOOST_AUTO_TEST_CASE(test_piped_transport_factory_preserves_configuration) {
 // failing on a truncated buffer.
 BOOST_AUTO_TEST_CASE(test_framed_transport_factory_enforces_custom_max_frame_size) {
   const int kMaxFrameSize = 64;
-  auto config = std::make_shared<TConfiguration>(TConfiguration::DEFAULT_MAX_MESSAGE_SIZE, kMaxFrameSize);
+  const int kMaxMessageSize = TConfiguration::DEFAULT_MAX_MESSAGE_SIZE;
+  auto config = std::make_shared<TConfiguration>(kMaxMessageSize, kMaxFrameSize);
```
Jens-G, line 109: Not sure how related that change is?

Jens-G:
### Code review

Found 5 issues:

1. Deleting javame's `generate_java_doc(ostream&, t_function*)` override silently regresses javame output. `t_javame_generator` still declares two other overloads of that name and has no `using t_oop_generator::generate_java_doc;`, so C++ name hiding removes the base `t_function*` overload from the candidate set. Since `t_function : public t_doc`, the call sites at [lines 1956](https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_javame_generator.cc#L1955-L1957) and 1981 now bind to the `t_doc*` overload instead. Javame loses the auto-generated `@param` tags it emitted before this PR and never gains the new `@throws` tags. It compiles without a warning.

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_javame_generator.cc#L188-L192

   Worth fixing together with this: [javame's `get_namespace()`](https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_javame_generator.cc#L253-L257) appends no trailing separator, unlike the base default (`"."`) and the cpp override (`"::"`), while the call site concatenates straight onto the type name. Once the overload resolution above is corrected, javame will emit `@throws thrift.testXception`. [Kotlin's override](https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L183-L191) has the same shape.

2. The unconditional `ss << '\n';` runs whether or not the function declares any exceptions, and the `@throws` loop below already prefixes each entry with `\n`. Functions with no `throws()` clause get a trailing blank ` * ` line; functions with one get a doubled blank line. This affects every documented service function emitted by the cpp and java generators, which are the two that reach this shared method. Guarding it with `if (!exceptions.empty())` fixes both cases.

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_oop_generator.h#L112-L121

3. `t_type::get_name()` returns an empty string for inline container types, because `t_list`/`t_set`/`t_map` construct through `t_container()` into the default `t_type()` ctor, which never sets `name_`. A field declared `1: list<string> items` renders as `- items ()`. Base types also come out as IDL names (`i32`, `string`) rather than Python ones. The file already has `type_to_py_type()` for exactly this, used by `arg_hint`/`member_hint`/`func_hint`, which renders `list[str]` and `dict[str, int]` correctly.

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_py_generator.cc#L2774-L2780

4. The `@throws` name is taken from `e->get_type()` without resolving typedefs. `validate_throws()` in `main.cc` checks the thrown type via `get_true_type()`, so `typedef SomeException Alias` followed by `throws (1: Alias e)` is legal IDL, and this would emit `@throws ns.Alias` — a name with no generated class behind it, since typedefs produce no class in Java or C++. The sibling `generate_java_doc(ostream&, t_field*)` two functions above already goes through `get_true_type()` for the same reason (THRIFT-4086, `4f63573f5`), and the netstd path in this PR gets it right via `type_name()`.

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_oop_generator.h#L118-L122

5. This change to `lib/cpp/test/TTransportFactoryConfigTest.cpp` is unrelated to THRIFT-6108 and is a no-op — it extracts `TConfiguration::DEFAULT_MAX_MESSAGE_SIZE` into a local and passes that instead of the constant. Worth dropping so the PR stays scoped to the doc generators.

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/lib/cpp/test/TTransportFactoryConfigTest.cpp#L106-L109

On the design, separately from the defects above: `get_gen_name()` looks like more machinery than the feature needs.

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_oop_generator.h#L39-L45

Making it pure virtual obliges all 18 `t_oop_generator` subclasses to implement it, each with its own `gen_name_` member, which is most of why this PR touches 21 generator files when the machinery only ever serves the shared doc path. It has three call sites and all of them are inside `get_namespace()`, whose only consumer is the new `@throws` line — reachable from cpp, java and javame alone. Since javame's override hardcodes `"java"`, only the cpp and java implementations are ever actually called.

The bigger issue is that the abstraction doesn't fit its users: four of the six `get_namespace()` overrides discard `get_gen_name()` outright — javame hardcodes `"java"`, c_glib returns `nspace`, and perl and php delegate to the existing `perl_namespace()`/`php_namespace()` helpers. It is only usable where a generator's registration name and its IDL namespace key happen to coincide, which is exactly the cpp and java case.

That string also already exists in the compiler, in one place. `THRIFT_REGISTER_GENERATOR` registers each generator under `#language`:

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/generate/t_generator_registry.h#L96-L104

and `t_program::set_namespace()` validates every `namespace <key>` directive against that same map:

https://github.com/apache/thrift/blob/b3754d44fe4bd6edce787f481bcf4c7ff8b95e95/compiler/cpp/src/thrift/parse/t_program.h#L297-L306

Keeping `get_namespace()` as the virtual hook and dropping `get_gen_name()` entirely would give the same result: `t_cpp_generator::get_namespace()` passes `"cpp"` directly, the base default passes `"java"`, and the other 16 generators need no change at all. That is the pattern the rest of the compiler already uses, e.g. `t_netstd_generator.cc:140` and `t_kotlin_generator.cc:202`.

lib/cpp/test/TTransportFactoryConfigTest.cpp
SLAVONchick: Otherwise I get an error during compilation:
```
[ 48%] Linking CXX executable ../../../bin/TTransportFactoryConfigTest
/usr/bin/x86_64-linux-gnu-ld.bfd: CMakeFiles/TTransportFactoryConfigTest.dir/TTransportFactoryConfigTest.cpp.o: in function `test_framed_transport_factory_enforces_custom_max_frame_size::test_method()':
/home/slavonchick/repos/cpp/thrift-fork/lib/cpp/test/TTransportFactoryConfigTest.cpp:107:(.text+0x3959): undefined reference to `apache::thrift::TConfiguration::DEFAULT_MAX_MESSAGE_SIZE'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/cpp/test/CMakeFiles/TTransportFactoryConfigTest.dir/build.make:112: bin/TTransportFactoryConfigTest] Error 1
make[1]: *** [CMakeFiles/Makefile2:2415: lib/cpp/test/CMakeFiles/TTransportFactoryConfigTest.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```
Although it works for me, I can delete this fix if necessary.

Jens-G:
### Code review

The `replace_all()` consolidation from this PR is now on master as `2ae9c11db`, so a rebase will drop that part. Reviewed the rest. The main path works — `testMultiException` in `test/ThriftTest.thrift` comes out as `@throws thrift.test.Xception Thrown when a bad thing happens`. Everything below was reproduced against a compiler built from `b80a1464d`.

Found 4 issues:

1. The `@throws` name is resolved through typedefs but the namespace is not, so an exception reached through a typedef gets a qualified name that no generated class has. `get_true_type()` is applied to the name half only, while `get_namespace()` still receives the raw type — and a `t_typedef` carries the program the *typedef* was declared in, not the one the underlying exception lives in. With `inc.thrift` (`namespace java com.inc`, `exception RealExc`) and `main.thrift` (`namespace java com.main`, `include "inc.thrift"`, `typedef inc.RealExc Alias`, `void go() throws (1: Alias e)`), the generated file contradicts itself three lines apart:

   ```
    * @throws com.main.RealExc
    */
   public void go() throws com.inc.RealExc, org.apache.thrift.TException;
   ```

   Same in C++ (`@throws main_ns::RealExc` for a class that is really `inc_ns::RealExc`) and in JavaME through the new delegation. The haxe and netstd paths added in this PR avoid it by going through `type_name()`, which resolves once and derives both halves from the resolved type.

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_oop_generator.h#L117-L119

2. The default `get_namespace()` appends the `.` separator unconditionally, so a file with no `namespace java` yields `@throws .SomeError`, which is not a resolvable Javadoc reference. `exception NoNsErr` plus `void doThing() throws (1: NoNsErr e)` with no namespace declared generates:

   ```
    * @throws .NoNsErr
    */
   public void doThing() throws NoNsErr, org.apache.thrift.TException;
   ```

   `type_name()` in the java generator guards this case with `if (!package.empty())` ([t_java_generator.cc:4687](https://github.com/apache/thrift/blob/2ae9c11db596a90fc8daa21c3bd319821a4da42b/compiler/cpp/src/thrift/generate/t_java_generator.cc#L4686-L4688)), and `get_enum_class_name()` in this same file sidesteps it by prefixing only for types from another program. C++ is unaffected — `::NoNsErr` is a valid global-namespace qualifier — but Java and JavaME are.

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_oop_generator.h#L39-L43

3. The `if (!exceptions.empty())` guard fixed the no-exceptions case from the last round, but for functions that *do* declare exceptions the guard's own `ss << '\n'` still runs immediately before a loop whose every entry already begins with `\n`. Since IDL doc text carries its own trailing newline, a function with no parameters, or one whose last parameter has its own doc comment, gets two blank ` * ` lines:

   ```
    * zero params, throws
    * 
    * 
    * @throws com.probe.MyError
   ```

   A function with undocumented parameters renders with a single blank line, which is why `testMultiException` looks right.

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_oop_generator.h#L112-L119

4. Python's `Raises:` section lists the throws-clause field label rather than the exception type, so in the default configuration the docstring never names the exception. It shows up only as a parenthetical under `py:type_hints,enum` (` - err1 (Xception)`). `test/DebugProtoTest.thrift` already demonstrates the effect — `void methodThatThrowsAnException() throws (1: ExceptionWithAMap xwamap)` gains a docstring whose entire content is the label:

   ```python
   def methodThatThrowsAnException(self):
       """
       Raises:
        - xwamap

       """
   ```

   The field label is the salient name for a parameter, since it is the real keyword argument, but in a throws clause it is an arbitrary tag — java, haxe and netstd all emit the type as the primary token. (Checked the rest of the refactor: plain `--gen py` over `DebugProtoTest.thrift` is otherwise byte-identical to master, so the `has_doc` pointer change did not disturb existing docstrings.)

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_py_generator.cc#L2734-L2737

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_py_generator.cc#L2774-L2778

Three suggestions, below the bar for the list above but verified:

- The `Client:` trailer lists `cpp,haxe,java,netstd,py` but omits `javame`, whose output this change alters — swapping its own implementation for a delegation to the base method gains it `@throws` lines it did not emit before. [AGENTS.md](https://github.com/apache/thrift/blob/2ae9c11db596a90fc8daa21c3bd319821a4da42b/AGENTS.md#L34-L38) asks for a "comma-separated list of affected languages", and this repo tags javame separately from java (e.g. `8e8e58a80`).

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_javame_generator.cc#L3059-L3062

- The new kotlin `get_namespace()` override is unreachable. Its only caller is the `@throws` block in `t_oop_generator::generate_java_doc(ostream&, t_function*)`, and kotlin emits function docs through its own `generate_kdoc_comment()` instead, so a kotlin service function with a documented `throws` clause generates no `@throws` at all — consistent with `Client:` omitting kotlin. Either wire kotlin up or drop the override.

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L176-L185

- The `gen_type_hints_` block annotates every parameter *and* every struct attribute with its Python type, which is a separate feature from documenting exceptions. It changes existing output for current `py:type_hints` users beyond the new `Raises` sections: `- arg` becomes `- arg (str)` in service docstrings and `- m` becomes `- m (str)` in `ttypes.py`. Worth its own commit so THRIFT-6108 stays about exception docs.

https://github.com/apache/thrift/blob/b80a1464d901fdf0c849d8f11d49b9c0d794c99d/compiler/cpp/src/thrift/generate/t_py_generator.cc#L2775-L2778

One design observation, no action needed for this PR: a base default that reads `get_namespace("java")` makes "java" the inherited namespace language for all 18 `t_oop_generator` subclasses, which is a little surprising in a class named for OOP rather than for Java. It is harmless today because the shared doc path is the only caller and only cpp, java and javame reach it. Both fixes for issues 1 and 2 land inside that same two-line default.

Jens-G:
FYI I extracted the `replace_all()` refactoring into a separate commit, keeping all the credits etc.

SLAVONchick:
> FYI I extracted the `replace_all()` refactoring into a separate commit, keeping all the credits etc.

Yeah, thanks! Really appreciate that.

lib/cpp/test/TTransportFactoryConfigTest.cpp
SLAVONchick: Removed this from this PR

SLAVONchick:
@Jens-G Hi! Just a friendly reminder to check on the PR :-)

Jens-G:
### Code review

Thanks for the update, and sorry for the slow turnaround. Re-reviewed at `c554a5997`; everything below was reproduced against compilers built from the PR head and from the merge base (`64601a565`).

From the last round: issues 1 (typedef namespace) and 4 (Python listing the field label) are fixed, and so are suggestions 1 (`Client:` trailer) and 3 (the `gen_type_hints_` block). Issue 2 is fixed in the base method but reintroduced in the new Kotlin override, and issue 3 is fixed in the base but present in the new haxe copy — both are below as issues 2 and 6. Suggestion 2 was addressed by wiring Kotlin up, which is where issues 2 and 3 come from.

Found 7 issues:

1. The whole `@throws` block sits inside `if (tfunction->has_doc())`, so documenting only the throws clause and not the method emits nothing at all. Python is the only one of the seven bindings where the feature fires in that case — its rewritten path builds `has_doc` up from the sections, so `Raises:` can stand alone. cpp, java, javame, kotlin, haxe (`t_haxe_generator.cc:3139`) and netstd (`t_netstd_generator.cc:3945`) all generate zero comment lines for:

   ```thrift
   void f() throws (/** this exception doc is dropped */ 1: E1 e)
   ```

   Documenting the exceptions rather than the method is the natural way to write this, and it silently produces nothing in six of seven languages.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_oop_generator.h#L104-L106

2. The Kotlin override has two problems. It reads `namespace kotlin`, which is a key nothing else in the compiler uses — the Kotlin generator derives every real package from `namespace java` ([:195](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L195), :325, :408), and `namespace kotlin` appears in no `doc/` page and no `.thrift` in the tree. With both keys set, the generated file contradicts itself:

   ```kotlin
   package com.main          // from namespace java
   ...
    * @throws com.mainkt.LocalExc   // from namespace kotlin
   ```

   Second, it returns `namespace_str + "."` unconditionally, dropping the empty-package guard the base method grew in this round. `exception NoNsErr` with no namespace declared yields `@throws .NoNsErr` — exactly round-2 issue 2, fixed in the base and reintroduced here.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L177-L184

3. Swapping `generate_kdoc_comment` for `generate_java_doc` at this call site also adopts the base `@param` loop, which Kotlin did not emit before. `generate_kdoc_comment` ([:2022](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L2022-L2026)) writes the doc text and nothing else. Diffing the generated `ThriftTest.kt` before and after, every interface method gains a blank ` * ` line and a description-less `@param thing` — sitting directly under the hand-written `@param string thing - the string to print` the IDL already carries. Only the `@throws` loop was needed here.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L1430-L1434

4. `get_true_type(...)->get_name()` emits the raw IDL name, bypassing the target language's identifier escaping, so the `@throws` reference does not match the class the same generator emits three lines below it. This is the THRIFT-5927 family — java routes every class name through `make_valid_java_identifier()` and kotlin through `kotlin_safe_name()` ([:410](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_kotlin_generator.cc#L410)), and neither is reached from here:

   ```
   exception native  →  class file $native.java, signature `throws $native`,
                        but the doc says @throws kw.pkg.native
   ```

   javadoc with default doclint reports one error (`unexpected text  * @throws kw.pkg.native`) plus `warning: no @throws for kw.pkg.$native`. Kotlin has the same shape with `exception object`. netstd gets it right via `type_name()`.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_oop_generator.h#L121-L124

5. Newlines are erased from the exception doc without a space in their place, so the last word of one line is glued to the first of the next:

   ```
   @throws com.nasty.MultiLineExc First line of the explanationsecond line continues here
   ```

   Same in the netstd copy at `t_netstd_generator.cc:3971`. The erase is also unnecessary: the `@param` loop ten lines above does not strip, and `generate_docstring_comment()` ([t_generator.cc:170-192](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_generator.cc#L170-L192)) already splits on newlines and re-prefixes each line.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_oop_generator.h#L125-L128

6. The haxe copy emits the throws-clause field name between the type and the description, so under the `@throws <Type> <description>` grammar the field label becomes the first word of the description. The PR's own fixture change shows it: `@throws Xception err1 Thrown when a bad thing happens`. For an undocumented exception the result is a dangling `@throws Boom e` whose entire description is a field label. The base, netstd, php and py paths all omit the name.

   The same loop does not strip the trailing newline `get_doc()` carries, so consecutive documented entries are separated by a stray blank line — round-2 issue 3, in the haxe copy:

   ```
    * @throws Xception err1 Thrown when a bad thing happens
    * 
    * @throws Xception2 err2 Thrown when the input is in incorrect format, for example
   ```

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_haxe_generator.cc#L3152-L3160

7. Author-controlled doc text goes verbatim into an XML doc element. `/** thrown if a < b && c > d */` produces `/// <exception cref="global::Ns.XmlExc">thrown if a < b && c > d</exception>`; building that with `GenerateDocumentationFile=true` gives four `CS1570` warnings and drops the entire member from the generated `.xml` (`<!-- Badly formed XML comment ignored for member "M:..." -->`). The `<param>`/`<summary>` text in the same function has the same hole today, so this is a new injection site rather than a new class of bug — `t_delphi_generator.cc:477` already carries the `&`/`<`/`>` helper.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_netstd_generator.cc#L3963-L3975

On the design: a correct implementation of this exact feature already exists in a sibling `t_oop_generator` subclass, and the PR leaves it untouched.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_php_generator.cc#L2833-L2874

`t_php_generator::generate_php_doc(ostream&, t_function*)` emits `@throws <qualified type> <doc>` and avoids issues 1, 4 and 5 structurally rather than by getting three separate details right. `has_doc()` wraps only the function's own text, so an exception-only doc still renders. It reuses the generator's own `type_to_phpdoc()`, so qualification and escaping come for free instead of via a new namespace lookup. And it does not strip newlines. Its `@param` and `@return` sections are built the same way.

After this PR there are four parallel `@throws` loops in four formats — `t_oop_generator.h:121`, `t_haxe_generator.cc:3156`, `t_netstd_generator.cc:3968`, `t_php_generator.cc:2866` — so every fix above reaches exactly one of them, and php gets none of them. If the shared loop delegated the type rendering to a per-generator hook the way php delegates to `type_to_phpdoc()`, issues 4 and 5 would not be expressible, and haxe and netstd would not need copies at all.

Three suggestions, below the bar for the list above but verified:

- The new C++ `get_namespace()` re-implements `namespace_prefix()` ([:4724](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_cpp_generator.cc#L4724-L4746)) and already disagrees with it: `namespace_prefix()` returns ` ::a::b::` with a leading global qualifier and space, and carries a comment explaining why; the new one returns `a::b::`. So `@throws main_ns::LocalExc` where the rest of the same generator writes ` ::main_ns::LocalExc`. `type_name(e->get_type())` would replace both this override and the `get_namespace` + `get_name` concatenation at the call site.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_cpp_generator.cc#L459-L463

- The shared params helper branches on `is_method_xcepts()` to choose type-vs-name rendering, using an AST property as a proxy for which of its two callers invoked it — both of which already pass the mode explicitly as `subheader`. The flag does not mean "this is a throws clause": [t_struct.h:162](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/parse/t_struct.h#L162) documents it as "struct holds the exceptions declared at a service method", and `t_delphi_generator.cc:2624` sets it on a `<fn>_result` struct that also carries `success`. Passing a bool alongside `subheader` removes the coupling.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_py_generator.cc#L2771-L2780

- A few style points, since the checklist asks for `make style`: line 117 is whitespace-only, so `git diff --check` exits non-zero; line 123 is 120 characters against the 100-column limit in `.clang-format` and [doc/coding_standards.md:34](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/doc/coding_standards.md#L34), and evaluates `get_true_type(e->get_type())` twice — hoisting a local fixes both; and `t_cpp_generator.cc:459` / `t_kotlin_generator.cc:177` write `t_type *type` where their own declarations and `PointerAlignment: Left` say `t_type* type`.

https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/src/thrift/generate/t_oop_generator.h#L116-L124

Last thing — the only change under `test/` is two doc comments in a cross-language interop fixture that asserts nothing about generated documentation, and it exercises none of the paths above: not the exception-only doc, not the multi-line doc, not the empty namespace, not a reserved identifier, not XML escaping. `compiler/cpp/test/compiler/markdown_doc_test.py` with `DocTest.thrift`/`DocTest.md` is a working golden-output test for generated docs, already wired up at [compiler/cpp/test/CMakeLists.txt:36](https://github.com/apache/thrift/blob/c554a5997061e7cb1f89cb122c47fdedc8ae99cd/compiler/cpp/test/CMakeLists.txt#L36) — it passes unchanged against this build, which is to say it covers none of the new code. That harness looks like the right home for a case or two here.

SLAVONchick:
@Jens-G Hi! Could you please re-run the AppVeyor CI job? It exceeded the build time limit while compiling the C++ tests, before any tests were run, so the failure does not appear related to this PR. Thanks!

Jens-G:
### Code review

Re-reviewed at `d8ec9325e`. All seven issues and the three suggestions from the last round are fixed, and the design point landed as the `get_doc_type_name()` hook — the parallel `@throws` loops are down from four to three (the shared one, netstd's, which has to build a different element, and php's pre-existing one). Everything below was reproduced against compilers built from the PR head and from the merge base (`3dece7e9e`). The netstd output was compiled with `GenerateDocumentationFile=true`: 0 warnings, and every member survives into the `.xml` — the `cref` values resolve as `T:Main.Ns.object` / `T:Main.Ns.native`.

Found 1 issue:

1. JavaME doubles the package for an exception declared in an included file. `type_name()` already qualifies a type whose program is not the one being generated, so prefixing `get_namespace("java")` on top of it yields a name no class has. With `inc.thrift` (`namespace java com.inc`, `exception RealExc`) included from `main.thrift`, the generated interface contradicts itself on adjacent lines:

   ```java
       /**
        * doc
        * 
        * @throws com.inc.com.inc.RealExc
        */
       public void viaInclude() throws com.inc.RealExc, TException;
   ```

   Both the direct-include and the typedef path hit it; a same-file exception is unaffected. `t_java_generator::get_doc_type_name()` avoids this by letting `type_name(..., force_namespace = true)` qualify once — javame's `type_name()` has no such parameter, so the extra prefix needs to be conditional on the type being local.

https://github.com/apache/thrift/blob/d8ec9325e8bbf9f6b8bb505e912bd2bcf9445eae/compiler/cpp/src/thrift/generate/t_javame_generator.cc#L2837-L2843

   The qualification it duplicates:

https://github.com/apache/thrift/blob/d8ec9325e8bbf9f6b8bb505e912bd2bcf9445eae/compiler/cpp/src/thrift/generate/t_javame_generator.cc#L2825-L2835

Three suggestions, below the bar for the list above but verified:

- `exception_doc_test.py` covers five of the seven languages in the `Client:` trailer — py and javame are missing, and javame is where the issue above lives. Both fit the existing harness unchanged (`--gen py` writes a `Raises:` section, `--gen javame` writes `@throws`), so it is two more `_generate()` calls.

- The new `t_netstd_generator::xml_encode()` is the same three replacements as [`t_delphi_generator::xml_encode()`](https://github.com/apache/thrift/blob/d8ec9325e8bbf9f6b8bb505e912bd2bcf9445eae/compiler/cpp/src/thrift/generate/t_delphi_generator.cc#L472-L487). Both classes derive from `t_oop_generator`, which is already where `replace_all()` lives, so one copy there would serve both.

- A method whose exceptions are documented but which carries no doc of its own opens its comment with an empty line, because `generate_throws_doc()` unconditionally leads with `"\n@throws"`. cpp, java, javame, kotlin, haxe and netstd all emit `/**\n * \n * @throws …`. Cosmetic, but it lands on exactly the case this round made work.

https://github.com/apache/thrift/blob/d8ec9325e8bbf9f6b8bb505e912bd2bcf9445eae/compiler/cpp/src/thrift/generate/t_oop_generator.h#L44-L57

On scope: delphi does generate method documentation — `--gen delphi:xmldoc` emits `<summary>` and `<param>` from [`generate_delphi_doc(ostream&, t_function*)`](https://github.com/apache/thrift/blob/d8ec9325e8bbf9f6b8bb505e912bd2bcf9445eae/compiler/cpp/src/thrift/generate/t_delphi_generator.cc#L536-L559) — and is not in the `Client:` list. Same shape as the netstd path. Not a blocker here; a follow-up ticket would be fine.

Jens-G:
### Code review

Re-reviewed at `19109264a`, against compilers built from this head and from `d8ec9325e`.

The issue from the last round is fixed. JavaME now writes `@throws com.inc.RealExc` for a directly included exception, for a typedef in the including file and for a typedef in the included file. That matches the signature's `throws` clause. The new javame test fails against `d8ec9325e` and passes now. The first two suggestions are done too: py and javame are covered by the test, and moving `xml_encode()` into `t_oop_generator` leaves the `delphi:xmldoc` and netstd output byte-identical.

The third suggestion is fixed for cpp, java, javame, kotlin and haxe through `generate_throws_doc()`. netstd builds its `<exception>` elements in its own loop, which still writes the separator before the first one. So a method without a doc of its own still opens with an empty line, e.g. from `test/DebugProtoTest.thrift`:

```csharp
    /// 
    /// <exception cref="ExceptionWithAMap"></exception>
    global::System.Threading.Tasks.Task<bool> declaredExceptionMethod(bool shouldThrow, CancellationToken cancellationToken = default);
```

Cosmetic. Writing the `'\n'` only when `ps` already has content, as `generate_throws_doc()` now does, would fix it. A netstd version of the new java `/**\n     * @throws` assertion would keep it covered.

https://github.com/apache/thrift/blob/19109264a5386de3dcd7f6d85397a5841248db24/compiler/cpp/src/thrift/generate/t_netstd_generator.cc#L3962-L3979

Nothing else changed in the generated output: for `ThriftTest.thrift` and `DebugProtoTest.thrift` across all affected generators, the only difference is the removed empty lines.

SLAVONchick:
Hi @Jens-G  — checking in on this PR. The test failure after your push looks unrelated to the changes. Is there anything I need to do on my end, or is this waiting on your side?

Jens-G:
can you rebase pls?

SLAVONchick:
> can you rebase pls?

done!

PR #3669, opened 2026-07-24
THRIFT-6120: Bound Ruby TLS accept handshakes

kpumuk:
`SSLServerSocket` previously applied its client timeout only after OpenSSL had completed the TLS handshake. A TCP peer that remained silent or stopped partway through negotiation could therefore keep `accept` blocked beyond the configured timeout.

This change separates TCP acceptance from TLS negotiation and drives the handshake nonblockingly against one monotonic deadline. Connections that time out, fail, or are interrupted before ownership transfers to the returned transport are closed. Successful connections continue to be returned as `Thrift::Socket` instances with the configured client timeout.

`NonblockingServer` and `ThreadPoolServer` now treat TLS handshake errors as recoverable accept failures, matching the existing behavior of the simple and threaded servers. The thread-pool server retries from the same worker instead of reporting the handshake failure as a worker exception. The optional OpenSSL check remains guarded so directly loaded plain-TCP server components preserve unrelated accept errors when OpenSSL has not been loaded.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? [THRIFT-6120](https://issues.apache.org/jira/browse/THRIFT-6120)

### DIFF
--- a/lib/rb/lib/thrift/server/nonblocking_server.rb
+++ b/lib/rb/lib/thrift/server/nonblocking_server.rb
@@ -53,7 +53,12 @@ module Thrift
             break
           end
           next if rd.nil?
-          socket = @server_transport.accept
+          begin
+            socket = @server_transport.accept
+          rescue => e
+            next if defined?(OpenSSL::SSL::SSLError) && e.is_a?(OpenSSL::SSL::SSLError)
+            raise
+          end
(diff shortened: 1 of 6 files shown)


kpumuk:
Copilot, you are not helping.

PR #3686, opened 2026-07-29
THRIFT-6129: Stream Ruby HeaderTransport ZLIB output

kpumuk:
Ruby HeaderTransport could buffer a complete ZLIB inflate result before checking its configured decompressed-size limit. Process the stream incrementally and reject a chunk before it would exceed that limit.

Add coverage for a highly compressible Header frame with a 100-byte limit, while retaining normal within-limit decoding.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([THRIFT-6129](https://issues.apache.org/jira/browse/THRIFT-6129), not required for trivial changes)

### DIFF
--- a/lib/rb/lib/thrift/transport/header_transport.rb
+++ b/lib/rb/lib/thrift/transport/header_transport.rb
@@ -77,9 +77,6 @@ module Thrift
     # Default decompressed-size cap for ZLIB transform (~15.6 MB, matches other Thrift bindings)
     DEFAULT_MAX_DECOMPRESSED_SIZE = 16_384_000
 
-    # Chunk size for streaming inflate loop
-    ZLIB_INFLATE_CHUNK_SIZE = 16_384
-
     # Binary protocol version mask and version 1
     BINARY_VERSION_MASK = 0xffff0000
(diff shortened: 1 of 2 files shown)


kpumuk:
@Jens-G this is a follow-up to https://github.com/apache/thrift/pull/3547, which moves the check to _before_ we use the uncompressed buffer, and also adds the same check to the final block.

kpumuk:
Follow-up: removed `ZLIB_INFLATE_CHUNK_SIZE`, which previously only sliced the compressed input passed to `Inflate#inflate`. One such input slice could still expand into the complete decompressed payload before `HeaderTransport` checked its limit.

The streaming `Zlib::Inflate` API instead delivers decompressed output incrementally. Ruby’s current implementation emits callback chunks up to 16 KiB, and `HeaderTransport` now checks the cumulative size before appending each chunk. This avoids materializing an unbounded full inflate result in `HeaderTransport`.

The configured limit remains an application-buffer limit rather than a strict process-allocation limit: Ruby Zlib may allocate its internal output chunk before invoking the callback.

PR #3689, opened 2026-07-30
Add configurable compatibility allowances to thrift audit

zsy056:
## Summary

- add `--audit-allow-optional-field-removal` so explicitly optional fields can be removed when intentionally enabled
- add `--audit-allow-required-field-to-default` as a binding-dependent audit suppression
- keep strict behavior by default and retain negative checks for default/required removal and unsupported requiredness transitions
- document binding differences, required reader-first rollout, and service-argument scope
- keep audit fixtures as standalone `.thrift` files
- register the audit suite with Automake and CMake and exercise it in GitHub Actions

## Compatibility behavior

Removing an explicitly optional field is wire-compatible.

Changing `required` to default requiredness is not universally wire-compatible. Local generated-code validation showed:

- ordinary C++ default-requiredness strings, containers, and nested structs are written, and an empty default `string` round-tripped into a required reader
- a Java default-requiredness `string` may remain `null` and be omitted; an older required reader then rejects the payload

Users enabling `--audit-allow-required-field-to-default` must verify every binding and field type and upgrade all readers away from explicit `required` before deploying writers that may omit it. The option also applies to service arguments; omission may fail before the handler runs. Explicit `required` in `throws` is illegal and normalized by the parser, so throws clauses are not affected.

## Regression proof

In a disposable WSL2 source copy, changing the old service fixture from `required string` to default requiredness made the suite fail:

```
TEST FAILURE: service argument required to default is rejected by default: expected exit code 2, got 0
```

Restoring the fixture made it pass.

## Validation

- WSL2 Autotools `make -C test/audit check`: passed (`1 test passed`)
- WSL2 CMake `ThriftAuditTest`: passed
- `perl -c test/audit/thrift_audit_test.pl`: passed
- targeted spelling checks for the new documentation and fixtures: passed
- `git diff --check`: passed
- repository-wide `make style`: blocked by pre-existing codespell findings outside this PR

## Checklist

- [ ] Apache Jira ticket (not created for this contribution)
- [x] Single commit
- [x] No default audit behavior changed; allowances require explicit options
- [x] Test coverage uses standalone Thrift files

## AI assistance


Jens-G:
>  required fields can move to default requiredness when intentionally enabled

But you are aware that this breaks your contract? IOW that is what audit is designed to check.

zsy056:
> > required fields can move to default requiredness when intentionally enabled
> 
> But you are aware that this breaks your contract? IOW that is what audit is designed to check.

Hi Jens

I was using thrift as the internal RPC implementation of a micro-service, for this scenario, the contract is the client API of our micro-service, not the thrift definition itself.

I found audit to be externally helpful to prevent issues on wire protocol incompatibility, it is used as part of a check-in gate. However, over the years, things have changed enough that we have to remove some optional and slowly deprecate some required fields. The current audit mode doesn’t provide the flexibility and we also definitely don’t want to drop audit entirely in the check-in gate.

So I suppose if these relaxations are behind some command line option and is default to off, it should be safe for all existing users while still provide flexibility?

Thanks

Jens-G:
### Code review

Found 3 issues:

1. The new "wire-compatible changes" framing holds for `--audit-allow-optional-field-removal`, but not for `--audit-allow-required-field-to-default`. Dropping `required` is only safe in one rollout direction: a writer built from the new schema may omit the field entirely, while a peer still built from the old schema rejects the message at deserialization.

https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/test/audit/README.md#L8-L18

Reproduced with 0.25.0 on `struct Record { 1: required string value }` -> `struct Record { 1: string value }`: the new-schema Java writer skips the field when unset (`if (struct.value != null)`), while the old-schema reader ends `read()` with `struct.validate()`, which throws `Required field 'value' was not present!`. That check is emitted here, and the equivalent exists in every binding I looked at:

- Java `validate()`: https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/compiler/cpp/src/thrift/generate/t_java_generator.cc#L2235-L2242 (called from `read()` at https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/compiler/cpp/src/thrift/generate/t_java_generator.cc#L5615-L5618)
- C++: https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/compiler/cpp/src/thrift/generate/t_cpp_generator.cc#L1826-L1836
- Python: https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/compiler/cpp/src/thrift/generate/t_py_generator.cc#L1274-L1279

The compatibility validation in the PR description used C++ with an `i32` field, a primitive that is always written, so it exercises the one case that cannot break. Nullable-typed fields are about half of `test/ThriftTest.thrift` (36 of 77, counting only string/binary/containers).

Suggest scoping "wire-compatible" to the first bullet and documenting the required rollout order for the second: every reader must be off `required` before any writer stops setting the field.

2. This needs a JIRA ticket. Two new public compiler flags, changed audit pass/fail semantics, and new build/CI wiring is a significant change, not a trivial one, so `CONTRIBUTING.md` applies: the PR title and commit subject both need the `THRIFT-NNNN:` prefix. The commit's `Client: cpp` and `Generated-by:` trailers are already correct, so amending the subject line is all that is needed.

https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/CONTRIBUTING.md#L15-L26

3. `--audit-allow-required-field-to-default` also relaxes service method argument lists and `throws` clauses, which neither the README nor the `help()` text mentions. Both are compared through `compare_single_struct`, so the new `requiredToDefaultAllowed` gate in `compare_struct_field` applies to them as well. Unlike `optional` — which the parser downgrades to default requiredness inside a service body, warning `optional keyword is ignored in argument lists` — `required` is preserved there, so `void doIt(1: required string arg)` is in scope. The consequence is larger than for a plain struct field: the server throws while reading `doIt_args`, so the entire call fails before the handler runs. None of the new fixtures cover a `service`.

https://github.com/apache/thrift/blob/194dcab82a69101a42e4f91dc4e494728b92febb/compiler/cpp/src/thrift/audit/t_audit.cpp#L360-L365

Jens-G:
What's the issue with those appveyor errors?

Jens-G:
ping

zsy056:
Seems to be error during configuration, let's see if the new run will pass.
```
-- Building with unittests
CMake Error at C:/Program Files/CMake/share/cmake-4.1/Modules/FindFLEX.cmake:295 (message):
  Command "C:/ProgramData/chocolatey/lib/winflexbison3/tools/win_flex.exe
  --version" failed with output:
Call Stack (most recent call first):
  compiler/cpp/CMakeLists.txt:26 (find_package)
-- Found FLEX: C:/ProgramData/chocolatey/lib/winflexbison3/tools/win_flex.exe
CMake Error at C:/Program Files/CMake/share/cmake-4.1/Modules/FindBISON.cmake:245 (message):
  Command "C:/ProgramData/chocolatey/lib/winflexbison3/tools/win_bison.exe
  --version" failed with output:
Call Stack (most recent call first):
  compiler/cpp/CMakeLists.txt:27 (find_package)
-- Found BISON: C:/ProgramData/chocolatey/lib/winflexbison3/tools/win_bison.exe
CMake Error at C:/Program Files/CMake/share/cmake-4.1/Modules/FindFLEX.cmake:295 (message):
  Command "C:/ProgramData/chocolatey/lib/winflexbison3/tools/win_flex.exe
  --version" failed with output:
Call Stack (most recent call first):
  compiler/cpp/tests/CMakeLists.txt:41 (find_package)
CMake Error at C:/Program Files/CMake/share/cmake-4.1/Modules/FindBISON.cmake:245 (message):
  Command "C:/ProgramData/chocolatey/lib/winflexbison3/tools/win_bison.exe
  --version" failed with output:
Call Stack (most recent call first):
  compiler/cpp/tests/CMakeLists.txt:42 (find_package)
```

PR #3733, opened 2026-08-20
Update supported go versions to 1.26+1.27

fishy:
Follow the release of go 1.27.0 today, go 1.25 is no longer supported by
upstream.

Also run `go fix` on the code, which changed all the `-Ptr` usage to
`new`, and update compiler to also generate `//go:fix inline` line above
generated `-Ptr` functions.

Depending on https://github.com/apache/thrift/pull/3734.


test/keys/keygen/make-serverkey.sh
```diff
@@ -48,7 +48,6 @@ echo '' >> my.cfg
 echo '[ alternate_names ]' >> my.cfg
 echo 'IP.1=127.0.0.1' >> my.cfg
 echo 'IP.2=::1' >> my.cfg
-echo 'IP.3=::ffff:127.0.0.1' >> my.cfg
```
fishy: @Jens-G I have to remove this and regenerate the certs, as this is no longer supported by go 1.27 (https://github.com/golang/go/commit/a5b2a19336774bf0f59a0d43eafdb66b3f1c1c5b). With this we can't even get to crosstest.

But after removed this I see all the crosstests with java/kotlin as client failed the ssl ones, so maybe this was added to make java/kotlin clients to pass ssl crosstests?

fishy: Moved that change to https://github.com/apache/thrift/pull/3734 instead.

jimexist (approved):

jimexist:
i wonder if the go mod should use newest rather than the eldest go version

fishy:
> i wonder if the go mod should use newest rather than the eldest go version

that would break people using go 1.26.

dcelasun (approved):

PR #3735, opened 2026-08-20
Add Windows ARM64 wheel distribution

ndabas:
As the title says. It's mostly straightforward.

One of the slightly hacky bits is the change in `lib/py/setup.py` to not include `compat/win32/stdint.h` when building in CI. That shim is not needed anyway for most versions of Visual C++/Build Tools found in the wild today. However I did not get rid of it altogether because someone, somewhere might have a version that does need it. The AMD64 build seems to work fine in any case, but on ARM64 the `arm64_neon.h` file chokes on it. Let me know if there is a better idea to handle this.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -109,6 +109,10 @@ jobs:
             runner: windows-latest
             archs: AMD64
             build: "cp310-win_* cp311-win_* cp312-win_* cp313-win_* cp314-win_*"
+          - platform: windows-arm64
+            runner: windows-11-arm
+            archs: ARM64
+            build: "cp310-win_* cp311-win_* cp312-win_* cp313-win_* cp314-win_*"
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/lib/py/setup.py
+++ b/lib/py/setup.py
@@ -35,7 +35,9 @@ if 'vagrant' in str(os.environ):
 
 include_dirs = ['src']
 if sys.platform == 'win32':
-    include_dirs.append('compat/win32')
+    # Windows ARM64 builds require the real stdint.h.
+    if os.environ.get('CIBUILDWHEEL') != '1':
+        include_dirs.append('compat/win32')
     ext_errors = (CompileError, ExecError, PlatformError, IOError)
 else:
     ext_errors = (CompileError, ExecError, PlatformError)
--- a/lib/py/pyproject.toml
+++ b/lib/py/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=61.0", "wheel"]
+requires = ["setuptools>=69", "wheel"]
 build-backend = "setuptools.build_meta"


doc/ReleaseManagement.md
```diff
@@ -440,7 +440,7 @@ See https://thrift.apache.org/lib/ for the current status of each external packa
     `.github/workflows/pypi.yml` workflow, and the `release` environment.
   * After publishing the GitHub release, verify that the workflow completed and
     that the PyPI release contains both the `.tar.gz` source distribution and
-    the `.whl` Linux manylinux, Linux musllinux, macOS, and Windows wheel
```
fishy, line 443: isn't the old description more "correct" with this change? we don't list archs for linux/macOS wheels either.

ndabas: True, although it does say " Linux manylinux, Linux musllinux" -- so yes, I can just remove this change and it will be consistent with the macOS bit at least.

.github/workflows/pypi.yml
```diff
@@ -109,6 +109,10 @@ jobs:
             runner: windows-latest
             archs: AMD64
             build: "cp310-win_* cp311-win_* cp312-win_* cp313-win_* cp314-win_*"
+          - platform: windows-arm64
+            runner: windows-11-arm
+            archs: ARM64
+            build: "cp310-win_* cp311-win_* cp312-win_* cp313-win_* cp314-win_*"
```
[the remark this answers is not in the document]
ndabas: cibuildwheel does support `win_arm64` builds with the cp310 target, as a binary distribution is available in NuGet, which is where cibuildwheel gets it from.

doc/ReleaseManagement.md
ndabas: Done.

fishy:
the MSVC build on CI failed: https://github.com/apache/thrift/actions/runs/32422316847/job/96597632215

ndabas:
> the MSVC build on CI failed: https://github.com/apache/thrift/actions/runs/32422316847/job/96597632215

That's odd, because I haven't touched any of the C++ sources, and looking at the build logs, the `compat/win32` dir is still included as it was before on the cl command line. Do you have any ideas why this might be?

fishy:
then maybe it's just flaky and you can retry it to see if it works.

ndabas:
> then maybe it's just flaky and you can retry it to see if it works.

I don't see the option to re-run the workflow myself, I have squashed and force pushed the commits, so let's try again when someone is able to approve the workflows. Thanks!

lib/py/setup.py
```diff
@@ -35,7 +35,9 @@
 
 include_dirs = ['src']
 if sys.platform == 'win32':
-    include_dirs.append('compat/win32')
+    # Windows ARM64 builds require the real stdint.h.
+    if os.environ.get('CIBUILDWHEEL') != '1':
+        include_dirs.append('compat/win32')
```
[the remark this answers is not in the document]
ndabas: The Python versions we are building wheels for [need Visual C++ 14 or higher to build](https://wiki.python.org/moin/WindowsCompilers), which includes a proper `stdint.h` -- therefore, the shim included from `compat/win32` is never going to be needed for any (AMD64 or ARM64) builds when compiled from cibuildwheel.

fishy (approved):
looks good to me, but we'll only know whether it works when we actually cut a release next time.

xieofxie:
> looks good to me, but we'll only know whether it works when we actually cut a release next time.

When could be the next release time? From doc https://github.com/apache/thrift/blob/master/doc/ReleaseManagement.md it said twice a year so early next yesr? Thanks

fishy:
it would likely happen in the next month or so.

PR #3736, opened 2026-08-23
Fix TSSLSocket building with LibreSSL

brad0:
LibreSSL does not have the function OPENSSL_thread_stop().

### DIFF
--- a/lib/cpp/src/thrift/transport/TSSLSocket.cpp
+++ b/lib/cpp/src/thrift/transport/TSSLSocket.cpp
@@ -157,7 +157,7 @@ void cleanupOpenSSL() {
 #endif
   EVP_cleanup();
   CRYPTO_cleanup_all_ex_data();
-#if OPENSSL_VERSION_NUMBER >= 0x10100000
+#if OPENSSL_VERSION_NUMBER >= 0x10100000 && !defined(LIBRESSL_VERSION_NUMBER)
   // Do nothing unless an openssl derivative is detected
 #  if !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
   // https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_thread_stop.html
@@ -405,7 +405,7 @@ void TSSLSocket::close() {
     SSL_free(ssl_);
     ssl_ = nullptr;
     handshakeCompleted_ = false;
-#if OPENSSL_VERSION_NUMBER >= 0x10100000
+#if OPENSSL_VERSION_NUMBER >= 0x10100000 && !defined(LIBRESSL_VERSION_NUMBER)
     // Do nothing unless an openssl derivative is detected
 #  if !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
     // https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_thread_stop.html


uros-b (approved):

uros-b:
+1, LGTM

Jens-G:
### Code review

Found 1 issue:

1. Skipping `OPENSSL_thread_stop()` leaves LibreSSL with no per-thread error-state cleanup at all. LibreSSL hardcodes `OPENSSL_VERSION_NUMBER` to `0x20000000L`, so it always takes the `>= 0x10100000` branch and can never reach the `ERR_remove_state(0)` fallback in the `#else`. Excluding it from the inner guard therefore removes the only cleanup call on both paths — the same shape of gap THRIFT-5482 closed in 98be76fc. Unlike BoringSSL and AWS-LC, LibreSSL does ship a working equivalent: `ERR_remove_state()` -> `ERR_remove_thread_state()` -> `err_thread_del_item()` ([lib/libcrypto/err/err.c](https://github.com/openbsd/src/blob/master/lib/libcrypto/err/err.c)), and there is no pthread_key destructor, so the per-thread `ERR_STATE` otherwise stays in the global hash keyed by tid. Substituting rather than skipping would preserve the cleanup:

```c
#if defined(LIBRESSL_VERSION_NUMBER)
    ERR_remove_state(0);
#elif !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
    OPENSSL_thread_stop();
#endif
```

This is clear-cut at the `cleanupOpenSSL()` site. At the `TSSLSocket::close()` site it is more of a judgement call, since `ERR_remove_state(0)` there would also discard error state the calling thread has not read yet — skipping is defensible if that is the intent.

https://github.com/apache/thrift/blob/b1b9a606aa88f5d7045cda4f73ee7eddc412bf54/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L160-L171

https://github.com/apache/thrift/blob/b1b9a606aa88f5d7045cda4f73ee7eddc412bf54/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L408-L419

Everything else checked out: both `OPENSSL_thread_stop()` call sites are covered, the guard is on the inner `#if` (avoiding the `#else` fallthrough caught in #3055), `<openssl/opensslv.h>` is included at line 47 so `LIBRESSL_VERSION_NUMBER` is visible, and `CONF_modules_unload` is correctly left un-guarded since LibreSSL exports it.

brad0:
The original patch although it did build was based on a mistake. After conferring with a LibreSSL developer the updated patch was the intent.

Jens-G:
### Code review

Re-reviewed after the force-push. The revised patch resolves the earlier finding: moving `!defined(LIBRESSL_VERSION_NUMBER)` from the inner guard to the outer one means LibreSSL now falls through to the `#else` and calls `ERR_remove_state(0)`, so it gets real per-thread cleanup instead of none. Verified against current LibreSSL sources: it still exports `ERR_remove_state()` (kept deliberately — `err.h` notes "Still used in 2023"), has never shipped `OPENSSL_thread_stop()`, and `LIBRESSL_VERSION_NUMBER` comes in via the `<openssl/opensslv.h>` already included at line 47. The outer-guard placement also matches the AWS-LC precedent in `61080c673f`, while correctly leaving BoringSSL/AWS-LC on the inner guard, since those two have no `ERR_remove_state()` at all.

Found 2 issues:

1. Merge sequencing: this fix now depends on the `ERR_remove_state(0)` call that #3752 deletes

#3752 removes `ERR_remove_state(0)` from both of these `#else` branches on the grounds that it is a no-op since OpenSSL 1.1 — true for mainline, but after this PR LibreSSL is the branch's main consumer. If #3752 lands after #3736, LibreSSL silently goes back to no cleanup at all, which is exactly the defect the force-push just fixed. Worth sequencing the two, or keeping a LibreSSL carve-out in whichever lands second. (#3761 removes `ERR_remove_state` only from the c_glib file, so it does not collide here.)

https://github.com/apache/thrift/blob/45364c031a18e4f63035a7972bd6351f3eb3ee55/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L159-L172

https://github.com/apache/thrift/blob/45364c031a18e4f63035a7972bd6351f3eb3ee55/lib/cpp/src/thrift/transport/TSSLSocket.cpp#L407-L420

2. Commit message is missing the `Client:` trailer ([AGENTS.md](https://github.com/apache/thrift/blob/45364c031a18e4f63035a7972bd6351f3eb3ee55/AGENTS.md#L33-L38) says "Commit message includes affected `Client:` languages")

The commit is a bare subject line with no body; expected `Client: cpp`.

https://github.com/apache/thrift/blob/45364c031a18e4f63035a7972bd6351f3eb3ee55/AGENTS.md#L33-L38

Jens-G:
I want to add that there is an upcoming conflict between #3736 and #3761. 
Therefore I would like to suggest that you both discuss this and come up with **one** patch.

Jens-G:
Any news here?

brad0:
> Any news here?

My patch was integrated into the other PR and as I said there that PR builds fine with LibreSSL.

Jens-G:
Superseded by #3850, credits retained.

PR #3739, opened 2026-08-23
THRIFT-6159 Use faster varint library in Rust

Leo-Tinkeam:
Changed the Rust varint library from integer-encoding to varint-rs, my [tests](https://github.com/Leo-Tinkeam/bench-rust-varint) on differents library reveal that varint-rs can be up to 2 times faster.

Also tested thrift before and after [here](https://github.com/Leo-Tinkeam/bench-thrift), 20% improvements in release mode (`cargo run --release`) but 60% slower with dev (`cargo run`). This may be an issue but I think that dev time don't matters that much.

The generated output.bin are exactly the same before and after (tested with `cmp master.bin before.bin`).

I did not created a JIRA ticket, I may if it's mandatory.

My usage is not the basic usage of the library, integer-encoding is using `write_all` once on `transport` but varint-rs is doing it for each byte, which is slower with our implementation of `transport.write_all`, I am writing to a buffer with varint-rs in order to use `write_all` only once at the end.

I think that transport should use a fixed size buffer (something like 1ko ?) and the buffer that is up to 10o that I implemented here will became irrelevant. I'm not sure about this last observation but I may investigate if you want me to.

### DIFF
--- a/lib/rs/Cargo.toml
+++ b/lib/rs/Cargo.toml
@@ -14,12 +14,12 @@ keywords = ["thrift"]
 
 [dependencies]
 byteorder = "1.3"
-integer-encoding = "3.0.3"
 uuid = "1"
 log = {version = "0.4", optional = true}
 ordered-float = "3.0"
 threadpool = {version = "1.7", optional = true}
 rustls = { version = "0.23.42", default-features = false, features = ["std", "tls12"], optional = true }
+varint-rs = "2.2.1"
 
 [features]
 default = ["server"]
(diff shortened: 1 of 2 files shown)


Jens-G:
> I did not created a JIRA ticket, I may if it's mandatory.

Pls do that

Leo-Tinkeam:
I asked for a Jira account. I will also fix the formatting error and link the Jira issue once created

Leo-Tinkeam:
JIRA Ticket : https://issues.apache.org/jira/browse/THRIFT-6159

Jens-G:
### Code review

~Found 1 issue: single commit message is missing the required `Client:` line~

EDIT: I think that's still fine, don't worry about that one.

PR #3753, opened 2026-08-28
THRIFT-4623 Fix python complex nested struct includes

thomasbruggink:
Fix the bug where the following:
```
struct C {
    1: i32 value
}

struct A {
    1: C nested = {}
}

struct B {
    1: A itm = {}
}
```
would generate the following invalid python code:
```
class C:
    thrift_spec = None

    def __init__(self, value=None):
        self.value = value

class A:
    thrift_spec = None

    def __init__(self, nested=C()):
        if nested is self.thrift_spec[1][4]:
            nested = C()
        self.nested = nested

class B:
    thrift_spec = None

    def __init__(self, itm=A()):
        if itm is self.thrift_spec[1][4]:
            itm = A()
        self.itm = itm
```
The problem with the above code is that `self.thrift_spec` was initialized and set after the module was loaded. In python this would work when B is instantiated but since B immediately invokes A() it happens before the thrift_spec is set and this would result in `TypeError: 'NoneType' object is not subscriptable`. By first initializing a temporary placeholder with a unique address to compare to called `_THRIFT_DEFAULT` the generated code would look like:
```
_THRIFT_DEFAULT = object()

class C:
    def __init__(self, value=None):
        self.value = value

class A:
    def __init__(self, nested=_THRIFT_DEFAULT):
        if nested is _THRIFT_DEFAULT:
            nested = C()

class B:
    def __init__(self, itm=_THRIFT_DEFAULT):
        if itm is _THRIFT_DEFAULT:
            itm = A()
```
Here the instantiation of the default object is defered until after the module has been fully loaded resulting in the correct value being assigned.
  

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_py_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_py_generator.cc
@@ -452,6 +452,7 @@ void t_py_generator::init_generator() {
   } else {
     f_types_ << "all_structs = []" << '\n';
   }
+  f_types_ << "_THRIFT_DEFAULT = object()" << '\n';
 
   f_consts_ <<
     py_autogen_comment() << '\n' <<
@@ -907,10 +908,9 @@ void t_py_generator::generate_py_struct_definition(ostream& out,
 
     for (m_iter = members.begin(); m_iter != members.end(); ++m_iter) {
       // Initialize fields
-      t_type* type = (*m_iter)->get_type();
+      t_type* type = get_true_type((*m_iter)->get_type());
       if (!type->is_base_type() && !type->is_enum() && (*m_iter)->get_value() != nullptr) {
-        indent(out) << "if " << maybe_escape_identifier((*m_iter)->get_name()) << " is "
-                    << "self.thrift_spec[" << (*m_iter)->get_key() << "][4]:" << '\n';
+        indent(out) << "if " << maybe_escape_identifier((*m_iter)->get_name()) << " is _THRIFT_DEFAULT:" << '\n';
         indent_up();
         indent(out) << maybe_escape_identifier((*m_iter)->get_name()) << " = " << render_field_default_value(*m_iter)
                     << '\n';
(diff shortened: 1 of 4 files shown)


Jens-G:
### Code review

A few suggestions, below the bar for an issue but verified:

- A small formality for when you next push: CONTRIBUTING.md asks for a colon after the ticket id in the commit subject (`THRIFT-4623: Fix python complex nested struct includes`, the PR title can follow suit) and a `Client: py` line below it.

https://github.com/apache/thrift/blob/bb4bf37da9c565e5e5d473d6ffb3562817541a07/CONTRIBUTING.md#L18-L27

- `type` now goes through `get_true_type()`, and the same variable feeds the unchanged immutable branch below it. With `-gen py:enum`, a field typed as a typedef of an enum, in an exception or a `python.immutable` struct, now gets the enum coercion (`x if hasattr(x, 'value') else Color.__members__.get(x)`) instead of a plain assignment, so an int passed there is stored as `None`. Non-typedef enum fields already behave that way, so this may be acceptable, but it is a behaviour change outside the struct-default fix. A separate variable for the new check would avoid it.

https://github.com/apache/thrift/blob/bb4bf37da9c565e5e5d473d6ffb3562817541a07/compiler/cpp/src/thrift/generate/t_py_generator.cc#L910-L926

- With `-gen py:type_hints`, the new default produces signatures like `nested: typing.Optional[C] = _THRIFT_DEFAULT`. A type checker rejects that, because `object()` is neither `C` nor `None`. Runtime behaviour is not affected.

https://github.com/apache/thrift/blob/bb4bf37da9c565e5e5d473d6ffb3562817541a07/compiler/cpp/src/thrift/generate/t_py_generator.cc#L2786-L2794

Jens-G:
@thomasbruggink Coulkd you please have a look? Thanks!

thomasbruggink:
> @thomasbruggink Coulkd you please have a look? Thanks!

@Jens-G i can take a look at the end of next week, currently on vacation without access to a computer :bow: thanks for the review

PR #3778, opened 2026-09-01
THRIFT-2063: Generate Go maps with container keys as entry slices

slachiewicz:
Go maps cannot be keyed by a slice or a map, so the Go generator aborted on any thrift `map` keyed by a `list`, `set`, `map`, or a typedef of one of those. This change generates such fields as `[]thrift.MapEntry[K, V]`, a slice of key/value pairs written to the wire in slice order. As with `set` fields, the generated writer rejects a slice with two equal keys with an `INVALID_DATA` protocol exception; `Equals` compares entries position by position. Maps whose keys were already representable keep their current Go type, so existing generated code is unchanged.

The Go test Makefiles no longer strip `CrazyNesting.list_field` from `ThriftTest.thrift`, and the cross-test client, server and stress binaries build against the full IDL. Struct-keyed maps still generate as `map[*K]V`; changing that is left for a separate ticket because it alters working output.

Verified: `go test ./gopath/src/tests ./gopath/src/dontexportrwtest` in lib/go/test and `go test ./...` in lib/go/thrift → ok; `go build`/`go vet` on all generated test packages → clean.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -1244,13 +1244,29 @@ string t_go_generator::render_const_value(t_type* type, t_const_value* value, co
     t_type* ktype = ((t_map*)type)->get_key_type();
     t_type* vtype = ((t_map*)type)->get_val_type();
     const map<t_const_value*, t_const_value*, t_const_value::value_compare>& val = value->get_map();
+    map<t_const_value*, t_const_value*, t_const_value::value_compare>::const_iterator v_iter;
+    if (is_container_keyed_map(type)) {
+      string entry = map_entry_type((t_map*)type);
+      if (val.empty()) {
+        out << "[]" << entry << "{}";
+        return out.str();
+      }
+      out << "[]" << entry << "{" << '\n';
+      indent_up();
+      for (v_iter = val.begin(); v_iter != val.end(); ++v_iter) {
+        out << indent() << "{Key: " << render_const_value(ktype, v_iter->first, name)
+            << ", Value: " << render_const_value(vtype, v_iter->second, name) << "}," << '\n';
+      }
+      indent_down();
+      out << indent() << "}";
+      return out.str();
+    }
     if (val.empty()) {
       out << "map[" << type_to_go_key_type(ktype) << "]" << type_to_go_type(vtype) << "{}";
       return out.str();
     }
     out << "map[" << type_to_go_key_type(ktype) << "]" << type_to_go_type(vtype) << "{" << '\n';
     indent_up();
-    map<t_const_value*, t_const_value*, t_const_value::value_compare>::const_iterator v_iter;
     size_t max_key_len = 0;
     for (v_iter = val.begin(); v_iter != val.end(); ++v_iter) {
       string rendered_value = render_const_value(vtype, v_iter->second, name);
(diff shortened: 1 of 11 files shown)


fishy:
>Key uniqueness is not enforced, which matches how thrift sets are already represented as slices in Go

this is not true. while go sets don't guarantee uniqueness at wire level, it's checked at write time and duplicates will cause a write failure.

slachiewicz:
Corrected. The generated writer now runs the same pairwise key check that sets use and returns INVALID_DATA on a duplicate key; covered by TestContainerKeyWriteRejectsDuplicateKeys. PR description updated to match.

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -4484,16 +4567,35 @@ string t_go_generator::type_to_enum(t_type* type) {
   throw "INVALID TYPE IN type_to_enum: " + type->get_name();
 }
 
+/**
+ * Returns true when the map cannot be a Go map because its key type is a
+ * container (list, set or map, possibly behind a typedef). Such maps are
+ * represented as a slice of thrift.MapEntry.
+ */
+bool t_go_generator::is_map_entry_slice(t_type* type) {
```
fishy: the name is misleading, we return true not only for slices, but also for maps and sets. call it `is_map_entry_container` or something like that instead?

fishy:
looks good to me besides the `is_map_entry_slice` name. although I'm not sure splitting the generated map types is going to be a good idea longer term. cc @dcelasun for second opinion

slachiewicz:
A real-world IDL that hits this: Apache Hive's `hive_metastore.thrift` (tag `rel/release-4.2.1`) declares `SkewedInfo.skewedColValueLocationMaps` as `map<list<string>, string>`, and the released Go generator aborts on the whole file:

```
[FAILURE:generation:1] Error: Cannot produce a valid type for a Go map key: []string - aborting.
```

A pure-Go metastore client currently has to strip that field from the IDL before generating. With this branch's compiler the pristine IDL generates, and the field becomes `[]thrift.MapEntry[[]string, string]`.

Verified: `thrift -r --gen go -out gen hive_metastore.thrift` (built from this PR's head) → exit 0; `go build ./gen/...` against this branch's `lib/go` → clean.

The only other thing that keeps that IDL from building for Go is unrelated to this change: three fields literally named `isSetQueryParallelism`, `isSetDefaultPoolPath` and `isSetSchedulingPolicy` collide with the generated `IsSetX()` accessors. I will file that separately.

slachiewicz:
Filed the accessor collision as THRIFT-6176.

compiler/cpp/src/thrift/generate/t_go_generator.cc
slachiewicz: Renamed to `is_container_keyed_map`: it answers whether the map's key is a list, set or map, which is what the call sites branch on. Pushed as a separate commit (f07ab0b).

slachiewicz:
On the two-representation point: Go has no hashable form for a list, set or map key, so the only single representation would be to emit every map as an entry slice, which changes the type of every existing generated field. The per-key-type split keeps generated code for today's IDLs byte-identical and is confined to one predicate in the generator. There is precedent for the Go representation following what Go can express rather than the Thrift kind: `set` has always generated as a slice while `map` generates as a map. If a uniform slice representation is wanted later, it fits as a generator option layered on the same entry type without changing this default. Happy to go either way if you and @dcelasun prefer the option now.

lib/go/thrift/map_entry.go
```diff
@@ -0,0 +1,32 @@
...
+
+package thrift
+
+// MapEntry is one key/value pair of a Thrift map whose key type cannot be a
+// Go map key: list, set and map keys, and typedefs resolving to one of those.
+//
+// The generator represents such a field as []MapEntry[K, V]. Entries are
+// written to the wire in slice order. Uniqueness of keys is not enforced on
+// read or write, which mirrors how this library represents Thrift sets as
+// slices.
```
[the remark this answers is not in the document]
slachiewicz: Updated in 5fb0eeb: the comment now states the INVALID_DATA write check and that reads do not check duplicates.

slachiewicz:
Copilot's second finding was real and pre-existing: the generated map `Equals` skipped the `ok` check, so maps with different key sets compared equal when the zero value matched, and the new write check then rejected map-keyed entries such as `{a:0}` and `{b:0}`. Fixed in 14165b1 with tests for both cases.

fishy (approved):

fishy:
I don't think an uniformed slice implementation for maps is better, it's definitely better to use builtin map whenever we can :sweat_smile:

PR #3781, opened 2026-09-02
THRIFT-6178: Read a WebSocket payload to the end in the C++ library

Jens-G:
> **Stacked on #3780 (THRIFT-6177).** Only the second commit belongs to this ticket. Merge #3780 first and this rebases to a single commit.

`TWebSocketServer::readFrame` reads a frame's payload with a single call to the transport underneath and treats anything short of the whole payload as end of stream:

```cpp
read = transport_->read(buffer, length);
readBuffer_.wroteBytes(read);
if (read < length) {
  return false;
}
```

`TSocket::read` performs one `recv()` and returns whatever that produced, and the transport factory hands `TWebSocketServer` the accepted `TSocket` itself, so the payload arrives in as many pieces as the network chose. Returning `false` makes `readAll_virt` return 0, which the caller reads as a closed connection.

So no frame whose payload does not arrive in one read can be received at all. Demonstrated with a `TServerSocket`, a `TWebSocketServer` over the accepted socket, and a client writing a well-formed 4000-byte frame as 2000 + 2000 bytes with a pause between them:

```
client: payload 4000 bytes, sent as 2000 + 2000 with 300 ms between
server: readAll returned 0 of 4000
```

and with this change, over the same socket:

```
server: readAll returned 4000 of 4000
server: payload intact: yes
```

In practice this is any frame past the path MTU, so the transport could not carry a Thrift message beyond roughly 1.4 kB over an ordinary Ethernet path.

### The fix

Read the payload with `readAll()`, which is how `TFramedTransport::readFrame` reads its own frames, and keep the existing signal for a peer that leaves in the middle of one: `readAll` raises `END_OF_FILE`, which is caught and turned back into the same `return false` the caller sees today. Nothing else about the contract changes.

### Why it is stacked

`readAll()` waits for the number of bytes the frame declared. That is only safe once that number is bounded, which is what #3780 does — without it, a header declaring `0xFFFFFFFF` would make the server wait for 4 GB rather than return. With it, the wait is bounded by `TConfiguration::maxFrameSize` and by the socket's own receive timeout, the same two limits that bound `TFramedTransport`.

### Tests

Two more cases in `lib/cpp/test/TWebSocketServerTest.cpp`:

- a frame delivered as 8 + 1000 + the rest arrives whole — fails before the change with `0 != 4000`;
- a frame whose payload stops after 500 bytes is still reported as end of stream rather than handed over half-read — passes either way, and is there so that waiting for the rest of a frame cannot quietly become handing a half-read one to the protocol.

Full `bin/UnitTests`: 108 of 109. The one failure, `TServerSocketTest/test_bind_to_address`, is a pre-existing environment failure on this host and is present on `master`.

### DIFF
--- a/lib/cpp/src/thrift/transport/TWebSocketServer.h
+++ b/lib/cpp/src/thrift/transport/TWebSocketServer.h
@@ -270,8 +270,17 @@ private:
       }
     }
 
-    // size_t is smaller than a ulong on a 32-bit system
-    if (payloadLength > UINT32_MAX) {
+    // The read buffer below is sized from this number before a single payload
+    // byte has arrived, so the frame header on its own decides how much memory
+    // the server commits unless something bounds it. Hold it to the configured
+    // maximum frame size, the same ceiling TFramedTransport applies to its own
+    // frames. The UINT32_MAX arm stays for its own reason: size_t is narrower
+    // than the length field on a 32-bit system.
+    //
+    // The cast is safe because the first arm has already ruled out anything
+    // wider than 32 bits by the time the second is evaluated.
+    if (payloadLength > UINT32_MAX
+        || static_cast<int64_t>(payloadLength) > getConfiguration()->getMaxFrameSize()) {
       failConnection(CloseCode::MessageTooBig);
       return false;
     }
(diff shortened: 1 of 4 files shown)


Jens-G:
Superseded by the rebase-merge of #3782, which carried this commit into master along with the rest of the stack. The change is on master as 3d98d7214, byte-identical to the commit here, so there is nothing left to merge.

PR #3784, opened 2026-09-02
THRIFT-6181: Bound the response frame size in the Java async client

Jens-G:
`TAsyncMethodCall` sizes the response frame from the four bytes the peer sends ahead of it, with no check of any kind:

```java
if (sizeBuffer.remaining() == 0) {
  state = State.READING_RESPONSE_BODY;
  frameBuffer = ByteBuffer.allocate(TFramedTransport.decodeFrameSize(sizeBufferArray));
}
```

`decodeFrameSize` returns a signed `int`, so the peer picks anything from `-2147483648` to `2147483647` and `ByteBuffer.allocate` is handed it. A negative one raises `IllegalArgumentException` — not a `TTransportException` — which reaches the callback as it is; a large positive one is allocated, and `ByteBuffer.allocate` zeroes the array, so the memory is resident rather than reserved.

### It is the odd one out

`lib/java` reads a frame size in four places. Three of them are guarded:

| | negative | over `maxFrameSize` | other |
|---|---|---|---|
| `TFramedTransport.readFrame` | ✅ | ✅ | |
| `TFastFramedTransport.readFrame` | ✅ | ✅ | |
| `AbstractNonblockingServer.FrameBuffer` | ✅ (`<= 0`) | ✅ | also defers past `MAX_READ_BUFFER_BYTES` |
| `TAsyncMethodCall.doReadingResponseSize` | ❌ | ❌ | |

This applies the same two checks, with the same messages, so the two ways of reading a framed response agree.

### Compatibility — worth a release note

`maxFrameSize` defaults to 16384000, so an async client now refuses a response frame larger than that where it used to allocate it. `TNonblockingTransport.setMaxFrameSize` raises the bound with it.

### Tests

`TestTAsyncMethodCallFrameSize` drives a real `TAsyncMethodCall` against a socket that answers with a frame size and no body. The assertions are about the buffer that was allocated, not merely that the call failed — a response body that never arrives ends the call either way, so *"did it fail?"* passes on the unmodified library too.

Two fail before the change:

```
allocated 65536 for a frame that is over the maximum
a negative size reached ByteBuffer.allocate: java.lang.IllegalArgumentException: capacity < 0: (-1 < 0)
```

The third reads a frame of exactly the maximum end to end and passes either way.

Full `gradle test spotlessCheck`: BUILD SUCCESSFUL.

### DIFF
--- a/lib/java/src/main/java/org/apache/thrift/async/TAsyncMethodCall.java
+++ b/lib/java/src/main/java/org/apache/thrift/async/TAsyncMethodCall.java
@@ -254,8 +254,25 @@ public abstract class TAsyncMethodCall<T> {
       throw new TTransportException(TTransportException.END_OF_FILE, "Read call frame size failed");
     }
     if (sizeBuffer.remaining() == 0) {
+      // The peer chose this number, and it sizes the buffer below before any of
+      // the response has arrived. Refuse what TFramedTransport.readFrame refuses,
+      // so that the two ways of reading a framed response agree.
+      int size = TFramedTransport.decodeFrameSize(sizeBufferArray);
+      if (size < 0) {
+        throw new TTransportException(
+            TTransportException.CORRUPTED_DATA, "Read a negative frame size (" + size + ")!");
+      }
+      if (size > transport.getMaxFrameSize()) {
+        throw new TTransportException(
+            TTransportException.CORRUPTED_DATA,
+            "Frame size ("
+                + size
+                + ") larger than max length ("
+                + transport.getMaxFrameSize()
+                + ")!");
+      }
       state = State.READING_RESPONSE_BODY;
-      frameBuffer = ByteBuffer.allocate(TFramedTransport.decodeFrameSize(sizeBufferArray));
+      frameBuffer = ByteBuffer.allocate(size);
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/java/src/main/java/org/apache/thrift/async/TAsyncMethodCall.java
```diff
@@ -254,8 +254,25 @@ private void doReadingResponseSize() throws TTransportException {
...
+      // the response has arrived. Refuse what TFramedTransport.readFrame refuses,
+      // so that the two ways of reading a framed response agree.
+      int size = TFramedTransport.decodeFrameSize(sizeBufferArray);
+      if (size < 0) {
+        throw new TTransportException(
+            TTransportException.CORRUPTED_DATA, "Read a negative frame size (" + size + ")!");
+      }
+      if (size > transport.getMaxFrameSize()) {
+        throw new TTransportException(
+            TTransportException.CORRUPTED_DATA,
```
fishy, line 267: is `CORRUPTED_DATA` newly added recently? in Go's implementation we are using `UNKNOWN_TRANSPORT_EXCEPTION` for this one (and there's no `CORRUPTED_DATA` defined)

fishy (approved):

lib/java/src/main/java/org/apache/thrift/async/TAsyncMethodCall.java
Jens-G: CORRUPTED_DATA was added in fe5330955f6e52c63ed76819e4b36b9f263a9218 over 10 years ago

PR #3786, opened 2026-09-02
THRIFT-4244: Explain literal backslash escaping in parser errors

sveneld:
## Summary

An unknown escape such as `\T` in an IDL string currently produces an unhelpful "Bad escape character" error. Keep unknown escapes invalid and report the offending escape with the hint `Use \\ for a literal backslash`. Authors can write `"Y-m-d\\TH:i:s.uP"` to obtain the intended format string.

Add a committed fixture covering literal backslashes and all six supported escapes in both quote styles. Negative lexer tests cover unknown T, x, 0 and u escapes and check the diagnostic. The test parser preserves diagnostic text in exceptions, and the negative tests clean up scanner state.

Distinguish a trailing backslash at EOF from a literal NUL using Flex's yywrap callback. Render non-printable and non-ASCII escape bytes in hexadecimal instead of writing raw bytes into diagnostics. Cover EOF, LF, CRLF, NUL, control bytes and UTF-8 input in both quote styles, using memory and file input and asserting complete error messages. The test error handler validates both formatting passes, writes diagnostics with explicit lengths and avoids duplicate newlines.

The lexer is shared by all language bindings; no generator changes are needed.

## Testing

- Test-first: the negative test failed against the previous PR behavior because no exception was thrown.
- Test-first: the trailing-backslash case reproduced the incorrect and truncated diagnostic before the fix.
- Test-first: literal NUL input reproduced the incorrect EOF diagnostic before the yywrap change.
- Full standalone compiler suite in Docker: 20 test cases, 314 assertions passed, also with randomized order (seed 4244).
- A standalone probe linked against the test parser verified embedded NUL preservation and single-newline output.
- git diff --check passed.
- make style attempted; blocked by the existing missing lib/swift/Makefile.am in the local build configuration.

## CI Investigation

On head 734522b5c, compiler and PHP jobs passed. Three jobs failed while downloading artifacts with HTTP 403, before their tests ran. The two failing Python cross-test jobs logged client certificate address mismatches for ::1 and ::ffff:127.0.0.1. These failures occur outside the lexer changes; no Python TLS behavior is changed by this PR. Results from the updated head are pending.

Generated-by: OpenAI Codex GPT-6

### DIFF
--- a/compiler/cpp/src/thrift/thriftl.ll
+++ b/compiler/cpp/src/thrift/thriftl.ll
@@ -66,6 +66,8 @@
 #include "thrift/globals.h"
 #include "thrift/parse/t_program.h"
 
+static bool yyinput_reached_eof = false;
+
 /**
  * Must be included AFTER parse/t_program.h, but I can't remember why anymore
  * because I wrote this a while ago.
@@ -98,11 +100,6 @@ void error_no_longer_supported(const char* text, const char* replace_with) {
  * Provides the yylineno global, useful for debugging output
  */
 %option lex-compat
-
-/**
- * Our inputs are all single files, so no need for yywrap
- */
-%option noyywrap
 
 /**
  * We don't use it, and it fires up warnings at -Wall
@@ -287,17 +284,33 @@ literal_begin (['\"])
   std::string result;
   for(;;)
   {
+    yyinput_reached_eof = false;
     int ch = yyinput();
     switch (ch) {
-      case EOF:
-        yyerror("End of file while read string at %d\n", yylineno);
-        exit(1);
+      case 0:
+        if (yyinput_reached_eof) {
+          yyerror("End of file while reading string at %d\n", yylineno);
+          exit(1);
+        }
+        result.push_back('\0');
+        continue;
       case '\n':
-        yyerror("End of line while read string at %d\n", yylineno - 1);
+        yyerror("End of line while reading string at %d\n", yylineno - 1);
         exit(1);
       case '\\':
+        yyinput_reached_eof = false;
         ch = yyinput();
         switch (ch) {
+          case 0:
+            if (yyinput_reached_eof) {
+              yyerror("End of file while reading string at %d\n", yylineno);
+              exit(1);
+            }
+            yyerror("Invalid escape byte 0x00. Use \\\\ for a literal backslash.\n");
+            return -1;
+          case '\n':
+            yyerror("End of line while reading string at %d\n", yylineno - 1);
+            exit(1);
           case 'r':
             result.push_back('\r');
             continue;
@@ -317,7 +330,12 @@ literal_begin (['\"])
             result.push_back('\\');
             continue;
           default:
-            yyerror("Bad escape character\n");
+            if (ch >= 0x20 && ch <= 0x7e) {
+              yyerror("Invalid escape sequence '\\%c'. Use \\\\ for a literal backslash.\n", ch);
+            } else {
+              yyerror("Invalid escape byte 0x%02X. Use \\\\ for a literal backslash.\n",
+                      static_cast<unsigned int>(ch));
+            }
             return -1;
         }
         break;
@@ -338,6 +356,12 @@ literal_begin (['\"])
 }
 
 %%
+
+// A zero from yyinput() can also be a literal NUL; yywrap marks actual EOF.
+int yywrap(void) {
+  yyinput_reached_eof = true;
+  return 1;
+}
 
 #ifdef _MSC_VER
 #pragma warning( pop )
--- a/compiler/cpp/tests/cpp/t_cpp_parser_string_constant_tests.cc
+++ b/compiler/cpp/tests/cpp/t_cpp_parser_string_constant_tests.cc
@@ -1,0 +1,144 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "t_cpp_generator_test_utils.h"
+
+#include <cstdio>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+struct yy_buffer_state;
+extern yy_buffer_state* yy_scan_string(const char*);
+extern yy_buffer_state* yy_scan_bytes(const char*, int);
+extern yy_buffer_state* yy_create_buffer(FILE*, int);
+extern void yy_switch_to_buffer(yy_buffer_state*);
+extern int yylex_destroy();
+
+using cpp_generator_test_utils::join_path;
+using cpp_generator_test_utils::parse_thrift_for_test;
+using cpp_generator_test_utils::source_dir;
+
+TEST_CASE("parser preserves supported string escapes", "[parser]")
+{
+    const std::string path = join_path(source_dir(), "test_string_escapes.thrift");
+    std::unique_ptr<t_program> program(new t_program(path, "test_string_escapes"));
+    parse_thrift_for_test(program.get());
+
+    const std::vector<std::pair<std::string, std::string>> expected = {
+        {"DOUBLE_QUOTED", "Y-m-d\\TH:i:s.uP"},
+        {"SINGLE_QUOTED", "Y-m-d\\TH:i:s.uP"},
+        {"ALL_DOUBLE", "\r\n\t\"'\\"},
+        {"ALL_SINGLE", "\r\n\t\"'\\"}
+    };
+    REQUIRE(program->get_consts().size() == expected.size());
+    const t_scope& scope = *program->scope();
+    for (const auto& entry : expected) {
+        INFO(entry.first);
+        const t_const* constant = scope.get_constant(entry.first);
+        REQUIRE(constant != nullptr);
+        REQUIRE(constant->get_value()->get_type() == t_const_value::CV_STRING);
+        REQUIRE(constant->get_value()->get_string() == entry.second);
+    }
+}
+
+TEST_CASE("lexer explains how to escape a literal backslash", "[parser]")
+{
+    // Clear scanner state left by earlier parser tests.
+    yylex_destroy();
+    for (const std::string& escape : {"T", "x41", "0", "u00e9"}) {
+        for (const char quote : {'"', '\''}) {
+            const std::string source = std::string(1, quote) + "\\" + escape + quote;
+            INFO(source);
+            // Reset scanner state even when the expected parser exception is thrown.
+            auto cleanup = [](yy_buffer_state*) { yylex_destroy(); };
+            std::unique_ptr<yy_buffer_state, decltype(cleanup)> buffer(
+                yy_scan_string(source.c_str()), cleanup);
+            const std::string expected = "Invalid escape sequence '\\" + escape.substr(0, 1)
+                + "'. Use \\\\ for a literal backslash.\n";
+            REQUIRE_THROWS_WITH(yylex(), expected);
+        }
+    }
+}
+
+TEST_CASE("lexer diagnoses incomplete and non-printable escapes", "[parser]")
+{
+    yylex_destroy();
+    const std::vector<std::pair<std::string, std::string>> cases = {
+        {"", "End of file while reading string at 1\n"},
+        {"\n", "End of line while reading string at 1\n"},
+        {"\r\n", "Invalid escape byte 0x0D. Use \\\\ for a literal backslash.\n"},
+        {"\t", "Invalid escape byte 0x09. Use \\\\ for a literal backslash.\n"},
+        {"\x01", "Invalid escape byte 0x01. Use \\\\ for a literal backslash.\n"},
+        {"\x7f", "Invalid escape byte 0x7F. Use \\\\ for a literal backslash.\n"},
+        {"\xc3\xa9", "Invalid escape byte 0xC3. Use \\\\ for a literal backslash.\n"},
+        {"\xff", "Invalid escape byte 0xFF. Use \\\\ for a literal backslash.\n"},
+        {std::string(1, '\0'), "Invalid escape byte 0x00. Use \\\\ for a literal backslash.\n"},
+        {std::string("\0tail", 5), "Invalid escape byte 0x00. Use \\\\ for a literal backslash.\n"}
+    };
+    for (const auto& test : cases) {
+        for (const char quote : {'"', '\''}) {
+            const std::string source = std::string(1, quote) + "abc\\" + test.first;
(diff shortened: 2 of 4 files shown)


Jens-G:
### Code review

Reviewed at `97efe4d3d`. Built compilers from the PR head and from the merge base (`59a949c80`) and compared the generated constant across 24 generators. The compiler test suite is 17/17 green on the base and 18/18 on the PR head.

Two things first, since the ticket leaves them open.

**The existing escapes are not affected.** `\r \n \t \" \' \\` all still convert, and every input that parsed before parses to the same bytes: for a constant exercising all six, the output of all 24 generators is byte-identical between the two compilers. The change reaches only input that used to be a hard error.

**`\\` already does this today, with the same result.** On the merge base, `const string ISO8601U = "Y-m-d\\TH:i:s.uP"` compiles, and `gen-php/Example/Constant.php` comes out byte-identical to what this PR produces for the ticket's unescaped spelling — `return "Y-m-d\\TH:i:s.uP";`, which PHP evaluates to `Y-m-d\TH:i:s.uP`, exactly the format string the reporter wanted. The same holds for the other 23 generators. So the capability is not missing; the PR removes the need to double the backslash. That is worth having, but it is an ergonomics change rather than a bug fix, and it has costs `\\` does not:

- **It burns the escape namespace.** After this, `"\x41"`, `"\0"` and `"\u00e9"` have a defined meaning — the literal characters. Adding `\xNN`, `\0` or `\uXXXX` later becomes a silent behaviour change for anyone who wrote them expecting pass-through.
- **It makes the escape set silently context-dependent.** `"C:\dir\file.txt"` now yields `C:\dir\file.txt`, but `"C:\new\table.txt"` yields `C:<LF>ew<TAB>able.txt` — same shape, opposite meaning, no diagnostic either way. Before, the first was rejected, and that rejection was the signal that the string needed escaping.

Found 3 issues:

1. Four generators cannot carry a literal backslash, so the value the parser now accepts does not survive code generation. Against the PR compiler, with the ticket's own constant:

   - **rust** — `pub const ISO8601U: &str = "Y-m-d\TH:i:s.uP";` → `error: unknown character escape: 'T'`. The generated crate does not compile.
   - **c_glib** — `#define ISO8601U g_strdup ("Y-m-d\TH:i:s.uP")` → gcc `warning: unknown escape sequence: '\T'`, and the value silently becomes `Y-m-dTH:i:s.uP`. The backslash is gone.
   - **kotlin** — `const val ISO8601U: kotlin.String = "Y-m-d\TH:i:s.uP"`, the same shape as rust. No Kotlin toolchain here, so unlike the two above this one is by inspection, not compiled.
   - **st** — emits `"Y-m-d\\TH:i:s.uP"` into a Smalltalk literal, which has no backslash escapes, so the value gains a second backslash.

   All four are reachable on master today by writing `\\`, so these are pre-existing generator bugs rather than regressions. They matter here because the point of the change is to make that input natural to write: after it, the ticket's literal example silently produces a Rust crate that does not compile and a C constant with the wrong value.

https://github.com/apache/thrift/blob/97efe4d3db67637e867bbe3c573d02d06ee5ce2b/compiler/cpp/src/thrift/thriftl.ll#L324-L329

2. This changes IDL string-literal semantics with no spec change. [`doc/specs/idl.md`](https://github.com/apache/thrift/blob/97efe4d3db67637e867bbe3c573d02d06ee5ce2b/doc/specs/idl.md#L186-L190) defines `Literal ::= ('"' [^"]* '"') | ("'" [^']* "'")` and documents no escape mechanism at all — so neither the six escapes the lexer already honours nor the new "everything else is literal" rule is written down anywhere. Since the reporter's argument in THRIFT-4244 is precisely a reading of that grammar, the grammar is the thing that needs to say what the answer is.

3. The regression test asserts the new behaviour but not the behaviour most at risk from editing this switch. There is no assertion that `\n`, `\t`, `\r`, `\"`, `\'` and `\\` still convert — a one-line constant plus one `REQUIRE` would pin all six, and it is the only guard against a future edit to the same `default:` arm swallowing them.

https://github.com/apache/thrift/blob/97efe4d3db67637e867bbe3c573d02d06ee5ce2b/compiler/cpp/tests/cpp/t_cpp_parser_string_constant_tests.cc#L43-L49

Three suggestions, below the bar for the list above but verified:

- The test writes its fixture into the process CWD, unlike every sibling in that directory, which loads a committed `.thrift` via `source_dir()` + `join_path()` (`test_enum_class.thrift`, `test_forward_setter.thrift`, …). It also leaks the file whenever an assertion fails, since `std::remove()` sits after the `REQUIRE`s and Catch unwinds past it.

https://github.com/apache/thrift/blob/97efe4d3db67637e867bbe3c573d02d06ee5ce2b/compiler/cpp/tests/cpp/t_cpp_parser_string_constant_tests.cc#L29-L42

- `namespace php Example` in the fixture makes the suite print `No generator named 'php' could be found!` on every run — the standalone test binary registers cpp/netstd/ocaml/rb only — and nothing in the test depends on it.

- `Client: cpp,php` understates the blast radius: this is a lexer change, so it affects every binding, and the trailer feeds the CHANGES generator. Also, `CONTRIBUTING.md` asks for a single squashed commit and there are currently two.

The `case EOF` / `case '\n'` additions inside the escape are a straight improvement — `"abc\` at end of file used to fall into `default:` and report `Bad escape character`.

Jens-G:
> `\\` already does this today, with the same result. On the merge base, const string ISO8601U = `Y-m-d\\TH:i:s.uP` compiles, and gen-php/Example/Constant.php comes out byte-identical to what this PR produces for the ticket's unescaped spelling — return `Y-m-d\\TH:i:s.uP`, which PHP evaluates to `Y-m-d\TH:i:s.uP`, exactly the format string the reporter wanted. The same holds for the other 23 generators. So the capability is not missing; the PR removes the need to double the backslash. That is worth having, but it is an ergonomics change rather than a bug fix, and it has costs `\\` does not:
> 
> It burns the escape namespace. After this, "\x41", "\0" and "\u00e9" have a defined meaning — the literal characters. Adding \xNN, \0 or \uXXXX later becomes a silent behaviour change for anyone who wrote them expecting pass-through.
>
> It makes the escape set silently context-dependent. "C:\dir\file.txt" now yields C:\dir\file.txt, but "C:\new\table.txt" yields C:<LF>ew<TAB>able.txt — same shape, opposite meaning, no diagnostic either way. Before, the first was rejected, and that rejection was the signal that the string needed escaping.

That was exactly my first thought when I read the ticket, so I instructed Claude to verify that hunch and back either result up. 

FWIW I am not really happy with that change. Maybe we can improve the error message to lead the user to the real solution, bu´t other than that I see not much room for improvement here. 

What do you say? @sveneld

sveneld:
@Jens-G  I am agree with your proposal, my fix was not good. I updated code and description. Take a look

Jens-G:
### Code review

Re-reviewed at `7dcc58ca0`. Built the compiler and the standalone test suite from the PR head and from `da18006b9`, and compared behaviour on crafted inputs.

The rewrite closes everything actionable from the last round: the fixture is a committed `.thrift` loaded through `source_dir()`/`join_path()`, the `php` namespace is gone, all six supported escapes are pinned by `ALL_DOUBLE`/`ALL_SINGLE`, and since accepted input is unchanged the `doc/specs/idl.md` question is moot. The suite goes from 17 cases / 216 assertions green on the base to 19 / 234 on the head, order-independent (`--order rand`, six seeds) and clean under ASan/UBSan. The ticket's escape now reports what it should: `Invalid escape sequence '\T'. Use \\ for a literal backslash.`

Found 2 issues:

1. The new `case EOF:` in the escape switch is unreachable, so the end-of-file case it was added for still lands in `default:`. flex's `yyinput()` returns `0` at end of input — the generated scanner does `if (yywrap()) return 0;`, and `%option noyywrap` (line 105) makes `yywrap()` the constant `1` — and every other return path is `c = *(unsigned char *)(yy_c_buf_p)`, so the value is always 0..255 and never `EOF` (-1). A file ending `const string A = "abc\` prints, at the PR head, `Invalid escape sequence '\^@'. Use \\ for a literal backslash.` — `^@` being a NUL byte written straight to stderr by the `%c`. `case '\n':` two lines down does work: `\` before a newline correctly reports `End of line while reading string`. `case 0:` would make both the new branch and the identical, equally dead outer one at line 292 live. (For context, the outer one being dead is why `const string A = "abc` with no closing quote spins forever in the `for(;;)` pushing NUL bytes — I timed it out at 8s on both `da18006b9` and the head, so that part is pre-existing and not this PR's to fix.)

https://github.com/apache/thrift/blob/7dcc58ca09d52b8aec91cd9d990e6971920eae6e/compiler/cpp/src/thrift/thriftl.ll#L299-L307

2. The rewritten test-harness `yyerror()` truncates its own message at the first embedded NUL, which drops exactly the hint this PR adds. `message` is a `std::vector<char>` correctly sized by the two-pass `vsnprintf`, but line 60 hands `message.data()` to `%s` and line 63 to `std::runtime_error(const char*)`, and both stop at the first NUL. Together with issue 1, the backslash-at-EOF case formats `%c` with `0`, so `what()` comes back as the 26 bytes `Invalid escape sequence '\` and the `Use \\ for a literal backslash.` clause is gone — verified by calling `yylex()` on `yy_scan_string("\"abc\\")` against a build of the head. `main.cc`'s real `yyerror()` is unaffected, since it `vfprintf`s straight off the va_list; this is specific to the harness. The new test cannot see it: the loop only probes `{"T","x41","0","u00e9"}`, all of which put a printable byte after the backslash, and it asserts on the hint substring only, never on the `'\%c'` part.

https://github.com/apache/thrift/blob/7dcc58ca09d52b8aec91cd9d990e6971920eae6e/compiler/cpp/tests/thrift_test_parser_support.cc#L57-L64

Three suggestions, below the bar for the list above but verified:

- `%c` writes one raw byte, so the other two non-printables land in the message as-is. On a CRLF file, `\` before the line break gives `Invalid escape sequence '` + CR, which returns the cursor to column 0 and overwrites the rest of the line on a terminal; a multi-byte character gives a lone continuation byte (`"ab\éc"` → `Invalid escape sequence '\M-C'`, not valid UTF-8). Escaping non-printables would cover those and the NUL in one go.

- Still 4 commits, up from 2 last time, and `34ac5e314`/`97efe4d3d` still carry `Client: cpp,php` while the two newer ones carry the correct `Client: compiler`. `build/generate-changes.py` reads that trailer per commit and a default squash-merge concatenates all four messages, so as it stands this would also be filed under C++ and PHP. `34ac5e314`'s subject also states the opposite of what now ships.

- `extern int yylex();` at line 27 of the new test duplicates the `extern "C"` declaration already visible from `main.h` through `t_cpp_generator_test_utils.h`.

Jens-G:
### Code review

Re-reviewed at `09c89d9cf`. Built the compiler and the standalone test suite from the PR head and from the merge base (`59a949c80`), compared behaviour on crafted inputs, and probed flex 2.6.4 directly with the scanner options this file uses.

The rewrite does what the last round asked for, and it fixes one thing more than the ticket needed:

- Unknown escapes stay invalid, and the diagnostic now names the offender and the way out: `"a\Tb"` → `Invalid escape sequence '\T'. Use \\ for a literal backslash.`
- The dead `case EOF:` in the *escape* switch is genuinely fixed. `const string B = "abc\` at end of file reports `Bad escape character` on the base and `End of file while reading string at 1` at the head.
- Dropping `%option noyywrap` is behaviour-neutral: flex expands that option to `#define yywrap() 1`, so the new `yywrap()` returns exactly what the macro returned. It is defined in the same file, so no `-lfl` dependency and no checked-in generated scanner to go stale.
- The fixture and the new test file need no build-file changes — `file(GLOB ${name}/*.c* ${name}/*.thrift)` in `tests/CMakeLists.txt` picks them up, and `EXTRA_DIST = … tests` covers `make dist`.
- Accepted input is unchanged: `tutorial.thrift` through all 25 generators is byte-identical between the base compiler and the PR head.

Found 1 issue.

**The *outer* `case EOF:` (line 289) is still unreachable, and the loop it guards never terminates.**

`yyinput()` returns `0` at end of input, never `-1` — I measured that with a standalone flex 2.6.4 scanner using this file's options, with and without `noyywrap`, and it keeps returning `0` on every subsequent call. So an unterminated string literal at end of file falls through to `default:`, pushes a NUL into `result`, and comes back for more. Two manifestations of the one defect:

| input | base `59a949c80` | head `09c89d9cf` |
|---|---|---|
| `const string A = "abc` at EOF, from a file | hangs (10s timeout) | hangs (10s timeout) |
| the same source via `yy_scan_bytes` | — | SIGSEGV |

The crash is the memory-buffer path: with no `yyin`, the next `yyinput()` reaches `yy_get_next_buffer()`, which calls `fread` on a null `FILE*`.

```
#0  __GI__IO_fread (buf=..., size=1, count=3, fp=0x0) at ./libio/iofread.c:37
#1  yy_get_next_buffer () at thriftl.cc:1708
#2  yyinput () at thriftl.cc:1841
#3  yylex () at thriftl.ll:287
```

That second row matters for this PR specifically: the new tests drive the lexer through `yy_scan_bytes`, so the first person who adds a case for an unterminated string takes the whole suite down with it.

This is pre-existing — the base hangs identically, so it is not a regression from this PR. But the PR edits that exact line (`while read` → `while reading`, on a message that can never print), and `yyinput_reached_eof` is already there to tell the two cases apart. Four lines finish it:

```diff
   for(;;)
   {
+    yyinput_reached_eof = false;
     int ch = yyinput();
     switch (ch) {
-      case EOF:
-        yyerror("End of file while reading string at %d\n", yylineno);
-        exit(1);
+      case 0:
+        if (yyinput_reached_eof) {
+          yyerror("End of file while reading string at %d\n", yylineno);
+          exit(1);
+        }
+        result.push_back('\0');
+        continue;
       case '\n':
```

`result.push_back('\0')` keeps today's behaviour for a literal NUL inside a *closed* string — it lands in `result` and `strdup` truncates there, exactly as now.

I have pushed that onto this branch as f8d01479b, together with two tests (an unterminated string at end of input, and a literal NUL inside a closed string). With it applied, `const string A = "abc` reports `End of file while reading string at 1` and exits 1, `const string N = "a<NUL>b"` still compiles unchanged, `tutorial.thrift` through all 25 generators stays byte-identical, and the suite is 22 cases / 326 assertions green, also under `--order rand`. Both tests fail (SIGSEGV) without the lexer change. Please squash it into your commit before this merges — and do shout if you would rather have it out again.

Note the same dead `case EOF:` sits in the doc-comment block (line 139) and the multiline-comment block (line 182): `/* x` and `/** x` at end of file hang the compiler as well. Those are untouched by this PR and belong in their own ticket.

### Copilot's open comments

Of the eight threads still marked current, one was worth acting on, and it is the finding above — reported twice, at `thriftl.ll:295` and `:305`, and now addressed by f8d01479b. The rest:

- `parser_support.cc:68` (unchecked second `vsnprintf`, double newline) and `:54` (prefix printed before formatting is known to succeed) — both already fixed in this head; Copilot is re-reporting its own earlier round.
- `parser_support.cc:55` and `:57` (`vsnprintf(nullptr, 0, …)` portability) — not a real constraint. That behaviour is required by C99 §7.19.6.5 and by C++11, and MSVC has conformed since VS2015. Test-only code besides.
- `thriftl.ll:333` (`ch` signedness) — theoretical. `yyinput()` provably returns 0…255 here and `0` is handled separately, so `ch` in `default:` is 1…255. Harmless to add the `unsigned char` cast if you want the belt as well as the braces.
- `parser_support.cc:68` (newline normalised for stderr but not for the thrown message) — true, and the call sites in `thrifty.yy` and `main.cc` do lack the trailing `\n`, but nothing asserts the difference and it is test-only.

Worth knowing when you read those: Copilot's last four runs all report *"couldn't run its full agentic review … timeout"*, so they ran degraded. That is why the same half-resolved finding came back three times.

PR #3806, opened 2026-09-06
THRIFT-6195: Compare Go set and entry-slice map fields without regard to order

slachiewicz:
Fixes THRIFT-6195.

Thrift defines `set` and `map` as unordered. The Go binding represents a `set`, and a `map` whose key type cannot be a Go map key, as a slice, and the generated `Equals` compared those slices position by position. Two values holding the same members in a different order therefore compared unequal.

This affects three constructs, because they share one code path in `generate_go_equals_container`:

| Thrift type | Go type | Since |
|---|---|---|
| `set<T>` | `[]T` | long-standing |
| `map<container, V>` | `[]thrift.MapEntry[K, V]` | THRIFT-2063 |
| `map<struct, V>` with `go:struct_key_entries` | `[]thrift.MapEntry[*K, V]` | THRIFT-6175, #3788 |

Java compares these with `HashSet` and `HashMap` equality and Python with `set` and `dict`, so both are order-insensitive without extra work. Go cannot use a native map or set here, because a slice, a map, or a struct holding either is not a valid Go map key. The slice representation is forced by the language; the positional comparison is not.

### What changed

Sets and entry slices now go through a new `thrift.UnorderedEqual`. Lists keep the positional comparison, because a list is ordered, so the list and set arm is split.

The helper compares position by position up to the first pair that differs, then matches only the elements after that against the unmatched elements of the other side. Values that agree on order, which includes anything that has just been deserialized, cost one pass and allocate nothing, and a prefix that matched in order is never walked twice.

Matching consumes an element of the other side at most once. Without that a set holding `{a, a}` compares equal to one holding `{a, b}`, since both copies of `a` match the same element. `TestUnorderedEqual` pins that case and the symmetric one.

### Cost

The fallback is quadratic. A hashed index would make it linear for element types that can be Go map keys, but it needs a type predicate, a separate nil path, and per-type projections, and it is roughly twelve times slower than the positional pass on the common path when applied unconditionally. Trying positional first avoids paying that, and the hashed tier can be added later if a profile asks for it.

`TestUnorderedEqualOrderedPathDoesNotAllocate` pins the zero-allocation property of the fast path, and `TestUnorderedEqual` covers the boundary between the ordered prefix and the pairwise tail.

### Compatibility

This changes observable behaviour. Code whose tests depend on positional comparison of a set or an entry-slice map field will start seeing values compare equal that previously did not. Nothing semantically correct should break, since the change only makes `Equals` agree with what the Thrift type system already says these collections mean.

### Notes for the reviewer

- Supersedes #3805, which GitHub closed when its branch was renamed to match the ticket.
- Rebased on master now that #3788 has landed, so struct-keyed maps generated with `go:struct_key_entries` pick this up automatically.
- The `-remote` packages for `includestest` fail to build on master already, with `undefined: includestest.Numberz`. That is unrelated and not part of the `check` target.

### DIFF
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -4276,6 +4276,19 @@ void t_go_generator::generate_go_equals_struct(ostream& out,
 }
 
 /**
+ * Emits the head of a thrift.UnorderedEqual call, up to the opening brace of
+ * the element comparison. The caller fills in the body, which must return
+ * true when the two elements are equal, and closes the call.
+ */
+void t_go_generator::generate_go_equals_unordered(ostream& out,
+                                                  const string& go_type,
+                                                  const string& tgt,
+                                                  const string& src) {
+  out << indent() << "if !thrift.UnorderedEqual(" << indexable_go_expr(tgt) << ", "
+      << indexable_go_expr(src) << ", func(_tgt, _src " << go_type << ") bool {" << '\n';
+}
+
+/**
  * Compares any container type
  */
 void t_go_generator::generate_go_equals_container(ostream& out,
@@ -4288,15 +4301,20 @@ void t_go_generator::generate_go_equals_container(ostream& out,
   indent_down();
   out << indent() << "}" << '\n';
   if (is_container_keyed_map(ttype)) {
+    // An entry slice stands in for a map, which is unordered, so the entries
+    // are compared as a collection rather than position by position.
     t_map* tmap = (t_map*)ttype;
-    out << indent() << "for i, _tgt := range " << tgt << " {" << '\n';
+    generate_go_equals_unordered(out, map_entry_type(tmap), tgt, src);
     indent_up();
-    string element_source = tmp("_src");
-    out << indent() << element_source << " := " << indexable_go_expr(src) << "[i]" << '\n';
-    generate_go_equals(out, tmap->get_key_type(), "_tgt.Key", element_source + ".Key");
-    generate_go_equals(out, tmap->get_val_type(), "_tgt.Value", element_source + ".Value");
+    generate_go_equals(out, tmap->get_key_type(), "_tgt.Key", "_src.Key");
+    generate_go_equals(out, tmap->get_val_type(), "_tgt.Value", "_src.Value");
+    out << indent() << "return true" << '\n';
     indent_down();
-    indent(out) << "}" << '\n';
+    out << indent() << "}) {" << '\n';
+    indent_up();
+    out << indent() << "return false" << '\n';
+    indent_down();
+    out << indent() << "}" << '\n';
   } else if (ttype->is_map()) {
     t_map* tmap = (t_map*)ttype;
     out << indent() << "for k, _tgt := range " << tgt << " {" << '\n';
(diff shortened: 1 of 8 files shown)


slachiewicz:
Rebased onto master now that #3788 has landed, which is what the two `lib-go` failures were about.

THRIFT-6175 added `TestStructKeyEqualsDetectsKeyDifference/entry_order_only`, asserting that reordering entries changes equality, plus a README paragraph saying the same. Both describe exactly the behaviour this change removes. The assertion is now its opposite, as `TestStructKeyEqualsIgnoresEntryOrder`, and the entry-slice section of the README says a map is compared as a collection.

I missed it locally because the branch was cut before #3788 merged, so neither the test nor the paragraph existed in my tree.

lib/go/thrift/unordered_equal.go
```diff
@@ -0,0 +1,71 @@
...
+	ordered := true
+	for i := range a {
+		if !eq(a[i], b[i]) {
+			ordered = false
+			break
+		}
+	}
+	if ordered {
+		return true
+	}
```
fishy: do we even need this part? we don't care if they are matching in order in this function do we?

in worst case scenario, a and b has elements 0 to n-3 matching in order, only the last 2 elements are swapped, this part would waste n-2 iterations.

lib/go/thrift/unordered_equal_test.go
```diff
@@ -0,0 +1,122 @@
...
+func BenchmarkUnorderedEqualReordered(b *testing.B) {
+	x := make([]int, 1000)
+	y := make([]int, 1000)
+	for i := range x {
+		x[i] = i
+		y[len(y)-1-i] = i
+	}
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
fishy: use `b.Loop()` instead: https://go.dev/doc/go1.26#testingpkgtesting

and we no longer need to call `b.ResetTimer` here.

lib/go/thrift/unordered_equal_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func BenchmarkUnorderedEqualSameOrder(b *testing.B) {
+	x := make([]int, 1000)
+	y := make([]int, 1000)
+	for i := range x {
+		x[i], y[i] = i, i
+	}
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
fishy: same here

compiler/cpp/src/thrift/generate/t_go_generator.cc
```diff
@@ -4278,6 +4278,19 @@ void t_go_generator::generate_go_equals_struct(ostream& out,
 /**
  * Compares any container type
  */
```
fishy: this comment should be right above `generate_go_equals_container`

lib/go/thrift/unordered_equal.go
slachiewicz: Kept, but restructured so it costs nothing: the first loop stops at the first differing pair and the pairwise phase starts from that index, so the n-2 leading matches you describe are consumed rather than repeated.

Removing it outright would make the common case quadratic and allocating — for values that agree on order, the inner loop still walks the matched prefix before `matched[j]` short-circuits it. `BenchmarkUnorderedEqualSameOrder`, 1000 elements: 3646 ns/op, 0 allocs.

Consuming the prefix is safe because `eq` is an equality: a pair it accepted can belong to any matching of the whole.

lib/go/thrift/unordered_equal_test.go
slachiewicz: Done — `for b.Loop()`, `ResetTimer` dropped.

lib/go/thrift/unordered_equal_test.go
slachiewicz: Same here.

compiler/cpp/src/thrift/generate/t_go_generator.cc
slachiewicz: Moved back above `generate_go_equals_container`.

fishy (approved):

PR #3809, opened 2026-09-07
THRIFT-5420: Drop the testing import from the Go library

slachiewicz:
Fixes THRIFT-5420.

`TestLogger` declared its parameter as `testing.TB`, which made `testing` a build dependency of `lib/go/thrift` itself: `go list -deps ./lib/go/thrift` listed `flag`, `runtime/trace` and `testing` before this change and lists none of them after.

The parameter is now the single method the function calls:

```go
func TestLogger(tb interface {
	Errorf(format string, args ...any)
}) Logger
```

A `*testing.T`, a `*testing.B` and a `testing.TB` all satisfy that, so existing callers compile unchanged. No exported identifier is added, which matters because the whole `Logger` API is already marked deprecated.

`TestLibraryDoesNotImportTesting` parses the package's non-test files and fails if any imports `testing`, so the dependency cannot come back unnoticed.

The alternative is to delete the deprecated `Logger` API outright, since it says it is no longer used by any library code. That is a breaking change and a separate decision, so this PR keeps the API and removes only the dependency.

### DIFF
--- a/lib/go/thrift/logger.go
+++ b/lib/go/thrift/logger.go
@@ -22,7 +22,6 @@ package thrift
 import (
 	"log"
 	"os"
-	"testing"
 )
 
 // Logger is a simple wrapper of a logging function.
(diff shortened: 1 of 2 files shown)


fishy (approved):

Kunal8954:
what ai do you used for the code creation?

PR #3819, opened 2026-09-08
THRIFT-5224: Replace the deprecated Buffer() constructor in the Node.js library

slachiewicz:
Replace deprecated `new Buffer()` constructor calls with `Buffer.alloc()` and `Buffer.from()` across `lib/nodejs/lib` and `lib/nodejs/test`.

- `new Buffer(size)` -> `Buffer.alloc(size)`
- `new Buffer(data[, encoding])` -> `Buffer.from(data[, encoding])`

Eliminates the Node.js `DEP0005` runtime deprecation warning emitted whenever the library or tests run. Supersedes stale/closed PR #2163.

- [X] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? (THRIFT-5224)
- [X] If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?
- [X] Did you squash your changes to a single commit?
- [X] Did you do your best to avoid breaking changes?
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.

*This change was created with AI assistance.*

PR #3821, opened 2026-09-08
THRIFT-6203: Replace deprecated url.parse and fs.exists in the Node.js library

slachiewicz:
JIRA: [THRIFT-6203](https://issues.apache.org/jira/browse/THRIFT-6203)
Client: nodejs

The library called two APIs Node has deprecated, so an application embedding it printed `DEP0169` and `DEP0006` on every supported runtime: `url.parse()` and `fs.exists()` in `web_server.js`. They become the WHATWG `URL` class and `fs.stat()`.

The third API the ticket names, `require("constants")` (`DEP0063`), is already gone from master: ec1066ab4 removed it from `server.js` and `connection.js` and leaves the TLS defaults to Node.

Two behaviour changes come with that and are worth reviewing on their own terms.

**A malformed request target now gets a 400.** `url.parse()` accepts one and returns a partly-parsed path; `new URL()` throws. `processPost` and `processGet` catch it and answer `400 Bad Request` rather than letting the exception reach the request handler. Only the pathname is used, so the base is a fixed `http://localhost`: built from the `Host` header, a value that is not a valid authority (`example.com:abc`, an unbracketed IPv6 address) made `new URL()` throw and produced a 400 that `url.parse()` never did.

**`options.services` defaults to `{}`.** A server serving only static files has no reason to pass it, and `createWebServer` read it without a default. The registration loop does not catch that, because `for...in` over `undefined` is a no-op rather than an error, so the first POST failed instead:

```
TypeError: Cannot read properties of undefined (reading '/')
```

The WebSocket upgrade listener is unaffected: since `31f5897cb` on master it refuses an empty `services` object with a 403, and `web_server_ws.test.js` covers that case.

The `fs.exists()` replacement also drops a redundant `fs.statSync()` that ran immediately after the async check.

Also carried here: the comment in `thrift_4987_xhr_protocol.test.mjs` names `Buffer.from(string)`, matching the `Buffer()` replacement from #3819.

The new cases in `web_server_ws.test.js` cover static file serving, a 404, and a request routed to an unregistered service.

Verified: `node -e "for (var u in undefined) {}"` is a no-op, while indexing `undefined` throws the `TypeError` above, so the failure could only surface on a request, not at construction. `web_server_ws.test.js` → 10 passed; `eslint lib/nodejs/test` clean.

- [X] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? (THRIFT-6203)
- [X] Did you do your best to avoid breaking changes?


Jens-G:
### Code review

Found 1 issue:

1. Defaulting `services` to `{}` changes how the WebSocket `upgrade` listener handles a server created without `services`. That listener only notices "no services" when `Object.keys(services)` throws inside its try block. `Object.keys(undefined)` throws, so today such a request gets `403 No Apache Thrift Service available`. `Object.keys({})` does not throw, so with this change `svc` stays `undefined`, the listener writes `101 Switching Protocols` and then throws `TypeError: Cannot read properties of undefined (reading 'transport')` at `svc.transport.DEFAULT_MAX_LENGTH`. The test this PR adds ("HTTP static file serving and URL handling") creates exactly such a server (`files` only, no `services`) but only sends GET and POST requests, so the suite does not see this. I reproduced it with that configuration: 403 on the merge base, the `TypeError` on this branch. An explicit `if (!svc)` check before the handshake would bring back the 403, and an upgrade request in the new test would cover it.

https://github.com/apache/thrift/blob/7aab2fd27007074b0199215ccae4e3acbf62fda6/lib/nodejs/lib/thrift/web_server.js#L373-L375

https://github.com/apache/thrift/blob/7aab2fd27007074b0199215ccae4e3acbf62fda6/lib/nodejs/lib/thrift/web_server.js#L667-L675

- Only the `pathname` of the parsed URL is used, so the base does not need to come from the `Host` header. Built from `"http://" + request.headers.host`, a Host value that is not a valid authority (for example `example.com:abc`, or an IPv6 address without brackets) makes `new URL` throw, and the request gets a 400; `url.parse` never looked at that header. A fixed base such as `http://localhost` avoids this.

https://github.com/apache/thrift/blob/7aab2fd27007074b0199215ccae4e3acbf62fda6/lib/nodejs/lib/thrift/web_server.js#L450-L460

Jens-G:
Follow-up to the review above: master now refuses a WebSocket upgrade when no service is configured, an empty `services` object included (31f5897cb, with tests in `web_server_ws.test.js`). After a rebase onto master, the `services || {}` default no longer reaches `svc.transport`, so point 1 needs no change in this PR.

slachiewicz:
Branch rewritten (70cae7bfb): rebased on current master, which carries the upgrade-listener fix from 31f5897cb, so point 1 needs nothing here as you said. The suggestion is taken: both `new URL()` calls now use the fixed base `http://localhost`, so a `Host` header that is not a valid authority no longer turns into a 400. The comment cleanup from #3819 (`Buffer.from(string)` in `thrift_4987_xhr_protocol.test.mjs`) rides along.

Verified: `web_server_ws.test.js` → 10 passed on the rebased branch; `eslint lib/nodejs/test` clean.

PR #3822, opened 2026-09-08
THRIFT-5806: Count zero set fields on a nil union rather than dereferencing it

slachiewicz:
JIRA: [THRIFT-5806](https://issues.apache.org/jira/browse/THRIFT-5806)
Client: go

`CountSetFields` read the union's fields straight off the receiver, so a struct holding a nil union field panicked the moment it was serialized:

```
panic: runtime error: invalid memory address or nil pointer dereference
```

Returning 0 for a nil receiver lets the union's own arity check speak instead:

```
*Descendant error writing struct: *Descendant write union: exactly one field must be set (0 set)
```

## Scope

This PR is only the nil guard: no generated write path changes, no wire bytes change.

Omitting an unset default-requiredness pointer field from the output, the way Python, Java, Node.js and C# omit it, is a wire change rather than a panic fix and is [THRIFT-6204](https://issues.apache.org/jira/browse/THRIFT-6204), #3827, which is stacked on this PR. The measurements of what that guard reaches are in that PR.

## Tests

`StructWithUnsetUnion` in `UnionDefaultValueTest.thrift` plus `TestNilUnion` and `TestStructWithUnsetUnion`.

Verified: with the master generator `TestNilUnion` fails with `panic: runtime error: invalid memory address or nil pointer dereference`; with this change `go test ./tests/ -count=1` in `lib/go/test` passes.

*This change was created with AI assistance (Claude Opus 5).*


Jens-G:
### Code review

Found 1 issue:

1. The commit has no AI attribution trailer, although the PR description says the change was created with AI assistance (AGENTS.md says "**Always** label AI-assisted commits and PRs. Use one or both of: `Co-Authored-By: <AI tool name and version>` / `Generated-by: <AI tool name and version>`"). The PR text names no tool or version either. The commits of #3816 and #3817 carry such a trailer.

https://github.com/apache/thrift/blob/3eeb97b2f4f8326c7c8ee5a9d1f2c2cce3c40e41/AGENTS.md#L60-L74

slachiewicz:
Rebased on master and added the `Co-Authored-By: Claude Opus 5` trailer to the commit; the PR description now names the tool as well. Patch content is unchanged (8517fd87f).

fishy (approved):

PR #3823, opened 2026-09-08
THRIFT-5807: Format out-of-range enums as EnumName(val) instead of <UNSET>

slachiewicz:
JIRA: [THRIFT-5807](https://issues.apache.org/jira/browse/THRIFT-5807)
Client: go

An enum holding a value the IDL does not define printed `<UNSET>`, which reads as "no value was sent" when in fact the integer arrived intact and was written back out unchanged. Rust prints `Enum0(10)`, and C++, Python and Node.js print the integer. Go now prints `Color(999)`, the shape `stringer` produces.

A generated `IsDefined() bool` comes with it, so callers have something to ask instead of comparing `String()` against a sentinel. `go_validator_generator` was doing exactly that comparison for `vt.defined_only` and now calls `IsDefined()`.

`lib/go/README.md` gains a "A note about undefined enum values" section covering the `String()`, `MarshalText` and JSON change and the new `IsDefined()`, in line with the notes for THRIFT-2063, THRIFT-6175 and THRIFT-6195.

## What else moves with String()

`MarshalText` returns `[]byte(p.String())`, so this changes text and JSON encoding for out-of-range values:

```
master:  json.Marshal(Color(999)) -> "<UNSET>"
here:    json.Marshal(Color(999)) -> "Color(999)"
```

Neither round-trips: `UnmarshalText` goes through `FromString`, which rejects both with `not a valid Color string`, so no encoding that used to survive a round trip stops surviving one. The new form at least says which value failed.

No import risk: the generated `FromString` already references `fmt` for every enum, so `fmt.Sprintf` needs no new import in any file.

## Blast radius

Regenerating every IDL under `test/`, `lib/go/test/` and `tutorial/` with both compilers, 73 of 650 generated files change, and the change is additive apart from one line per enum. Across the whole corpus the only removed lines are:

| Removed | Count |
|---|---|
| `return "<UNSET>"` | 153 |
| `if (p.Enum0).String() == "<UNSET>" {` | 1 |
| `if (*p.Enum1).String() == "<UNSET>" {` | 1 |

Everything else added is the new `IsDefined()` method, 153 of them, one per enum in the corpus. No existing generated logic is restructured.

## What can break

Code comparing `String()` to the literal `"<UNSET>"` to test for an undefined value. The two call sites of that pattern in this repository are the validator lines above; `IsDefined()` is the replacement, and it says what the caller meant.

Verified: `Color(999).String()` returns `Color(999)` and `Foo_One.String()` returns `One`; `json.Marshal` output as quoted above, read from a run against both compilers.

### DIFF
--- a/compiler/cpp/src/thrift/generate/go_validator_generator.cc
+++ b/compiler/cpp/src/thrift/generate/go_validator_generator.cc
@@ -239,7 +239,7 @@ void go_validator_generator::generate_enum_field_validator(std::ostream& out,
       }
     } else if (key == "vt.defined_only") {
       if (values[0]->get_bool()) {
-        out << indent() << "if (" << context.tgt << ").String() == \"<UNSET>\"";
+        out << indent() << "if !(" << context.tgt << ").IsDefined()";
       } else {
         continue;
       }
--- a/compiler/cpp/src/thrift/generate/t_go_generator.cc
+++ b/compiler/cpp/src/thrift/generate/t_go_generator.cc
@@ -817,7 +817,7 @@ void t_go_generator::generate_typedef(t_typedef* ttypedef) {
  */
 void t_go_generator::generate_enum(t_enum* tenum) {
   begin_types_declaration();
-  std::ostringstream to_string_mapping, from_string_mapping, known_values_mapping;
+  std::ostringstream to_string_mapping, from_string_mapping, known_values_mapping, is_defined_mapping;
   std::string tenum_name(publicize(tenum->get_name()));
   generate_go_docstring(f_types_, tenum);
   generate_deprecation_comment(f_types_, tenum->annotations_);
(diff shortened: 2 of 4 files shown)


Jens-G:
### Code review

Found 1 issue:

1. The commit has no AI attribution trailer, although the PR description says the change was created with AI assistance (AGENTS.md says "**Always** label AI-assisted commits and PRs. Use one or both of: `Co-Authored-By: <AI tool name and version>` / `Generated-by: <AI tool name and version>`"). The PR text names no tool or version either. Your earlier generator commits for THRIFT-2063 and THRIFT-6175 carry such a trailer.

https://github.com/apache/thrift/blob/aed6b47d3f31f219420297ebb104711c459c644a/AGENTS.md#L60-L74

- This changes what generated code produces: `String()`, and with it `MarshalText`/JSON, of an unknown value, plus a new exported `IsDefined()` on every enum. For THRIFT-2063, THRIFT-6175 and THRIFT-6195 you added a "A note about …" section to lib/go/README.md for comparable generated-code changes; a short note here would keep that consistent.

https://github.com/apache/thrift/blob/aed6b47d3f31f219420297ebb104711c459c644a/lib/go/README.md#L141-L143

slachiewicz:
Rewrote the branch after the review above: the commit carries the attribution trailer, and lib/go/README.md gains a "A note about undefined enum values" section covering the `String()`, `MarshalText` and JSON change and the new `IsDefined()`, in line with the notes for THRIFT-2063, THRIFT-6175 and THRIFT-6195.

PR #3826, opened 2026-09-08
THRIFT-5242: Add Homebrew installation instructions for macOS

slachiewicz:
JIRA: [THRIFT-5242](https://issues.apache.org/jira/browse/THRIFT-5242)
Client: doc

`doc/install/os_x.md` now covers the pre-built compiler first, `brew install thrift`, and then building the compiler from source with Homebrew-provided tools.

- The Homebrew line installs `automake libtool pkg-config cmake bison flex`, the set the macOS CI job uses, so `./bootstrap.sh` finds libtoolize and `configure.ac` finds the pkg-config macros.
- Both recipes build only the compiler, which is what the page promises: `--disable-libs --disable-tests` for Autotools and `-DBUILD_LIBRARIES=OFF -DBUILD_TESTING=OFF -DBUILD_TUTORIALS=OFF` for CMake. Neither needs Boost, libevent or OpenSSL; one sentence tells readers who also want the C++ library what to add.
- The CMake recipe builds into `cmake-build`, since `cmake -B build` would reuse the repository's own `build/` directory, which holds the CMake modules.
- A note explains that the `bison` shipped with macOS is 2.3 and cannot process the grammar, and that Homebrew's is keg-only, so each recipe points at it explicitly.
- External links move to HTTPS.

Verified: both recipes run as written on macOS 15 (Apple Silicon) → `thrift -version` prints 0.25.0 from each build. Limits of that check: the `sudo … install` steps ran into a scratch prefix, so the `/usr/local` permissions were not exercised; `flex` on the test machine came from Xcode rather than Homebrew; and the unchanged section on building Boost and libevent by hand was not run. The macOS CI job covers the Autotools path on a fresh runner; the CMake path has no CI coverage.


Jens-G:
### Code review

Found 2 issues:

1. The Autotools recipe does not work with the dependencies the page lists. `brew install cmake boost libevent openssl bison flex` installs none of autoconf, automake, libtool or pkg-config, and macOS does not ship them. The recipe starts with `./bootstrap.sh`, which exits with "Couldn't find libtoolize!" when libtool is missing. It also requires automake 1.13 or later and runs aclocal/autoconf/automake, and `configure.ac` needs the pkg-config macros. The macOS CI job installs `automake bison flex boost libevent openssl libtool pkg-config` before running the same steps. (bug due to doc/install/os_x.md: `brew install cmake boost libevent openssl bison flex`, followed by `./bootstrap.sh`)

https://github.com/apache/thrift/blob/362d95457d960060130f142670a9b3cf794237dd/doc/install/os_x.md#L23-L25

https://github.com/apache/thrift/blob/362d95457d960060130f142670a9b3cf794237dd/.github/workflows/build.yml#L131-L133

2. The commit has no AI attribution trailer, although the PR description says the change was created with AI assistance (AGENTS.md says "**Always** label AI-assisted commits and PRs. Use one or both of: `Co-Authored-By: <AI tool name and version>` / `Generated-by: <AI tool name and version>`").

https://github.com/apache/thrift/blob/362d95457d960060130f142670a9b3cf794237dd/AGENTS.md#L60-L74

Two suggestions, below the bar for the list above but verified:

- The page is about installing the compiler, but the CMake recipe runs with the default options. Those also build the libraries, tests and tutorials (`BUILD_LIBRARIES`, `BUILD_TESTING` and `BUILD_TUTORIALS` default to ON), and the tests need compiled Boost libraries. The compiler-only options used in doc/ReleaseManagement.md (`-DBUILD_TESTING=OFF -DBUILD_TUTORIALS=OFF -DWITH_CPP=OFF … -DBUILD_COMPILER=ON`) would match what the page promises. Also, the macOS CI job only runs the Autotools path, so the recipe marked "(Recommended)" is the one that is not tested.

https://github.com/apache/thrift/blob/362d95457d960060130f142670a9b3cf794237dd/doc/install/os_x.md#L28-L37

- `export LDFLAGS="-L$(brew --prefix bison)/lib"` is not needed. bison only runs as a program and nothing links against it; CI only adds bison's `bin` directory to `PATH`. The export also replaces any `LDFLAGS` the reader already has set.

https://github.com/apache/thrift/blob/362d95457d960060130f142670a9b3cf794237dd/doc/install/os_x.md#L41-L44

slachiewicz:
Rewrote the branch after the review above:

- The Homebrew line now installs `automake libtool pkg-config` as the macOS CI job does, so `./bootstrap.sh` finds libtoolize.
- Both recipes build only the compiler, which is what the page promises: `--disable-libs --disable-tests` for Autotools and `-DBUILD_LIBRARIES=OFF -DBUILD_TESTING=OFF -DBUILD_TUTORIALS=OFF` for CMake. Neither needs Boost, libevent or OpenSSL, so those moved to a sentence for readers who also want the C++ library.
- Dropped the `LDFLAGS` export and the "(Recommended)" label.
- The CMake recipe now builds into `cmake-build`. `cmake -B build` reuses the repository's own `build/` directory, which holds the CMake modules.
- The commit carries the attribution trailer.

Verified: both recipes run as written on macOS 15 (Apple Silicon) → `thrift -version` prints 0.25.0 from each build.

slachiewicz:
A limit on the verification above, since I had no clean macOS environment for it:

- The `sudo … install` steps ran into a scratch prefix (`cmake --install --prefix`, `make install DESTDIR=`), so the install target is tested but not the `/usr/local` permissions.
- `flex` on the test machine came from Xcode, not Homebrew, so the `brew install` line was not exercised from a fresh machine. All six formula names resolve.
- The unchanged section on building Boost and libevent by hand was not run.

The macOS CI job covers the Autotools path on a fresh runner; the CMake path has no CI coverage.

Jens-G:
### Code review

Re-reviewed at 640e71fc3 after the rewrite: the points from the earlier review are addressed.

One suggestion, below the bar for an issue but verified:

- Line 29 says that building the C++ library as well only takes `boost libevent openssl` and dropping the compiler-only options. On the CMake path that also enables the JavaScript and NodeJS libraries: `BUILD_JAVASCRIPT` and `BUILD_NODEJS` default to ON without any check for Node.js, and both add targets to the default build that run `npm install`. Node.js is not on the page's `brew install` line, so without it `cmake --build cmake-build` fails at the first `npm install`. I reproduced this with these CMake files on Linux, as no Mac was available; the conditions involved do not depend on the platform. Adding `-DWITH_JAVASCRIPT=OFF -DWITH_NODEJS=OFF` to that variant removes both targets. Java and Python are switched on as well when CMake finds their tools.

https://github.com/apache/thrift/blob/640e71fc37551735dc63395bf537d77cec4fe51d/doc/install/os_x.md#L28-L30

https://github.com/apache/thrift/blob/640e71fc37551735dc63395bf537d77cec4fe51d/build/cmake/DefineOptions.cmake#L99-L108

https://github.com/apache/thrift/blob/640e71fc37551735dc63395bf537d77cec4fe51d/lib/nodejs/CMakeLists.txt#L34-L40

slachiewicz:
Right, dropping `BUILD_LIBRARIES=OFF` switches on every library CMake finds tools for. 2e838c8b4 replaces that sentence: the CMake path now names `-DWITH_JAVASCRIPT=OFF -DWITH_NODEJS=OFF -DWITH_JAVA=OFF -DWITH_PYTHON=OFF`, and the Autotools path `--without-<lang>` for each other language whose tools are installed.

PR #3820, opened 2026-09-08
Update LANGUAGES.md with current versions and supported features

slachiewicz:
### Description

This PR updates `LANGUAGES.md` to bring the language support matrix and documentation in step with recent releases (0.23.0, 0.24.0, and 0.25.0 development on master):

- **Header & Navigation**:
  - Update guidance header to `0.25.0`
  - Add version links for `0.24.0`, `0.23.0`, and `0.22.0`
  - Update CI documentation reference from Travis CI to GitHub Actions and fix typo (`Windows..`)
- **Missing Languages**:
  - Add **Kotlin** row (supported since 0.17.0 via `THRIFT-5548`, with autoconf/cmake builds, UUID, and CI/cross-test coverage)
- **Field Types**:
  - Update **UUID** support status to green for **Java (SE)** (`THRIFT-5587`, `THRIFT-5925`), **PHP** (`THRIFT-5587`), **Rust** (`b6b6dc715`), and **Lua** (`THRIFT-4386`)
- **Tested Language / Library Levels**:
  - **PHP**: Update to `8.1` - `8.5` (tested in CI workflow; PHP 7.x dropped)
  - **Python**: Update to `3.10` - `3.14` (tested in CI workflow; Python 2.x dropped)
  - **node.js**: Update to `20.x` (tested in CI and pinned in Dockerfiles)
  - **.NET**: Update to `.NET Standard 2.x, net8, net9, net10` (`THRIFT-5902`, `THRIFT-6002`)
  - **Lua**: Update to `5.1.5` - `5.4` (`THRIFT-4386`)
  - **Rust**: Update to `1.83.0` - `1.85.1`
- **Build Systems**:
  - Mark **CMake** as supported (green) for **Javascript** and **node.js** (`BUILD_JAVASCRIPT` and `BUILD_NODEJS` in `CMakeLists.txt`)
- **Transport Wrappers**:
  - Mark **zlib** transport wrapper as supported for **C (glib)** (`THRIFT-5265`)
- **Hyperlinks**:
  - Fix broken `README.md` links for `javame`, `lua`, and `nodets` to directory trees

*This change was created with AI assistance.*

PR #3834, opened 2026-09-09
Correct LANGUAGES.md entries flagged after the matrix update

slachiewicz:
Follow-up to #3820, addressing the two review comments there. Every entry was checked against the source at 5c5e93e1e; nothing here is carried over from the earlier matrix.

| Entry | Was | Now | Evidence |
|---|---|---|---|
| Java SE levels | 11 / 19 | 8 / 17 | `build.yml` builds on JDK 17 and runs the cross-tests on Java 8; `sourceConfiguration.gradle` sets a JDK 17 toolchain with `options.release = 8` |
| Rust TLS | no | yes | the `rustls` feature in `Cargo.toml`, with `TTlsClientChannel::connect` and `TServer::listen_tls` |
| Rust Domain | no | yes | `TServer::listen_uds`, under `#[cfg(unix)]` |
| Rust ThreadPool | no | yes | `TServer` holds a `ThreadPool` sized by `num_workers` |
| Python Memory, Framed | no | yes | `TMemoryBuffer` and `TFramedTransport` |
| Python Threaded, ThreadPool | no | yes | `TThreadedServer` and `TThreadPoolServer` |
| Kotlin level | 1.8 | 2.4.10 | `settings.gradle.kts` uses `kotlin("jvm") version "2.4.10"`; 1.8 is the bytecode target, not the Kotlin version |
| Kotlin Multiplex, zlib, Nonblocking, ThreadPool | no | yes | the cross-test server uses `TMultiplexedProcessor`, `TZlibTransport`, `TNonblockingServer` and `TThreadPoolServer` over the Java runtime |
| Haxe Memory | no | yes | `TMemoryStream`, with round-trip coverage in `StreamTest.hx` |

The table has no way to carry a condition, so this adds a short notes section for three of them: the `rustls` feature is off by default, `listen_uds` is Unix-only, and the Java and Kotlin rows report tested bounds rather than a maximum supported release.

One entry is deliberately left alone. Rust has a single server type, a fixed-size worker pool, and the row marks it as both Threaded and ThreadPool. Marking ThreadPool is right; whether Threaded should stay is a separate judgement, and I did not want to fold it into a corrections PR.

Thanks @HTHou for the two lists. The claims held up on inspection.

### DIFF
--- a/LANGUAGES.md
+++ b/LANGUAGES.md
@@ -185,7 +185,7 @@ Thrift's core protocol is TBinary, supported by all languages except for JavaScr
 <!-- Build Systems ---------><td><img src="/doc/images/cgrn.png" alt="Yes"/></td><td><img src="/doc/images/cred.png" alt=""/></td>
 <!-- Language Levels -------><td>4.1.5</td><td>4.2.5</td>
 <!-- Field types -----------><td><img src="/doc/images/cgrn.png" alt="Yes"/></td>
-<!-- Low-Level Transports --><td><img src="/doc/images/cred.png" alt=""/></td><td><img src="/doc/images/cgrn.png" alt="Yes"/></td><td><img src="/doc/images/cred.png" alt=""/></td><td><img src="/doc/images/cred.png" alt=""/></td><td><img src="/doc/images/cgrn.png" alt="Yes"/></td><td><img src="/doc/images/cred.png" alt=""/></td>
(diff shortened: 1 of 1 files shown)


slachiewicz:
While table looks nice maybe it coudl be somehow reorganized to some capabilities and more verbose clarification that this capability means for specific language. For example Go port is in many places marked as "not supported" due to different architecture.
But i'm too new to make decisions here, simpy found that we have outdated page with features while reviewing AI tool output.

HTHou (approved):
Thanks for following up! The changes address all the entries raised in the two comments on #3820, and the added notes clarify the platform, feature and version conditions.

Reviewed the documentation changes against the source and build configuration with AI assistance; I did not rerun the test suite locally. No blocking issues found in this diff. The Rust Threaded/ThreadPool terminology can be discussed separately.

PR #3835, opened 2026-09-09
THRIFT-6211: Add native Go fuzz targets that run their seed corpus in CI

Jens-G:
Implements [THRIFT-6211](https://issues.apache.org/jira/browse/THRIFT-6211).

## The gap

The Go tree has only [go-fuzz](https://github.com/dvyukov/go-fuzz) style targets — `func FuzzX(data []byte) int` behind the `gofuzz` build tag, which is what the OSS-Fuzz build consumes. Nothing in this repository fuzzes them: `make -C lib/go check` recurses into `lib/go/test/fuzz` and runs `go test -tags gofuzz`, and the only test there is

```go
func TestFuzz(t *testing.T) {
	FuzzTutorial([]byte{1, 2, 3})
}
```

one call, one three-byte input. That is a compile check against generated-code drift, worth keeping, but it finds nothing. There is also no corpus anywhere in the tree, so an input OSS-Fuzz finds and we fix has nowhere to land as a regression test.

## What this adds

Native `testing.F` targets **alongside** the go-fuzz ones — the go-fuzz targets are untouched and OSS-Fuzz keeps consuming them.

The lever is that **`go test` without `-fuzz` replays every `f.Add` seed and every file under `testdata/fuzz/<Target>/` as an ordinary test case.** So the existing `make check` picks up regression coverage on both Go versions in the CI matrix at no extra runtime and with no fuzzing infrastructure in the PR path, and a committed failing input keeps a fixed bug fixed.

**`lib/go/thrift/fuzz_test.go`** — twelve targets that need no generated code. They drive the read paths through `ReadMessageBegin` plus `Skip`, which is what generated struct-read code amounts to: dispatch on the wire-supplied type, read containers at wire-supplied sizes. Because they need only the library, they live with it and run in `go test -race ./thrift`, which `make check` already invokes.

They reach ground no existing target does: `THeaderProtocol` (client-type detection, transform-ID list, info-header parsing), `TFramedTransport`, `TSimpleJSONProtocol`, a server-side dispatch hop through `TMultiplexedProcessor`, and `ParseTuuid`.

**`lib/go/test/fuzz/fuzz_native_test.go`** — native equivalents of the existing struct-read and round-trip targets, which do need `gen-go`. The file carries no build tag, so a single `go test -tags gofuzz` builds and runs both sets and the make target needs no change. Only `EXTRA_DIST` gains the new file.

**Docs** — `FUZZING.md` and `lib/go/test/fuzz/README.md` describe both kinds, how to fuzz one for real, and that a failing input belongs in `testdata/fuzz/<Target>/` alongside the fix.

## Verification

- 18 targets fuzzed 60s each: **~19M executions, no failures.**
- Seed-corpus replay: 147 cases in `lib/go/thrift`, 60 in `lib/go/test/fuzz`, all passing.
- `go vet -stdmethods=false ./thrift` and `go test -race ./thrift` clean.
- `gofmt` clean; `codespell` (what `make style` runs) clean.

### DIFF
--- a/FUZZING.md
+++ b/FUZZING.md
@@ -48,6 +48,13 @@ To ensure fuzzing can find issues as soon as possible, we will enable fuzzing su
 
 Currently the only convenient, formally supported build with fuzzing support enabled is via the oss-fuzz workflow. For languages where local fuzzing is practical, documentation is provided alongside the fuzzers. For example, C++ builds libFuzzer binaries directly, while Ruby exposes `make` targets that wrap Ruzzy.
 
+Go additionally has native `testing.F` targets, in `lib/go/thrift/fuzz_test.go` and
+`lib/go/test/fuzz/fuzz_native_test.go`. Those need no external tooling, and run without `-fuzz`
+they replay their seed corpus as ordinary test cases -- so `make check` gets regression coverage
+from them for free, and a fixed input stays fixed once its file under `testdata/fuzz/<Target>/`
+is committed. To fuzz one for real, `go test -run '^$' -fuzz <Target> -fuzztime 2m`. See
+`lib/go/test/fuzz/README.md`.
+
 ## OSS-Fuzz Integration
 
 Our fuzzers run continuously on OSS-Fuzz. To view build status:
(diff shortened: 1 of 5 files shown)


fishy (approved):
I think this is fine for manual fuzzing. The issue is that fuzzing is something needs to be done continuously (or for a long time), and we lack some environment to do that automatically. So for CI process this is essentially just some added test cases (with fuzz seedings in the code). Which is still a good thing.

The old `lib/go/test/fuzz` I believe is something that's run against more continuously through some google project so we definitely want to keep that, until we can switch that to the native fuzzing (I think there's a ticket/issue for that somewhere but I don't believe there's any movements).

PR #3854, opened 2026-09-14
THRIFT-6290: Rust: skip string/binary fields without heap-allocating

lbarthon:
Client: rs

JIRA: https://issues.apache.org/jira/browse/THRIFT-6290

`skip(TType::String)` allocated a `Vec` the size of the field via `read_bytes()` and immediately dropped it. Binary and compact now read the length prefix and discard the payload through a 256-byte stack buffer (`skip_binary` / `discard_exact`). Negative binary sizes and `max_string_size` still apply. Custom `TInputProtocol` impls keep the allocating default.


lbarthon:
cc @Jens-G when you have a moment 🙏

santiagomed (approved):

Jens-G:
### Code review

Found 1 issue:

1. There is no JIRA ticket for this change, and the PR title and the commit subject do not start with `THRIFT-NNNN:`. The change adds a method to the public `TInputProtocol` trait and changes how the binary and compact protocols skip string and binary fields, so it is not a trivial one (AGENTS.md says a ticket is "Required for all non-trivial PRs" and "All significant changes need a JIRA ticket.")

https://github.com/apache/thrift/blob/7d6be866ba569aed496b2f263f6278cf747b1bf7/AGENTS.md#L27-L29

https://github.com/apache/thrift/blob/7d6be866ba569aed496b2f263f6278cf747b1bf7/AGENTS.md#L48-L50

- None of the new tests notices if `skip()` goes back to `read_bytes()`, or if the `skip_binary` forward in `Box<P>` or in `TStoredInputProtocol` is dropped, because both paths consume the same bytes and return the same errors. A test transport that records the largest buffer it is asked to fill does notice: while skipping a 1024-byte string it sees a single 1024-byte request on the old path and no request over 256 bytes on the new one. I checked this against all three changes.

https://github.com/apache/thrift/blob/7d6be866ba569aed496b2f263f6278cf747b1bf7/lib/rs/src/protocol/mod.rs#L1213-L1224

https://github.com/apache/thrift/blob/7d6be866ba569aed496b2f263f6278cf747b1bf7/lib/rs/src/protocol/mod.rs#L407-L411

lbarthon:
@Jens-G Filed THRIFT-6290 and updated the PR title and commit subject to start with it:
https://issues.apache.org/jira/browse/THRIFT-6290

On the test note: agreed that the current tests would still pass if skip() went back to read_bytes(), or if the skip_binary forward on Box<P> or TStoredInputProtocol were dropped. Happy to add a transport that records the largest buffer it is asked to fill, and assert that skipping a 1024-byte string never asks for more than 256, if you want that in this PR.

Jens-G:
Thanks — THRIFT-6290 settles the ticket point.

Yes please, add the recording transport in this PR. "Skipping a 1024-byte string never asks for more than 256 bytes" is exactly the two-state signal the current tests are missing: they pass either way, because the old and the new path consume the same bytes and return the same errors.

Worth covering all three places the forward can be lost, not just the protocols themselves: `TBinaryInputProtocol` / `TCompactInputProtocol`, the `Box<P>` impl, and `TStoredInputProtocol`. If a `skip_binary` forward is dropped in either wrapper it silently falls back to the allocating default, and nothing else in the suite would notice.

lbarthon:
@Jens-G Added that recording transport in e54863c.

Skipping a 1024-byte string now asserts the transport is never asked to fill more than 256 bytes, on `TBinaryInputProtocol`, `TCompactInputProtocol`, `Box<dyn TInputProtocol>`, and `TStoredInputProtocol`. Dropping either `skip_binary` forward fails only that wrapper's test with a 1024-byte read; pointing the protocol methods back at `read_bytes()` fails all four the same way.

PR #3856, opened 2026-09-15
THRIFT-5463: Take the address of a container literal in a Go constant

slachiewicz:
JIRA: [THRIFT-5463](https://issues.apache.org/jira/browse/THRIFT-5463)
Client: go

An optional container field with a default value is a pointer field in the generated Go, but a constant of the enclosing struct rendered the container as a plain literal, so the generated consts file did not compile:

```
cannot use []int32{…} (value of type []int32) as *[]int32 value in struct literal
```

`render_const_value` already addresses base-type and struct values for pointer fields. The list, set and map branches now do the same, and use the typedef name when the field has one so the address has the field's type. This is the first item of the ticket. The second, strings with defaults staying values while containers with defaults become pointers, renames generated fields and is not touched.

The test IDLs gain optional list, set and map fields with defaults, plain and typedef'd, plus a struct-keyed map for the entry-slice path, each set in a constant. That last shape exposed one more defect on the same pointer-field path: the duplicate-key check for a struct-keyed map ranged over `(*p.Field)`, which gofmt rewrites, so the generated-code format check failed. It ranges over `*p.Field` now.

Blast radius: regenerating every IDL under `test/`, `lib/go/test/` and `tutorial/` with master and with this branch, 136 inputs and 582 files, only the consts files of the two extended test IDLs differ.

Verified: `go build` of the ConstOptionalField package fails with eight errors on master once the new fields are present and passes here; the `lib/go/test` check target builds and `go test` passes for `tests` and `dontexportrwtest`; `gofmt -l -e` is clean on every generated file under `lib/go/test`, `test` and `tutorial`.


Jens-G:
### Code review

PR #3869, opened 2026-09-16
THRIFT-6277: List only tickets fixed in the release in the CHANGES draft

Jens-G:
Fixes [THRIFT-6277](https://issues.apache.org/jira/browse/THRIFT-6277).

The "Generate CHANGES draft" workflow lists every THRIFT ticket that a commit message in the release range mentions, whatever JIRA says about the ticket. The current 0.25.0 draft lists THRIFT-1337 from 2011, because the THRIFT-6183 commit mentions it as history. It also lists other old tickets cited the same way, a Won't Do ticket, and tickets that are still open.

### What changes

- Tickets found in git are now fetched with Fix Version/s, status and resolution. A ticket gets a JIRA line only if it passes the same test as the `--jira-version` query: the release version is among its Fix Version/s, it is resolved, and it is not Open. The version is `--jira-version` if given, otherwise `--version` or `configure.ac`. The test applies in git-only mode, which CI runs on every push, and to the extra lookup in `--jira-version` mode.
- A commit whose tickets all fail the test still appears, as a PR or commit line under its `Client:` section. The skipped tickets are named on stderr, so Fix Version/s can be corrected in JIRA where a ticket does belong to the release.
- The fixVersion query and the new test both leave out tickets resolved without a change, even when they carry the fix version. Excluded: Won't Do, Won't Fix, Not A Problem, Not A Bug, Cannot Reproduce, Works for Me, Invalid, Incomplete, Information Provided, Later, Abandoned, Auto Closed. Duplicate, Done and Implemented still count. This changes nothing for 0.25.0 today. For 0.21.0 it would drop 2 tickets, and for 0.23.0 one.
- Commit lines drop a `Patch: Name` or `Autor: Name` trailer from the subject. Commits such as 2ae9c11db used to be hidden behind their ticket, so the trailer never showed in the draft.

### Effect on the current 0.25.0 draft

Git-only run over `v0.24.0..master`, as CI does it:

| | before | after |
|---|---|---|
| entries | 290 | 271 |
| JIRA entries | 170 | 146 |

21 tickets drop out, in 24 entries because some are filed under several sections:

- 15 old tickets that commit bodies only mention: THRIFT-892, 1337, 2644, 3175, 3532, 5166, 5186, 5370, 5880, 5927, 5998, 6053, 6056, 6083, 6096
- 3 open tickets that commit bodies only mention: THRIFT-6069, 6186, 6204
- 3 tickets named in commit subjects: THRIFT-5917 (Won't Do), THRIFT-6108 and THRIFT-6152 (both still open)

Their commits now show up as 5 entries: #3749 (Build Process), 79dc86d55 (D), 2ae9c11db (Dart and Delphi) and #3637 (Rust). Commits that also carry a 0.25.0 ticket, like the THRIFT-6183 one, stay covered by that ticket.

### Testing

- `python3 -m unittest build.test_generate_changes`: 38 tests pass on Python 3.13, and on Python 3.10 in `thrift:jammy`. The new end-to-end tests run `generate_changes()` against a temporary git repository and a fake JIRA that returns only the requested fields. They failed before the change.
- The before/after comparison above, run against live JIRA and GitHub. The new JQL was also checked against live JIRA for 0.21.0, 0.23.0 and 0.25.0.
- flake8 and codespell report nothing new for the two files. CI lint excludes `build/`.

Related: [THRIFT-6278](https://issues.apache.org/jira/browse/THRIFT-6278) is a separate, pre-existing problem found while testing. JIRA rejects a whole lookup of 25 keys or fewer when one key does not exist, and a `thrift-0.24.0/...` path in a commit body yields THRIFT-0. The `--jira-version` run prints that HTTP 400 today, before and after this change. With this change, that lookup can only add tickets the main query already returns, so nothing is lost.

- [x] Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket? THRIFT-6277
- [x] Did you do your best to avoid breaking changes? Only the generated draft changes; no library or compiler code is touched.
- [ ] If your change does not involve any code, include `[skip ci]` anywhere in the commit message. (n/a, this changes a script)


Jens-G:
Landed on master as a0f4ae48e together with #3871, which was stacked on this PR. Closing.

PR #3873, opened 2026-09-16
THRIFT-6281: Hold the frames TFramedTransport writes to the configured frame size

Jens-G:
JIRA: [THRIFT-6281](https://issues.apache.org/jira/browse/THRIFT-6281)
Client: go

`TFramedTransport.Flush` checked a frame against `math.MaxUint32` only. `readFrame` refuses a frame larger than the configured `MaxFrameSize` (default 16384000), but `Flush` wrote such frames. Measured on master with `MaxFrameSize: 1024`: `Flush` of a 2048-byte frame returned nil and wrote 2052 bytes, and a `TFramedTransport` with the same configuration refused that frame with "Incorrect frame size (2048)".

This is the plain framed counterpart of #3860 (THRIFT-6262), which does the same for `THeaderTransport`.

### Change

- `checkWriteFrameSize`, added for `THeaderTransport.Flush` in #3860, becomes a package-level function taking a `*TConfiguration`. `TFramedTransport.Flush` and the two write paths in `THeaderTransport.Flush` all go through it, so the frames the two transports write are held to the same limits: the configured `MaxFrameSize` and `THeaderMaxFrameSize`, both below what the 32-bit length word can carry.
- A larger frame is refused with the helper's `TProtocolException` of type `SIZE_LIMIT`, and nothing of it is written.
- The `defer` that returns the write buffer now runs before the check, so a refused frame is dropped. The next `Write`/`Flush` starts from an empty buffer rather than failing again on the same oversized one.
- `lib/go/README.md` gets a note for clients that send frames larger than the default, next to the one #3860 added.

### Compatibility

A client that writes frames larger than its own `MaxFrameSize` now gets an error from `Flush` instead of sending a frame that a peer with the same configuration refuses. The error is a `TProtocolException` of type `SIZE_LIMIT`, as on the THeader path; the `MaxUint32` check it replaces returned a `TTransportException`.

### Tests

`TestTFramedTransportFlushFrameSizeLimit`, with `requireFlushSizeLimit` and `flushCountingTransport` from `header_transport_test.go`:
- **At the limit:** a frame of exactly the limit is written and read back.
- **One byte over:** `Flush` returns a `TProtocolException` of type `SIZE_LIMIT` and writes 0 bytes. The next small frame then goes out on its own, 4 + 4 bytes.

Results:
- **Unmodified library:** the over-limit case fails, because `Flush` returned nil.
- **Mutations:** each fails one assertion:
  - `>` → `>=` in the shared check, which fails the framed test and both THeader flush tests;
  - removing the call from `TFramedTransport.Flush`, which fails the framed test alone;
  - moving the buffer release back behind the check.
- **Checks:** `go vet -stdmethods=false` and `go test -race ./thrift` pass with Go 1.26 and 1.27, and `GOARCH=386` vet and tests pass as well.


lib/go/thrift/framed_transport.go
```diff
@@ -202,11 +202,14 @@ func (p *TFramedTransport) WriteString(s string) (n int, err error) {
...
-	if uint64(size) > uint64(math.MaxUint32) {
-		return NewTTransportException(UNKNOWN_TRANSPORT_EXCEPTION, fmt.Sprintf("frame too large: %d bytes exceeds uint32 max", size))
-	}
-
 	defer bufPool.put(&p.writeBuf)
+
+	// readFrame refuses a frame larger than the configured maximum, and so does
+	// a peer holding the same configuration. The maximum is below what the
+	// 32-bit length can carry. A refused frame is dropped with the buffer.
+	if maxSize := p.cfg.GetMaxFrameSize(); int64(size) > int64(maxSize) {
```
fishy: should we reuse the same helper in THeader implementation? in that helper we also checked against `MaxUint32` which we didn't do here.

Jens-G: Done — `checkWriteFrameSize` is a package-level function taking a `*TConfiguration` now, and all three write paths go through it: the two in `THeaderTransport.Flush` (header frame, plain framed answer) and this one. It stayed in `header_transport.go` next to `THeaderMaxFrameSize`; glad to move it if you'd rather have it elsewhere.

On `MaxUint32`: I had dropped that check on purpose, because `GetMaxFrameSize()` returns `int32`, so `int64(size) > int64(maxSize)` already holds size below `MaxInt32` and the old check could no longer fire. What the shared helper really adds here is `THeaderMaxFrameSize`, and that does change behaviour, so two points for your call:

1. **The 0x3fffffff ceiling now applies to plain framed writes too.** `readFrame` checks the configured maximum only, so with `MaxFrameSize` and `MaxMessageSize` set above it — `math.MaxInt32`, say — a `TFramedTransport` can now read a 1073741824-byte frame that it refuses to write. A conservative refusal, and it takes a deliberate >1 GiB configuration to reach, but it is new.

2. **The error type changes with it.** `Flush` returned a `TTransportException` (`UNKNOWN_TRANSPORT_EXCEPTION`), matching `readFrame` in this file; through the helper it returns THeader's `TProtocolException` of type `SIZE_LIMIT`. In `TSimpleServer` that costs one log line and nothing else: the generated processor returns `ok=false` with a `ProcessorError` either way, so `if !ok { break }` and the goroutine's `defer client.Close()` end the connection in both cases, but only a `TTransportException` comes back out of `processRequests` to be logged as "error processing request". #3860 already made that trade for THeader, so I followed it; say the word if you would rather the framed path kept its transport exception.

Tests: the over-limit case reuses `requireFlushSizeLimit` and `flushCountingTransport` from `header_transport_test.go` instead of its own copies. Turning `>` into `>=` in the shared check fails the framed test and both THeader flush tests; dropping the call from `TFramedTransport.Flush` fails only the framed one.

Rebased on master with #3860 in, so the two README notes sit next to each other.

fishy (approved):

PR #3878, opened 2026-09-17
THRIFT-6329: Keep the TPipe implementation alive while a read or write uses it

jnordstrom-iar:
This another (old) patch we've had inhouse. The test is completely written by AI.

On Windows there is no supported way to stop a pipe server while clients are still connected: `TPipeServer` does not implement `interruptChildren()`, and `TPipe` has neither an `interrupt()` nor a read timeout, so `TServer::stop()` cannot wake a worker blocked in `read()`. See: https://issues.apache.org/jira/browse/THRIFT-3590

A workaround we've had is to track connected clients through `TServerEventHandler` and then `close()` their transports from the thread that stops the server. That is not safe on master. `close()` destroys the implementation object a concurrent `read()` is still using. This change guards the implementation pointer with a critical section and hands each caller its own reference, so the implementation stays alive for the duration of the call. `close()` now also detaches the implementation before releasing it and then cancels its outstanding I/O, so a reader waiting for data returns instead of waiting on an implementation that is already gone.

`stress_pipe_close_during_use` dies inside the transport in 19 of 20 runs without the change and passes 20 of 20 with it (MSVC, RelWithDebInfo).

  

- [ ] ~~Did you create an [Apache Jira](https://issues.apache.org/jira/projects/THRIFT/issues/) ticket?  ([Request account here](https://selfserve.apache.org/jira-account.html), not required for trivial changes)~~
- [ ] ~~If a ticket exists: Does your pull request title follow the pattern "THRIFT-NNNN: describe my issue"?~~
- [X] ~~Did you do your best to avoid breaking changes?  If one was needed, did you label the Jira ticket with "Breaking-Change"~~
- [ ] ~~If your change does not involve any code, include `[skip ci]` anywhere in the commit message to free up build resources.~~


Jens-G:
### Code review

Found 2 issues:

1. `close()` no longer unblocks a reader on the ordinary client path, because `cancel()` is a no-op for `TNamedPipeImpl` and `TAnonPipeImpl` (bug due to `TPipeImpl::cancel()` being overridden only by `TWaitableNamedPipeImpl`)

`close()` hands its reference to a concurrent reader and then calls `oldImpl->cancel()`:

https://github.com/apache/thrift/blob/4c6ceee0a401f49bec21bf07fae11db8a2032d7b/lib/cpp/src/thrift/transport/TPipe.cpp#L314-L323

`TWaitableNamedPipeImpl` is the only class that implements `cancel()`:

https://github.com/apache/thrift/blob/4c6ceee0a401f49bec21bf07fae11db8a2032d7b/lib/cpp/src/thrift/transport/TPipe.cpp#L117-L119

`TNamedPipeImpl` and `TAnonPipeImpl` inherit the base no-op:

https://github.com/apache/thrift/blob/4c6ceee0a401f49bec21bf07fae11db8a2032d7b/lib/cpp/src/thrift/transport/TPipe.cpp#L46-L48

and `TPipe::open()` (like the non-anonymous branch of `setPipeHandle()` at L392) installs `TNamedPipeImpl`:

https://github.com/apache/thrift/blob/4c6ceee0a401f49bec21bf07fae11db8a2032d7b/lib/cpp/src/thrift/transport/TPipe.cpp#L311-L313

On that path the reader now holds the last reference, so the `TAutoHandle` member is not destroyed and `CloseHandle()` is not called, while `cancel()` does nothing. A thread blocked in `pseudo_sync_read()`'s `GetOverlappedResult(pipe, &tempOverlap, &bytes, TRUE)` therefore stays blocked. Previously `close()` was `impl_.reset()`, which released the implementation and closed the handle, completing the pending operation and returning control to the reader. So for `TPipe::open()` clients and for anonymous pipes, `close()` no longer ends that wait, which is the opposite of what the comment on those lines states.

Only the `TPipe(TAutoHandle&)` and `TPipe(HANDLE)` constructors build `TWaitableNamedPipeImpl`, so the server-accepted named-pipe path the PR targets does get the intended wake-up. `stress_pipe_close_during_use` cannot catch this: it goes through `setPipeHandle()`, so every iteration uses `TNamedPipeImpl`, and its reads fail fast with `ERROR_PIPE_LISTENING` rather than pending. Either implement `cancel()` for the other two implementations or narrow the comment to the waitable case.

2. A non-trivial change with no JIRA ticket and no `THRIFT-NNNN:` prefix on the title or commit subject (AGENTS.md says "All significant changes need a JIRA ticket." and, for the JIRA row, "Required for all non-trivial PRs")

The GitHub-issues row is scoped to "Typos, trivial compiler warnings, etc."; this change adds a `TCriticalSection` plus two guarded accessors to `TPipe.h`, adds a `cancel()` virtual to the `TPipeImpl` hierarchy, and changes the concurrency semantics of `close()`/`read()`/`write()`. THRIFT-3590 is cited as background, but it is a separate open ticket that this PR does not resolve. The `Client: cpp` line and the AI attribution trailer are both present and correct.

https://github.com/apache/thrift/blob/4c6ceee0a401f49bec21bf07fae11db8a2032d7b/AGENTS.md#L27-L34

Jens-G:
@jnordstrom-iar Could you please adress the findings? Thank you!

jnordstrom-iar:
> @jnordstrom-iar Could you please adress the findings? Thank you!

Yes, I'll look into them!

jnordstrom-iar:
@Jens-G I'm not that familiar with low-level pipe programming in Windows and not the Thrift repo in general (and neither is anyone at the company). I've dispatched a few coding agents on trying to solve this, but the solutions are quite different and the diffs can be large, so it's difficult predict the implications of those changes.

What do you think the best approach for introducing these fixes? For example, (a) a minimal patch scoped tightly to our use case, with lower risk but narrower coverage, or (b) a more general fix that's more "correct" but touches more code and is harder to verify?

Jens-G:
Ok, I'll have a look but not today.

PR #3893, opened 2026-09-20
THRIFT-6310: Build the Thrift compiler on Windows in CI

Jens-G:
> **Now based directly on `master`.** [#3901](https://github.com/apache/thrift/pull/3901) (THRIFT-6320), which settled what this job's import check asserts, has been merged, so this PR no longer carries it.
>
> *(It previously also sat on #3894, which I withdrew — see the note there.)*

The compiler is built in CI on Linux and on macOS. Windows was never covered — the `compiler` job in `build.yml` still carried a `# TODO windows and macos` marker, and only the macOS half of it was ever done. The one Windows job that exists, `msvc.yml`, builds the whole library stack inside a large Windows container, which is a different and much slower thing.

## What this adds

A `compiler-windows` job in `cmake.yml`, next to the existing Linux `compiler` job. It runs natively on a `windows-2025` runner — no container:

- installs `winflexbison3`, pinned to the version `build/docker/msvc/Dockerfile` already pins, since flex and bison are not on the runner image;
- builds the `Release` configuration;
- runs `ctest` — **all 18 compiler tests, including `ThriftCompilerTests`, `StalenessCheckTest`, `MarkdownDocTest` and `ThriftAuditTest`, had never run on Windows before**;
- runs `thrift -version`;
- checks what the built `thrift.exe` imports;
- uploads it as a build artifact.

## The import check

The released `thrift.exe` is handed out as a single file, so it must depend on nothing but Windows itself and the Visual C++ runtime. Boost, OpenSSL, zlib or libevent finding their way into it is a real possibility in this CMake, and today it would only be noticed by hand, once per release, with Dependency Walker.

`build/windows/check-compiler-imports.ps1` reads the import table with `dumpbin`, which it locates through `vswhere` so it also runs outside a developer prompt. `ReleaseManagement.md` now points at it.

**It has to fail in two directions, and an allow list only covers one.** A statically linked build imports strictly *fewer* DLLs, every one of them permitted, so an allow list would wave it straight through — and that is not hypothetical: #3901 is about the release recipe still asking for a static build. `-RequiredImport` is what rejects it, naming the Visual C++ runtime whose absence is the symptom.

## Tests

`check-compiler-imports-tests.ps1` drives the check with canned `dumpbin` output:

- a release build passes (the fixture is the real import table from a Windows CI run);
- **a statically linked build is rejected** — and passes when `-RequiredImport ''` says that is wanted;
- a third-party dependency is rejected, including a delay-loaded one;
- an import table that could not be read fails rather than passing;
- the `Summary` section is not mistaken for imports;
- the wildcard matching that lets one entry stand for the whole UCRT works, and a wildcard that does *not* cover the UCRT still rejects it.

Twelve checks, no compiler and no Windows needed — any PowerShell runs them, including `pwsh` on Linux — and the job runs them before it builds anything. I verified they fail as intended by mutating the required-import check and the wildcard matching in turn.

On top of that, the job fails if CMake did not report `Build compiler: ON`, and `ctest` runs with `--no-tests=error`.

## Notes for review

- **Why not `build.yml`.** That workflow is the autotools path (`bootstrap.sh`, `configure`, `make`), which has no native Windows toolchain. The `# TODO windows` marker there is replaced by a pointer to `cmake.yml`.
- **Why not extend `msvc.yml`.** That job is the container-based full-library build. This one is compiler-only and fast.
- **One runner, not a matrix.** `windows-2025` only, to keep the added CI time down. Happy to add `windows-2022` if you would rather have both.
- No new third-party Actions — only `actions/checkout` and `actions/upload-artifact`, both already used here.

JIRA: [THRIFT-6310](https://issues.apache.org/jira/browse/THRIFT-6310)


.github/workflows/cmake.yml
```diff
@@ -69,3 +69,69 @@ jobs:
           name: cmake-LastTest-log
           path: cmake_build/Testing/Temporary/LastTest.log
           if-no-files-found: warn
+
+  compiler-windows:
+    runs-on: windows-2025
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
```
fishy: usually we comment which tagged version is the pinned sha, can we also comment it here?

fishy (approved):

.github/workflows/cmake.yml
Jens-G: Done — good catch, and thanks for the review.

Added `# v7.0.1` here, and also to the `actions/checkout` in the existing `compiler` job just above: it was the only other unannotated pin in the file, and leaving exactly one uncommented next to a commented one seemed worse than touching a line that is not strictly mine. Shout if you would rather I left it alone.

I verified the SHA resolves to `v7.0.1` against the tags API rather than trusting the comment I was copying.

Related: the security scanner caught two genuinely *wrong* version comments of mine on #3896 (`setup-dotnet` was v6.0.0 not v5.0.0, `NuGet/login` v1.2.0 not v1.0.0) — both fixed there. My local zizmor had reported clean because it defaults to offline mode, where `ref-version-mismatch` cannot resolve tags and never fires. With a token it is clean across all five workflows in this stack now.

Still unannotated elsewhere, but outside this change so I left them: `docker.yml` (two `checkout`, one `setup-go`) and `msvc.yml` (one `checkout`). Happy to send a separate PR for those if you want them tidied.

PR #3896, opened 2026-09-20
THRIFT-6313: Publish the Windows Thrift compiler as a .NET tool

Jens-G:
> **Based directly on `master`.** Everything this depended on ([#3901](https://github.com/apache/thrift/pull/3901), [#3893](https://github.com/apache/thrift/pull/3893)) is merged.

A .NET developer on Windows who wants the Thrift compiler has to find the download page, pick the right file and put it on `PATH` by hand. Every other build-time tool in that ecosystem comes from `dotnet tool install`, and can be pinned per repository in a `dotnet-tools.json` manifest so everyone building a project uses the same compiler version.

```shell
dotnet tool install --global Apache.Thrift.Compiler
thrift --version
```

## The package id

Checked against nuget.org:

| id | |
|---|---|
| `ApacheThrift` | the existing netstd runtime library — stays what it is |
| `Apache.Thrift` | **not available.** Held by a third party and delisted. NuGet ids are permanent, so a delisted id cannot be reused. |
| `Apache.Thrift.Compiler` | free, and follows the `Apache.Avro` / `Apache.Arrow` naming sibling ASF projects use |

`Apache.*` is not a reserved prefix on nuget.org (`Apache.Avro`, `Apache.Arrow` and `ApacheThrift` are all unverified), so a push under this id will be accepted. Only `Apache.NMS*` and `Apache.Ignite*` are reserved, and neither covers it.

## Why there is a managed launcher

A .NET tool package needs a managed entry point, and the payload is a native executable. So the package carries a small launcher that finds `thrift.exe` next to itself and hands over — arguments, standard streams and exit code all pass straight through.

.NET 10 *can* name a native executable as the entry point directly, with no managed code (`ToolCommandRunner=executable`). I tested it: a package built that way **cannot be installed by the .NET 8 or .NET 9 SDK at all** — `unsupported runner 'executable'` — and the two MSBuild properties involved are undocumented. The launcher costs one 72 KB assembly and works on every supported SDK, so the package targets `net8.0`.

## The Visual C++ redistributable

The compiler links the shared runtime, so it needs the redistributable, which the ASF cannot ship. The launcher catches the `Win32Exception` from `Process.Start` and says so with a download link, instead of surfacing Windows' own unhelpful error. The package README states the requirement too.

## Platforms

The package carries the Windows compiler only, and NuGet installs a tool package on any platform without gating. So the launcher checks and says so, pointing at the download page, rather than failing to start a Windows executable. If the compiler is built for other platforms later, .NET 10's RID-specific tool packages turn this into an install-time refusal without changing the install command.

## Testing

`dotnet-tool/test-dotnet-tool.ps1` looks inside the `.nupkg` and then **installs it into a throwaway directory and runs it**. Two things can go wrong with a tool package wrapping a native executable, and neither shows up before somebody installs it: the payload can be missing, and the layout can be such that the tool installs but does not run.

18 checks: package layout, `DotnetToolSettings.xml` command name and entry point, nuspec id/version/`DotnetTool` package type/licence expression, exactly one copy of the payload, non-empty `LICENSE` and `NOTICE`, install, and run.

It runs anywhere PowerShell and the .NET SDK do — I developed it against the real .NET 8 SDK on Linux, where it exercises the platform-guard path; on Windows it checks the compiler's actual output instead. It already earned its keep: it caught that `LICENSE`/`NOTICE` were **not** being packed by the `<None Pack="true">` items I first wrote, because a tool package is built from the *publish* output and does not use the ordinary pack file set. They now travel as content and land next to the executable, where the user actually gets them.

## Publishing

`release: published`, not a pre-release — the gate `pypi.yml`, `release_rust.yml` and `release_ruby.yml` use. The workflow refuses to publish when the release tag and the version in `CMakeLists.txt` disagree, as `release_rust.yml` does for the crate.

Because **a NuGet version can never be replaced**, the compiler's imports are checked *before* the package is built, not only afterwards in `cmake.yml`.

Authentication is NuGet **trusted publishing** via `NuGet/login` (on the ASF allowlist), matching PyPI in `pypi.yml` and crates.io in `release_rust.yml` — no long-lived key is stored in the repo.

### What is needed before the first release

- a trusted publishing policy on nuget.org for `apache/thrift`, the `.github/workflows/dotnet-tool.yml` workflow and the `release` environment;
- a repository variable `NUGET_USER` naming the nuget.org account that owns it.

Both are noted in `doc/ReleaseManagement.md`. If trusted publishing turns out not to be workable for a not-yet-existing package, swapping in an API-key secret is a three-line change — tell me and I will.

## Notes

- `.gitignore` gains the tool project's `bin`/`obj`.
- No version in the `.csproj`; it arrives on the command line, so `build/veralign.sh` gains nothing to keep in step.
- Building without `-Compiler` fails loudly rather than producing a package with no compiler in it.
- `zizmor` is clean on the new workflow.

JIRA: [THRIFT-6313](https://issues.apache.org/jira/browse/THRIFT-6313)


Jens-G:
Fixed — thanks, scanner.

Both pins were right; both of my version comments were wrong:

| | pinned SHA is actually | I had written |
|---|---|---|
| `actions/setup-dotnet` | `v6.0.0` | ~~`v5.0.0`~~ |
| `NuGet/login` | `v1.2.0` | ~~`v1.0.0`~~ |

The `setup-dotnet` one was doubly careless: `build.yml` already carries that exact SHA with the correct `# v6.0.0`.

I had run zizmor locally before opening this, and it reported clean — because it defaults to **offline mode**, where `ref-version-mismatch` cannot resolve tags and so never fires. Re-run with a token it catches both, and is clean after the fix. I also re-resolved every other pin across all five workflows in this stack against the tags API; `actions/checkout` (v7.0.1), `upload-artifact` (v7.0.1), `download-artifact` (v8.0.1) and `setup-python` (v7.0.0) are all correct.

fishy (approved):

PR #3898, opened 2026-09-20
THRIFT-2208: Package the Windows Thrift compiler for Chocolatey

Jens-G:
> **Based directly on `master`.** Everything this depended on ([#3901](https://github.com/apache/thrift/pull/3901), [#3893](https://github.com/apache/thrift/pull/3893), [#3895](https://github.com/apache/thrift/pull/3895)) is merged, so this no longer sits on another PR.

THRIFT-2208 has been open since 2013.

```shell
choco install thrift
```

## The package downloads, it does not carry

The package fetches the Windows installer published with the release and runs it silently for all users. It does **not** embed the compiler. So:

- what a user installs is the file that was voted on and signed;
- no binary is redistributed through a third-party CDN;
- the package stays a few KB.

The URL is `archive.apache.org`, for the same reason as the WinGet manifest: `downloads.apache.org` only carries the *current* release, so a package naming it stops installing at the next one — and Chocolatey keeps old versions installable. The checksum is computed from the file at that exact URL, so the package cannot record a checksum the published file does not have.

That makes it a post-release step: it cannot be built until the archive has the release. The workflow runs when the release is published and can be re-run from the Actions tab.

## Nothing is pushed yet — on purpose

**The `thrift` id is taken.** It belongs to a third-party package, not an ASF one:

| | |
|---|---|
| latest version | **0.12.0**, published February 2019 — thirteen releases behind |
| maintainers | `chaliy`, `Lite` |
| source | <https://github.com/Litee/chocolatey-packages> |

(Its "last updated" field on chocolatey.org moves with Chocolatey's automated re-verification, not with any maintainer action — it looks recent and is not.)

Per our discussion the plan is to **request a maintainer takeover** under Chocolatey's process for abandoned packages, so that `choco install thrift` keeps meaning what people expect. So this PR builds and tests the package but pushes nothing: the push step needs a `CHOCO_API_KEY` secret that does not exist, and when it is absent the workflow attaches the package to the run and says in its summary that it did not push. `doc/ReleaseManagement.md` documents the takeover steps.

If the takeover is refused, `apache-thrift` and `thrift-compiler` are both free — I verified there is exactly one package matching "thrift" on the whole community feed. Switching is an id change in `thrift.nuspec.in` plus the package file name.

## The Visual C++ redistributable

The package depends on `vcredist140`. The compiler links the shared runtime and the ASF cannot ship the redistributable, so Chocolatey installs it first — a `choco install thrift` does not end in a compiler that will not start.

## Testing

`chocolatey/test-chocolatey-package.ps1`, 19 checks: the URL, checksum and silent-install arguments reach the install script; no placeholder survives into a file that would be published verbatim; the nuspec parses and carries the right id, version and licence; `LICENSE` and `NOTICE` are staged and non-empty; a malformed version or checksum is refused.

Given `-Package`, it also looks inside the packed `.nupkg` — including asserting it carries **no executable**, since a package that did would be shipping an unsigned copy of the compiler.

It runs anywhere PowerShell does (`-StageOnly` skips the `choco pack` step), so I developed and ran it on Linux; CI runs it on Windows both before and after packing. It already earned its keep — it caught an over-broad assertion of mine that matched an explanatory comment rather than the URL value.

## Notes

- `/ALLUSERS` and `/TASKS="modifypath"` are what make Chocolatey's machine-wide install put `thrift.exe` on the system `PATH`; both are asserted.
- `zizmor` is clean on the new workflow.

JIRA: [THRIFT-2208](https://issues.apache.org/jira/browse/THRIFT-2208)

---

**Merge note.** [#3897](https://github.com/apache/thrift/pull/3897) (WinGet) and [#3898](https://github.com/apache/thrift/pull/3898) (Chocolatey) are siblings off #3895 and both append a subsection to `doc/ReleaseManagement.md` at the same anchor. Whichever merges second will conflict there; the resolution is to keep both subsections. Nothing else overlaps.


fishy (approved):

chaliy:
@Jens-G is this correct that you would like to take over of thift chocolatey package?

Jens-G:
> @Jens-G is this correct that you would like to take over of thift chocolatey package?

@chaliy Yes, correct, that was the idea of my request.

PR #3919, opened 2026-09-21
THRIFT-4564: Correctly reset the transport in NodeJS

KarboniteKream:
Client: nodejs
Jira: [THRIFT-4564](https://issues.apache.org/jira/browse/THRIFT-4564)

This PR is a follow-up to #1558, where the `reset()` method was added to `TBufferedTransport`. I discovered that the code was trying to call `reset()` on the protocol in the NodeJS branch, instead of on the transport, as intended. The missing `.getTransport()` was added, with an appropriate test.

The `reset()` method is also added to `TFramedTransport`, to unify the behavior.


Jens-G:
### Code review

Two suggestions, below the bar for an issue but verified:

- A small formality for when you next push: CONTRIBUTING.md asks for a `Client: nodejs` line below the subject of the commit message. At the moment it is only in the PR description.

https://github.com/apache/thrift/blob/e8b37a096737f0042c4aab64aa959c256c3802e8/CONTRIBUTING.md#L22-L27

- The generated code now calls `reset()` on the transport if it has one. `TBufferedTransport` does, but `TFramedTransport` does not, so framed clients still send the half-written message in front of the next request: after `testString(42)` throws, the next `testVoid()` goes out as one 50-byte frame that starts with the 25 leftover bytes. This was the same before this PR. Adding a `reset()` to `TFramedTransport` that clears `outBuffers` and `outCount` fixes it; with that, the next frame contains only `testVoid`. That could go into this PR or into a follow-up.

https://github.com/apache/thrift/blob/e8b37a096737f0042c4aab64aa959c256c3802e8/compiler/cpp/src/thrift/generate/t_js_generator.cc#L2277-L2281

https://github.com/apache/thrift/blob/e8b37a096737f0042c4aab64aa959c256c3802e8/lib/nodejs/lib/thrift/buffered_transport.js#L39-L47

PR #3926, opened 2026-09-21
THRIFT-6334: Require PHP 8.2 and document the version support policy

Jens-G:
Raises the minimum PHP to 8.2 for 0.26.0 and adds a version support policy to `LANGUAGES.md`.

### Why now
- PHP 8.1 reached end of life on 2025-12-31; PHP 8.2 still receives upstream security support until 2026-12-31.
- 8.2 therefore stays the floor until 2027. Once it has reached its end of life, the policy makes 8.3 the next floor.
- PHP 8.1 additionally carries end-of-life engine defects that upstream will not fix.

### Changes
- `composer.json`: `"php": "^8.1"` → `"^8.2"`; `lib/php/README.md` minimum → 8.2.
- CI: PHP matrix `[8.1, 8.2, 8.3, 8.4, 8.5]` → `[8.2, 8.3, 8.4, 8.5]`; static-analysis job → 8.2.
- Build images: `ubuntu-jammy` and `ubuntu-focal` → PHP 8.2 (both via `ppa:ondrej/php`); `ubuntu-noble` already ships 8.3.
- `LANGUAGES.md`: PHP `Min` → 8.2, plus a new **Version support policy** section (a guideline, not a hard rule — keep EOL versions only while the cost is manageable, drop when no longer practical).

**Breaking change:** raises the minimum PHP version; users on 8.1 must upgrade.

Reviewer note: no CI job builds the images under `build/docker`, so this PR's CI does not cover the image changes. The jammy PHP install was checked in an `ubuntu:jammy` container, where PHP 8.2.33 becomes the default `php`. The focal image cannot be built at the moment: `ppa:ondrej/php` no longer publishes packages for focal, so this was already the case for the 8.1 it installed before.

Prepared with AI assistance (Claude Opus 4.8, Claude Opus 5); reviewed by the submitter.


sveneld:
Based on the support policy proposed in LANGUAGES.md, I don't see an urgent need to raise the minimum PHP version to 8.3. Maintaining compatibility with PHP 8.1 and 8.2 currently seems to require little effort, and we don't have version-specific branches in the runtime code to accommodate them.

I could agree to raising the minimum to 8.2 as a compromise. It still receives upstream security updates until the end of 2026, and I would expect it to remain in use throughout 2027  —even after upstream support ends—as upgrades in real-world projects tend to take a while. 🙂

sveneld (changes requested):
I could agree to raising the minimum to 8.2 as a compromise. It still receives upstream security updates until the end of 2026, and I would expect it to remain in use throughout 2027 —even after upstream support ends—as upgrades in real-world projects tend to take a while. 🙂

Jens-G:
> upgrades in real-world projects tend to take a while.

Strange argument. That applies to Thrift updates as well then, so where is the problem?

>  8.2 as a compromise

Agreed but it is not really a compromise: 8.1 has known segfault-causing bugs, those are fixed in 8.2 and will not be backported. So 8.1 is only good for scrap metal I guess.

Jens-G:
Updated: the floor is 8.2 now. It stays there until 2027; once 8.2 has reached its end of life, the policy makes 8.3 the next floor.

